PR #3, opened 2017-11-01
Fix typo

thecrypticace:
border color -> border width


reinink:
Thanks @thecrypticace! ❤️

PR #6, opened 2017-11-01
Fix typo on border radius docs

belisarh:
Small mistake on border radius, referencing `.shadow-sm` instead of `.rounded-sm`


reinink:
Ooops, thanks @belisarh! 😄

PR #7, opened 2017-11-01
Fix installation link path

jericopulvera:


reinink:
Thanks @jericopulvera! 💟

PR #17, opened 2017-11-01
Change visiblity property value to hidden

heycesar:
Fix value for property visibility.
`visiblity: hidden`


reinink:
Awesome, not sure how that happened. 🙊

Thanks @cesaramirez!

PR #24, opened 2017-11-01
Add responsive sample to default config

pedroborges:
To be consistent with other sections on the same file.


lodenrogue (approved):

reinink:
Good idea @pedroborges! I did rename "breakpoint" to "screen" since that's the word we've used throughout our docs, and since you can technically use any value media query here, not just breakpoints (ie. print).

PR #28, opened 2017-11-01
Dash in comments is wrong

HellPat:
For margin/padding utilities there is a dash in the comment.
The rendered css-classes look like that: ```.mt-4``` so i think the comments are wrong.


reinink:
You're absolutely right, thanks @psren! 😊

PR #114, opened 2017-11-06
Introduce black and white w/ alpha colors

### LINKED ISSUE #111: Feature request: Black and white w/ alpha colours
I propose we introduce the black colour with different alpha values, so that, e.g. `bg-black-50` sets a background-color of `rgba(0, 0, 0, 0.5)`.

The scale could be `black-10`, `black-20`... `black-90`, as `black-0` would actually be transparent, and `black-100` is already represented by the class `black`.

We will probably need the same for white colours.

I'll be working on a pull request.

1player:
Fixes #111 

I would update the docs, but I have no idea how to run them locally...


adamwathan:
Hey @1player! We considered this originally but decided to leave them out because we have hand-picked light and dark variations of all of the colors in the color scheme which tend to look better than using semi-transparent black/white on a colored background.

We also figured for text situations at least, it's easy to combine the existing opacity helpers to get this sort of effect.

Is there a particular situation you're hitting that these solve better? Open to discussing it, just hesitant to add new default colors if we can avoid it because the impact on file size is pretty significant.

adamwathan:
Just as a quick follow-up, this navigation example from our docs was done entirely with the existing color palette without using the semi-transparent black/white trick and does a good job achieving that same look I think:

image

1player:
Makes sense, Adam, but in my case if I decide to change the colour scheme, I would have the alpha'd black/white at my disposal instead of spending far too much time w/ a color tool to generate lighter and darker shades.

With your example, if I wanted to change the background colour, I could easily replicate the same effect on the "Docs", "Examples" and "Blog" links with a bg-white-70, instead of going on Adobe Kuler to generate the shades, and I don't really like the idea of changing the element opacity just to recreate the same effect.

In any case, the pull request is really simple, and if you prefer not to include it, I can just add the colours in my config file and be done with it, so no problem at all.

reinink:
@1player Thanks for understanding. While we'll pass on this PR, we very much appreciate your effort  here.

And to be clear, there are definitely use-cases for a color system like this. In fact, as @adamwathan  hinted, this is how our original color scheme worked. However, in our testing we found that using hand picked colors tended to look better than "dynamic" alpha colors, which is why they were removed.

And while we could add them back for those who prefer this approach, we always have to be careful how many utilities we ship with, as they all contribute to the overall CSS file size. But there is nothing stopping you from adding them to your project! 💪

marflow:
Sorry @adamwathan, only had been looking at the open issues. My use case would have been divs that have underlying img or colors that you would like to come through. Opacity also changes text and icons that my be inside the div with applied opacity so that´s not a perfect solution. But like @1player I will include it in my config then. Thanks.

adamwathan:
@marflow Totally, I run into those situations a lot too. It's tempting to include them by default but every new color we add generates 30 new classes so it has a really big impact on the default file size. Even though I don't think that's a huge issue in practice, it hurts peoples' perception of the framework to see a default file size of 300kb or more.

We might still change our minds on this down the road but safer to say no for now; it's much easier to add this later than it is to remove it if we add it prematurely.

aparajita:
IMO non-optimized file size shouldn't be as important as functionality out of the box. If we're using PurgeCSS, default file size really doesn't matter in the end.

adamwathan:
@holic:

> If I were able to use black at a few different opacity levels for text colors within the card, I wouldn't need to worry about changing every text element as the colors would blend on their own.

You can totally do this already; just add the colors you want to the `colors` object in your config file. Almost every site anyone ever builds is going to require modifying the default colors; don't be afraid to get in there and customize it for your project.

If you're just setting the text color, there's also nothing stopping you from combining the text color utilities with the opacity utilities. I do this all the time:

`<span class="text-white opacity-75">...</span>`

PR #143, opened 2017-11-07
Add sticky utility

sebastiandedeyne:
Adds a `.sticky` class for `position: sticky`.

https://developer.mozilla.org/en-US/docs/Web/CSS/position

### DIFF
--- a/__tests__/fixtures/tailwind-output.css
+++ b/__tests__/fixtures/tailwind-output.css
@@ -2192,6 +2192,10 @@ button,
   position: relative;
 }
 
+.sticky {
+  position: sticky;
+}
+
 .pin-t {
   top: 0;
 }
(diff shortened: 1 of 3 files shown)


sebastiandedeyne:
Just realised there are larger positioning examples in the docs later on the page. If you guys are interested in merging this I'd be happy to add a a section there too in another commit or follow up PR, let me know!

reinink:
Hey @sebastiandedeyne, thanks for this contribution. One issue is see that sticky doesn't have IE11 browser support, which is a browser we've been trying to support with Tailwind.

image

sebastiandedeyne:
Yeah, can't be solved with pure CSS either... I personally wouldn't mind having utilities that aren't supported everywhere, with a note in the docs where it is or isn't compatible.

I'd totally understand it if you guys think it's a bit early and don't want to deal with stuff like this though, in that case I'll just declare it in my own projects when necessary :)

adamwathan:
Going to close this for now; if we ever add any other utilities by default that don't support IE 11 we can revisit this then. Thanks though!

adamwathan:
Reopening and adding this because position sticky is great, thanks!

PR #154, opened 2017-11-08
Move errant SVG and extract nav links to a partial

mattstauffer:
Note: This only extracts the child links, not the parent; we could make that a separate partial or just keep it in the code like it is now.


reinink:
Real interesting. Almost makes me think we should go full on and do this for the entire nav. Ie. One big data structure. Was there a reason you didn't do that?

mattstauffer:
@reinink I'd love to!! just didn't want to go overboard :) If y'all are up for it I will

reinink:
I think it might actually be better that way, so I say go for it! 🏄 

Thanks buddy 😍

mattstauffer:
@reinink OK, ready for review. Thanks!

reinink (approved):
giphy

reinink:
Hmm, I think we lost at least three pages in the nav:

- Border Radius
- Container
- Display

reinink (changes requested):
Need to add back in missing pages 😂

mattstauffer:
@reinink brought back--thanks for catching! 😳

reinink:
YOU ROCK @mattstauffer

reinink (approved):

PR #170, opened 2017-11-09
Add the word "flexbox" to all flexbox doc pages

reinink:
Right now when you search for "flexbox" in the Tailwind docs you don't see any of the flexbox pages, since they don't actually say "flexbox" anywhere.

This PR simply changes "flex" to "flexbox" in the description of each page, which will fix this. I thought the description might be the best place to add this because the snippet shown in the search will likely end up being very useful.

### DIFF
--- a/docs/source/docs/flexbox-align-content.blade.md
+++ b/docs/source/docs/flexbox-align-content.blade.md
@@ -1,7 +1,7 @@
 ---
 extends: _layouts.documentation
 title: "Align Content"
-description: "Utilities for controlling how lines are positioned in multi-line flex containers."
+description: "Utilities for controlling how lines are positioned in multi-line flexbox containers."
 features:
   responsive: true
   customizable: false
--- a/docs/source/docs/flexbox-align-items.blade.md
+++ b/docs/source/docs/flexbox-align-items.blade.md
@@ -1,7 +1,7 @@
 ---
 extends: _layouts.documentation
 title: "Align Items"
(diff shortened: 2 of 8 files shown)


reinink:
@maxiloc @LukyVj Gents, I hope you don't mind me bugging you one more time about the Algolia DocSearch on our site. As you can see from this proposed PR, we're running into a situation where searching for "flexbox" does not return any of our flexbox pages:

- https://tailwindcss.com/docs/flexbox-display
- https://tailwindcss.com/docs/flexbox-direction
- https://tailwindcss.com/docs/flexbox-wrapping
- https://tailwindcss.com/docs/flexbox-justify-content
- https://tailwindcss.com/docs/flexbox-align-items
- https://tailwindcss.com/docs/flexbox-align-content
- https://tailwindcss.com/docs/flexbox-align-self
- https://tailwindcss.com/docs/flexbox-flex-grow-shrink

The reason, as noted above, is quite simple: the word "flexbox" does not actually exist on those pages.

The problem with this PR is that we're actually not using the proper language for flexbox. For example, while the spec is called "flexbox", "flexbox containers" are not a thing—they are actually "flex containers".

My question is simple: is there a way for us to add keywords to a page via meta tags or something, without actually having those words on the page? Or maybe you have another recommendation on how to handle this?

Thanks!

maxiloc:
I updated the config so it take the infos from the sidebar. That should fix the issue. Tell me if that's good for you.

reinink:
@maxiloc Incredible! 👯

PR #173, opened 2017-11-10
Change the config.js to tailwind.js tht is recommended in step 3.

ericlbarnes:
In step 2 of installation you say: 

> We recommend creating a `tailwind.js` file in your project's root.

Then in Step 4: 

``` 
./node_modules/.bin/tailwind build styles.css [-c ./config.js] [-o ./output.css]
```

Why config.js and not tailwind.js that you recommended earlier? This PR I just tried to unify the docs. 😄

### DIFF
--- a/docs/source/docs/installation.blade.md
+++ b/docs/source/docs/installation.blade.md
@@ -110,7 +110,7 @@ To avoid specificity issues, we highly recommend structuring your main styleshee
 For simple projects or just giving Tailwind a spin, you can use the Tailwind CLI tool to process your CSS:
 
 <div class="bg-smoke-lighter font-mono text-sm p-4">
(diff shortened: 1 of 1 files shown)


reinink:
Hey there Eric! So, the reason it's different here is because when you use `tailwind build` it will name the config file `tailwind.js` by default. The `-c` option allows you to change this to something else. This is why we had something different here, to illustrate that you can use a non-default value. But maybe that's confusing and it still makes sense to show the default value?

adamwathan:
Actually without the -c option it uses no config file, so it probably makes sense to be consistent with the name. The reason for using config there though was to make it clear that the argument is your config file, but maybe that’s still clear even without the word config there?

reinink:
Right! I'm getting confused with `init`. I think originally we actually had `your-tailwind-config.js` here, but I shortened it to make it fit in the docs without scrolling:

image

I'd still like to keep it from scrolling, but also agree that using `tailwind.js` is better. I'll merge this in.

PR #202, opened 2017-11-14
Feature/add facebook meta tags

christophrumpel:
Hey Adam,

like requested on Twitter here is the PR :-)

This PR adds the Facebook Open Graph Meta Tags to the site. This way the right content will be shown when the site is shared on Facebook. I have used the given Twitter Card image. Please change it if you want to use another image.

Greets


reinink:
Does it make sense to have a more dynamic title and description, like we've done for the Twitter meta tags?

christophrumpel:
Yeah you're right. I'll add it. 👍

reinink:
Thanks Chris! 🎉

reinink:
Hmmm, grim, I don't think this got added to the home page. Might have to also update the master layout. Feel liek it Chris? :)

PR #282, opened 2017-12-02
Add variants section to docs

MichaelDeBoey:
As requested by @adamwathan [on Twitter](https://twitter.com/adamwathan/status/936749392014348288) 🙂

Not all is done yet, but thinking about making a partial for it?

### DIFF
--- a/docs/source/docs/flexbox-align-content.blade.md
+++ b/docs/source/docs/flexbox-align-content.blade.md
@@ -270,36 +270,13 @@ For more information about Tailwind's responsive design features, check out the 
 
 ## Customizing
 
-### Responsive, Hover, and Focus Variants
-
-By default, only responsive variants are generated for flexbox utilities.
-
-You can control which variants are generated for the flexbox utilities by modifying the `flexbox` property in the `modules` section of your Tailwind config file.
-
-For example, this config will _also_ generate hover and focus variants:
-
-```js
-{
-    // ...
-    modules: { 
-        // ...
-        flexbox: ['responsive', 'hover', 'focus'],
-    }
-}
-```
-
-Note that modifying the `flexbox` property will affect what variants are generated for _all_ flexbox modules, not just the align-content utilities.
-
-### Disabling
-
-If you aren't using the flexbox utilities in your project, you can disable them entirely by setting the `flexbox` property to `false` in the `modules` section of your config file:
-
-```js
-{
-    // ...
-    modules: {
-        // ...
-        flexbox: false,
-    }
-}
-```
+@include('_partials.variants-and-disabling', [
+    'utility' => [
+        'name' => 'flexbox',
+        'property' => 'flexbox',
+    ],
+    'variants' => [
+        'responsive',
+    ],
+    'extraMessage' => 'Note that modifying the <code>flexbox</code> property will affect what variants are generated for <em>all</em> flexbox modules, not just the align-content utilities.'
+])
--- a/docs/source/docs/flexbox-align-content.blade.md
+++ b/docs/source/docs/flexbox-align-content.blade.md
@@ -267,3 +267,39 @@ For more information about Tailwind's responsive design features, check out the 
 </div>
 @endslot
 @endcomponent
+
+## Customizing
+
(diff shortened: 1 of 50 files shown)


MichaelDeBoey:
Also the `flexbox` docs aren't in 1 place, so what would be a good solution for them?
copy/paste in each file?
Add a specific "global flexbox" file? If so, how do we handle the navigation for this one?

adamwathan:
You are a madman! I was going to be happy to even see a PR come through that updated a single page 😄

The only change I would love to make if you don't mind is replacing this language:

> By default, no hover, focus, or group-hover variants are generated for...

I think phrasing it that way made sense for the SVG page where no variants are generated at all, but for modules that do have variants I think it's more useful to say:

> By default, only responsive variants are generated for...

...or:

> By default, only responsive and hover variants are generated for...

Thanks man, this is amazing!

adamwathan:
Oh, re: flexbox stuff, it is admittedly a bit of a tricky one. We considered making each section it's own module but it seemed unnecessary; I think people want all of flexbox or none of it.

I think we can just copy and paste the same language on each page, and maybe add a note like:

> Note that modifying the `flexbox` property will affect what variants are generated for all Flexbox modules, not just the (align items|justify content|etc) utilities.

...after the code sample block or something?

MichaelDeBoey:
Figured I just do every doc page, so this one is done 🙂

Will make the extra note for the `flexbox` utilities 🙂

What about extracting into a separate component?
Maybe this can be done for the hoverable piece also (will make a new PR for that if approved 🙂)

adamwathan:
I don't mind the idea of creating a partial for it but would want to try and make it flexible so we could still tweak the language per-page if necessary, or provide a different code sample if we wanted to.

MichaelDeBoey:
@adamwathan I think we can start with a partial for almost all the utility pages, but have a custom piece on the special pages (`svg`, ...)?

docs/source/docs/flexbox-align-content.blade.md
```diff
@@ -267,3 +267,39 @@ For more information about Tailwind's responsive design features, check out the
...
+{
+    // ...
+    modules: { 
+        // ...
+        flexbox: ['responsive', 'hover', 'focus'],
+    }
+}
+```
+
+Note that modifying the `flexbox property will affect what variants are generated for _all_ flexbox modules, not just the align-content utilities.
```
adamwathan: Missing closing backtick here after `flexbox`

MichaelDeBoey:
Fixed in 923a5a2 🙂

MichaelDeBoey:
@adamwathan Do you prefer the use of the Prism library for the syntax highlighting or do I create a custom version like in the `customized-config` partial?

adamwathan:
Yeah I’m ok with using a custom one here, it makes it easier to do a nice diff-ish format where we can show both the default variants and a customized version for each example 👍🏻

MichaelDeBoey:
tbh, I was more thinking about something like (I like the highlight of the strings):
schermafbeelding 2017-12-03 om 01 09 22
But you prefer more something like
schermafbeelding 2017-12-03 om 01 09 15
Can try to figure out the string highlighting inside the + part if you want?

MichaelDeBoey:
I like the diff-ish idea, but also like the Prism highlighting 😛
Will keep the diff-ish custom partial we already use 🙂

schermafbeelding 2017-12-03 om 02 05 48

MichaelDeBoey:
@adamwathan All is extracted to a partial
Only the `svg` and `spacing` docs aren't extracted to the partial yet, because they're special cases with multiple utilities that get changed, and that's not (yet) supported by the partial.

MichaelDeBoey:
@adamwathan Don't know if it's a good idea to make the partial more complicated to also support docs like the `svg` and `spacing` ones?

adamwathan:
I think just keep those as done manually with no partial for now 👍🏻

MichaelDeBoey:
OK then this one can be merged I think 🙂

Maybe @reinink or @davidhemphill have some thoughts about this one yet? 🙂

adamwathan:
Merged, thanks!

PR #348, opened 2018-01-04
Add Position top offset 100%

stefanoruth:
When creating dropdown menuens and tooltip actions, this can be useful. when combining with position relative and absolute. If this is interesting, i will add the documentation aswell :)

About the naming, maybe there is a better name?

Example:
capture

### DIFF
--- a/__tests__/fixtures/tailwind-output.css
+++ b/__tests__/fixtures/tailwind-output.css
@@ -3481,6 +3481,10 @@ button,
   left: 0;
 }
 
+.pin-t-100 {
+  top: 100%;
+}
+
 .resize-none {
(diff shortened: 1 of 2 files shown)


stefanoruth:
Can anyone guide me to, whats failing? do i need to add the final css output somewhere too?

tlgreg:
It's the sanity test to check if the css output of the whole tailwind process is what it should be.
When you add a new utility also update the tailwind-output.css fixture.

stefanoruth:
@tlgreg could you guide me a bit close to whats missing now, i have sometrouble running the test locally, it just won't finish.

capture2

PR #353, opened 2018-01-11
Provide a safer and future-proofed system font stack

nternetinspired:
Hey guys!

This PR brings two discrete additions:

1. By transposing `system-ui` and `BlinkMacSystemFont` keywords, it eliminates the potential for issue which _may_ occur if the font-stack were used in a shorthand declaration, as that causes IE and Edge to choke at `-apple-system` and ignore the entire declaration. See [Implementing system fonts on Booking.com — A lesson](https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f) learned for full details.

2. Future proofs the stack by adding support for the newer, unprefixed, SystemUI font-family keyword that was [introduced with Chrome 56](https://www.chromestatus.com/feature/5640395337760768) with the aim of providing a standard, cross-browser implementation.

Thanks for your consideration. 

Absolutely loving Tailwind so far, keep up the great work!

### DIFF
--- a/__tests__/fixtures/tailwind-output.css
+++ b/__tests__/fixtures/tailwind-output.css
@@ -2378,7 +2378,7 @@ button,
 }
 
 .font-sans {
-  font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
+  font-family: system-ui, BlinkMacSystemFont, -apple-system, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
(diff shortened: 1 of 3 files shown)


sgraewe:
system-ui is not supported in IE11 and Edge

https://caniuse.com/#search=system-ui

nternetinspired:
Correct @sgraewe but thats's kinda irrelevant, and why it's sensible to declare a font stack rather than a single family in the first place; a browser that doesn't understand it will just skip it and move on to the next one.

It's perfectly valid stack declaration, and causes no issues, for either Edge or IE;

screen shot 2018-01-13 at 16 48 05

The only issues with using such a stack in IE is with using the -apple-system keyword first, the issue that this PR will resolve by transposing the keywords, as described above.

adamwathan:
Thanks!

PR #490, opened 2018-06-14
Add new outline module

reinink:
This adds a new outline module, which can be configured in a very similar way to shadows.

```js
outline: {
  default: `2px solid rgba(52,144,220,0.5)`,
  'none': 'none',
},
```

However, now that I've built this, I sort of think shadows are a better choice regardless, read more about that here: https://github.com/tailwindcss/tailwindcss/issues/56#issuecomment-397363973

I'm thinking we simply create a hardcoded `outline-none` reset instead, and be done with it.

### DIFF
--- a/__tests__/fixtures/tailwind-output.css
+++ b/__tests__/fixtures/tailwind-output.css
@@ -3191,6 +3191,22 @@ button,
   opacity: 1;
 }
 
+.outline {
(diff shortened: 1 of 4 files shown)


m1guelpf:
@reinink I think the `outline-none` utility would be enough.

reinink:
Closing in favour of #491

PR #502, opened 2018-06-21
Added more grid sizes (up to 12)

niclashedam:
Since many new users of Tailwind comes from Bootstrap, I think it makes perfectly good sense to use a grid of size 1/12 instead of just 1/6. I also often come across needing such precision.

Let me know what you think about this.


adamwathan:
Hey @NiclasHedam! Since this is so easy to change in your own config file we're going to hold off on this at least for now.

NiclasHedam:
Hi @adamwathan. I certainly do agree this is an easy change, but to know where to change it, you would have to understand how Tailwind compiles itself. I think that is more than any 'new' developer knows.

NiclasHedam:
On a side note: This is also true for the reduceable parts of the grid. I would fx expect that `2/6` was usable, even though it is the same as `1/3`. It's not that the current state doesn't make any sense. It's just a steeper learning curve for newcomers in my opinion.

adamwathan:
The thing with Tailwind is that every user is expected to generate and
maintain their own tailwind.js config file; the defaults are supposed to be
a useful general starting point but were not at all attempting to cater to
people who are unwilling to use the framework without a build step and
without understanding how to customize it.

The main reason Tailwind exists is because customization of a framework is
inevitable for almost any project, and other frameworks make it needlessly
painful to do or don’t provide enough guidance. With Tailwind we’re trying
to say, “look, I know you’re going to need to customize this, so we’ve
tried to make it as easy as possible and have put together a ton of
guidance on how to do it.”

Thanks again for the contribution though and I totally understand where
you’re coming from — we may still add more down the line but the thing
about open source is that “yes” is forever, “no” is for now, so I’d rather
avoid adding new stuff until I don’t have any doubts that I want it to be
there permanently.

PR #777, opened 2019-03-17
Update defaultConfig.stub.js

kevinruscoe:
The serif array seems to be on a single line whereas every other array is multiline, this just adds consistency.


MichaelDeBoey:
@kevinruscoe Prettier is formatting it like that

kevinruscoe:
Ah, in that case should "printWidth" be lower? The actual prettier docs says "For readability we recommend against using more than 80 characters".

adamwathan:
I find 80 too low personally, but I’m cool with this change in general,
just not sure how we can prevent Prettier from “fixing” it all the time
without lowering the line length across the board, which results in some
really nasty looking code.

MattStypa:
it looks like `.prettierignore` could be used to ignore the stubs. We already do it for eslint

kevinruscoe:
Yeah, either using `.prettierignore` to globally ignore it, or using inline comments `// prettier-ignore`

adamwathan:
Looks like the stubs are already ignored in our `.eslintignore` so just killed the `.prettierignore` file and merged, thanks!

PR #796, opened 2019-03-20
Remove comment text

zaknesler:
Inspired by Laravel, I guess. Just to keep the default config cleaner 🤷‍♂️


devcircus:
Those seem like placeholders until the actual comments get added when 1.0 is released.

zaknesler:
@devcircus You know what yeah that would make sense. I'm sure Adam has plans for that 😅

PR #807, opened 2019-03-28
add the option to configure classes names and sides

trivago-vgarcia:
Optional configuration for classes names and sides.

The reason to create this PR is because we use a different naming convention in our utilities library, it was written without any generator. We think there is a lot of potential by using Tailwind CSS to generate that library, and we will love to colaborate with the Tailwind CSS community with this PR.


hacknug:
Interested on getting this (or an alternative) merged in. Also like it because having `sides` will make it easier to add support for [Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) which is something I've wanted to release a plugin for for a while to better support non ltr-ttb languages.

Will probably create a plugin to generate Tachyons-like classes if this gets merged since I have quite a few projects where I'd love to remove it from and start fully using Tailwind (and not just using it behind the scenes with `@apply`).

gc-victor:
Thanks for your reply!

No problem.

We will keep working in our fork. Maybe in the future we will merge it ;)

Great work!

PR #900, opened 2019-05-07
feat: extend list container reset to <menu> tag

alexsasharegan:
I just noticed that the menu tag has some strange defaults and extending the magin reset from other list types seems to fix this.


alexsasharegan:
It looks like this tag isn't well supported. Not sure if this is an archaic tag, or new school, but I think I should probably remove it in my application vs. add it to tailwind.

PR #999, opened 2019-07-03
Add default border-width to hr elements

### LINKED ISSUE #979: Bug? Horizontal rules are invisible by default

aggreggator:
Resolves  tailwindcss/tailwindcss#979

The resulting horizontaler ruler has a height of 2 pixels.

### DIFF
--- a/__tests__/fixtures/tailwind-output-important.css
+++ b/__tests__/fixtures/tailwind-output-important.css
@@ -470,6 +470,10 @@ img {
   border-style: solid;
 }
 
+hr {
+  border-width: 1px;
+}
+
 textarea {
   resize: vertical;
 }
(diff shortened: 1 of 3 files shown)


aggreggator:
Close duplicate PR.
Solution in #991

PR #1059, opened 2019-08-01
Added display: 'contents'

afuno:
More details: https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents

In the future, it would be great to add this to the configuration file.


iskrisis:
Why was this closed?

maurer2:
Hello, probably for the same reason that `display: flow-root` has not been added yet. Support in Safari is a bit buggy (https://github.com/tailwindcss/tailwindcss/pull/1247 & https://caniuse.com/#feat=css-display-contents).
But it would definitely be useful to have  `display: contents` and `display: flow-root` in a future release.

jlsorak:
+1 for `display: contents` support

PR #1095, opened 2019-08-25
Add vertical align sub & super

hansott:
- sub: Aligns the baseline of the element with the subscript-baseline of its parent
- super: Aligns the baseline of the element with the superscript-baseline of its parent


hansott:
I need to fix tests first, will open PR again when fixed.

PR #1107, opened 2019-09-02
Add support for multiple config resolving

serg-plusplus:
Problem:
I have Tailwind Config in my design-system(one repo) and I want to extend it in my app(another repo). How can I do this?

Example of my solution:
```js
  const resolveConfig = require('tailwindcss/resolveConfig')
  const designSystemConfig = require('@my-org/my-design-system/tailwind.config')

  module.exports = resolveConfig(designSystemConfig, {
    theme: { colors: { kek: '#123456' } }
  })
```

(NO breaking changes!)

### DIFF
--- a/resolveConfig.js
+++ b/resolveConfig.js
@@ -1,6 +1,6 @@
 const resolveConfigObjects = require('./lib/util/resolveConfig').default
 const defaultConfig = require('./stubs/defaultConfig.stub.js')
 
-module.exports = function resolveConfig(config) {
-  return resolveConfigObjects([config, defaultConfig])
+module.exports = function resolveConfig(...configs) {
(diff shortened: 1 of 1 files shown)


hacknug:
Would love to see `resolveConfig()` being able to handle more scenarios. In my case I could use it to merge multiple configs but I need it not to use the default config sometimes. This is how I worked around it: https://github.com/hacknug/tailwindcss-plugin-utils/blob/master/src/index.js#L104-L125

Note `variants` can either be an object or an array, not sure this PR would cover that scenario or not (haven't tested, sorry).

serh11p:
'tailwindcss/resolveConfig' works differently that I want

PR #1191, opened 2019-10-26
Add touch-action utilities

innocenzi:
This PR adds 10 new utilities for [`touch-action`](https://developer.mozilla.org/fr/docs/Web/CSS/touch-action).


tlgreg:
IMO if it's implemented, it should use the `theme`, because with separate `touch-action` utilities one can't combine multiple (the different axes and pinch-zoom can be combined). If it's in theme one can just add their combination of touch actions and it's not necessary to generate all by default.

hawezo:
That's a good point. I'll update my PR.

PR #1479, opened 2020-03-23
Support for justify-self

### LINKED ISSUE #868: [Feature Request] add support for `justify-self` flexbox utilities
It seems kind of random that Tailwind provides utilities for `align-self` declarations, but not for `justify-self`. Also the current `self-*` classes are very ambiguous as one would not know if `self-center` applies `align-self: center` or `justify-self: center` until they have tried it. If it doesn't have the desired behavior, they are left to come up with some kind of custom solution, which doesn't feel very satisfying given that each plugin of Tailwind is meant to be comprehensive.

I propose that the current `self-*` classes be renamed to something less ambiguous (e.g., `align-self-*` or `self-align-*`) and that `justify-self` classes be introduced (e.g., `justify-self-*` or `self-justify-*`).

fex80:
I added the most needed properties of justify-self. I did not include all, to keep css size reasonable. Is this what you envisioned, or should I change it in some way?

 Resolves tailwindcss/tailwindcss#868

### DIFF
--- a/__tests__/fixtures/tailwind-output-important.css
+++ b/__tests__/fixtures/tailwind-output-important.css
@@ -3560,6 +3560,26 @@ video {
   align-content: space-around !important;
 }
 
+.justify-self-start {
+  justify-self: start !important;
+}
+
+.justify-self-end {
+  justify-self: end !important;
+}
+
+.justify-self-center {
+  justify-self: center !important;
+}
+
+.justify-self-stretch {
+  justify-self: stretch !important;
+}
+
+.justify-self-baseline {
+  justify-self: baseline !important;
+}
+
 .flex-1 {
   flex: 1 1 0% !important;
 }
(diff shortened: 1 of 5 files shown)


ekazda:
Any way you can add 'Justify-Items' as well? I believe this is also missing.

fex80:
@ekazda I surely could. But as this point I'd like to first hear back from @adamwathan or @reinink if they will be merging this PR. Should I update it? It's been waiting for some time now while other PRs have been merged, so it might just not be what the maintainers want for tailwind at this time.

corysimmons:
Easy/helpful one (lots of thumbs up on [the associated issue](https://github.com/tailwindcss/tailwindcss/issues/868)). Worth a quick review/merge.

adamwathan:
Planning to review and likely merge this just haven't had a chance. Nothing quick about it I'm afraid, making decisions about permanent public APIs takes a lot of consideration.

adamwathan:
Does `justify-self-baseline` actually work? I noticed no browsers support `baseline` for `justify-content` but the compatibility chart for `justify-self` is not very detailed at all on MDN:

image

It's literally _wrong_ even, because `justify-self` does nothing in flex layouts — it certainly doesn't have "full support" hah.

corysimmons:
>`justify-self` does nothing in flex layouts

True https://codepen.io/corysimmons/pen/yLejXVO?editors=1100 😬 

But just because flex doesn't work, `justify-self-center` should still be a class right? For the thing that does work?

It might introduce confusion, but no more than `justify-self: center` already does.

FredyC:
Shouldn't support considerations be left to browsers and developers? I mean, it's in the spec, so it should be part of the API.

simonswiss:
Linking this with #2241 👍

Dozorengel:
Until it's not officially supported I use this package for grids, it works well for me
https://github.com/subhero24/tailwindcss-grid-placement

adamwathan:
Merged this to another branch for now as I'm going to also add the other missing alignment properties that only work in grid there 👍  Thanks!

PR #2000, opened 2020-07-05
use percentage value for rounded-full

JaKXz:
Curious to hear if there's a really good reason to use `9999px` - I think the percentage value is more intuitive to see in the inspector tools.

I perhaps would also go as far as renaming this to `rounded-circle` but I didn't want to make a breaking change :)


estevanmaito:
`50%` is not the same as `9999px`
See https://github.com/tailwindcss/tailwindcss/issues/1949

JaKXz:
Thanks! that issue didn't come up in my search 👍

PR #2224, opened 2020-08-21
Fix multiline @apply values

RobinMalfait:
This PR will allow you to use spaces, tabs and newlines in the values of `@apply` rule.

This is a bit of an obscure use case so not 100% sure it will be accepted. However it was not a lot of code 🙃


Evertt:
I'm super happy this got merged. @adamwathan would you be willing to create a new release already? Because at the moment I have some *really* long `@apply` lines that I'd like to format a bit prettier. 😇

PR #2225, opened 2020-08-21
add tailwind -p init option to generate a postcss file

RobinMalfait:
As seen here https://twitter.com/adamwathan/status/1296950213907820545 👀


mikemand:
I was having troubles with PostCSS complaining something about Object is not valid for config. Turns out even though I had (manually, since it wasn't added when I ran `tailwind init -p`) installed `postcss-cli@8.0`, it didn't install PostCSS itself at `v8.0`, it was still 7.something. After adding `"postcss": "^8.0"` to my `package.json`, I can use this new object-based plugin syntax.

PR #2422, opened 2020-09-22
Remove slash at the beginning of href prop

marco-carvalho:
With the slash, it was trying to reference the file outside of the root folder.
Without it, it reference the dist folder inside the root folder.


adamwathan:
Hey thanks! The current instructions are actually correct though, when using a proper web server (like `live-server` which we recommend in the same documentation) the absolute path works correctly, so not going to change this I don't think. I would definitely use the forward slash + a real web server process myself, so I want to recommend the approach I'd use.

marco-carvalho:
Hello @adamwathan, thanks for the reply. Sorry to bother you with something so silly like this, but I created an `index.html` file in the root of the Tailwind project, used the `live-server` and it doesn't work with slash. Maybe it is something with `live-server`, IDK.

With slash:
image

Without slash:
image

PR #2444, opened 2020-09-24
proposal: Add marker utilities

neupauer:
This PR proposal adds a new set of utilities for the marker box of a list item.

| Class                     | Description                   |
| ------------------------- |:-----------------------------:|
| marker-{type}             | Marker type (shape)           |
| marker-{color}-{shade}    | Applies color to the marker   |
| marker-opacity-{modifier} | Applies color opacity         |
| marker-space-{modifier}   | Space between marker and item |

### Usage

```html
<ol class="marker-decimal marker-blue-500 marker-space-3">
  <li>Hello</li>
  <li>Hello</li>
  <li>Hello</li>
</ol>

<!-- Works on any elements, not just lists -->
<div class="marker-disc marker-blue-500 marker-space-3">
  <span>Hello</span>
  <span>Hello</span>
  <span>Hello</span>
</div>
```

Since the `::marker` pseudo-element is not yet supported I came up with solution which uses `display: list-item` on `::before` pseudo element and should work even in IE11.

### Design details

The general idea here is that we change the actual `display` property  on list item to something generic e.g., `display: block`, and we use this hack of setting a `display: list-item` property on list item `::before` pseudo-element.

So we can customize the marker with `list-style-type` ( because the marker will be rendered inside ::before element instead of list item ),  also we can customize the marker with `content: "something"` ( because why not ;) ), and we can customize the marker with properities like `width`, `height`, `background-color`... ( because it is a pseudo-element  )

**Still WIP**, so any feedback is more than welcome.

See my proof of concep on [CodeSandbox](https://codesandbox.io/s/tailwindcss-plugin-marker-proof-of-concept-ie37g)


neupauer:
Closed in favor of https://github.com/tailwindlabs/tailwindcss/pull/2472

PR #2616, opened 2020-10-19
Refactor util functions

20lives:
following #2610 this is the first step at removing Lodash

### DIFF
--- a/src/util/getAllConfigs.js
+++ b/src/util/getAllConfigs.js
@@ -1,21 +1,19 @@
 import defaultConfig from '../../stubs/defaultConfig.stub.js'
 import { flagEnabled } from '../featureFlags'
-import { flatMap, get, isFunction } from 'lodash'
 
 export default function getAllConfigs(config) {
-  const configs = flatMap([...get(config, 'presets', [defaultConfig])].reverse(), (preset) => {
-    return getAllConfigs(isFunction(preset) ? preset() : preset)
-  })
+  const configs = (config?.presets ?? [defaultConfig])
+    .slice()
+    .reverse()
+    .flatMap((preset) => getAllConfigs(preset instanceof Function ? preset() : preset))
 
   const features = {
     // Add experimental configs here...
   }
 
-  Object.keys(features).forEach((feature) => {
-    if (flagEnabled(config, feature)) {
-      configs.unshift(features[feature])
-    }
-  })
+  const experimentals = Object.keys(features)
+    .filter((feature) => flagEnabled(config, feature))
+    .map((feature) => features[feature])
 
-  return [config, ...configs]
+  return [config, ...experimentals, ...configs]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
     "jest": true
   },
   "parserOptions": {
-    "ecmaVersion": 2018,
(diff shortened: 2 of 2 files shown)


adamwathan:
Hey sorry, mind fixing the conflict then happy to merge? Can fix myself if needed.

20lives:
:wink:

20lives:
btw, I also ran additional tests where the experimental features array isn't empty

src/util/getAllConfigs.js
```diff
@@ -1,21 +1,18 @@
 import defaultConfig from '../../stubs/defaultConfig.stub.js'
 import { flagEnabled } from '../featureFlags'
-import { flatMap, get } from 'lodash'
 
 export default function getAllConfigs(config) {
-  const configs = flatMap([...get(config, 'presets', [defaultConfig])].reverse(), (preset) => {
-    return getAllConfigs(preset)
-  })
+  const configs = (config && config.presets ? config.presets : [defaultConfig])
+    .reverse()
```
adamwathan, line 6: Should be careful here because `reverse` mutates the array, so we are directly mutating `config.presets` which could have side-effects outside of Tailwind itself.

20lives: Good catch.

src/util/getAllConfigs.js
```diff
@@ -1,21 +1,19 @@
 import defaultConfig from '../../stubs/defaultConfig.stub.js'
 import { flagEnabled } from '../featureFlags'
-import { flatMap, get } from 'lodash'
 
 export default function getAllConfigs(config) {
-  const configs = flatMap([...get(config, 'presets', [defaultConfig])].reverse(), (preset) => {
-    return getAllConfigs(preset)
-  })
+  const configs = (config && config.presets ? config.presets : [defaultConfig])
```
adamwathan, line 5: Personally I find the reliance of truthiness and shortcircuiting stuff in these type of statements a bit obfuscating. We rely on `get` from lodash quite a lot throughout the codebase and I think making this sort of change everywhere is going to make the code quite a bit less declarative and harder to read.

I'd like some sort of abstraction around these sorts of "grab a key from an object but use this default if it's not there" operations which happen so much in the codebase. One very robust and very small library I've used for this in the past is `dlv`, perhaps we can just pull that in and start swapping these over?

https://github.com/developit/dlv

20lives: by combining es2020 **optional Chaining** and **nullish coalescing** you get something as declerative as get or dlv:
```JS
config?.presets ?? [defaultConfig]
```

Since you already have **Babel** configured, only need to update **eslint** ecmaVersion.


Anyhow it's your decision.

adamwathan: I'm good with that if you want to update the eslint config in this PR too then 👍🏻

20lives:
gone

Another conflict, I can't keep up :joy:

20lives:
Dropped `flattenColorPalette` commit, will add it in another PR.

adamwathan:
Sorry another conflict 😬

adamwathan:
Finally ;)

20lives:
Shocking, I can't even remember what I did here.
Was pretty sure this will not be relevant with all the new JIT stuff.

PR #2710, opened 2020-10-30
Use esbuild for building Tailwind

RobinMalfait:
This PR will use `esbuild` instead of `babel` for compiling the js files. We will also use esbuild as a transformer for our jest tests. This allows us to make everything a bit faster!

### DIFF
--- a/jest/customMatchers.js
+++ b/jest/customMatchers.js
@@ -2,7 +2,6 @@ import prettier from 'prettier'
 import diff from 'jest-diff'
 
 function format(input) {
-  return input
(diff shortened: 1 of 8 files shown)


reslear:
hi, thank you for experience, very good idea for fast compilation, what is the state of migration, maybe need help?)

RobinMalfait:
This PR is a bit outdated, and esbuild was only used to build tailwind itself. Esbuild was not used for end-users directly. Going to close this for now, we may want to do this again later, but not that important currently.

PR #2750, opened 2020-11-10
🎨 Add CSS custom property definitions for color utilities

brandonpittman:
Not a significant feature, but a nice thing to have. While screwing around with bg-gradient and custom underlines, I found a need to create a text-shadow that referenced the current background color. Having that as a custom property would be super useful.

Output winds up looking like this:

```css
.text-blue-100 {
  --text-opacity: 1;
  --tw-text-color: rgba(219, 234, 254, var(--text-opacity));
  color: rgba(219, 234, 254, var(--text-opacity));
}
```

Since Tailwind 2 won't support IE11, you could probably even take it down to this.

```css
.text-blue-100 {
  --text-opacity: 1;
  --tw-text-color: rgba(219, 234, 254, var(--text-opacity));
  color: var(--tw-text-color);
}
```

## Todo

- [ ] update fixtures to ensure sanity tests pass

### DIFF
--- a/__tests__/withAlphaVariable.test.js
+++ b/__tests__/withAlphaVariable.test.js
@@ -2,9 +2,15 @@ import withAlphaVariable from '../src/util/withAlphaVariable'
 
 test('it adds the right custom property', () => {
   expect(
-    withAlphaVariable({ color: '#ff0000', property: 'color', variable: '--tw-text-opacity' })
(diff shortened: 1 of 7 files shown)


brandonpittman:
TODO: update fixtures to ensure sanity tests pass

PR #2773, opened 2020-11-16
Postcss7 compatibility

RobinMalfait:
We want to use `PostCSS 8` in Tailwind, however a lot of tools are not compatible with `PostCSS 7` yet. To account for this we can will create a separate tag on npm:

```js
npm install tailwindcss@compat
```

---

To get the repo into `PostCSS 7` mode we have to run `yarn compat` or `npm run compat`. This will prepare the repo for PostCSS 7 mode. Essentially it will do multiple things:

1. Run `yarn babelify`, to ensure that we have the latest and greatest build
2. Update `package.json` to include dependencies from the `package.postcss7.json` file, it will deeply merge all the values
3. Use the `lib/index.postcss7.js` as entry point, this will allow us to use the "old" PostCSS Plugin register system.

Once you run this command, you will get the following output:

```
You can safely publish `tailwindcss` in PostCSS 7 compatibility mode:

  npm version
  npm publish --tag compat
  npm run compat:restore
```

2 important parts here are the publish with the `--tag` of `compat`, this allows for the `npm install tailwindcss@compat`.

To restore everything to normal, you can use the `yarn compat:restore` or `npm run compat:restore`. This does the inverse and uses the old `package.json` and `lib/index.js` entry point.

This will output:

```
Restored from PostCSS 7 mode to latest PostCSS mode!
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,10 @@
     "rebuild-fixtures": "npm run babelify && babel-node scripts/rebuildFixtures.js",
     "prepublishOnly": "npm run babelify && babel-node scripts/build.js",
     "style": "eslint .",
-    "test": "jest && eslint ."
+    "test": "jest && eslint .",
+    "precompat": "npm run babelify",
+    "compat": "node scripts/compat.js --prepare",
+    "compat:restore": "node scripts/compat.js --restore"
   },
   "files": [
     "dist/*.css",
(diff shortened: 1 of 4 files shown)


andresgutgon:
Awesome, thanks!

PR #2790, opened 2020-11-18
Add `justify-content: normal`

virgofx:
The `justify-content` property is missing `normal` which is a completely different from all other positioning/distributions. This PR adds it.

```
normal | <content-distribution> | <overflow-position>? [ <content-position> | left | right ]
```

## References
- [W3 Justify-Content](https://www.w3.org/TR/css-align-3/#propdef-justify-content)
- [Mozilla Developer](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)

### DIFF
--- a/src/plugins/justifyContent.js
+++ b/src/plugins/justifyContent.js
@@ -20,6 +20,9 @@ export default function () {
         '.justify-evenly': {
           'justify-content': 'space-evenly',
         },
+        '.justify-normal': {
+          'justify-content': 'normal',
+        },
       },
       variants('justifyContent')
     )


adamwathan:
Hey thanks for the PR! Can you explain the difference between normal and just using stretch which seems to be the default value?

image

virgofx:
Hey @adamwathan . It's been a month or two since I have the exact reference. I can tell you with certainty I was designing a UI block with various components and needed it to be exactly `normal` as the `stretch` didn't achieve the same behavior.  I believe I was working on a grid component. I'll try to find the reference to see if I can replicate.

adamwathan:
Going to close for now, if you run into a situation like this again happy to discuss again, thanks.

PR #2805, opened 2020-11-19
replaced the discord link on README.md by a permalink

abusarah-tech:
I have noticed that the link was leading to an invalid invite and as I am already in the server, I just wanted to contribute with this insignificant change so I created a permalink and replaced it with the dead link.


RobinMalfait:
The first link shows 
image

The second links shows
image


Both work, so not going to replace this.

Issue #3360, opened 2021-01-15
`@apply` does not work with multiple css definitions of the same selector


RobinMalfait:
### Describe the problem:

When you use `@apply`, and the rules contain multiple definitions for the same classes/key. For example:

```css
.aspect-w-1 {
  position: relative;
}

.aspect-w-1 {
  --tw-aspect-w: 1;
}
```

Then only the contents of the last one is applied. I am assuming that this is because our lookup table has the key of `aspect-w-1` and it just overrides it with the new data.

This behaviour results in issues where you can't `@apply` aspect ratio related classes: https://github.com/tailwindlabs/tailwindcss-aspect-ratio/issues/2

### Link to a minimal reproduction:

I'll open a PR with a broken test soon, and will update this PR!

RobinMalfait:
When I manually add the aspect-ratio plugin styles in a test, I do get the expected behavior:

<details>

<summary>Input:</summary>

```css
    .aspect-w-16 {
      position: relative;
      padding-bottom: calc(var(--tw-aspect-h) / var(--tw-aspect-w) * 100%);
    }

    .aspect-w-16 > * {
      position: absolute;
      height: 100%;
      width: 100%;
      top: 0;
      right: 0;
      bottom: 0;
      left: 0;
    }

    .aspect-w-16 {
      --tw-aspect-w: 16;
    }

    .aspect-h-9 {
      --tw-aspect-h: 9;
    }

    .sixteen-by-nine {
      @apply aspect-w-16 aspect-h-9;
    }
```

</details>

<details>

<summary>Output:</summary>

```css
    .aspect-w-16 {
      position: relative;
      padding-bottom: calc(var(--tw-aspect-h) / var(--tw-aspect-w) * 100%);
    }

    .aspect-w-16 > * {
      position: absolute;
      height: 100%;
      width: 100%;
      top: 0;
      right: 0;
      bottom: 0;
      left: 0;
    }

    .aspect-w-16 {
      --tw-aspect-w: 16;
    }

    .aspect-h-9 {
      --tw-aspect-h: 9;
    }

    .sixteen-by-nine {
      position: relative;
      padding-bottom: calc(var(--tw-aspect-h) / var(--tw-aspect-w) * 100%);
    }

    .sixteen-by-nine > * {
      position: absolute;
      height: 100%;
      width: 100%;
      top: 0;
      right: 0;
      bottom: 0;
      left: 0;
    }

    .sixteen-by-nine {
[...]

PR #3362, opened 2021-01-15
Add tests to verify the `@apply` with `aspect-ratio` classes

RobinMalfait:
Could help with #3360


RobinMalfait:
Closing because we already have tests for applying multiple definitions of the same class.

PR #3563, opened 2021-02-10
Added support for display : inline-table

erikverbeek:
I was missing a class for the `display : inline-table` property. I've added it underneath the other display classes and updated the tests to get back to green.

If you accept this request, hopefully it can also be added to V1 as the project I was missing it in still has to support IE11.

### DIFF
--- a/__tests__/fixtures/tailwind-output-flagged.css
+++ b/__tests__/fixtures/tailwind-output-flagged.css
@@ -10804,6 +10804,10 @@ video {
   display: table;
 }
 
+.inline-table {
+  display: inline-table;
+}
+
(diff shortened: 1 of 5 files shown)


adamwathan:
Hey thanks happy to merge but can you insert it between `.table` and `.table-caption` instead? We have `.hidden` at the very end on purpose because people tend to toggle that with JS and we'd get lots of issues being opened when people expected it to override other display classes.

image

erikverbeek:
I just updated my request with your suggestion. Thanks for the merge and the great framework.

adamwathan:
Thanks!

PR #3925, opened 2021-04-04
Add integration testing tooling setup for JIT

thecrypticace:
Tools:
- [ ] Hugo
- [x] Laravel Mix v6
- [ ] Next
- [ ] Nuxt
- [x] Parcel 2
- [x] PostCSS CLI
- [ ] Snowpack
- [x] Vite
- [x] webpack 5 + postcss
- [ ] webpack 5 + sass/less/etc…
- [ ] Webpacker

Actions:
- [x] Initial Build
- [x] Add utility to html files
- [ ] Add utility to vue files
- [ ] Add utility to react/jsx files
- [x] Add @​apply
- [x] Update @​apply
- [x] Change config

There are a few things I have to figure out:
- [x] Proper process cleanup — if a process is spawned that is not an apparent child process Jest does not exit but killing the process that spawned it doesn't clean up either. This affects the Laravel Mix which is otherwise also "working".
- [ ] Handling known / expected tooling failures

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,8 @@
 /docs
 /__tests__/fixtures/cli-utils.js
 /stubs/*
+
+# Generated tooling jit files during testing
+/jit/tooling/*/src/*
+/jit/tooling/*/dist/*
+/jit/tooling/*/tailwind.config.js
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,63 @@
+# This workflow runs integration tests across some known tools
+
+name: Integration Tests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  integration:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [15.x]
+        tool: [laravel-mix-v6, parcel-v2, postcss-cli-v8, vite-v2, webpack-v5]
+
+    steps:
+      - uses: actions/checkout@v2.3.3
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1.4.4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Use cached node_modules (top-level)
+        id: cache
+        uses: actions/cache@v2.1.1
+        with:
+          path: node_modules
+          key: nodeModules-tailwind-${{ hashFiles('**/package-lock.json') }}-${{ matrix.node-version }}
+          restore-keys: |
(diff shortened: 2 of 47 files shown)


thecrypticace:
I need to adjust the actions scripts to install npm dependencies for the tooling directories.

@RobinMalfait Do you know of / can think of a good way to handle these?

RobinMalfait:
I think there are a few things you can do. We could create individual workflows for each tool. We could also setup steps for each tool.

If you look at the current GitHub actions, then you will see this:

```yml
- name: Use cached node_modules
  id: cache  # This is the ID used below
  uses: actions/cache@v2.1.1
  with:
    path: node_modules
    key: nodeModules-${{ hashFiles('**/package-lock.json') }}-${{ matrix.node-version }}
    restore-keys: |
      nodeModules-
- name: Install dependencies
  if: steps.cache.outputs.cache-hit != 'true'  # This is referring to steps.ID.cache.outputs.cache-hit
  run: npm install
  env:
    CI: true
```

So we could create  a step for each tool, one way to do it is to define a matrix for each tool. We already do that for the node-version:

```yml
strategy:
  matrix:
    node-version: [12.x, 14.x]
```

Then we use that **node-version** inside the steps below. For example we can define a matrix for each node-version (currently there) and for each tool. Example: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-running-with-more-than-one-operating-system

Note, this could result in **a lot** of combinations 😅 

If we go with this approach, then we should also be able to run them in parallel (I think that's the default behaviour). 
image

I think the only thing left to do is to also encode the `tool` in the `key` field when caching:

```yml
- name: Use cached node_modules
  id: cache-${{ matrix.tool }}  # This is the ID used below
  uses: actions/cache@v2.1.1
  with:
    path: node_modules
    key: nodeModules-${{ hashFiles('**/package-lock.json') }}-${{ matrix.node-version }}-${{ matrix.tool }}
    restore-keys: |
      nodeModules-
- name: Install dependencies
  if: steps.cache-${{ matrix-tool }}.outputs.cache-hit != 'true'  # This is referring to steps.ID.cache.outputs.cache-hit
  run: npm install
  env:
    CI: true
```

I haven't tested it, and we might need to change this approach if we have a lot of tools to test, but I think this can work. What do you think?

thecrypticace:
Yeah a separate GitHub action would work. We wouldn't necessarily need to to a matrix build (but could). It could end up being really easy to inch close to the 2,000 minute / month limit running all these integration tests and I want to be mindful of that.

thecrypticace:
Seems like CI is having problems with the main tests. They're taking a really really long time to even start up. Perhaps related to test coverage checking?

PR #4005, opened 2021-04-07
Update calc operator regex to one that works in Safari

bradlc:
https://github.com/tailwindlabs/play.tailwindcss.com/issues/51

Lookbehinds are not supported in Safari, so this PR updates the way that `calc` operators are transformed.


tymsai:
need more light

tymsai:
there are mistakes in the written code

PR #4088, opened 2021-04-15
Implement createSimpleStaticUtilityPlugin for gridTemplateColumns

nathanielop:
Enable JIT styling for grid-cols-X.

Figure there's likely a reason certain things were excluded from the initial JIT release but if the difference between JIT classes and non-JIT classes is really just their implementation of the JIT functions or not then I don't see it being much of a stretch for me to expand off of it and convert some of the other utilities over.

Hence why only this one file is included in this PR, it's the one I need to kill off styled components from my codebase, and because it's a small enough diff that if there's a legitimate reason as to why every component is not being translated into JIT, I wouldn't feel any significant loss from seeing this PR declined.

### DIFF
--- a/jit/corePlugins/gridTemplateColumns.js
+++ b/jit/corePlugins/gridTemplateColumns.js
@@ -1,15 +1,17 @@
-const { asList, nameClass } = require('../pluginUtils')
+const { createSimpleStaticUtilityPlugin } = require('../pluginUtils')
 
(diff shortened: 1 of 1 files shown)


nathanielop:
This is pointless due to the issue mentioned in https://github.com/tailwindlabs/tailwindcss/discussions/4056

PR #4530, opened 2021-06-01
Register PurgeCSS content array as PostCSS dependencies

bradlc:
This pull requests ensures that files and directories in the PurgeCSS content array are registered as PostCSS dependencies, so that runner caches are correctly invalidated when these change.

### DIFF
--- a/src/lib/purgeUnusedStyles.js
+++ b/src/lib/purgeUnusedStyles.js
@@ -3,6 +3,8 @@ import postcss from 'postcss'
 import purgecss from '@fullhuman/postcss-purgecss'
 import log from '../util/log'
 import htmlTags from 'html-tags'
+import parseGlob from 'parse-glob'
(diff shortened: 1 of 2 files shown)


foxted:
Hi there! 

For some obscure reason, when using tailwindcss with Nuxt.js, affects Nuxt.js itself. 

With `tailwindcss@2.1.2`, Nuxt.js was still behaving properly, watching only the files needed for the project (vue files, javascript and css). 

After upgrading to `tailwindcss@2.1.3`, now Nuxt.js starts watching every single file in the project (including node modules, markdown files, IDE specific files (.idea/.vscode) etc..).

Not really sure how yet, I'll keep digging though.

AlexanderBelokon:
I've created a minimal reproduction repo for this issue: https://github.com/AlexanderBelokon/nuxt-tailwind-watch

PR #4580, opened 2021-06-08
implement purge safelist

RobinMalfait:
- fix --help output in tests
- add tests to ensure we can use `purge.safelist`
- implement the `purge.safelist` for strings

### DIFF
--- a/integrations/tailwindcss-cli/tests/cli.test.js
+++ b/integrations/tailwindcss-cli/tests/cli.test.js
@@ -172,7 +172,7 @@ describe('Build command', () => {
          -o, --output             Output file
          -w, --watch              Watch for changes and rebuild as needed
              --jit                Build using JIT mode
-             --files              Template files to scan for class names
+             --purge              Content paths to use for removing unused classes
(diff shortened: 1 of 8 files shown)


chris-morgan:
I just happened to read the 2.2.0 release notes, and just want to say: `safelist` is a rather bad name here, honestly worse to my Australian eyes than `whitelist`, which would still be a poor name for the same reason: it doesn’t convey the meaning well. I recommend deprecating this newly-released name in favour of something semantically superior.

Some relevant words: “retain”, “keep”, “preserve”; as an outsider, I think my favourites from this couple of minutes of thought are `keepClasses` and `retain`.

(Feel free to do something or ignore me as you desire; I don’t intend to engage any further; I don’t even use Tailwind at this time.)

xiCO2k:
@RobinMalfait Hey Robin, checking out the safelist stuff, there is any particular reason to not support an array with regex's?

image

Checked the documentation on the PurgeCSS docs (https://purgecss.com/safelisting.html#patterns) and does looks like is supported.

Thanks

PR #4596, opened 2021-06-09
Move context dependency modified tracking after class candidate resolution

bradlc:
This PR moves the `trackModified(contextDependencies)` call after class candidates have been resolved.

`contextDependencies` contains the source file, and the source file may be included in the list of purge content files. A concrete example of this is when using a `<style>` block in [Astro](https://github.com/snowpackjs/astro). In this case `result.opts.from` is the template file that the `<style>` appears in, and this template file is likely included in your purge content.

The problem is that the source file's modified time is updated when creating the context, so when we resolve the class candidates it thinks this file has not changed. This PR resolves that by updating the modified time of context dependencies (including the source file) at a later point.


bradlc:
This breaks the integration tests actually, but I think this is along the right lines? 🤔

bradlc:
> This breaks the integration tests actually

Fixed! There was a missing import 🤦‍♂️

PR #4600, opened 2021-06-09
run integration tests on CI

RobinMalfait:


RobinMalfait:
Will investigate this further in the future.

PR #4601, opened 2021-06-09
Add `target:` variant

seanpdoyle:
The `:target` pseudo-selector can be used to style an element whose
`[id]` attribute matches the current URL's [hash][] fragment.

For example:

```html
<div id="section-1" class="border-0 target:border-1"><!-- ... --></div>

<div id="section-2" class="border-0 target:border-1"><!-- ... --></div>
```

When the URL's path is `/#section-1`, the `#section-1` element's
`target:` variant will be active (i.e. `border-1`), and the `#section-2`
element's will be inactive (i.e. `border-0`).

[:target]: https://developer.mozilla.org/en-US/docs/Web/CSS/:target
[hash]: https://developer.mozilla.org/en-US/docs/Web/API/URL/hash


seanpdoyle:
Whoops! Closing in favor of https://github.com/tailwindlabs/tailwindcss/pull/4482 and https://github.com/tailwindlabs/tailwindcss/pull/4409.

PR #4627, opened 2021-06-11
Add `[open]` variant

seanpdoyle:
The `[open]` pseudo-selector can be used to style an element that
declares the `[open]` attribute (like `<details>` or `<dialog>`).

For example:

```html
<details class="border-0 open:border-1" open>
  <!-- ... -->
</details>

<details class="border-0 open:border-1">
  <!-- ... -->
</details>
```

The first `<details>` element is expanded, and so the `open:` variant
will be active (i.e. `border-1`), and the second element's will be
inactive (i.e. `border-0`).

### DIFF
--- a/src/jit/corePlugins.js
+++ b/src/jit/corePlugins.js
@@ -211,6 +211,17 @@ export default {
       )
     }
   },
+  booleanAttributeVariants: function ({ addVariant }) {
+    let booleanAttributeVariants = ['open']
+
+    for (let variant of booleanAttributeVariants) {
+      addVariant(variant, ({ config, modifySelectors }) => {
+        modifySelectors(({ className }) => {
+          return `.${e(`${variant}${config('separator')}${className}`)}[${variant}]`
+        })
+      })
+    }
+  },
   directionVariants: function ({ config, addVariant }) {
     addVariant(
       'ltr',
--- a/src/jit/lib/setupContextUtils.js
+++ b/src/jit/lib/setupContextUtils.js
@@ -404,7 +404,11 @@ function resolvePlugins(context, tailwindDirectives, root) {
 
   // TODO: This is a workaround for backwards compatibility, since custom variants
   // were historically sorted before screen/stackable variants.
-  let beforeVariants = [corePlugins['pseudoElementVariants'], corePlugins['pseudoClassVariants']]
+  let beforeVariants = [
+    corePlugins['pseudoElementVariants'],
+    corePlugins['pseudoClassVariants'],
+    corePlugins['booleanAttributeVariants'],
+  ]
   let afterVariants = [
     corePlugins['directionVariants'],
     corePlugins['reducedMotionVariants'],
--- a/tests/variantsAtRule.test.js
+++ b/tests/variantsAtRule.test.js
@@ -590,6 +590,27 @@ test('it can generate read-only variants', () => {
   })
 })
 
+test('it can generate open variants', () => {
+  const input = `
+    @variants open {
+      .banana { color: yellow; }
+      .chocolate { color: brown; }
+    }
+  `
+
+  const output = `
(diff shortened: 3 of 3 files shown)


src/jit/corePlugins.js
```diff
@@ -211,6 +211,17 @@ export default {
       )
     }
   },
+  booleanAttributeVariants: function ({ addVariant }) {
+    let booleanAttributeVariants = ['open']
+
+    for (let variant of booleanAttributeVariants) {
+      addVariant(variant, ({ config, modifySelectors }) => {
+        modifySelectors(({ className }) => {
+          return `.${e(`${variant}${config('separator')}${className}`)}[${variant}]`
```
seanpdoyle, line 220: @adamwathan in addition to styling the elements that declare `[open]` themselves (e.g. `<details>`), it'd be _really_ valuable to also integrate the `open:` variant with the `.group` and `.peer` utilities.

For example:

```html
<details class="group peer" open>
  <summary>
    <span class="hidden group-open:inline">Close me, I'm open</span>
    <span class="inline group-open:hidden">Open me, I'm closed</span>
  </summary>
</details>

<span class="hidden peer-open:inline">I'm next to an open element</span>
<span class="inline peer-open:hidden">I'm next to a closed element</span>
```

Since the `booleanAttributeVariants` key is new, I wasn't sure how to best integrate with the existing `group` and `peer` support from the `pseudoElementVariants` without copy-pasting. Is there an existing utility we could re-use, or would this require extracting a new one?

MichaelAllenWarner: Native `[open]` support would be great.

Not sure if this will be helpful, but I use a single custom variant to target both an open `details` and any of its descendants:

```javascript
addVariant('open', ({ modifySelectors, separator }) => {
  return modifySelectors(({ selector }) => {
    return selectorParser(selectors => {
      const clonedSelectors = selectors.clone();
      selectors.walkClasses(classNode => {
        classNode.value = `open${separator}${classNode.value}`;
        classNode.parent.insertBefore(classNode, selectorParser().astSync('[open]'));
      });
      clonedSelectors.walkClasses(classNode => {
        classNode.value = `open${separator}${classNode.value}`;
        classNode.parent.insertBefore(classNode, selectorParser().astSync('[open] '));
      });
      selectors.append(clonedSelectors);
    }).processSync(selector);
  });
});
```

In my experience it works well, including in combination with other variants (I'm using JIT), though I guess having a single variant for both purposes could be a problem if you have nested `details` elements.

MichaelAllenWarner:
Native `[open]` support would be great.

Not sure if this will be helpful, but I use a single custom variant to target both an open `details` and any of its descendants:

```javascript
addVariant('open', ({ modifySelectors, separator }) => {
  return modifySelectors(({ selector }) => {
    return selectorParser(selectors => {
      const clonedSelectors = selectors.clone();
      selectors.walkClasses(classNode => {
        classNode.value = `open${separator}${classNode.value}`;
        classNode.parent.insertBefore(classNode, selectorParser().astSync('[open]'));
      });
      clonedSelectors.walkClasses(classNode => {
        classNode.value = `open${separator}${classNode.value}`;
        classNode.parent.insertBefore(classNode, selectorParser().astSync('[open] '));
      });
      selectors.append(clonedSelectors);
    }).processSync(selector);
  });
});
```

In my experience it works well, including in combination with other variants (I'm using JIT), though I guess having a single variant for both purposes could be a problem if you have nested `details` elements.

reinink:
@seanpdoyle Thanks for this contribution! A lot has changed to the Tailwind CSS codebase since you originally submitted this, so instead of refactoring this PR, I just created a new one (#5627) and marked you as a co-author. I also got the `group` and `peer` versions working. 👍 

Thanks again! 🙌

PR #4644, opened 2021-06-14
Improve production build performance for the case of many small non-tailwind stylesheets

SevInf:
Hi there!
We decided to use tailwind as a base of our design system. Our codebase is legacy-ish Vue 2 SPA with pretty liberal use of `<style>` blocks in our components, built with webpack (this info will be important later). Tailwind worked great and was delightful to work with. That is, until we got to production build part. 

Turns out, just adding tailwind to our postcss plugins list added a couple of minutes to our build times. Actual usage did not seem to matter: commenting out every `@apply` and `@tailwind` rule did not seem to change anything: build with tailwind plugin included still was couple of minutes slower than without it. Further investigation pointed out at `purge` option as the culprit. Removing it seemed to bring out build times in line with what we expected, but we'd very much like to keep it. Of course, purging unused styles takes time, but minutes still seemed excessive, especially considering that `tailwind-cli` with the same config successfully generates the stylesheet in a couple of seconds.

As it turns out, tailwind plugin runs purgecss on every source stylesheet. Which, in case of legacy-ish code base with a lot of tiny stylesheets, is quite a lot. Sure, if said file does not have `@tailwind` directive, it will get wrapped in `purgecss ignore` block, but extraction of used rules from html and js will still be done from scratch for every one of those tiny stylesheets.

This PR changes the plugin to completely skip `purgecss` in `layers` mode if the source file did not include any tailwind layers. For that, I had to switch from `postcss-purgecss` plugin to calling `purgecss` manually — it does not seem like there is a way to decorate `postcss-purgecss` plugin to call it conditionally for both PostCSS 7 and 8. I guess, the upside of this is that latest purgecss could also be used in compat build. 

Some benchmarks of this PR on our codebase:
Before: `pnpm build  421.86s user 29.51s system 169% cpu 4:25.98 total`
After: `pnpm build  285.95s user 16.38s system 214% cpu 2:21.21 total`

As you can see, it saves 2 minutes for us. It might be also very useful for other people in similar situations: integrating tailwind into existing codebase with modularized css.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "2.1.2",
       "license": "MIT",
       "dependencies": {
-        "@fullhuman/postcss-purgecss": "^4.0.3",
         "arg": "^5.0.0",
         "bytes": "^3.0.0",
         "chalk": "^4.1.1",
@@ -36,6 +35,7 @@
         "postcss-selector-parser": "^6.0.6",
         "postcss-value-parser": "^4.1.0",
         "pretty-hrtime": "^1.0.3",
+        "purgecss": "^4.0.3",
         "quick-lru": "^5.1.1",
         "reduce-css-calc": "^2.1.8",
         "resolve": "^1.20.0"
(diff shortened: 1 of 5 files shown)


SevInf:
@adamwathan is there anything I can do to get this PR merged? This is quite an important and important issue for us and I'd be happy to do any changes you deem necessary.

RockSharabShrestha:
Unsubscribe me from this I am not understanding anything.

adamwathan:
Just haven't had time to review, have higher priorities unfortunately. Will try to look soon, have to be really careful with changes like this though so it takes a bit of a commitment to get through it. Hopefully soon, can you use your own fork though in the mean time?

bradlc:
Hey @SevInf. Comparing this to [PurgeCSS itself](https://github.com/FullHuman/purgecss/blob/bf89ff7179de4782d62d77b0b54b068bcff976cb/packages/purgecss/src/index.ts#L636-L638), are we missing these few lines?

```js
if (this.options.variables) {
  this.variablesStructure.safelist = safelist.variables || [];
}
```

SevInf:
@bradlc this code is taken from [`purgecss-postcss`](https://github.com/FullHuman/purgecss/blob/master/packages/postcss-purgecss/src/index.ts#L13) plugin and there those lines are also missing. That, however, does not look intentional to me, so I'll add them in a moment.

SevInf:
Hey @bradlic, so I am back with the fix. You were right, those lines were missing and were breaking `safelist.variables` functionality. This is also something that was already broken when using `purgecss-postcss`. Along with the fix I've also added 2 tests: one for purging the variables and another for respecting safelist for variables. You can see that the latter one also fails on master branch.

adamwathan:
@SevInf thanks man will put this out in the next patch!

PR #4667, opened 2021-06-16
Use `purge.options.safelist` as a fallback in JIT mode

RobinMalfait:
Currently, for AOT mode you can use the following:

- purge.safelist
- purge.options.safelist

But for JIT mode, you can only use:

- purge.safelist

This commit will ensure that you can use these in both cases:

- purge.safelist
- purge.options.safelist

This should make transitioning to JIT a little bit smoother.


danstans:
Hi Robin, will this enable Regex patterns in purge.safelist? Unable to find it or get it working just yet.

PR #4692, opened 2021-06-18
Add support for PurgeCSS contentFunction

TheDoctor0:
PurgeCSS plugin allows to specify **contentFunction** in [options](https://purgecss.com/plugins/postcss.html#options) as an alternative for default **content** option.
It can be used to return different content paths for every processed CSS file.
This is a really useful feature for projects with a more modular approach to further reduce the size of the production build.

This PR aims to close the gap between AOT and JIT modes by adding **contentFunction** support to the latter.

### DIFF
--- a/src/jit/lib/setupTrackingContext.js
+++ b/src/jit/lib/setupTrackingContext.js
@@ -7,6 +7,7 @@ import normalizePath from 'normalize-path'
 
 import hash from '../../util/hashConfig'
 import getModuleDependencies from '../../lib/getModuleDependencies'
+import getPurgeContent from '../../lib/getPurgeContent'
 
 import resolveConfig from '../../../resolveConfig'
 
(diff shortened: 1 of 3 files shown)


reinink:
Hey @TheDoctor0, thanks for this contribution. Can you share a specific example of how you are using this in your projects? I know with PurgeCSS this can be good for performance, however I'm not sure that's going to matter all that much with JIT because of all the caching we do. Is there a good use-case for this outside of the performance benefits?

TheDoctor0:
Hi @reinink. The contentFunction available in the PurgeCSS plugin can be used to better optimize unused CSS in production build when using multiple CSS files because you have control over what files should be considered as a source of CSS class references for every one of those files.

That said, the current version of JIT does a pretty good job by itself and we actually migrated away from using the contentFunction to one global CSS file generated with a content array declaration. We could achieve ~30% smaller production build but at a cost of maintaining a sizable list of files with CSS class references for every separate CSS file that we used and it was just not worth it.

PR #4936, opened 2021-07-09
Add variant for `::file-selector-button` pseudo element

lukewarlow:
This PR adds a variant for the [`::file-selector-button`](https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button)  pseudo element, which is supported by the latest versions of all three engines.

I think I've added to the relevant tests but feel free to point out any others that are required.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -73,13 +73,10 @@ export let pseudoElementVariants = ({ config, addVariant }) => {
   ])
 
   addVariant(
-    'file-selector-button',
+    'file',
     transformAllSelectors((selector) => {
       return updateAllClasses(selector, (className, { withPseudo }) => {
-        return withPseudo(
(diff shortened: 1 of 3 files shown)


mainej:
+1 I'd love to have this too. Would address https://github.com/tailwindlabs/tailwindcss-forms/issues/25

reinink:
Hey thanks for this PR @lukewarlow, didn't know about this feature! Let's go with just `file` since `file-selector-button` is a bear to type 😅 I've also added a test for combining this with `hover`, which works perfect in the new JIT mode 👍 Thanks again!

colbyhemond:
For those that end up here looking for the solution (like myself), here's a link to the section of the tailwind docs that mentions styling file input buttons: https://tailwindcss.com/docs/hover-focus-and-other-states#file-input-buttons

PR #5336, opened 2021-08-30
[WIP] Ensure `updateLastClasses` updates the correct class

### LINKED ISSUE #5152: [JIT] Class based variant bug when chaining

v2.2.7

postcss v8.3.6

v14.15.5

chrome

macOS

### Reproduction repository

https://github.com/minimit/twbug-class-variant-chain

### Describe your issue

When using **custom variants** with classes instead of pseudo selectors, and chaining multiple variants, the output is wrong.

This is the code of the `addVariant`.

```
    addVariant('class', ({ modifySelectors, separator }) => {
        return `.${e(`class${separator}${className}`)}.class`
      })
    })
```

This is the code of the apply or html class.

```
.test-class {
  @apply sm:class:hidden;
}

```

This is the output.

```
@media (min-width: 640px) {
    {
         display: none
    }
}
```

This is the desired output.

```
@media (min-width: 640px) {
    .test-class.class {
         display: none
    }
}
```

Seems also html class does the same.

RobinMalfait:
While looking at #5152 it looked like this was an `@apply` issue, however it turns out that it was an issue with our internal `updateLastClasses`.

If you have a custom plugin that looks like this:

```js
function ({ addVariant, e }) {
  addVariant('bar', ({ modifySelectors, separator }) => {
      return `.${e(`bar${separator}${className}`)}.bar-parent`
    })
  })
}
```

Then you will notice the `.bar-parent`:
```
    ┌─[./example-plugin.js]
    │
  1 │   function ({ addVariant, e }) {
  2 │     addVariant('bar', ({ modifySelectors, separator }) => {
  3 │       modifySelectors(({ className }) => {
> 4 │         return `.${e(`bar${separator}${className}`)}.bar-parent`
  5 │       })                                            ─────┬─────
  6 │     })                                                   ╰─────── This is the issue
  7 │   }
    │
    └─
```

The reason why this is the issue is if you have `bar:hidden`, then this will eventually resolve to:
```css
.bar\:hidden.bar-parent {
  display: none;
}
```

However, if you have `sm:bar:hidden`, then this will eventually resolve to:
```css
.bar\:hidden.sm\:bar-parent {
  display: none;
}
```
But it should resolve to this instead:
```css
.sm\:bar\:hidden.bar-parent {
  display: none;
}
```

This is because the `sm` responsive modifier will use `updateLastClasses`, internally this will search for the last class. If you have `.a .b .c` then the last class is `.c`, however, if you have `.a .b.c` then the last class is **also** `.c`, even though you expected it to be `.b` because this is the last "group" of combined classes. This PR will ensure that we get `.b` instead. So that in our example the `sm:` is applied to `.bar` and not to `.bar-parent`.

A solution I mentioned in userland, would be to add the `.bar-parent` to the front instead of the end of the return statement in `modifySelectors` but that's more of a workaround for the actual problem.

---

Fixes: #5152

### DIFF
--- a/src/util/pluginUtils.js
+++ b/src/util/pluginUtils.js
@@ -47,7 +47,16 @@ export function updateAllClasses(selectors, updateClass) {
 export function updateLastClasses(selectors, updateClass) {
   let parser = selectorParser((selectors) => {
     selectors.each((sel) => {
-      let lastClass = sel.filter(({ type }) => type === 'class').pop()
+      // Find the last class that is not part of a "combined" selector.
+      // E.g.:
+      // .a .b .c -> this should result in `.c`
+      // .a .b.c -> this should result in `.b`
+      let lastClass = sel
+        .filter(({ type }) => type === 'class')
+        .reverse()
+        .find(({ source }, idx, all) => {
+          return !all[idx + 1] || source.start.column - all[idx + 1].source.end.column > 1
+        })
 
       if (lastClass === undefined) {
         return
--- a/tests/jit/apply.test.js
+++ b/tests/jit/apply.test.js
@@ -231,3 +231,56 @@ test('@apply error with nested @anyatrulehere', async () => {
     '@apply is not supported within nested at-rules like @genie'
   )
 })
+
+test('@apply with custom variant that contain an additional `.class` at the end', async () => {
+  let config = {
+    mode: 'jit',
+    darkMode: 'class',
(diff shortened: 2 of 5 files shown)


RobinMalfait:
This specific PR introduces the opposite issue, where now adding classes at the beginning doesn't work anymore but only at the end.
Will close it for now and re-open a new one if we have a better solutions.

Already provided a workaround in the original issue that works for this scenario.

PR #5340, opened 2021-08-30
Remove AOT

RobinMalfait:
The start of something beautiful. Let's remove AOT mode!


TrySound:
Does it mean we don't need large bundles inside dist anymore?

adamwathan:
@TrySound Yeah we are removing the CDN builds from this package for v3.0, whatever we make available CDN-wise we'll put in a separate package so we stop wasting everyone's bandwidth 👀

RobinMalfait:
For people interested, this basically has a 10x reduction in size...
| Before | After |
| -- | -- |
| 34.6 MB | 3.7 MB |

image

src/cli.js
```diff
@@ -420,39 +418,21 @@ async function build() {
     let resolvedConfig = resolveConfigInternal(config)
 
     if (args['--purge']) {
-      resolvedConfig.purge = {
-        enabled: true,
-        content: args['--purge'].split(/(?<!{[^}]+),/),
+      log.warn(['The `--purge` flag has been deprecated.', 'Please use `--content` instead.'])
+      if (!args['--content']) {
+        args['--content'] = ['--purge']
```
binaryoverload, line 423: This should probably be `args['--content'] = args['--purge']`

The current version will just assign an array with `--purge` as an element

binaryoverload: Whoops, just realised I commented on a merged branch 🤦‍♂️

RobinMalfait: No worries, and thanks for notifying! I've fixed it here: https://github.com/tailwindlabs/tailwindcss/pull/5646

lightyen:
How can I get all available classname from tailwind.config after this change?

RobinMalfait:
@lightyen you cannot get *all* the available classes because there will be millions of those. We do have a way to get all the **base** classes, so that means something like `bg-red-500` without all variants but the API is a bit complex right now and you do require an internal `context` to access these. 

However, I'm more interested in your use case, can you talk more about that and maybe open an issue/discussion so that we can talk about the use case and find solutions?

lightyen:
@RobinMalfait I just need all variants and utilities(or components) for list, if there is a way, it is good!

RobinMalfait:
@lightyen can you open a discussion, then we can move the conversation to its dedicated spot. Thanks! https://github.com/tailwindlabs/tailwindcss/discussions/new

PR #5359, opened 2021-09-01
Add native aspect ratio support

reinink:
This PR adds native aspect ratio support. Once Safari 15 is popularized, this will have [full support in modern browsers](https://caniuse.com/mdn-css_properties_aspect-ratio).

By default, we include these utilities:

| Class | CSS |
| --- | --- |
| `aspect-auto` | `aspect-ratio: auto` |
| `aspect-square` | `aspect-ratio: 1 / 1` |
| `aspect-video` | `aspect-ratio: 16 / 9` |

We considered adding a bunch more by default, but in practice it's easier to just use arbitrary values instead of providing an exhaustive list of numeric values in your config file. For example:

```html
<img class="aspect-[16/9] ... "/>
```

This utility can be customized under the `aspectRatio` property in your `tailwind.config.js` file.

This addition does not conflict with the existing [aspect ratio plugin](https://github.com/tailwindlabs/tailwindcss-aspect-ratio), as it uses different utility names.


say8425:
When you plan to release this feature?

reinink:
@say8425 The plan is to release a v3.0 alpha of Tailwind CSS in the next couple of weeks. You can, however, also use an [insiders release](https://www.npmjs.com/package/tailwindcss?activeTab=versions) if you want it today.

ChazUK:
I have tailwindcss version ^3.1.6 and I'm trying to do class="aspect-video" but no styling is being applied. Is it not yet available in 3.1.6? because the docs say it is https://tailwindcss.com/docs/aspect-ratio

PR #5378, opened 2021-09-03
Support `@apply` for classes outside of a `@layer`

RobinMalfait:
This PR will allow you to apply classes that are not inside a `@layer` directive. This is an important part of making JIT the default for backwards compatibility reasons. In short:

The following already worked:
```css
@tailwind utilities;

.foo {
  @apply bg-red-500 hover:bg-red-600;
}
```

This got converted to the following output:
```css
.foo {
  --tw-bg-opacity: 1;
  background-color: rgba(239, 68, 68, var(--tw-bg-opacity));
}
.foo:hover {
   --tw-bg-opacity: 1;
  background-color: rgba(220, 38, 38, var(--tw-bg-opacity));
}
```

However, it was not possible to do this:

```css
@tailwind utilities;

.foo {
  @apply bg-red-500 hover:bg-red-600;
}

.bar {
  @apply hover:foo;
}
```

This resulted in the following output:
```
CssSyntaxError: The `hover:foo` class does not exist. If `hover:foo` is a custom class, make sure it is defined within a `@layer` directive.
```

But with this PR, it results in the following output:
```css
.foo {
  --tw-bg-opacity: 1;
  background-color: rgba(239, 68, 68, var(--tw-bg-opacity));
}
.foo:hover {
  --tw-bg-opacity: 1;
  background-color: rgba(220, 38, 38, var(--tw-bg-opacity));
}
.bar:hover {
  --tw-bg-opacity: 1;
  background-color: rgba(239, 68, 68, var(--tw-bg-opacity));
}
.bar:hover:hover {
  --tw-bg-opacity: 1;
  background-color: rgba(220, 38, 38, var(--tw-bg-opacity));
}
```

There is this weird issue that `.bar:hover:hover` exists, but I will fix that in another PR. (Note: this is an existing issue, and not a newly introduced issue)

---

Apart from the backwards compatibility aspect, the biggest difference is the behaviour whether you are using the class or not.

```css
@tailwind utilities;

.foo { /* This will _always_ exist in your output, even if you didn't use the "foo" class in your content. */
  
}

@layer components {
  .bar { /* This will _only_ exist in your output if you used the "bar" class in your content. */
  }
}
```

### DIFF
--- a/tests/apply.test.js
+++ b/tests/apply.test.js
@@ -292,29 +292,26 @@ test('@apply classes from outside a @layer', async () => {
       }
 
       .bar {
-        font-weight: 700;
         --tw-text-opacity: 1;
         color: rgba(239, 68, 68, var(--tw-text-opacity));
+        font-weight: 700;
       }
 
       .bar:hover {
         --tw-text-opacity: 1;
-        color: rgba(16, 185, 129, var(--tw-text-opacity));
-      }
-
-      .baz {
-        font-weight: 700;
-        --tw-text-opacity: 1;
-        color: rgba(239, 68, 68, var(--tw-text-opacity));
-      }
-
-      .baz:hover {
-        --tw-text-opacity: 1;
-        color: rgba(16, 185, 129, var(--tw-text-opacity));
+        color: rgba(34, 197, 94, var(--tw-text-opacity));
       }
 
       .baz {
         text-decoration: underline;
+        --tw-text-opacity: 1;
+        color: rgba(239, 68, 68, var(--tw-text-opacity));
+        font-weight: 700;
+      }
+
+      .baz:hover {
+        --tw-text-opacity: 1;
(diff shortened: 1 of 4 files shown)


tests/apply.test.js
```diff
@@ -292,29 +292,26 @@ test('@apply classes from outside a @layer', async () => {
...
-        font-weight: 700;
         --tw-text-opacity: 1;
         color: rgba(239, 68, 68, var(--tw-text-opacity));
+        font-weight: 700;
       }
 
       .bar:hover {
         --tw-text-opacity: 1;
-        color: rgba(16, 185, 129, var(--tw-text-opacity));
+        color: rgba(34, 197, 94, var(--tw-text-opacity));
```
RobinMalfait, line 302: The color value looks odd, but that's because of a rebase and the colors in #5384 has been merged since. 
The "old" value is `text-emerald-500` and the new value is `text-green-500`

PR #5387, opened 2021-09-04
Add `accent-color` utilities

lukewarlow:
This PR adds utilities for the [`accent-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) property.

Example usage:

`<input type="checkbox" class="accent-red-500">`

### DIFF
--- a/src/plugins/accentColor.js
+++ b/src/plugins/accentColor.js
@@ -5,7 +5,7 @@ export default function () {
   return function ({ matchUtilities, theme, variants }) {
     matchUtilities(
       {
-        caret: (value) => {
+        accent: (value) => {
           return { 'accent-color': toColorValue(value) }
         },
       },
--- a/src/plugins/accentColor.js
+++ b/src/plugins/accentColor.js
@@ -1,0 +1,19 @@
+import flattenColorPalette from '../util/flattenColorPalette'
+import toColorValue from '../util/toColorValue'
+
+export default function () {
+  return function ({ matchUtilities, theme, variants }) {
+    matchUtilities(
+      {
+        caret: (value) => {
+          return { 'accent-color': toColorValue(value) }
+        },
+      },
+      {
+        values: flattenColorPalette(theme('accentColor')),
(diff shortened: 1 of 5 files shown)


lukewarlow:
As this only has support in 2/3 browsers, I understand this may not be merged soon.

Here's the accent-color utilities in plugin form: https://github.com/lukewarlow/tailwind-accent-color

lukewarlow:
The test is failing locally but the css output doesn't make much sense to me, seeing as this is basically a direct copy of the caret color plugin.

image

src/plugins/accentColor.js
```diff
@@ -0,0 +1,19 @@
+import flattenColorPalette from '../util/flattenColorPalette'
+import toColorValue from '../util/toColorValue'
+
+export default function () {
+  return function ({ matchUtilities, theme, variants }) {
+    matchUtilities(
+      {
+        caret: (value) => {
```
adamwathan, line 8: This right here is why the tests are failing, you are telling Tailwind to match classes starting with the word "caret" when it should be "accent" 👍🏻

stubs/defaultConfig.stub.js
```diff
@@ -172,6 +172,7 @@ module.exports = {
       none: 'none',
     },
     caretColor: (theme) => theme('colors'),
+    accentColor: (theme) => theme('colors'),
```
adamwathan, line 175: Should probably include `auto` as a value here in addition to all of the colors:

https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color#values

src/plugins/accentColor.js
lukewarlow: 🤦  That'd certainly do it. Fixed now.

stubs/defaultConfig.stub.js
lukewarlow: Added

adamwathan:
Thanks @lukewarlow!

PR #5390, opened 2021-09-04
Remove lodash

RobinMalfait:
This PR will get rid of the `lodash` dependency since it is rather big. I replaced most of the functions with plain built-ins. I've added a few extra utilities to mimic lodash' behaviour.
Some files (like the `resolveConfig`) can use some refactoring, however I tried to mainly focus on `lodash`'s removal.


TrySound (approved):
❤️

PR #5635, opened 2021-09-29
Enable optimize universal defaults by default

RobinMalfait:


src/featureFlags.js
```diff
@@ -1,18 +1,24 @@
 import chalk from 'chalk'
 import log from './util/log'
 
-const featureFlags = {
+let defaults = {
+  optimizeUniversalDefaults: true,
```
RobinMalfait, line 5: The default for all `feature` and `experimental` flags are still `false` if you did not configure it yourself in your config.
However, with this new `defaults` object, we can change the default value. In this case the `optimizeUniversalDefaults` will be `true` even if you didn't configure it yourself in your own `tailwind.config.js`.

The nice part about this is that everybody gets the new `optimizeUniversalDefaults` by default, but the important part is that they can still turn it off in case something does break.

PR #5638, opened 2021-09-29
Add fit-content utilities for (min/max) height and width

lukewarlow:
This PR adds [fit-content](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content) classes for (min/max) height and width.

I noticed only width seemed to have min or max content utilities currently. I'm not sure why height is missing them but I left it that way to keep this PR focussed. Happy to make a follow up if they are wanted too.

Couldn't find relevant tests for the existing utils for width and height so haven't added any for this. Feel free to point me in the right direction and I can add them.

### DIFF
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -478,6 +478,7 @@ module.exports = {
       '5/6': '83.333333%',
       full: '100%',
       screen: '100vh',
+      fit: 'fit-content',
     }),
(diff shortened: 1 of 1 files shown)


RobinMalfait:
@samheyman yes it's there by default. It won't be there if you overrode the values in your tailwind config file so make sure to double check that. Checkout the generated tab at the bottom: https://play.tailwindcss.com/rzPIBlnfH2

PR #5639, opened 2021-09-29
Add `border-x` and `border-y` width and color utilities

reinink:
This PR adds new `border-x-{width}`, `border-y-{width}`, `border-x-{color}`, and `border-y-{color}` utilities to style the border on two sides of an element at the same time.

Added @DavydeVries as a co-author for his original work on #1159 which we closed at the time due to file size considerations 🙏


DavydeVries:
Cool, thank you! 🥰

PR #5653, opened 2021-09-30
Reset button's border-radius to 0 for Edge

targos:
I tested with the following playground: https://play.tailwindcss.com/ZcBGbScwwM

On Mac:
- Edge (version 94) has rounded corners by default.
image
 
- Safari, Chrome and Firefox do not.

On Windows:
- Edge has rounded corners
- Chrome and Firefox do not.

I have not been able to find when Chromium-based browsers changed the default. I only found an issue in the chromium tracker about the devtools that don't show it: https://bugs.chromium.org/p/chromium/issues/detail?id=1146551
Refs: https://github.com/tailwindlabs/tailwindcss/commit/331454f62dd1bdaf57dae266097255fc8f8c70b1

### DIFF
--- a/src/css/preflight.css
+++ b/src/css/preflight.css
@@ -162,6 +162,14 @@ textarea {
 }
 
 /*
+Remove the default border radius in Edge.
+*/
+
+button {
+  border-radius: 0;
+}
+
+/*
 Remove the inheritance of text transform in Edge and Firefox.
 */
 


reinink:
Hey thanks for reporting this @targos! I am able to reproduce this in Edge on MacOS, but not on Chrome (94). Going to do a little more digging into the implications of this change before doing anything, but I appreciate you making us aware of it.

targos:
I'm not on my computer anymore but I think for Chrome I tested with canary

targos:
I was wrong. Today I can only reproduce with Edge.

Edge 94:
image

Chrome 94 and 96:
image

adamwathan:
Heard from someone on the Edge team and this will be fixed in the next Edge canary release 👍🏻 

This is some code they have in their UA stylesheet that they forgot to follow up on:

image

What should we do, should we update this reset to target all of the elements in the screenshot or just wait and let this sort of resolve itself over the next couple months?

reinink:
> What should we do, should we update this reset to target all of the elements in the screenshot or just wait and let this sort of resolve itself over the next couple months?

Even if we did this, it's not going to help anyone who doesn't update their current Tailwind version, and it would also mean doing another 2.2.x release, which I guess is totally fine, I just wonder how much it actually matters, if a fix is already in the works.

On the flip side, if there are no obvious downsides to this, and it's just a simple addition to Preflight, there isn't really any harm, besides having to slice a new release (or two). We could even leave a comment referencing the issue, and saying something like "this can be removed by January 1, 2022", or something like that just to remind us to remove it when it's no longer needed.

PR #5662, opened 2021-10-01
Rename `completions` and remove color information

bradlc:
- Rename `context.completions` to `context.getClassList` – I think this better (1) indicates that it's a function, and (2) describes what the function returns
- Remove colour information from the returned list. While the colour information is useful to an extent, it is complicated to use this information to resolve the colour for classes like `hover:bg-red-500` or `focus:bg-red-500/30`. I have found it simpler to resolve the CSS for a class (using `generateRules`) and extract the colour from there.


RobinMalfait (approved):
Removing unused code is always fun! Looking good!

PR #5762, opened 2021-10-12
Text decoration color #2050

dance2die:
refer to #2050 

This PR adds utilities for [text-decoration-color](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color)

```
decoration-{color}
```

where {color} is all theme colors.

image


dance2die:
> I think this is doing the same as #5760 is trying to.

Ty @lukewarlow I will discard the Draft :)

PR #5775, opened 2021-10-13
Fix CLI `--content` option

bradlc:
This PR fixes the CLI `--content` option. Following resolution the content config looks like this, so it's actually `content.files` that needs updating:

```js
{
  files: [],
  extract: {
    DEFAULT: undefined
  },
  transform: {}
}
```

~~Side note: is `extract.DEFAULT` supposed to be `undefined` here? 🤔~~ See #5778


RobinMalfait (approved):
Ah that's a good catch! Also thanks for adding a test!

PR #5778, opened 2021-10-13
Neaten `extract.DEFAULT` resolution

bradlc:
This PR updates the `extract` resolution so that by default, instead of `{ DEFAULT: undefined }` you get `{}`


RobinMalfait (approved):
Ah yep, I honestly didn't bother with it because I know we had the checks further along, but it is a bit cleaner indeed!

PR #5853, opened 2021-10-22
Allow `_` inside `url()` when using arbitrary values

RobinMalfait:
This PR allows you to have `_` in `url()` while using arbitrary values.
Normally we convert `_` to spaces ` `. If you want to opt-out of this behaviour, you have to escape the `\_` to keep the `_`. 

It does  feel a bit awkward to do that for your URLs as well, so this PR allows you to keep `_` in URLs without escaping.


jean343:
What if we want a space in the URL in the `background-image`?
For example:
```
background-image: url("data:image/svg+xml,<svg width='100%' height='100%' xmlns='http://www.w3.org/2000/svg'><rect width='100%' height='100%' fill='none' stroke='%234B7AE7' stroke-dasharray='2,2'/></svg>");
```

PR #5970, opened 2021-11-03
Add `text-decoration-style` utilities

olivermrose:
Adds [`text-decoration-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-style) utilities.

I've left the name as `decoration-*` but I'm thinking it should be changed to disambiguate from the existing [box decoration break](https://tailwindcss.com/docs/box-decoration-break) utilities. I had `decoration-style-*` in mind but I'm open to suggestions.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -1,16 +1,16 @@
 import fs from 'fs'
 import * as path from 'path'
 import postcss from 'postcss'
-import createUtilityPlugin from './util/createUtilityPlugin'
-import buildMediaQuery from './util/buildMediaQuery'
-import parseAnimationValue from './util/parseAnimationValue'
-import flattenColorPalette from './util/flattenColorPalette'
-import withAlphaVariable, { withAlphaValue } from './util/withAlphaVariable'
-import toColorValue from './util/toColorValue'
-import isPlainObject from './util/isPlainObject'
-import transformThemeValue from './util/transformThemeValue'
 import { version as tailwindVersion } from '../package.json'
+import buildMediaQuery from './util/buildMediaQuery'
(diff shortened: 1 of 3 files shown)


neupauer:
Hey, I like that one.

How about adding the whole `text-decoration` module?

| CSS Property                                                                                              | Example                                         | Notes          |
| --------------------------------------------------------------------------------------------------------- | ----------------------------------------------- | -------------- |
| [`text-decoration-line`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-line)           | `.decoration-underline`, `.decoration-overline` | can be chained |
| [`text-decoration-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color)         | `.decoration-red-500`                           |                |
| [`text-decoration-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-style)         | `.decoration-dashed`                            |                |
| [`text-decoration-thickness`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-thickness) | `.decoration-2`                                 |                |
| [`text-underline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset)         | `.decoration-offset-2`                          |                |

I like the `decoration` prefix, so maybe in v 3.0 the `box-decoration-break` could be renamed to `box-decoration`? It doesn't have to be breaking-change because `.decoration-slice` and `.decoration-clone` doesn't collide with anything (can be marked as an deprecated alias).

I don't have time to prepare a PR right now, so I'm just giving my opinion and maybe inspiration here.

xiBread:
Continued in #6004.

PR #6003, opened 2021-11-07
Add `text-emphasis-*` utilities

olivermrose:
Adds utilities for:

- [`text-emphasis-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style)
- [`text-emphasis-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-color)
- [`text-emphasis-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position)

The property for `text-emphasis-style` is `text-emphasis` to be consistent with [`text-decoration`](https://tailwindcss.com/docs/text-decoration).

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -1646,6 +1646,38 @@ export let corePlugins = {
     )
   },
 
+  textEmphasis: ({ addUtilities }) => {
+    addUtilities({
+      '.emphasis-filled': { 'text-emphasis': 'filled' },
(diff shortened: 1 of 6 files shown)


lukewarlow:
I personally think it should use the long hand properties for all of them. Otherwise there's unintuitive override behaviour when using multiple properties together. Which someone's trying to fix for `text-decoration` here https://github.com/tailwindlabs/tailwindcss/pull/6378

PR #6006, opened 2021-11-08
Prevent failing at virtual files

sibbng:
Vite can serve virtual css files and those files are processed through postcss. If those files includes some "at rules" that Tailwind can parse Tailwind will throw error while trying to find them on filesystem.


thecrypticace:
@sibbng Hey, thanks for the PR! That sounds interesting but I'm not 100% sure I understand. Is there a reproduction you can provide that shows that this would be needed?

sibbng:
> 

Hi, it was a very experimental thing that I can't put into minimal reproduction. I looked back to it and found a better way. Thanks for your interest.

PR #6011, opened 2021-11-08
Improve nesting detection

RobinMalfait:
This PR improves the nesting detection.

Currently we only detected rules inside of rules, and if we did then we would trigger a PostCSS warning.

However, some people try to do something like this:
```css
.namespace {
  @tailwind base;
  @tailwind components;
  @tailwind utilities;
}
```

While in theory this could work~ish, this is not something we support. The biggest reason is that `base` styles (and thus also the preflight/reset) would end up looking like this: `.namespace html {}` for example which doesn't make much sense.

This PR also detects `@tailwind` at rules, with a parent that is not of type `root` and shows a similar message:
> Nested @tailwind rules detected, instead try to use a prefix: https://tailwindcss.com/docs/configuration#prefix or scope it using the important selector strategy https://tailwindcss.com/docs/configuration#selector-strategy

(@reinink can you double check the wording used in the message? Or do we keep it as is and improve all warning messages before v3 launch?)

### DIFF
--- a/src/lib/detectNesting.js
+++ b/src/lib/detectNesting.js
@@ -2,6 +2,23 @@ export default function (_context) {
   return (root, result) => {
     let found = false
 
+    root.walkAtRules('tailwind', (node) => {
+      if (found) return false
+
+      if (node.parent && node.parent.type !== 'root') {
+        found = true
+        node.warn(
+          result,
+          [
+            'Nested @tailwind rules were detected, but are not supported.',
+            "Consider using a prefix to scope Tailwind's classes: https://tailwindcss.com/docs/configuration#prefix",
+            'Alternatively, use the important selector strategy: https://tailwindcss.com/docs/configuration#selector-strategy',
+          ].join('\n')
+        )
+        return false
+      }
+    })
+
     root.walkRules((rule) => {
       if (found) return false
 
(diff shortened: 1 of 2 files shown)


reinink:
Here are some ideas.

First, for nested Tailwind at-rules:

```
warn - Nested @tailwind rules were detected, but are not supported.
warn - Consider using a prefix to scope Tailwind's classes: https://tailwindcss.com/docs/configuration#prefix
warn - Alternatively, use the important selector strategy: https://tailwindcss.com/docs/configuration#selector-strategy
```

And for just plain nesting:

```
warn - Nested CSS was detected, but CSS nesting has not been configured properly.
warn - Please enable a CSS nesting plugin *before* Tailwind in your configuration.
warn - See here for more information: https://tailwindcss.com/docs/using-with-preprocessors#nesting
```

RobinMalfait:
Thanks @reinink, I'll merge this with the provided changes. If we need further cleanup in warning messages (which we probably will if we want to add links to documentation pages).

ziyingjing:
https://stackoverflow.com/questions/63761312/how-to-scope-tailwind-css
like this question, 
I need to scope @tailwind base; ,   
(important: ".tailwind" ) can't limit @tailwind base

PR #6018, opened 2021-11-08
Enforce the order of some variants

### LINKED ISSUE #6016: In v3 alpha, variant combinations like `hover:before:` aren't working

v3.0.0-alpha.2

webpack 5

v16

Tested in Chrome and Firefox

macOS

**Reproduction URL**

https://play.tailwindcss.com/9hQ7AdVmRy

**Describe your issue**

In v2 I often used variant-combinations like `hover:before:`, but in v3 the `:before` and `:after` pseudo-element variants don't seem to work in combination with pseudo-class variants.

RobinMalfait:
The order of variants can be important for certain scenario's.

For example, if you have:
```html
<div class="group-hover:dark:text-center"></div>
```

This would result in:
```css
.group:hover .dark .group-hover\:dark\:text-center {
  text-align: center;
}
```
This requires the following DOM structure:
```html
<div class="group">
  <div class="dark">
    <div class="group-hover:dark:text-center"></div>
  </div>
</div>
```

However, switching the order might be useful.

```html
<div class="dark:group-hover:text-center"></div>
```

This would result in:
```css
.dark .group:hover .dark\:group-hover\:text-center {
  text-align: center;
}
```

This requires the following DOM structure:
```html
<div class="dark">
  <div class="group">
    <div class="group-hover:dark:text-center"></div>
  </div>
</div>
```

This is a contrived example, but there are scenarios (see our Typography plugin) where it does make sense to switch the order sometimes.
However, in some scenario's it doesn't make sense because you would generate invalid css selectors (well.. not invalid, but selectors that won't work).

Such an example is when you have `hover:before:text-center` vs `before:hover:text-center`. This would result in:
```css
.hover\:before\:text-center::before:hover { /* This is useless, because you can't hover a ::before element */
  content: var(--tw-content);
  text-align: center;
}

.before\:hover\:text-center:hover::before {
  content: var(--tw-content);
  text-align: center;
}
```

To fix this, we move all the pseudo elements to the back of the selector, otherwise the selector wouldn't even work. This is implemented on a selector level and not on a variant API level. Which means that this would also work for third party plugins.
With this PR, the examples we used earlier (`hover:before:text-center` vs `before:hover:text-center`) would result in:

```css
.hover\:before\:text-center:hover::before { /* This now has :hover::before, even if the class in the HTML is different */
  content: var(--tw-content);
  text-align: center;
}

.before\:hover\:text-center:hover::before {
  content: var(--tw-content);
  text-align: center;
}
```

The cool part here is that we didn't change the base selector found in the HTML. This is because of the new `addVariant` API introduced in #5809.
We only append and prepend selector parts to the existing base class (the candidate found in the HTML), instead of re-building the selector from scratch.

This PR does mean that if you used `hover:before:text-center` in one file, and `before:hover:text-center` in the other, that there is a bit of duplication going on. Maybe that's fine, maybe in the future we can warn about this to enforce consistency. But at least it will work now instead of introducing a bug.

Fixes: #6016

### DIFF
--- a/src/lib/generateRules.js
+++ b/src/lib/generateRules.js
@@ -277,9 +277,27 @@ function splitWithSeparator(input, separator) {
   return input.split(new RegExp(`\\${separator}(?![^[]*\\])`, 'g'))
 }
 
+// A list of variants that are forced to the end. This is useful for variants
+// that have pseudo elements which can't really be combined with other variant
+// if they are in the incorrect order.
+//
+// E.g.:
+//  - `before:hover:text-center` would result in `.before\:hover\:text-center:hover::before`
+//  - `hover:before:text-center` would result in `.hover\:before\:text-center:hover::before`
+//
+// `::before:hover` doesn't work, which means that we can make it work for you by flipping the order.
+let forcedVariantOrder = ['before', 'after']
+
 function* resolveMatches(candidate, context) {
   let separator = context.tailwindConfig.separator
   let [classCandidate, ...variants] = splitWithSeparator(candidate, separator).reverse()
+
+  // Sort the variants if we used a forced variant.
+  // Note: this will not sort the others, it would only sort the forced variants.
+  if (variants.some((variant) => forcedVariantOrder.includes(variant))) {
+    variants.sort((a, z) => forcedVariantOrder.indexOf(a) - forcedVariantOrder.indexOf(z))
+  }
+
   let important = false
 
   if (classCandidate.startsWith('!')) {
--- a/src/lib/generateRules.js
+++ b/src/lib/generateRules.js
@@ -277,27 +277,9 @@ function splitWithSeparator(input, separator) {
   return input.split(new RegExp(`\\${separator}(?![^[]*\\])`, 'g'))
 }
 
-// A list of variants that are forced to the end. This is useful for variants
-// that have pseudo elements which can't really be combined with other variant
-// if they are in the incorrect order.
-//
-// E.g.:
-//  - `before:hover:text-center` would result in `.before\:hover\:text-center:hover::before`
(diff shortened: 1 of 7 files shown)


src/lib/generateRules.js
```diff
@@ -277,9 +277,27 @@ function splitWithSeparator(input, separator) {
...
+//
+// `::before:hover` doesn't work, which means that we can make it work for you by flipping the order.
+let forcedVariantOrder = ['before', 'after']
+
 function* resolveMatches(candidate, context) {
   let separator = context.tailwindConfig.separator
   let [classCandidate, ...variants] = splitWithSeparator(candidate, separator).reverse()
+
+  // Sort the variants if we used a forced variant.
+  // Note: this will not sort the others, it would only sort the forced variants.
```
RobinMalfait, line 296: This is because the `indexOf()` checks below would result in `-1 - -1` which will be `0` and therefore the order would be consistent.

src/lib/generateRules.js
```diff
@@ -277,9 +277,27 @@ function splitWithSeparator(input, separator) {
...
+// A list of variants that are forced to the end. This is useful for variants
+// that have pseudo elements which can't really be combined with other variant
+// if they are in the incorrect order.
+//
+// E.g.:
+//  - `before:hover:text-center` would result in `.before\:hover\:text-center:hover::before`
+//  - `hover:before:text-center` would result in `.hover\:before\:text-center:hover::before`
+//
+// `::before:hover` doesn't work, which means that we can make it work for you by flipping the order.
+let forcedVariantOrder = ['before', 'after']
```
RobinMalfait, line 289: We can add more variant names to this list.

PR #6038, opened 2021-11-10
Replace cosmiconfig with lilconfig

TrySound:
It's already replaced in postcss-load-config

See https://github.com/postcss/postcss-load-config/blob/main/package.json#L26


adamwathan:
Hey sorry any chance you could rebase this to resolve the conflicts? Then can definitely merge 👍🏻 Thank you!

TrySound:
Hey, will rebase in a couple of hours! Btw having more ideas how to improve install size.

bradlc:
Hey @TrySound, thanks for updating the PR! What's the reason for updating the `postcss` version in `package.json`?

thecrypticace:
it probably happened automatically and wasn't noticed. NPM is doing the same thing for me any time I npm install a new dependency into the tailwindcss repo. I have absolutely no idea why.

TrySound:
yeah, npm does a bad job with lockfiles. Yarn and pnpm are more stable.

bradlc:
Ah I see, all good 👍 Thanks again!

MichaelDeBoey:
@adamwathan @bradlc This would be a breaking change, as files without an extension are now handled differently
https://github.com/antonk52/lilconfig#difference-to-cosmiconfig

Schermafbeelding 2022-03-06 om 22 06 54

TrySound:
[postcss-load-config](https://github.com/postcss/postcss-load-config) migrated in 3.1.

MichaelDeBoey:
@TrySound They provide the `yaml` loaders themselves instead, so for them it's not breaking at all

Since we don't do that in this PR, it's a breaking change for Tailwind

PR #6039, opened 2021-11-10
Replace chalk with picocolors

TrySound:
Already used in latest postcss, autoprefixer and browserslist versions.

See https://github.com/alexeyraspopov/picocolors


TrySound:
Hey @RobinMalfait, what do you think about this change?

bradlc:
Thank you! 🙏

PR #6111, opened 2021-11-16
Simplify CI

RobinMalfait:
It could be that this won't work, either first try or just at all.
However I saw that the actions/setup-node repo has a native `cache`
option.

So... let's try it?


RobinMalfait:
Welp, didn't have the effects I wanted. Our CI is already pretty nice so going to close this 👋

PR #6178, opened 2021-11-22
Add aspect ratio `dimensions` value

neupauer:
This PR add new `dimensions` value for the aspect ratio utility.

### DIFF
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -115,7 +115,7 @@ module.exports = {
       auto: 'auto',
       square: '1 / 1',
       video: '16 / 9',
-      dimensions: 'attr(width) / attr(height)'
+      attrs: 'attr(width) / attr(height)'
     },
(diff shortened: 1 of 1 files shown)


adamwathan:
Great idea! I think the name `dimensions` is a tricky one because I worry people might think it's the images intrinsic dimensions (which might be different than the dimensions specified by the width/height attributes).

I'm going to rename it to `aspect-attrs` which I think is a little more direct and makes it clearer that you actually have to include the attributes, and nice because it's shorter too. Feel free to ping me if you really disagree though and happy to discuss it before it's tagged in a release ❤️

neupauer:
I am OK with new name. Thanks.

adamwathan:
So looks like this doesn't work at all, no browsers support `attr(width)` or `attr(height)`:

https://caniuse.com/css3-attr

image

Too bad because this would have been cool.

Let's make sure to actually test features before PRing them in the future please and thank you! ❤️

PR #6233, opened 2021-12-02
Add `0` as an arbitrary value for utilities (jit)

AviAvinav:
Tried adding`0` as an arbitrary value for utilities.

I just added `value === 0` to the `length` function as I think it might not require a regular expression for just `0`.

I am not sure if this will solve the issue, because I am not sure how to test it for jit, can someone help me to test it out.

For issue: #6227


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #6289, opened 2021-12-07
Partially revert tuple syntax PR for `screens`: #6104

RobinMalfait:
This PR will revert the PR to enable tuple syntax: #6104

The reason this will be reverted is because we currently can't guarantee non-breaking changes and it doesn't guarantee screens order in all cases. If we want to redo this, then it will probably require a bit more thought.
Some things that are confusing:
- When you are using the `theme()` function, then you will receive a normalised object instead of what you have in your config.
- We have no good way to sort the internal `screens` object with an extended tuple syntax in your config (order can't be guaranteed)
  - Even if we added the tuple syntax to our internal default config. Because how should we merge the 2 arrays?
    - Append `user` config to `default` config -> might not be correct
    - Sort by first argument in tuple (the `key`) -> this could result in the `user` tuple to be in a different order.
    - Sort by second argument in tuple (the `value`) -> this could result in the `user` tuple to be in a different order. What if you are using `raw` screens, what if you are using mixed values (`min`, `max`), what if you are using units that rely on runtime information (`var()`, `rem`, `vmin`, ...)

To many confusing questions with no great answer right now.

The reason that it is **partially** reverted, is because it still contained usefull cleanup internally by moving the normalisation to a shared and tested function instead of copying the same logic in multiple places.


Bobakanoosh:
Any updates on this syntax coming back? I just finally got around to wanting to use it and was disappointed to find that it was reverted.

PR #6378, opened 2021-12-10
use text-decoration-line instead of text-decoration

o-t-w:
Currently the following code would not lead to the underline being green, which is a bug imo: `class="decoration-green-700 hover:underline hover:underline-offset-4"`

This is because `text-decoration` is a shorthand for `text-decoration-line`, `text-decoration-color`, `text-decoration-style`, and `text-decoration-thickness` so overrides any of those values (if they have been set) to `initial`.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -1660,10 +1660,10 @@ export let corePlugins = {
 
   textDecoration: ({ addUtilities }) => {
     addUtilities({
-      '.underline': { 'text-decoration': 'underline' },
-      '.overline': { 'text-decoration': 'overline' },
-      '.line-through': { 'text-decoration': 'line-through' },
-      '.no-underline': { 'text-decoration': 'none' },
+      '.underline': { 'text-decoration-line': 'underline' },
+      '.overline': { 'text-decoration-line': 'overline' },
+      '.line-through': { 'text-decoration-line': 'line-through' },
+      '.no-underline': { 'text-decoration-line': 'none' },
     })
   },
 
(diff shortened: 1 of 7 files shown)


src/corePlugins.js
```diff
@@ -1491,8 +1491,8 @@ export let corePlugins = {
   padding: createUtilityPlugin('padding', [
     ['p', ['padding']],
     [
-      ['px', ['padding-left', 'padding-right']],
-      ['py', ['padding-top', 'padding-bottom']],
+      ['px', ['padding-inline']],
+      ['py', ['padding-block']],
```
lukewarlow, line 1495: I would undo this change it's unrelated to the PR

src/corePlugins.js
```diff
@@ -379,8 +379,8 @@ export let corePlugins = {
     [
       ['m', ['margin']],
       [
-        ['mx', ['margin-left', 'margin-right']],
-        ['my', ['margin-top', 'margin-bottom']],
+        ['mx', ['margin-inline']],
+        ['my', ['margin-block']],
```
lukewarlow, line 383: I would also undo this change it's unrelated to the PR

lukewarlow:
Worth being aware this is technically a breaking change.

reinink:
@o-t-w I like this change.

@lukewarlow can you give an example of how this could break things? Seems pretty unlikely, but maybe I'm missing something?

lukewarlow:
@reinink I agree it's unlikely and someone would have to be relying on what I would agree is buggy behaviour but technically someone could be relying on the fact that text-decoration does override existing values. 

I just thought it was worth pointing out. I'm generally of the opinion these kinds of "breaking changes" are perfectly reasonable though.

reinink:
@o-t-w Hey, I forgot to say thanks when I merged this in earlier. Much appreciated! 🙏

PR #6405, opened 2021-12-11
Add 'text-decoration-color' to default 'transition' and 'transition-colors' utilities

fekle:
Hi there!

In this pull request I added the `text-decoration-color` property to the default options for `transition` and `transition-colors` utilities, as tailwindcss v3 introduced support for underline styles. 

As it is right now when the underline color differs from the text color, say it is transparent by default and black on hover, `transition` and `transition-colors` only enable transition on the text color itself, but not on the underline color. 

I'm not sure if this was intended, but in my opinion it makes sense to also transition the `text-decoration-color` property by default.
Or maybe it would be better to add a new `transition-text-decoration` property?

I updated all test files to reflect those changes, and all tests pass.


lucalves (approved):

reinink:
@fekle Good catch! Thanks for submitting this PR 🙏

fekle:
No worries, thanks for merging!
Sorry about those `tailwind-output-*` files in the fixtures, they were a part of master when i created the PR.

reinink:
@fekle All good, it was a good reminder to actually remove these, as they are no longer used: 0783a4275ea8d3dd374c976d357e900c6767e729

Also, FYI, I also switched the order of the `color` and `text-decoration-color ` properties: 1c7a793c9432e284d353b06c3339a520bde124ef

fekle:
@reinink ah, good to know. I wasn't sure how to order them, since not the entire list is ordered alphabetically, so I assumed that the 'longer' color attributes like `text-decoration-color` and `border-color` came before the base `color` attribute. But I guess it doesn't matter much in the end :)

reinink:
Yeah, oops, I meant to put `color` first, and then do alphabetical after that. Updated that now: ec911c73f08c81b7088246484379f2efce97a667

And you're right, doesn't really matter in the end.

PR #6526, opened 2021-12-15
Insert always-on `defaults` layer in correct spot

### LINKED ISSUE #6524: From 3.0.2 to 3.0.3 borders color in config is lost

"@tailwindcss/forms": "0.4.0",
"@tailwindcss/line-clamp": "0.3.0",
"tailwindcss": "3.0.3",

"postcss": "8.4.5",

16

Chrome

Windows

**Reproduction URL**

COMING SOON

**Describe your issue**

After upgrade from "tailwindcss": "3.0.2" to "tailwindcss": "3.0.3" my borders color is now black instead of `colors.gray.300`.

I'm using this in config:

```js
theme: {
  extend: {
    borderColor: (theme) => ({
      DEFAULT: theme("colors.gray.300", "currentColor"),
    }),
  },
},
```

RobinMalfait:
We were injecting the always on `@tailwind defaults` layer at the beginning of
the file. However, if a `@tailwind base` layer is available, then that
will now be injected _after_ the defaults layer. The base layer does
contain some reset that are now overriding the defaults we set.

So now we will:
- Insert the `@tailwind defaults` layer at the beginning of the file
  _if_ there is no `@tailwind base`
- Insert the `@tailwind defaults` layer after the `@tailwind base` layer
  if it exists.

Fixes: #6524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Insert always on defaults layer in correct spot ([#6526](https://github.com/tailwindlabs/tailwindcss/pull/6526))
 
 ## [3.0.3] - 2021-12-15
 
--- a/src/lib/expandTailwindAtRules.js
+++ b/src/lib/expandTailwindAtRules.js
@@ -221,15 +221,22 @@ export default function expandTailwindAtRules(context) {
 
     // Replace any Tailwind directives with generated CSS
 
+    if (layerNodes.base) {
+      layerNodes.base.before(cloneNodes([...baseNodes], layerNodes.base.source))
+    }
+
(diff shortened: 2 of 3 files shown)


productdevbook:
```
 Error: Can't find stylesheet to import.
    ╷
  1 │ @import "tailwindcss/defaults";
    │         ^^^^^^^^^^^^^^^^^^^^^^
    ╵
    src/assets/main.scss 1:9  root stylesheet

```

RobinMalfait:
Hey @productdevbook ! Thank you for your bug report!
Much appreciated! 🙏

But, this is a private API and you shouldn't use it yourself. If you remove it again then everything should keep working as expected.
What is the reason that you've added this?

productdevbook:
> Hey @productdevbook ! Thank you for your bug report! Much appreciated! 🙏
> 
> But, this is a private API and you shouldn't use it yourself. If you remove it again then everything should keep working as expected. What is the reason that you've added this?

😅 When I read the description, I thought we should add it too. Yes, it works when you remove it.

Issue #6321, opened 2021-12-10
Utility classes can't be built with @apply and custom props


RedRiderX:
**What version of Tailwind CSS are you using?**

v3.0.0

**What build tool (or framework if it abstracts the build tool) are you using?**

PostCSS

**What version of Node.js are you using?**

v14.13.0

**What browser are you using?**

Chrome

**What operating system are you using?**

Windows

**Reproduction URL**

https://play.tailwindcss.com/qm3e5FDSEr

**Describe your issue**

I've made utility classes that have both Tailwind tokens and custom css. But these classes could no longer be used with `@apply` after upgrading to v3.

This seems to be a [regression from v2->v3](https://play.tailwindcss.com/CwBZvIVuPx?file=css).

I can mitigate the issue for now by [duplicating some classes](https://play.tailwindcss.com/7FFOWiZ88S?file=css), but that's not a great solution.

JoseGarciaM:
same for me =(

reinink:
Hey! So we're going to dig into this a little deeper, but as a workaround you can wrap these classes in an `@layer utilities` directive to get this working:

```css
@layer utilities {
  .a {
    @apply b;
  }

  .b {
    @apply uppercase;
    color: red;
  }
}
```

Here's an updated Play illustrating this: https://play.tailwindcss.com/ksUrE9fW01?file=css

Tony2:
Hi, I have a lot of custom css classes that I reuse via `@apply my-css-class`. Recently I upgraded to v3 and  things stopped working. I'd like to ask if there is any workaround apart from putting those classes into an `@layer utilities` as above? Since I took for granted that I can reuse those classes with `@apply`, they are over all places and it's not easy to find them all.

garymathews:
@Tony2 with `3.0.7` you don't need to use `@layer`, but each css that uses `@apply` will need to include a reference to `@tailwind`. I have a PR that resolves this to `2.X.X` behaviour #6580

Tony2:
thank you @garymathews for the hint!

I rely heavily on twcss for my work (an in-house component framework). This bug made me somewhat worried if my bet on twcss is the correct decision. Till v2 it has appeared rock-solid and very well-maintained. I understand that v3 is still relatively new, however being able to use `@apply my-class` is IMO a vital feature to keep css maintainable.

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

[...]

PR #6580, opened 2021-12-17
Allow all classes for @apply

garymathews:
Relates to https://github.com/tailwindlabs/tailwindcss/issues/6321

- Basically, _apply_ only works for Tailwind generated classes under JIT; which is default when moving to Tailwind 3.
  - This issue is also present in Tailwind 2 when using JIT.


src/lib/expandApplyAtRules.js
```diff
@@ -122,6 +122,27 @@ function processApply(root, context) {
...
+      if (node.selector && node.selector[0] === '.') {
+        const name = node.selector.substring(1)
+
+        // Ignore unused candidates.
+        if (applyCandidates.has(name) && !applyClassCache.has(name)) {
+          applyClassCache.set(name, [
+            [
+              {
+                sort: BigInt(context.layerOrder.user),
+                layer: 'user',
```
garymathews, line 136: I'm not too sure about these

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

If I run your tests against the current `master` branch, then it already passes. I'll look into the actual issue and mark you as a contributor if I have a fix.

garymathews:
@RobinMalfait, thanks for fixing this 👍  https://github.com/tailwindlabs/tailwindcss/pull/6589

However, I noticed it's required to reference `@tailwind;` to the particular css file with the `@apply` for this to work (don't even need to define a directive); where before this was not necessary. (I should only need to define directives in my main css file)

```JS
it('should pickup all classes', () => {
  let config = {
    content: [{ raw: html`<div class="foo"></div>` }],
    plugins: [],
  }

  let input = css`
    @tailwind; /* remove me I break */

    .bop {
      color: red;
    }
    .bar {
      background-color: blue;
    }
    .foo {
      @apply absolute bar bop;
    }
  `

  return run(input, config).then((result) => {
    return expect(result.css).toMatchFormattedCss(css`
      @tailwind;
      .bop {
        color: red;
      }
      .bar {
        background-color: blue;
      }
      .foo {
        position: absolute;
        color: red;
        background-color: blue;
      }
    `)
  })
})
```
```
The `bar` class does not exist. If `bar` is a custom class, make sure it is defined within a `@layer` directive.
```

garymathews:
I found its due to an optimization where css files without a directive are ignored:
[setupTrackingContext.js#L127](https://github.com/tailwindlabs/tailwindcss/blob/master/src/lib/setupTrackingContext.js#L127)
[setupWatchingContext.js#L244](https://github.com/tailwindlabs/tailwindcss/blob/master/src/lib/setupWatchingContext.js#L244)

garymathews:
@RobinMalfait I've updated my PR with a solution that maintains the optimization for css without @tailwind or @apply

PR #6588, opened 2021-12-17
Improve circular dependency detection when using `@apply`

RobinMalfait:
This PR will improve the detection of circular dependencies.
We incorrectly had a false positive in this case:

HTML:
```html
<div class="bg-gray-500"></div>
```

CSS:
```css
.focus\:bg-gray-500 {
  @apply bg-gray-500;
}
```

While you should not do this, this resulted in an incorrect circular dependency error.
Now it will result in the correct css:

```css
.bg-gray-500 {
  --tw-bg-opacity: 1;
  background-color: rgb(107 114 128 / var(--tw-bg-opacity));
}

.focus\:bg-gray-500 {
  --tw-bg-opacity: 1;
  background-color: rgb(107 114 128 / var(--tw-bg-opacity));
}
```

I also changed the circular dependency message to the same message we used in v2.
Before:
```
Circular dependency detected when using: `@apply text-red-500`
```

After:
```
You cannot `@apply` the `text-red-500` utility here because it creates a circular dependency.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Don't mutate custom color palette when overriding per-plugin colors ([#6546](https://github.com/tailwindlabs/tailwindcss/pull/6546))
+- Improve circular dependency detection when using `@apply` ([#6588](https://github.com/tailwindlabs/tailwindcss/pull/6588))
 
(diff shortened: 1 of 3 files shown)


driskull:
I'm still seeing this issue as well...

```
You cannot `@apply` the `sticky` utility here because it creates a circular dependency.
```

```
:host {
  &.sticky {
    @apply sticky;
  }
}
```

driskull:
https://play.tailwindcss.com/YxcQz4gSop?file=css

PR #6589, opened 2021-12-17
Only generate variants for non-`user` layers

RobinMalfait:
This PR fixes a bug that generated variants for custom user css which is something we don't want. We only want this for css that's defined in a layer.

E.g.:

CSS:
```css
@tailwind utilities;

.foo {
  color: red;
}
```

HTML:
```html
<div class="hover:foo"></div>
```

This should not generate (but it did):
```css
.hover\:foo:hover {
  color: red;
}
```

This PR will disallow that. If you _do_ want this behaviour, then you would have to put the `.foo` class inside a layer. E.g.:
CSS:
```css
@tailwind utilities;

@layer utilities {
  .foo {
    color: red;
  }
}
```

HTML:
```html
<div class="hover:foo"></div>
```

This **should** generate:
```css
.hover\:foo:hover {
  color: red;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Don't mutate custom color palette when overriding per-plugin colors ([#6546](https://github.com/tailwindlabs/tailwindcss/pull/6546))
 - Improve circular dependency detection when using `@apply` ([#6588](https://github.com/tailwindlabs/tailwindcss/pull/6588))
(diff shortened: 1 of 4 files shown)


mnaderian:
**Shouldn't** it generate this?

```css
.foo:hover {
  color: red;
}
```

mnaderian:
@RobinMalfait, shouldn't it generate this? 🤔 

```CSS
.foo:hover {
  color: red;
}
```

mnaderian:
@RobinMalfait, shouldn't it generate this? 🤔 

```CSS
.hover\:foo:hover {
  color: red;
}
```

PR #6594, opened 2021-12-17
Ensure `@apply` of a rule inside an AtRule works

### LINKED ISSUE #6246: @apply with @supports produces extra commas

2.2.19

tailwind cli

v14.18.1

N/A

macOS

**Reproduction URL**

https://github.com/leevigraham/tailwind-apply-bug-repro

**Describe your issue**

* Issue: Output contains additional `,`'s breaking css.
* Desired output: No extra commas

I'll admit this is a bit of a weird one but I ran into it writing some funky css. This test case is the minimal repro (really minimal… no actual css attributes in rules).

The following css:

```css
@tailwind components;

@layer components {

    .grid--cards {
    }

    .grid-X,
    .grid-Y,
    .grid-Z {
        @apply grid--cards;
    }

    @supports (grid-template-rows: subgrid) {
        .grid--cards {}
        .some-other-rule {}
    }
}
```

Produces:

```css
.grid-X, .grid-Y, .grid-Z {
}

@supports (grid-template-rows: subgrid) {
  .grid-X, .grid-Y, .grid-Z {
  }

  , ,  {
  }
}
```

Reproduction:

1. Clone the repo and run `npm install`.
2. Run `npx tailwindcss -i input.css -o output.css`
3. check `output.css`

Issue also exists in tailwind@next: `npx tailwindcss@next -i input.css -o output.css`

Notes:

Removing `.some-other-rule {}` results in:

```css
.grid-X, .grid-Y, .grid-Z {
}

@supports (grid-template-rows: subgrid) {
  .grid-X, .grid-Y, .grid-Z {
  }
}
```

Adding more rules results in more commas:

```css
/* input.css */
@tailwind components;

@layer components {

    .grid--cards {
    }

    .grid-X,
    .grid-Y,
    .grid-Z,
    .grid-Z,
    .grid-Z,
    .grid-Z {
        @apply grid--cards;
    }

    @supports (grid-template-rows: subgrid) {
        .grid--cards {}
        .some-other-rule {}
    }
}
```
```css
/* output.css */
.grid-X, .grid-Y, .grid-Z, .grid-Z, .grid-Z, .grid-Z {
}

@supports (grid-template-rows: subgrid) {
  .grid-X, .grid-Y, .grid-Z, .grid-Z, .grid-Z, .grid-Z {
  }

  , , , , ,  {
  }
}
```

RobinMalfait:
This PR will ensure that if you are using `@apply something` and that `something` happens to be in an AtRule and that AtRule contains an unrelated node that we only apply the related node... 🥵

Let's imagine you have the following structure:

```css
.foo {
  @apply bar;
}

@supports (a: b) {
  .bar {
    color: blue
  }

  .something-unrelated {}
}
```

In this case we want to apply `.bar` but it happens to be in an AtRule node. We clone that node instead of the nested one because we still want that `@supports` rule to be there once we applied everything.

However it happens to be that the `.something-unrelated` is also in that same shared `@supports` atrule. This is not good, and this should not be there. The good part is that this is a clone already and it can be safely removed. The question is how do we know we can remove it. Basically what we can do is match it against the applyCandidate that you want to apply. If it doesn't match the we can safely delete it.

If we didn't do this, then the `replaceSelector` function would have replaced this with something that didn't exist and therefore it removed the selector altogether. In this specific case it would result in `{}` instead of `.something-unrelated {}`

Fixes: #6246

### DIFF
--- a/src/lib/expandApplyAtRules.js
+++ b/src/lib/expandApplyAtRules.js
@@ -266,6 +266,42 @@ function processApply(root, context) {
 
           if (canRewriteSelector) {
             root.walkRules((rule) => {
+              // Let's imagine you have the following structure:
+              //
+              // .foo {
+              //   @apply bar;
+              // }
+              //
+              // @supports (a: b) {
+              //   .bar {
+              //     color: blue
+              //   }
+              //
+              //   .something-unrelated {}
+              // }
+              //
+              // In this case we want to apply `.bar` but it happens to be in
+              // an atrule node. We clone that node instead of the nested one
+              // because we still want that @supports rule to be there once we
+              // applied everything.
+              //
+              // However it happens to be that the `.something-unrelated` is
+              // also in that same shared @supports atrule. This is not good,
+              // and this should not be there. The good part is that this is
+              // a clone already and it can be safely removed. The question is
+              // how do we know we can remove it. Basically what we can do is
+              // match it against the applyCandidate that you want to apply. If
+              // it doesn't match the we can safely delete it.
+              //
+              // If we didn't do this, then the `replaceSelector` function
+              // would have replaced this with something that didn't exist and
+              // therefore it removed the selector altogether. In this specific
+              // case it would result in `{}` instead of `.something-unrelated {}`
+              if (!extractClasses(rule).some((thing) => thing === applyCandidate)) {
+                rule.remove()
+                return
+              }
+
               rule.selector = replaceSelector(parent.selector, rule.selector, applyCandidate)
 
               rule.walkDecls((d) => {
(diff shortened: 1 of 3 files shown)


driskull:
Not sure that this is related, but with the latest `3.0.7` release I'm now seeing this error:

```
Cannot read properties of undefined (reading 'parent')
```

Seems like this PR is the only thing that might be related to reading parent. Is that correct?

RobinMalfait:
@driskull hmm can you create a minimal reproduction repo and open a new issue?

src/lib/expandApplyAtRules.js
```diff
@@ -266,6 +266,42 @@ function processApply(root, context) {
...
+              // how do we know we can remove it. Basically what we can do is
+              // match it against the applyCandidate that you want to apply. If
+              // it doesn't match the we can safely delete it.
+              //
+              // If we didn't do this, then the `replaceSelector` function
+              // would have replaced this with something that didn't exist and
+              // therefore it removed the selector altogether. In this specific
+              // case it would result in `{}` instead of `.something-unrelated {}`
+              if (!extractClasses(rule).some((thing) => thing === applyCandidate)) {
+                rule.remove()
```
driskull, line 301: @RobinMalfait I'm not sure what exactly the issue is. I'll keep investigating.

My best guess is that something else is missing besides just removing the rule. Maybe the parent still has a reference to the rule which would cause it to be undefined?

driskull: If I do the following, it no longer errors but I'm just shooting in the dark here...

```
rule.before(rule.nodes)
rule.remove();
return;
```

RobinMalfait: Please open an issue with a reproduction.

driskull: https://github.com/tailwindlabs/tailwindcss/issues/6599

PR #6797, opened 2021-12-29
Improve `DEBUG` flag

### LINKED ISSUE #6769: DEBUG environment variable

2.x, 3.x

n/a

n/a

n/a

n/a

**Reproduction URL**

Any project can be used to reproduce the issue. Just add `DEBUG=something` in the environment.

**Describe your issue**

The JS ecosystem, pushed by the [`debug`](https://github.com/debug-js/debug) package (625M dl / month 🔥), defined the `DEBUG` environment variable as a comma-separated list of tags. This lets multiple tools write their own debug messages, letting the user pick the ones they want to show.

Tailwind doesn't follow this pattern (it just checks [whether the value is anything other than 0](https://github.com/tailwindlabs/tailwindcss/blob/de1bdb4f5df663c10fb641deeb331f36daf2d7ae/src/lib/sharedState.js#L4)), causing Tailwind to print a large amount of unwanted logs if we attempt to print debug logs for anything else.

For instance, I get the following logs after each rebuild in a Next.js application (which I run with `DEBUG='postgraphile:*'` in order to get some debug information I care about); as you can see, two lines are for Webpack, and 22 are for Tailwind:

image

Tailwind should instead respect the `DEBUG` comma-separated-tags convention, and only print debug logs if DEBUG contains something `tailwind` in its tags (ie `DEBUG=tailwind`, `DEBUG=postgraphile,tailwind`, etc).

RobinMalfait:
This PR will improve the parsing of the DEBUG flag. 
Currently we only check if you provided a DEBUG flag and that you didn't provide the value `0`. This worked so far, but it's time to improve upon this.

There is the `debug` package which is used by many applications and includes some conventions: https://github.com/debug-js/debug#conventions

So now we will ensure that take care of the following cases:
| Value                   | Debug enabled |
| ----------------------- | ------------- |
| 'true'                  | true          |
| '1'                     | true          |
| 'false'                 | false         |
| '0'                     | false         |
| '\*'                    | true          |
| ~'tailwind'~              | ~true~          |
| ~'tailwind:\*'~           | ~true~          |
| 'tailwindcss'           | true          |
| 'tailwindcss:\*'        | true          |
| ~'other,tailwind'~        | ~true~          |
| ~'other,tailwind:\*'~     | ~true~          |
| 'other,tailwindcss'     | true          |
| 'other,tailwindcss:\*'  | true          |
| ~'other,-tailwind'~       | ~false~         |
| ~'other,-tailwind:\*'~    | ~false~         |
| 'other,-tailwindcss'    | false         |
| 'other,-tailwindcss:\*' | false         |
| ~'-tailwind'~             | ~false~         |
| ~'-tailwind:\*'~          | ~false~         |
| '-tailwindcss'          | false         |
| '-tailwindcss:\*'       | false         |

Added a bit of code but the good part is that this only runs once at the beginning so shouldn't be too bad.

Fixes: #6769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Improve `DEBUG` flag ([#6797](https://github.com/tailwindlabs/tailwindcss/pull/6797))
 
 ## [3.0.8] - 2021-12-28
 
--- a/src/lib/sharedState.js
+++ b/src/lib/sharedState.js
@@ -1,10 +1,49 @@
 export const env = {
   TAILWIND_MODE: process.env.TAILWIND_MODE,
   NODE_ENV: process.env.NODE_ENV,
-  DEBUG: process.env.DEBUG !== undefined && process.env.DEBUG !== '0',
+  DEBUG: resolveDebug(process.env.DEBUG),
   TAILWIND_DISABLE_TOUCH: process.env.TAILWIND_DISABLE_TOUCH !== undefined,
   TAILWIND_TOUCH_DIR: process.env.TAILWIND_TOUCH_DIR,
 }
 export const contextMap = new Map()
 export const configContextMap = new Map()
 export const contextSourcesMap = new Map()
+
+export function resolveDebug(debug) {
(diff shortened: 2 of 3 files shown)


mustafa0x:
Is there a config option to tell tailwind to ignore `DEBUG=1`? I have it in my .env for my backend, but tailwind picks it up, since I'm using a direnv, which automatically loads .env files.

PR #6906, opened 2022-01-05
Enable optimize universal defaults

thecrypticace:
1. Re-enable `optimizeUniversalDefaults` by default
2. Re-enable apply handling for isolated CSS modules
3. Guard apply isolation to only when `optimizeUniversalDefaults` is enabled


thecrypticace:
Going with a different decision here. Will open another PR.

Issue #6690, opened 2021-12-22
TailwindCSS Standalone CLI not working in Alpine Linux (docker)


altdsoy:
The tailwindcss binary release is not working on Alpine Linux.  
Tested with the release tailwindcss-linux-x64 v3.0.7 on the latest docker alpine image.

If I run an alpine container (installing curl, and downloading the standalone release as described here https://tailwindcss.com/blog/standalone-cli#get-started ) I got nothing (binary not found).
Running the exact same steps in debian works.

Tell me if I need to provide more information.

Example steps to reproduce it:
```sh
#On the host
docker run -it --rm alpine ash

#Within the container
apk add curl
curl -sLO https://github.com/tailwindlabs/tailwindcss/releases/latest/download/tailwindcss-linux-x64
chmod +x tailwindcss-linux-x64
mv tailwindcss-linux-x64 tailwindcss
./tailwindcss init
# Got => ./tailwindcss: not found
```

The exact same steps works fine using debian for example:
```sh
#On the host
docker run -it --rm debian bash

#Within the container
apt update
apt install curl # Type Y to install
curl -sLO https://github.com/tailwindlabs/tailwindcss/releases/latest/download/tailwindcss-linux-x64
chmod +x tailwindcss-linux-x64
mv tailwindcss-linux-x64 tailwindcss
./tailwindcss init
# Generate tailwind.config.js
```

Edit:
Here is the output of `file`:
```
tailwindcss: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=2e882d60273a4031f72a88815a36cb574dc1090b, for GNU/Linux 2.6.32, stripped
```

Seems I need to install some libraries...

Edit 2:
This answer from SO seems relevant to the issue: https://stackoverflow.com/a/66974607

Executables built on glibc distributions depend on /lib/x86_64-linux-gnu/libc.so.6, for example, which is not available on Alpine (unless, they are statically linked).

So is it possible to statically link dependencies when compiling the binaries?

Edit 3:
Some workarounds are based on installing gcompat and/or libc6-compat, but those do not fix the tailwindcss issue.

Edit 4:
I just noticed that the standalone CLI is using a tool called `pkg` under the hood.

So maybe this issue does not belong here..
If so, feel free to close.

altdsoy:
[...]

PR #6914, opened 2022-01-05
Switch to using linuxstatic build target for the CLI

mveytsman:
👋 Hi this is a pull request to go with the discussion https://github.com/tailwindlabs/tailwindcss/discussions/6785

The current linux build is dynamically linked, so it requires libraries to be present on the system when running. This means it does not work on minimal Alpine containers (https://github.com/tailwindlabs/tailwindcss/issues/6690) unless the user knows to install two library packages. It also appears to not be compatible with NixOS (https://github.com/tailwindlabs/tailwindcss/discussions/6785#discussioncomment-1913331)

Packaging the linux binary statically will fix these issues and make it platform independent. It also will not take up any more space, based on my test builds below.

| Arch   | Dynamic | Static |
| -- | ------------- | ------------- |
| x64 | 41mb | 39mb |  
| arm64 | 43mb | 36mb |

### DIFF
--- a/standalone-cli/package.json
+++ b/standalone-cli/package.json
@@ -2,9 +2,9 @@
   "name": "tailwindcss-standalone",
   "version": "0.0.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


reinink:
@mveytsman Thanks so much for this PR! We've tested this out and it seems to work well. Nice improvement 💪

PR #7016, opened 2022-01-11
Ensure nesting plugins can receive options

lubomirblazekcz:
PostCSS nesting added some options that can be configured eg.
https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-nesting#noispseudoselector

Current implementation of Tailwind Nesting doesn't pass the options, becuase the `typeof` of the plugin is object with options, not function. This fixes it.

Example PostCSS Vite configuration

```js
import postcssImport from 'postcss-import';
import postcssNesting from 'postcss-nesting';
import tailwindcss from 'tailwindcss'
import tailwindcssNesting from 'tailwindcss/nesting'
import autoprefixer from 'autoprefixer'

export default {
  css: {
    postcss: {
      plugins: [postcssImport, tailwindcssNesting(postcssNesting({
        noIsPseudoSelector: true
      })), tailwindcss, autoprefixer]
    }
}
```

### DIFF
--- a/nesting/plugin.js
+++ b/nesting/plugin.js
@@ -14,7 +14,10 @@ module.exports = function nesting(opts = postcssNested) {
     })
 
     let plugin = (() => {
-      if (typeof opts === 'function' || typeof opts === 'object') {
+      if (
+        typeof opts === 'function' ||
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

PR #7163, opened 2022-01-22
Correctly parse and prefix animation names with dots

### LINKED ISSUE #7149: Wrong keyframe keys when generationg CSS classes with JS

For example: v3.0.15

postcss-cli 9.1.0, vite 2.7.13

16.13.2

Firefox

Ubuntu 20.04

**Reproduction URL**

https://github.com/Mastercuber/tailwindcss-minimal-example

**Describe your issue**

In the config file `tailwind.config.js` I'm generating some animations with floating point numbers in the class name like [this](https://github.com/Mastercuber/tailwindcss-minimal-example/blob/master/tailwind.config.js).  
The generated animation classes looks good, but the `@keyframe` directives have a different value than the animation classes itself, though the exact same key is used for both.

vitorrd:
Fixes #7149.

Adds prefix alone to animation names, such that the parser does not run on the rest of the identifier, possibly damaging class names like "zoom-1.5".
Although tests still pass, care must be taken as this effectively disables previous [albeit similar] behavior.


thecrypticace:
Appreciate you taking the time to figure this one out! 🎉

PR #7206, opened 2022-01-25
Feature: Add negative z-index to default config

jonnitto:
Negative z-indexes can be handy if you use images as backgrounds


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Negative values are a first class concept in Tailwind and doesn't require changes to the default stub. In other words, this already works! Checkout https://play.tailwindcss.com/KRdvzD7dqL

PR #7220, opened 2022-01-26
Remove false positive warning in CLI when using the `--content` option

### LINKED ISSUE #7210: CLI --content option shows warning

v3.0.16

v17.4.0

Windows 11

**Reproduction**

[Sample Download Link](https://github.com/tailwindlabs/tailwindcss/files/7939046/TW.Repro.zip)

1. Use npm to run css-build-dev in package.json, preferably on Windows.
2. Observe the warning.

**Describe your issue**

Tailwind v3.0 deprecated the CLI's `--purge` option and replaced it with `--content`. That's fine, but it doesn't work. I get the following:
```
warn - The `content` option in your Tailwind CSS configuration is missing or empty.
warn - Configure your content sources or your generated CSS will be missing styles.
warn - https://tailwindcss.com/docs/content-configuration
```
The issue can indeed be solved by adding `content` to `tailwind.config.js`, but since the option still exists in the CLI, it should also be possible there. Important to note is that `--purge` using the CLI worked fine in v2.x. There is clearly a bug with the new replacement `--content` option.

RobinMalfait:
This PR will remove the false positive warning about the empty content option in your config.

When you are using the CLI, then we try to resolve the `tailwind.config.js`, but it can happen that you don't have one and that you are using the `--content` flag instead.

E.g.:
```
npx tailwindcss -i input.css -o output.css --content ./index.html
```

We do make sure that the `./index.html` file is eventually present in your config, but it isn't initially and therefore you get the (incorrect) warning.

We fixed this by making sure that your `--content` is present immediately which removes the incorrect warning.

Note: You will still receive a warning if we didn't generate any utilities at all. This is a valid warning, and an indication that you might have linked the wrong content files, or that you didn't use any Tailwind classes yet.

Fixes: #7210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Remove false positive warning in CLI when using the `--content` option ([#7220](https://github.com/tailwindlabs/tailwindcss/pull/7220))
 
 ## [3.0.16] - 2022-01-24
 
--- a/src/cli.js
+++ b/src/cli.js
@@ -437,7 +437,6 @@ async function build() {
 
   function resolveConfig() {
     let config = configPath ? require(configPath) : {}
-    let resolvedConfig = resolveConfigInternal(config)
 
     if (args['--purge']) {
       log.warn('purge-flag-deprecated', [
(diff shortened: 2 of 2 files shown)


EatonZ:
@RobinMalfait Wow, fixed & released less than 24 hours later. Thanks so much! 🎉

RobinMalfait:
@EatonZ thank you for the bug report! If you run into other issues, feel free to open new issues, thanks!

PR #7268, opened 2022-01-31
add fontSize 2xs and 3xs.

kght6123:
add Front size for
- 2xs
- 3xs

It is often used mainly in the following situations.
- UI of the administration screen.
- Comments and annotation characters.

The image applied for trial with Tailwind Play is as follows. (Red letters, red frame part.)
スクリーンショット 2022-01-31 10 06 56

### DIFF
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -372,6 +372,8 @@ module.exports = {
       ],
     },
     fontSize: {
+      '3xs': ['0.5rem', { lineHeight: '0.5rem' }],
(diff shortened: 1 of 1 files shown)


reinink:
Hey! Thanks for this suggestion, but I think we're going to leave the default font sizes as they are. The main reason being that `text-xs` is already very small (`12px`), and from an accessibility perspective is already a little risky to use.

There is, of course, nothing stopping you from using `2xs` and `3xs` in your projects by simply adding these values to your own Tailwind config 👍

kght6123:
@reinink 
Also, if you have a good idea, create a new PR.
thank you.

PR #7272, opened 2022-01-31
Add desktop first media queries

minimit:
This is a simple change we can discuss it in the PR maybe?

I think it's super useful when someone wants to add inverse media queries (desktop first).

I'm not sure the order of media queries is right, which one should have precedence (mobile first or desktop first), for sure the desktop first media queries must be in inverse order.

### DIFF
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -9,6 +9,11 @@ module.exports = {
       lg: '1024px',
       xl: '1280px',
       '2xl': '1536px',
+      '-2xl': { max: '1535px' },
+      '-xl': { max: '1279px' },
+      '-lg': { max: '1023px' },
+      '-md': { max: '767px' },
+      '-sm': { max: '639px' },
     },
     colors: ({ colors }) => ({
       inherit: colors.inherit,


reinink:
@minimit Hey thanks for this contribution and for your interest in Tailwind CSS! However, I don't think this update make sense, at least not as part of Tailwind core. Tailwind intentionally takes a [mobile-first](https://tailwindcss.com/docs/responsive-design#mobile-first) approach to responsive design, and we believe this is a sensible default for most projects. Of course, there is nothing stopping you from doing something different in your own projects by [customizing your breakpoints](https://tailwindcss.com/docs/responsive-design#customizing-breakpoints). Thanks either way! 👍

minimit:
I understand but I think it's wrong to promote overriding [screen media query breakpoints](https://tailwindcss.com/docs/screens#max-width-breakpoints), I think you should add it without overriding default media queries. If people override custom media queries and find cool components online, they copy-paste the classes and all breakpoints are wrong.

reinink:
@minimit Yeah, it really depends on your own app whether you want to replace the default breakpoints, or add additional ones 👍

PR #7369, opened 2022-02-09
Add a `from` when postcss parsing preflight

### LINKED ISSUE #7583: Tailwind preflight doesn't produce sourcemap file deterministically

v3.0.11

react-scripts 5.0

v16.13.2

N/A

macOS

**Describe your issue**

When tailwind preflight is processed by postcss, it does not pass a `from` parameter to `postcss.parse`, so sourcemaps generated by postcss get a random `from` value for every build. This makes sourcemaps produced by postcss non-deterministic.

thyming:
Hi, I was tracking down today why builds for some of our apps were not deterministic. I believe it's because no `from` is provided to `postcss.parse`, so it generates a random ID every compilation. I am out of my depth with frontend tooling, so I wasn't sure how this `from` value is used and if what I've done is appropriate. Regardless, I believe it should be something stable and non-null.

See https://github.com/postcss/postcss/issues/1512 and https://github.com/tailwindlabs/tailwindcss/pull/3356 for a related issue and discussion.

Fixes #7583

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -209,7 +209,10 @@ let cssBackdropFilterValue = [
 export let corePlugins = {
   preflight: ({ addBase }) => {
     let preflightStyles = postcss.parse(
-      fs.readFileSync(path.join(__dirname, './css/preflight.css'), 'utf8')
+      fs.readFileSync(path.join(__dirname, './css/preflight.css'), 'utf8'),
+      {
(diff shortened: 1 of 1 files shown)


thecrypticace:
Thanks! Gonna take a look at this either today or tomorrow as I'll be making a few fixes to source map generation.

thyming:
Awesome, thanks for the fix.

PR #7412, opened 2022-02-10
Expose `context.sortClassList(classes)`

RobinMalfait:
This PR will expose `sortClassList` on the context

This function will be used by the `prettier-plugin-tailwindcss` plugin, this way the sorting happens within Tailwind CSS itself adn the `prettier-plugin-tailwindcss` plugin doesn't have to use internal / private APIs.

The signature looks like this:
```ts
function sortClassList(classes: string[]): string[]
```

E.g.:
```js
let sortedClasses = context.sortClassList(['p-1', 'm-1', 'container'])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-Nothing yet!
+### Added
+
+- Expose `context.sortClassList(classes)` ([#7412](https://github.com/tailwindlabs/tailwindcss/pull/7412))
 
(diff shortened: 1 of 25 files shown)


thecrypticace:
Can you provide details on how it broke?

thecrypticace:
nvm I think I can see the problem 🤦‍♂️

dargmuesli:
Sure, linting fails https://github.com/maevsi/maevsi/runs/5148080329?check_suite_focus=true#step:5:1288

thecrypticace:
Fixed! Just published v3.0.21. We added the prettier plugin for internal usage so our tests are consistent and it was mistakenly put in `dependencies`. Sorry about that and thanks for the report!

PR #7478, opened 2022-02-15
Don't duplicate layer output when scanning content with variants + wildcards

### LINKED ISSUE #7204: Bug: some text content can trigger multiple Tailwind builds in the same output css

v3.0.16 (also saw on v3.0.15)

Just Tailwind.

v16.13.0

N/A

macOS

**Reproduction URL**

https://github.com/matthewcc/tailwind3-duplicates

**Describe your issue**

Modifiers (such as pseudo-classes) followed by an asterisk appear to trigger additional instances of "tailwindcss v3.x.x" being inserted into the outputted css file.

In our case, we had things like `sm:*` and `focus:*` as text included in a Storybook mdx file as documentation. This mdx file also has jsx with actual Tailwind classes, so we need it added to the `content` property in the Tailwind config. This triggered some extra builds (located in the same css output file), and aside from just extra lines some of our own styles lost priority to re-inserted default Tailwind styles. There are additional notes in the Read Me in the reproduction link.

We've changed the text that triggered this issue in our files in the meantime and it's possibly an edge case, but we wanted to share it.

Big fans of Tailwind, thanks for your work.

thecrypticace:
Tailwind CSS treats `*` as special internally. Because of this if it scanned content which contained something like `focus:*` it would then generate duplicate CSS — including base layers. The best way to fix for this would be to use a `Symbol` in place of the special character — however the stringy-ness of class candidates is fairly well baked into assumptions across the codebase. Instead, by using `new String(…)`, we get an identical API surface to normal strings along with the behavior of something similar to a Symbol as identity comparisons against `"*"` will fail since we have a unique, non-interned object.

Fixes #7204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove opacity variables from `:visited` pseudo class ([#7458](https://github.com/tailwindlabs/tailwindcss/pull/7458))
 - Support arbitrary values + calc + theme with quotes ([#7462](https://github.com/tailwindlabs/tailwindcss/pull/7462))
+- Don't duplicate layer output when scanning content with variants + wildcards ([#7478](https://github.com/tailwindlabs/tailwindcss/pull/7478))
 
 ## [3.0.22] - 2022-02-11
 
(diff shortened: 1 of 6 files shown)


RobinMalfait (approved):
Yep the uniqueness of `new String()` is quite clever.
My only concern would be if you have your own custom rules that look something like:

```css
    @tailwind base;
    @tailwind components;
    @tailwind utilities;

    * {
      color: red;
    }

    .combined,
    * {
      text-align: center;
    }

    @layer base {
      * {
        color: blue;
      }
      .combined,
      * {
        color: red;
      }
    }
```

But it seems like that works as expected 👍

thecrypticace:
Ah, that's a good test to add though. I'll add that and then merge this in!

PR #7524, opened 2022-02-18
Use local user css cache for apply

### LINKED ISSUE #7480: HMR is really slow when there are a lot of Vue components with `<style>`

3.0.22

vite 2.8.2

16.13.0

Chrome

macOS

**Reproduction URL**

With `<style>` tags (slow): https://github.com/hivokas/laravel-vue-vite-tailwind-slow/tree/with-style-tags
With `<style>` tags, without Tailwind (fast): https://github.com/hivokas/laravel-vue-vite-tailwind-slow/tree/without-tailwind
Without `<style>` tags, with Tailwind (fast): https://github.com/hivokas/laravel-vue-vite-tailwind-slow/tree/without-style-tags

How to run the reproduction:
1. Clone the repo
2. Run `npm install`
3. Run `composer install`
4. Open `laravel-vue-vite-tailwind-slow.test` if you're using Valet
5. Run `npm run dev` to start the Vite dev server 
6. Switch between `with-style-tags`, `without-tailwind` and `without-style-tags` to see the difference 

**Describe your issue**

HMR works really slow when there are a lot of Vue components with `<style>`.

Here is the detailed demo:

https://user-images.githubusercontent.com/22997803/154132172-ba7933d6-2693-499a-bf34-90e2b0877dbb.mp4

thecrypticace:
## Background

The PR #6580 caused any file using apply to create a new context. This meant that any file that used apply would also carry along the entire content dependency tree. Updating any file covered by `content` would cause these files to recompile. The original cause of the problem that #6580 was fixing turned out to be a cache invalidation issue. Since user CSS was added to the context things could end up incorrectly shared across files when it comes to apply. However, this would only happen once and future updates would not invalidate the cache. This would manifest as an error about the applied classes being missing in some cases.

## Why did it seem to fix the original problem?

By creating a new context for any file containing apply it caused the context to get re-created every time that file was saved. Because of this the CSS was re-scanned and added to the internal user CSS cache that we had on the context.

## What are we going to do instead

1. I've partially reverted #6580 to ensure that contexts are reused when using @apply. The reason this is safe is because the config can affect apply but content files *cannot*.
2. I've added an additional cache for apply to look at the root postcss node that's being passed in instead of internally storing user CSS as individual tailwind plugins.

**This is still a work in progress. Some of the code here needs to be cleaned up first.**

Fixes https://github.com/tailwindlabs/tailwindcss/issues/7480


src/lib/expandApplyAtRules.js
```diff
@@ -35,6 +37,94 @@ function prefix(context, selector) {
...
+
+/**
+ * @param {import('postcss').Root} root
+ */
+function buildLocalApplyCache(root, context) {
+  /** @type {ApplyCache} */
+  let cache = new Map()
+
+  let reservedBits = 0n
+  let tmp = context.layerOrder.utilities >> 3n
```
RobinMalfait, line 88: Question, what's the reason for using `context.layerOrder.utilities >> 3n` instead of `context.layerOrder.user` instead?

Because I think the `>> 3n` was used because of this code:
```js
  context.layerOrder = {
    defaults: (1n << reservedBits) << 0n,
    base: (1n << reservedBits) << 1n,
    components: (1n << reservedBits) << 2n,
    utilities: (1n << reservedBits) << 3n, // The inverse of this part
    user: (1n << reservedBits) << 4n,
  }
```

and I wonder if it is save / better to just use the `context.layerOrder.user` since you want to put everything in that bucket.

RobinMalfait: It's true that this number will be larger but I think that is fine. Eventually we still want to do the heuristic based sorting and then we just require the order of each layer and within that layer we can sort based on heuristic which should eventually clean up a lot of the BigInt code.

tests/apply.test.js
```diff
@@ -1328,3 +1328,54 @@ it('should be possible to use apply in plugins', async () => {
     `)
   })
 })
+
+it('The apply class cache is invalidated when rules change', async () => {
```
RobinMalfait, line 1332: Nitpick: can you make this read like a sentence `it('should ...')` or `test('the apply class...')`

src/lib/expandApplyAtRules.js
```diff
@@ -35,6 +37,94 @@ function prefix(context, selector) {
...
+
+  node.nodes = children
+
+  return tmp
+}
+
+/**
+ * @param {import('postcss').Node} node
+ */
+function structuralCloneOfNode(node) {
```
RobinMalfait, line 62: I get what this is doing (cloning a node) but I think that it might be a bit misleading because if I read this correctly it also walks up the tree and eventually returns a the node (if it is already the root) or a parent of the node.

I don't immediately know a better name though 🤔

RobinMalfait (changes requested):
Functionally it seems to work great, nice!
I do have some questions / nitpicks.

src/lib/expandApplyAtRules.js
```diff
@@ -35,6 +37,94 @@ function prefix(context, selector) {
...
+  let cache = new Map()
+
+  let reservedBits = 0n
+  let tmp = context.layerOrder.utilities >> 3n
+  while (tmp > 1n) {
+    tmp = tmp >> 1n
+    reservedBits++
+  }
+
+  let highestOffset = 1n << reservedBits
```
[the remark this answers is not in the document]
thecrypticace: 👀 yes you're totally correct 🤦

src/lib/expandApplyAtRules.js
thecrypticace: Yeah, I couldn't think of a better name either. Would definitely welcome something better. It's purpose is to clone the singular, nested version of a node that can be used to represent it.

So, if we have the source:
```css
@media (min-width: 768px) {
  @supports (display: grid) {
    .foo {
      display: grid;
      grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
    }
  }

  @supports (backdrop-filter: blur(1px)) {
    .bar {
      backdrop-filter: blur(1px);
    }
  }

  .baz {
    color: orange;
  }
}
```

And we're currently looking at `.bar` it'll return a cloned version of what's required for just that single node:
```css
@media (min-width: 768px) {
  @supports (backdrop-filter: blur(1px)) {
    .bar {
      backdrop-filter: blur(1px);
    }
  }
}
```

src/lib/expandApplyAtRules.js
thecrypticace: I don't remember why I was using utilities instead of `user`. I think it was because there was an ordering issue in the first iteration but that's no longer a problem now. Updated.

src/lib/expandApplyAtRules.js
thecrypticace: idk maybe `nestedClone`?

RobinMalfait: Yeah just `nestedClone` is fine, not perfect but will do the job 👍

thecrypticace: Okay, I went with `nestedClone`. I still don't like it but it feels less ridiculous.

RobinMalfait (approved):

hivokas:
@thecrypticace thanks for working on that!

PR #7526, opened 2022-02-18
remove address italics in preflight

cbejensen:
I realize this is very easy to do in your own config, but it seems like a reasonable default to not have addresses italicized.

### DIFF
--- a/src/css/preflight.css
+++ b/src/css/preflight.css
@@ -136,6 +136,14 @@ sup {
 }
 
 /*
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

As you mentioned, I think for now it is better to keep this in your own css since this is technically a breaking change that we can't just introduce. Therefore going to close this for now and we can revalidate this in the future!

PR #7532, opened 2022-02-19
chore(stub): add inherit and screen defaults height/width

j-mendez:
* add inherit to height and width
* add screen to min-width max-width
Todo:
Update docs


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I think we can also add this value `inherit` but potentially others to way more plugins than the ones listed in this PR. However, I'm not sure if we want to do this by default.

I would recommend to add them to your own `tailwind.config.js` file if you need them in the meantime.

PR #7588, opened 2022-02-22
Preserve source maps for generated CSS

### LINKED ISSUE #5043: Enabling JIT breaks sourcemaps in webpack + PostCSS project

2.2.6

webpack 5.46.0, postcss 8.3.6

14.17.0

Chrome

Ubuntu

### Reproduction repository

⠀

### Describe your issue

When using Tailwind's JIT option for processing PostCSS in the webpack project, source maps are not being shown in the Chrome DevTools (but still generated in the output CSS file). The value of `devtool` option in the webpack config doesn't matter, the result is the same.

With config like this:
```
  purge: [
    './src/**/*.tsx',
  ],
  mode: 'jit',
```
Source maps don't get processed:
Screenshot from 2021-07-22 16-36-57

When disabling the JIT option:
```
  purge: false,
```
Source maps are being processed properly, and the original file is shown in Chrome DevTools:
Screenshot from 2021-07-22 16-30-51

I'll bootstrap a repro repo if a more thorough investigation is needed.

### LINKED ISSUE #7583: Tailwind preflight doesn't produce sourcemap file deterministically

thecrypticace:
Source maps were not preserved for CSS in many cases because:
1. We needed to recursively associate CSS nodes with the appropriate input source
2. Defaults expansion didn't preserve source maps
3. Apply didn't correctly override source maps leading to generated rules having no associated source
4. Some but not all preflight rules didn't have source maps because of the recursive copying problem.

This should fix all of those and we now have tests to verify that source maps are properly generated as well.

Apply source maps could be better. For instance it may make sense that the individual rules generated for `@apply text-black font-mono` would point to the columns for `text-black` and `font-mono` respectively. However, that is a lot of bookkeeping for not much gain. Having them all point to the apply line that generated them seems like a reasonable tradeoff.

Fixes #7583
Fixes #5043
Closes #7369


evenfrost:
@thecrypticace do you have an idea when it can be released?

thecrypticace:
We moved some stuff around today just so we could get a patch release out with a whole slew of bug fixes! You can upgrade to v3.0.24!

PR #7665, opened 2022-02-25
Add `rgb` and `hsl` color helpers for CSS variables

### LINKED ISSUE #3953: Suggestion: improve support for color css variables
### What version of @tailwindcss/jit are you using?

0.1.6

14

all

linux

### Reproduction repository

no need

Today with tailwind, it is not that easy to use the colors stored in a css variable.
If we do this `--color-primary: theme(colors.blue.500);`,  `--color-primary: #hexcolor;`,   
 `--color-primary: rgb (255,255,255)` or other color format,  we definitely lose the ability to add color opacity.  

So, we are forced to store the color in an unconventional way: `--color-primary: 59, 130, 246;`.
This brings some disappointment: hand conversion, no visual on colors in IDEs  

Imgur

I think it will be interesting to have a function available that allows at build time to convert the most common color format to one that takes advantage of opacity. Like that  

Imgur
```css
/* result to */
--color-primary: 59, 130, 246;
```
- After that, in the same way, we can declare the color in tailwind.config

```js
// ...
colors: {
                yellow: colors.orange,
                primary: varToRgba('--color-primary'),
}
// ...
```

- And / or, as in a recent similar library, proposing to create classes on the fly with a syntax like this: `border-$color-primary` (with opacity support of course)

Thanks a lot for your work!

thecrypticace:
This adds two new helpers that can be used to simplify using colors that are CSS variables. The colors MUST use the space separated syntax, and MUST NOT use an alpha channel in their value as Tailwind CSS adds that itself. The user can customize the alpha by using the opacity modifier syntax for colors (e.g. the `/50` in `text-primary/50`).

The `rgb` and `hsl` functions only expect the name of a CSS variable / custom property at present. We may revisit this in the future but this will give those who are using CSS variables a nicer, terse syntax for defining their config in a way that works with placeholders, borders, rings, etc…

```js
// tailwind.config.js
module.exports = {
  theme: {
    colors: ({ rgb, hsl }) => ({
      primary: rgb('--color-primary'),
      secondary: hsl('--color-secondary'),
      // ...
    })
  }
}
```

Closes #4659
Fixes #3953

### DIFF
--- a/tests/opacity.test.js
+++ b/tests/opacity.test.js
@@ -207,7 +207,7 @@ it('can use rgb helper when defining custom properties for colors (opacity plugi
   }
 
   return run('@tailwind utilities', config).then((result) => {
-    expect(result.css).toMatchCss(css`
+    expect(result.css).toMatchFormattedCss(css`
       .divide-primary > :not([hidden]) ~ :not([hidden]) {
         border-color: rgb(var(--color-primary) / 1);
       }
--- a/tests/opacity.test.js
+++ b/tests/opacity.test.js
@@ -95,3 +95,335 @@ test('colors defined as functions work when opacity plugins are disabled', () =>
     `)
   })
 })
+
+it('can use rgb helper when defining custom properties for colors (opacity plugins enabled)', () => {
+  let config = {
+    content: [
+      {
+        raw: html`
+          <div class="divide-primary"></div>
+          <div class="divide-primary divide-opacity-50"></div>
+          <div class="border-primary"></div>
+          <div class="border-primary border-opacity-50"></div>
+          <div class="bg-primary"></div>
+          <div class="bg-primary bg-opacity-50"></div>
+          <div class="text-primary"></div>
+          <div class="text-primary text-opacity-50"></div>
+          <div class="placeholder-primary"></div>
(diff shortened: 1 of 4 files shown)


bakura10:
This is an awesome addition. Would it be possible to add a second parameter to the RGB specifying the default opacity? This way it would allow us to define variables like this:

```
'border': rgb('--text-color', 0.15),
```

Thanks!

tests/opacity.test.js
```diff
@@ -95,3 +95,335 @@ test('colors defined as functions work when opacity plugins are disabled', () =>
...
+      }
+      .bg-primary {
+        --tw-bg-opacity: 1;
+        background-color: rgb(var(--color-primary) / var(--tw-bg-opacity));
+      }
+      .bg-opacity-50 {
+        --tw-bg-opacity: 0.5;
+      }
+      .text-primary {
+        --tw-text-opacity: 1;
```
bakura10, line 150: Is this actually expected? Shouldn't this generate:

```
color: rgb(var(--color-primary) / 1);
```

or even simpler:

```
color: rgb(var(--color-primary));
```

PR #7926, opened 2022-03-23
Types: allow for arbitrary theme values (for 3rd party plugins)

RobinMalfait:
This PR provides more improvements to the types config.

- Allows for arbitrary objects in the `theme` section for third party plugins.
- ~~Provides code completion for the `theme()` helper function~~


anosadal:
Looks like your django.root has a trailing slash on it, which is mentioned in the documentation as not the way to go.
If that's not the case, please print out what req.uri is in that case. I can't immediately see why that should be happening and your fix isn't the right approach (it's symptom patching, not fixing the problem).

RobinMalfait:
@anosadal everything alright buddy?

PR #7933, opened 2022-03-24
Allow for custom properties in `rgb`, `rgba`, `hsl` and `hsla` colors

### LINKED ISSUE #7909: arbitrary value for color needs hint but should arguably be detectable without it.

For example: v3.0.23

For example: Play

For example: N/A

For example: Chrome

For example: macOS

**Reproduction URL**

https://play.tailwindcss.com/Flb7YuMeyy

**Describe your issue**

You need to do `bg-[color:hsl(var(--h),50%,50%)]` when `bg-[hsl(var(--h),50%,50%)]` should arguably suffice.

RobinMalfait:
This PR will allow for custom properties to exist inside the `rgb`, `rgba`, `hsl` and `hsla` color functions.

This is useful if you want to do things like this:
```html
<div class="bg-[hsl(var(--foo),50%,50%)]"></div>
```

Previously this didn't work and wouldn't generate anything because we didn't correctly parse this as a color. You could force it by using a color type hint:
```html
<div class="bg-[color:hsl(var(--foo),50%,50%)]"></div>
```

The only issue with this is that it doesn't compose well with the opacity modifiers. This should now be fixed, where this:
```html
<div class="bg-[hsl(var(--foo),50%,50%)]/75"></div>
```
... resolves to:
```css
.bg-\[hsl\(var\(--foo\)\2c 50\%\2c 50\%\)\]\/75 {
  background-color: hsl(var(--foo) 50% 50% / 0.75);
}
```
While the class itself looks strange, it is valid and works as aspected: https://play.tailwindcss.com/ImyUO7n2XU?file=css

Fixes: #7909

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix extraction from template literal/function with array ([#7481](https://github.com/tailwindlabs/tailwindcss/pull/7481))
 - Don't output unparsable arbitrary values ([#7789](https://github.com/tailwindlabs/tailwindcss/pull/7789))
 - Fix generation of `div:not(.foo)` if `.foo` is never defined ([#7815](https://github.com/tailwindlabs/tailwindcss/pull/7815))
+- Allow for custom properties in `rgb`, `rgba`, `hsl` and `hsla` colors ([#7933](https://github.com/tailwindlabs/tailwindcss/pull/7933))
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


bakura10:
Hi @RobinMalfait ,

I am not sure if this is because I am using a prefix (I found a lot of issues when using prefix), but on the latest insider version, this does not work and no CSS is generated:

```
tw-bg-[rgb(var(--background))]
tw-text-[rgb(var(--text-color))]
```

I am still required to use the color prefix as shown below:

```
tw-bg-[color:rgb(var(--background))]
tw-text-[color:rgb(var(--text-color))]
```

Is there a known incompatibility with prefix?

bakura10:
I am opening an issue as it will be easier to track :)

PR #7955, opened 2022-03-26
Add CSS `content-visibility` utilities

lukewarlow:
This PR adds [`content-visibility`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) utilities.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -2367,6 +2367,14 @@ export let corePlugins = {
     })
   },
 
+  contentVisibility: ({ addUtilities }) => {
+    addUtilities({
+      '.content-visibility-visible': { 'content-visibility': 'visible' },
+      '.content-visibility-hidden': { 'content-visibility': 'hidden' },
+      '.content-visibility-auto': { 'content-visibility': 'auto' },
+    })
+  },
+
   transitionProperty: ({ matchUtilities, theme }) => {
     let defaultTimingFunction = theme('transitionTimingFunction.DEFAULT')
     let defaultDuration = theme('transitionDuration.DEFAULT')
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Right now it looks like the browser support for this is pretty limited: 

https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility#browser_compatibility

I think it's technically safe to use this even in browsers that don't support it since it's mostly a performance optimization technique if I'm understanding correctly, but to add it to core means making permanent decisions about the API, and I think I'd rather push-off those decisions until there's better support.

In the mean time though you can still use this in your own projects pretty easily using the  arbitrary properties feature we shipped with v3.0:

```html
<div class="[content-visibility:hidden]">
  <!-- ... -->
</div>
```

Thanks either way, looking forward to adding this to core in the future when the browser support is better and we have a clearer picture of exactly what the API should be! ❤️

yairEO:
As of October 2024 the support is:

```
Desktop ... 84.4% 
Global .... 76.69%
```

> looking forward to adding this to core in the future when the browser support is better

It is unspecified which usage-percentages are acceptable *"browser support is better"*.
I assume the good people who maintain Tailwind have a minimum usage threshold for which a CSS property should be considered as a candidate for addition in accordance to how "popular" that properly is used.

I also assume well-supported properties which are seldomly used are not good candidates for Tailwind, but this should also be measured somehow. 

So there are 2 numbers to consider: 
- global browser support percentage
- global usage (as percentage: "projects using the property"/"total projects" * 100)

https://caniuse.com/?search=content-visibility

PR #8040, opened 2022-04-04
Update `theme.fontSize` types

bradlc:
This PR updates the `theme.fontSize` types so so you can mix the different formats. Currently all of the values need to have the same format, e.g.

```js
// ✅
module.exports = {
  theme: {
    fontSize: {
      foo: '4rem',
      bar: '5rem',
    },
  },
}

// ✅
module.exports = {
  theme: {
    fontSize: {
      foo: ['4rem', '6rem'],
      bar: ['5rem', '7rem'],
    },
  },
}

// ❌
module.exports = {
  theme: {
    fontSize: {
      foo: '4rem',
      bar: ['5rem', '7rem'],
    },
  },
}
```

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -158,18 +158,18 @@ interface ThemeConfig {
   textIndent: ThemeConfig['spacing']
   fontFamily: ResolvableTo<KeyValuePair<string, string[]>>
   fontSize: ResolvableTo<
-    | KeyValuePair<string, string>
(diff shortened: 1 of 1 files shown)


RobinMalfait (approved):
Hah, I totally forgot that you could mix those values and that you don't have to choose between one syntax or another for _everything_. This makes sense, thanks!

PR #8201, opened 2022-04-26
Improve type detection for arbitrary color values

### LINKED ISSUE #8171: Color custom properties still require hint

latest insiders (21st April 2022)

This PR added the ability to use rgb/hsl custom properties: https://github.com/tailwindlabs/tailwindcss/pull/7933

This unfortunately does not work with prefixes.

The following class does not generate anything: ```tw-bg-[rgb(var(--background))]``` and still requires a color: hint.

bradlc:
Fixes #8171 

This PR improves type detection for arbitrary color values. Previously the following classes would not be generated because the type of the values could not be determined:

```
bg-[rgb(var(--bg-color))]
bg-[rgb(var(--red-green),40)]
bg-[rgb(12,var(--green-blue))]
```

Instead you would have to provide a type hint:

```
bg-[color:rgb(var(--bg-color))]
bg-[color:rgb(var(--red-green),40)]
bg-[color:rgb(12,var(--green-blue))]
```

The hint should not be necessary here because we can tell the values are colors based on the use of `rgb()`.

This PR adds a `loose` mode to the `parseColor` function that allows colors with fewer than 3 parts (like the above examples) as long as at least one of the parts is variable. We then use this parsing mode when looking for colors in arbitrary values.

### DIFF
--- a/src/util/color.js
+++ b/src/util/color.js
@@ -8,13 +8,15 @@ let ALPHA_SEP = /\s*[,/]\s*/
 let CUSTOM_PROPERTY = /var\(--(?:[^ )]*?)\)/
 
 let RGB = new RegExp(
-  `^rgba?\\(\\s*(${VALUE.source}|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
+  `^(rgb)a?\\(\\s*(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
 )
 let HSL = new RegExp(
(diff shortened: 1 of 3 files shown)


src/util/color.js
```diff
@@ -8,13 +8,15 @@ let ALPHA_SEP = /\s*[,/]\s*/
 let CUSTOM_PROPERTY = /var\(--(?:[^ )]*?)\)/
 
 let RGB = new RegExp(
-  `^rgba?\\(\\s*(${VALUE.source}|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
+  `^(rgb)a?\\(\\s*(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
 )
 let HSL = new RegExp(
-  `^hsla?\\(\\s*((?:${VALUE.source})(?:deg|rad|grad|turn)?|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
+  `^(hsl)a?\\(\\s*((?:${VALUE.source})(?:deg|rad|grad|turn)?|${CUSTOM_PROPERTY.source})(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
 )
```
bradlc, line 15: The second and third color parts have been made optional, and the amount of parts is validated later instead (line 55)

RobinMalfait (approved):
Oh beautiful, seems to work as expected even if you provide multiple `var()` values.

PR #8248, opened 2022-05-01
Remove default `[hidden]` style in preflight

stefanr:
As discussed in https://github.com/tailwindlabs/tailwindcss/discussions/8247


thecrypticace:
Thanks @stefanr! 🎉

PR #8299, opened 2022-05-07
Arbitrary variants

RobinMalfait:
With the new `addVariant` API, we have a beautiful way of creating new variants.

You can use it as:
```js
addVariant('children', '& > *')
```

Now you can use the `children:` variant. The API uses a `&` as a reference for the candidate, which means that:
```html
children:pl-4
```

Will result in:
```css
.children\:pl-4 > * { .. }
```

Notice that the `&` was replaced by `.children\:pl-4`.

We can leverage this API to implement arbitrary variants, this means that you can write those `&>*` (Notice that we don't have spaces) inside a variant directly. An example of this can be:
```html
<ul class="[&>*]:underline">
  <li>A</li>
  <li>B</li>
  <li>C</li>
</ul>
```

Which generates the following css:
```css
.\[\&\>\*\]\:underline > * {
}
```

Now all the children of the `ul` will have an `underline`. The selector itself is a bit crazy since it contains the candidate which is the selector itself, it is just escaped.

### DIFF
--- a/src/lib/defaultExtractor.js
+++ b/src/lib/defaultExtractor.js
@@ -19,7 +19,7 @@ export function defaultExtractor(content) {
 function* buildRegExps() {
   yield regex.pattern([
     // Variants
-    /((?=([^\s"'\\]+:))\2)?/,
+    /((?=((\[[^\s"'\[\]\\]+\]:|[^\s"'\[\\]+:)+))\2)?/,
 
     // Important (optional)
     /!?/,
--- a/src/lib/defaultExtractor.js
+++ b/src/lib/defaultExtractor.js
@@ -1,25 +1,34 @@
 import * as regex from './regex'
 
-let patterns = Array.from(buildRegExps())
+export function defaultExtractor(context) {
+  let patterns = Array.from(buildRegExps(context))
 
-/**
- * @param {string} content
- */
-export function defaultExtractor(content) {
-  /** @type {(string|string)[]} */
-  let results = []
+  /**
+   * @param {string} content
+   */
+  return (content) => {
+    /** @type {(string|string)[]} */
+    let results = []
 
-  for (let pattern of patterns) {
-    results.push(...(content.match(pattern) ?? []))
+    for (let pattern of patterns) {
+      results.push(...(content.match(pattern) ?? []))
+    }
+
+    return results.filter((v) => v !== undefined).map(clipAtBalancedParens)
   }
-
-  return results.filter((v) => v !== undefined).map(clipAtBalancedParens)
 }
 
-function* buildRegExps() {
+function* buildRegExps(context) {
+  let separator = context.tailwindConfig.separator
+
   yield regex.pattern([
     // Variants
-    /((?=((\[[^\s"'\\]+\]:|[^\s"'\[\\]+:)+))\2)?/,
+    '((?=((',
+    regex.any(
+      [regex.pattern([/\[[^\s"'\\]+\]/, separator]), regex.pattern([/[^\s"'\[\\]+/, separator])],
+      true
+    ),
+    ')+))\\2)?',
 
     // Important (optional)
     /!?/,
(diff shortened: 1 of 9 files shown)


src/lib/defaultExtractor.js
```diff
@@ -19,7 +19,7 @@ export function defaultExtractor(content) {
 function* buildRegExps() {
   yield regex.pattern([
     // Variants
-    /((?=([^\s"'\\\[]+:))\2)?/,
+    /((?=([^\s"'\\]+:))\2)?/,
```
RobinMalfait, line 22: This will require some work to make it exactly right

thecrypticace: lol yep I was thinking about this. I'll throw this change against my new candidate refactor to see what breaks…

I had already tested support for arbitrary variant syntax (at least detection wise) in the form of [foo]:bar

thecrypticace: ngl I love this idea so much

thecrypticace: I pushed a regex tweak to fix the tests 👍👍

simonswiss:
This is rad! 🤙

onmax:
Wow, just requested this feature about one month ago and now is coming here. Timing couldn't be better.  I have a few ideas in my head to create some plugins with this.

https://github.com/tailwindlabs/tailwindcss/discussions/8157

Keep up the hard work 🔥 🔥.  Amazing!

adamwathan:
Added a failing test for attribute selectors like:

`class="[&:[data-open]]:underline"`

I think this one is going to be kind of tricky — we need to reliable extract that candidate from the template but to do that we need to worry about bracket balancing. We'll need to support multiple pairs too so the implementation probably can't be as naive as I'd wish:

`class="[&:[data-foo][data-bar]:has([data-baz])]:underline"`

jipika:
Is it possible to specify a class name? Sometimes it is not possible to change all of them

calvo-jp:
woah! how cool :rocket:

ivanchenhz:
Cool, 👍
And seems we don't support class selector?!

[&_i]:text-pink-100      👌
[&_i.some_class]:text-pink-100     ...

RobinMalfait:
@ivanchenhz we do, but `_` is used for spaces, so you essentially have this selector:
```css
i .some class {}
```
You have to escape the `_` if you want to keep it as an actual underscore. See: https://tailwindcss.com/docs/hover-focus-and-other-states#using-arbitrary-variants

ivanchenhz:
@RobinMalfait  thanks for confirming, and yes it works, 👍 😄
And I finally figure it out, the problem is we have prefix 'tw-'

- So, when I targeting class like ```$_i.tw-font-medium``` it generate ```i.tw-tw-font-medium```
- Should we considering DON'T count prefix in arbitrary mode !
Screen Shot 2022-06-13 at 10 04 25 PM


Screen Shot 2022-06-13 at 10 04 12 PM

ivanchenhz:
@RobinMalfait 
Got a real case for this, we like to restyling the 'popover-body' which comes from some library
After add ```[&_.popover-body]:tw-pr-2```, it auto prefix to generate ```tw-popover-body``` ( doesn't actually exist )
Have to manually add the class to make it work, ( which not feasible obviously  😂 ）
Please considering don't count prefix in arbitrary mode


Screen Shot 2022-06-13 at 11 07 57 PM

Emre-Zero:
Nice work!

I suppose can shorten to `ch`

In `tailwind.config.js`:
```javascript
const plugin = require("tailwindcss/plugin");
plugins: [
      plugin(function({ addVariant }) {
        addVariant('ch', '& > *')
      })
]
```

clintfisher:
Having issues with Arbitrary variants not showing in Intellisense:
https://github.com/tailwindlabs/tailwindcss/discussions/9449

PR #8420, opened 2022-05-23
Use single-line syntax for private types comments

bradlc:
This prevents the comments from appearing in a user's editor.

**Before**

CleanShot 2022-05-23 at 20 38 44@2x

**After**

CleanShot 2022-05-23 at 20 43 15@2x

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -64,11 +64,11 @@ type ExperimentalConfig = Expand<'all' | Partial<Record<ExperimentalConfigValues
 
 // DarkMode related config
 type DarkModeConfig =
-  /** Use the `media` query strategy. */
+  // Use the `media` query strategy.
   | 'media'
-  /** Use the `class` stategy, which requires a `.dark` class on the `html`. */
+  // Use the `class` stategy, which requires a `.dark` class on the `html`.
   | 'class'
-  /** Use the `class` stategy with a custom class instead of `.dark`. */
+  // Use the `class` stategy with a custom class instead of `.dark`.
   | ['class', string]
 
 type Screen = { raw: string } | { min: string } | { max: string } | { min: string; max: string }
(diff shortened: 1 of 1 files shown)


types/config.d.ts
```diff
@@ -64,11 +64,11 @@ type ExperimentalConfig = Expand<'all' | Partial<Record<ExperimentalConfigValues
...
 // DarkMode related config
 type DarkModeConfig =
-  /** Use the `media` query strategy. */
+  // Use the `media` query strategy.
   | 'media'
-  /** Use the `class` stategy, which requires a `.dark` class on the `html`. */
+  // Use the `class` stategy, which requires a `.dark` class on the `html`.
   | 'class'
-  /** Use the `class` stategy with a custom class instead of `.dark`. */
+  // Use the `class` stategy with a custom class instead of `.dark`.
```
RobinMalfait, line 71: I think that for these it could still be useful to know what each type means. But maybe they don't show up at all 🤔

types/config.d.ts
```diff
@@ -242,15 +242,15 @@ type ValueType =
   | 'relative-size'
   | 'shadow'
 export interface PluginAPI {
-  /** for registering new static utility styles */
+  // for registering new static utility styles
```
RobinMalfait, line 245: These could also be useful to the user, or do you reckon that this is a bit redundant because the name `addUtilities` itself is self explanatory enough?

RobinMalfait (approved):
Looks good to me, just 2 questions

types/config.d.ts
bradlc: I agree that we could write some documentation comments that are useful to users, but it seems like the comments in this file right now have not been written for that purpose, and it was not the intention for these to be visible to users. Consider comments like `Utilities` or `Responsiveness` for example. I propose that we hide everything for now and consider adding proper documentation comments in the future.

RobinMalfait: Yep sounds good, let's merge it 💪

PR #8763, opened 2022-07-03
Support font-weight in font-size utilities

tmkx:
`font-weight` comes with a typography system often.


thecrypticace:
Thanks for the PR! This feature will go out in our next release (should be later this week) — you can test it ahead of time by installing our insiders build `npm install tailwindcss@insiders` which _should_ be published in a few minutes.

PR #8780, opened 2022-07-05
Add more explicit types for the default theme

### LINKED ISSUE #8728: Type definitions for defaultTheme is incorrectly too generic

v3.1.4

Next.js v12.1.6

v16.13.2

N/A

Windows 11 Pro

**Reproduction URL**

https://www.typescriptlang.org/play?filetype=js#code/PTAEAEBcGcFoGMAWBTeBrAUB+B7AdtJKACbIBmAhgK4A2kAKigLbKgC8oATsgI5UCW3ABQBySBX40A7vzzF40aMFKVaDZshEBKANxZgAKgMRIATwAOrAN78m5nJ0ijxkmXIXRtAOgDC+MvwA5gC+oAbA2PiEoLh4AYHsoFYYAJCxkMh4kABcoADaIl7A0JzwhuEGXlYwADQwAB41HsEiALo1qZAauckpKcj1GXI9qX1k+JAAYhRMkqYjfX3QFAS5BQCSWcicIjUk5NR0jMgsXuNZ07M0pl7LBO2jKcEdfc+pb8F6GEw4xLTIXgG9kc0ESsXieiAA

**Describe your issue**

Hello,

Since the latest update to Tailwind which includes built in TypeScript definitions, I have been encountering some trouble with the type definitions for `defaultTheme` which I did not encounter with the unofficial DefinitelyTyped type definitions. I use TypeScript for my project, but because Tailwind configs use JavaScript, I use an `@ts-check` comment in my config to enable type checking.

In my config, I want to replace the default sans-serif font with a custom font. A minimal config that does that looks like this:

```js
// @ts-check

const config = {
	content: ['./src/**/*.{ts,tsx,css}'],
	theme: {
		extend: {
			fontFamily: {
				sans: ['Inter', 'sans-serif'],
			},
		},
	},
};

module.exports = config;
```

While this works, I would ideally like to replace the `sans-serif` with the system font stack provided by Tailwind. This would be done like this:

```js
// @ts-check

const defaultTheme = require('tailwindcss/defaultTheme');

const config = {
	content: ['./src/**/*.{ts,tsx,css}'],
	theme: {
		extend: {
			fontFamily: {
				sans: ['Inter', defaultTheme.fontFamily.sans],
			},
		},
	},
};

module.exports = config;
```

Unfortunately, TypeScript throws a few errors when using `defaultTheme`. After looking at the type definitions, it appears that the issue is that the type of `defaultTheme` is based on the types used to annotate user configs, i.e. the same type definition referenced in line 5 of the second example. The issue with this is that the type of `defaultTheme` is too generic. While user configs do not need to include the theme property, `defaultTheme` is a defined object. In addition, `fontFamily` (and many other keys) are incorrectly optional in `defaultTheme`.

While I am not very well familiar with the current type definitions for Tailwind, my guess is that the solution to this issue would be to create a separate type definition for `defaultTheme`.

Thank you for adding TypeScript definitions to your framework, as they are very useful, and any help with this issue would be greatly appreciated.

thecrypticace:
Fixes #8728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don’t prefix selectors in arbitrary variants ([#8773](https://github.com/tailwindlabs/tailwindcss/pull/8773))
 - Add support for alpha values in safe list ([#8774](https://github.com/tailwindlabs/tailwindcss/pull/8774))
 - Support default `font-weight`s in font size utilities ([#8763](https://github.com/tailwindlabs/tailwindcss/pull/8763))
+- Add more explicit types for the default theme ([#8780](https://github.com/tailwindlabs/tailwindcss/pull/8780))
 
 ## [3.1.4] - 2022-06-21
 
--- a/defaultTheme.d.ts
+++ b/defaultTheme.d.ts
@@ -1,3 +1,4 @@
 import type { Config } from './types/config'
-declare const theme: Config['theme']
+import { DefaultTheme } from './types/generated/default-theme'
+declare const theme: Config['theme'] & DefaultTheme
 export = theme
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -53,14 +53,14 @@ fs.writeFileSync(
 )
 
 function typeCoveringAllValues(value) {
+  let union = (values) => [...new Set(values.map((v) => typeCoveringAllValues(v)))].join(' | ')
+
   if (Array.isArray(value)) {
-    let types = value.map(v => typeCoveringAllValues(v))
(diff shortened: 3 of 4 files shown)


RobinMalfait:
I wonder if we could generate types on the fly when we are in watch mode so that we also catch types from the user config (and deeply nested resolved configs) that would be kind of cool.

I got worried for a second that the types weren't using all the shapes they could use (e.g. fontSize allows multiple forms) and we only used a single record. But then I realized that this is "just" for the `tailwindcss/defaultTheme` import itself which I think is good right now.

For the functions we could try to resolve the default config first so that the functions are gone, but then the issue is that it can look like you can just do `defaultTheme.backdropBlur.lg` but that's not the case because of that function so right now I think this makes sense.

RobinMalfait (approved):
Looks good to me.

PR #8992, opened 2022-07-30
Remove class prefix in arbitrary variant that is used multiple times

wongjn:
### Discussed in #8932

Fixes the issue when a class prefix is configured and an arbitrary variant with a class name is used within the variant, the class within the variant would have the class prefix when the same arbitrary variant is used more than once.

Extension of PR #8773.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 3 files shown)


thecrypticace:
🤦 Thanks for the fix!

aUsABuisnessman:
> ### Discussed in #8932
> 
> Fixes the issue when a class prefix is configured and an arbitrary variant with a class name is used within the variant, the class within the variant would have the class prefix when the same arbitrary variant is used more than once.
> 
> Extension of PR #8773.
> 

Thank you sir!

PR #9032, opened 2022-08-05
Fix vite3 postcss statSync not found

jardenliu:
image
I use tailwind+vite3+tauri to build my application, and I will get the above error. It seems to be `cwd()` has changed when fs.Statsync is called. 
Using an absolute address can better ensure the correct path of the files


thecrypticace:
I haven't had a chance to play with Tauri yet so I used this as an excuse to fiddle around a bit. It seems super neat! That said, I can't reproduce this. Do you maybe have a repo I can look at? Tailwind and Vite both detect changes and don't throw errors:

Screen Shot 2022-08-05 at 09 58 27

thecrypticace:
Okay I've got a reproduction from the linked issue in the Tauri/solid/ts/tailwind starter repo. Gonna see if I can boil it down to a simpler repro so we can add a test for it and get this merged.

thecrypticace:
Okay, so I tracked this down and it appears that Tauri is actively changing the current working directory of the node process _outside of node_. This is a problem because the OS and Node disagree on what the current working directory for the process actually is:

Screen Shot 2022-08-08 at 10 32 22

I produced the above by tweaking the Tailwind CSS source:

Screen Shot 2022-08-08 at 10 32 58

We can sorta fix this by making sure resolvedChangeFiles always uses absolute paths but this has other implications. It can potentially break other things, even in Node JS, that assume that the current working directory of the node process is accurately returned by `process.cwd()` when it is not.

thecrypticace:
One of the problems with this is that this fix may not even work in the future if the internals of `fast-glob` change and step outside of node _at all_ to resolve paths. But at least for now it seems generally fine.

jardenliu:
First of all, thank you for merge my PR. Originally, I planned to create a new repo to show this situation. But because I'm busy recently, I don't have time to do it. 

Fortunately, you have found the cause of the problem. That's pretty great.

Lastofthefirst:
@thecrypticace I've run into this issue, how  can I make use of the merged fix?

PR #9054, opened 2022-08-08
fix cli output stream for performance message

excessivecoding:
I was fiddling with executing Tailwind CLI command from a script and realized the output was directed to `stderr` instead of the standard `stdout`.

This should fix it

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -660,8 +660,8 @@ async function build() {
         })
         .then(() => {
           let end = process.hrtime.bigint()
-          console.error()
-          console.error('Done in', (end - start) / BigInt(1e6) + 'ms.')
+          console.log()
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

This is done on purpose and is a common practice in CLIs, so that you can redirect the output to a file or another program for example:
```sh
npx tailwindcss --content ./index.html > out.css
```
If we write the informational messages to `stdout` as well, then that would result in 2 things:
1. The "Done" message wouldn't be shown in the terminal because it has been redirected
2. The `out.css` file would be invalid because it now includes `Done in 62ms.` at the end.

We don't have any plans to change this since this works as expected. Thanks for the PR though!

PR #9088, opened 2022-08-12
Add future flag to disable color opacity utility plugins

thecrypticace:
Since we have the opacity modifier syntax these are no longer necessary and this will become the default in Tailwind CSS v4.0

If you need to be able to control the opacity independently from the color channels you have two options:

1. Use the opacity modifier syntax with CSS variables (recommended):
```html
<div class="bg-blue-200/[var(--bg-opacity)] text-blue-900/[var(--text-opacity)]" style="--text-opacity: 0.75; --bg-opacity: 0.25;">Test</div>
```

2. Re-enable the appropriate `corePlugin` in your config file:
```js
corePlugins: {
  backgroundOpacity: true,
  textOpacity: true,
}
```


dcastil:
@thecrypticace I just noticed that ring opacity with classes like `ring-opacity-10` stopped working as well when setting `disableColorOpacityUtilitiesByDefault: true`. Is that intended? I don't know of any other way to set ring opacity than with those classes.

PR #9200, opened 2022-08-27
fix safelist variant key required by typings

acalvino4:
The typescript typings were requiring a `variant` key on the `safelist` object in the config. This pr fixes that.

Screen shot of false positive:
image
image

Docs that show `variant` key should not be required:
https://tailwindcss.com/docs/content-configuration#using-regular-expressions
image


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Good catch!

PR #9319, opened 2022-09-13
Don't emit utilities containing invalid theme fn keys

### LINKED ISSUE #9318: Tailwind compiler crashes when you try to `[--box-shadow:theme('boxShadow.doesNotExist')]`

For example: v3.1.8

For example: vite 3.0.9, postcss 8.4.14 postcss-cli 8.3.1

For example: v16.0.0

For example: Chrome

For example: PopOS, Ubuntu

**Reproduction URL**

https://stackblitz.com/edit/vitejs-vite-zy2efk?file=src/App.jsx

**Describe your issue**

I was thinkering with tailwind to find ways of modifying css variables with classes. Came up with something like

```javascript
  <button className="[--box-shadow:theme('boxShadow.sm')] bg-white p-5">
    Button
  </button>
```

But if I change the `theme` to something that will throw a runtime error:

```javascript
  <button className="[--box-shadow:theme('boxShadow.doesNotExist')] bg-white p-5">
    Button
  </button>
```

Still crashes if I change it back to:

```javascript
  <button className="[--box-shadow:theme('boxShadow.sm')] bg-white p-5">
    Button
  </button>
```

Well, maybe the Tailwind compiler crashes and can't recover?

It's a bug, I think!?

thecrypticace:
Fixes #9318

This fix isn't ideal — need to iterate on it some before merging. We shouldn't even generate the utility in the first place. The problem here is the utility and the evaluation of the `theme()` function are at two separate stages. This causes the utility to be stored in the map when it is otherwise technically invalid. As a perf optimization we don't clear out the utility list when "watching" — it's effectively append-only causing this to throw as soon as you accidentally introduce a utility w/ a theme key that doesn't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ isolate*.log
 
 # Generated files
 /src/corePluginList.js
+
+# Generated files during tests
+/tests/evaluate-tailwind-functions.test.html
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix ordering of parallel variants ([#9282](https://github.com/tailwindlabs/tailwindcss/pull/9282))
 - Handle variants in utility selectors using `:where()` and `:has()` ([#9309](https://github.com/tailwindlabs/tailwindcss/pull/9309))
 - Improve data type analyses for arbitrary values ([#9320](https://github.com/tailwindlabs/tailwindcss/pull/9320))
+- Don't emit generated utilities with invalid uses of theme functions ([#9319](https://github.com/tailwindlabs/tailwindcss/pull/9319))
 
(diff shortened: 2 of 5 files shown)


thecrypticace:
Hey @RobinMalfait how bad of an idea is this — or does this seem fine-ish?

RobinMalfait (commented):
Hmm, I see what you mean. Instead of silently ignoring I would expect an error/warning being thrown instead so that you have at least something to debug (or open an issue about).

I think in the past we've handled the resolving of the `theme` function with special cases for some plugins. Ideally we figure out a way to resolve the `theme` value value per-plugin but that's probably out of scope for this one.

I think what we can do here is:
1. Throw some form of error in the general case when theme function doesn't resolve to a proper (string?) value.
2. Handle the `boxShadow` properly if that's not the case already (doubt it).

I think we've used `throw node.error('Message')` before which should be recoverable (the process should not crash but an error should be shown in the terminal)

Issue #7997, opened 2022-03-30
Can't use % for background-position arbitrary value


dwkns:
**What version of Tailwind CSS are you using?**

3.0.23

**What build tool (or framework if it abstracts the build tool) are you using?**

postcss-cli 9.1.0

**What version of Node.js are you using?**

v17.7.2

**What browser are you using?**

Safari

**What operating system are you using?**

macOS

**Reproduction URL**
https://play.tailwindcss.com/dBpXEnLlgb

**Describe your issue**
Percentages are not output when using `bg-[ arbitrary values ]`
`bg-[top_right_75px]` → works just fine
`bg-[top_right_5rem]` → works just fine
`bg-[top_right_50%]` → doesn't output the expected CSS

Describe the problem you're seeing, any important steps to reproduce and what behavior you expect instead.
https://play.tailwindcss.com/dBpXEnLlgb

Update:
`bg-[top_right_calc(50%)]` → works just fine!!!!

RobinMalfait:
Hey! Thank you for your bug report!
Much appreciated! 🙏

This should be fixed by #8015, and will be available in the next release.

You can already try it by using the insiders build `npm install tailwindcss@insiders` or `yarn add tailwindcss@insiders`.

PR #9349, opened 2022-09-17
Improve type checking for formal syntax

### LINKED ISSUE #9352: `bg-[10px_10%]` is missing background-size type checking

v3.1.8 or Insiders (4fddd2d)

tailwind cli

v16

N/A

macOS

**Reproduction URL**

https://play.tailwindcss.com/tpx3OYFFsH

**Describe your issue**

As the Reproduction URL,
```html
<div bg-[10px_10%]></div>
<div bg-[10%_10%]></div>
<div bg-[10px_10px]></div>
```
can generate
```css
.bg-\[10px_10\%\] {
  background-position: 10px 10%;
}
```
which does not make sense, because all these values are valid  (background-size and background-position) so it's ambiguous.

I've created a PR to solve this problem. https://github.com/tailwindlabs/tailwindcss/pull/9349

lzt1008:
There are type validators in `dataType.js` referencing [CSS Types](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types), but splitting the basic data types([\<length\>](https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L62) and [\<percentage\>](https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L87) ) is not a good solution for further maintenance and development. We can just keep them as minimal validators for the css data type.

In my opinion, `familyName`  and `position` is something like [value definition syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax). And we should take them out as validators for formal syntax for better type checking, and `dataType.js` is only for basic css data types.

This PR created `validateFormalSyntax.js` and `size` type (or formal syntax) for better background-size checking based on [background-size MDN web docs](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size#formal_syntax).
https://github.com/tailwindlabs/tailwindcss/blob/b37a44a85d803ca353f299c1989c425874da9af5/src/util/validateFormalSyntax.js#L22
 and I think it's also a better solution for https://github.com/tailwindlabs/tailwindcss/issues/7997, which can also fix https://github.com/tailwindlabs/tailwindcss/issues/9352

Also, as [MDN web doc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) says, the `calc() min() max() clamp()` function can all be used anywhere a [\<length\>](https://developer.mozilla.org/en-US/docs/Web/CSS/length), [\<frequency\>](https://developer.mozilla.org/en-US/docs/Web/CSS/frequency), [\<angle\>](https://developer.mozilla.org/en-US/docs/Web/CSS/angle), [\<time\>](https://developer.mozilla.org/en-US/docs/Web/CSS/time), [\<percentage\>](https://developer.mozilla.org/en-US/docs/Web/CSS/percentage), [\<number\>](https://developer.mozilla.org/en-US/docs/Web/CSS/number), or [\<integer\>](https://developer.mozilla.org/en-US/docs/Web/CSS/integer) is allowed. So we can basically just check the value is wrapped by css functions in `number`, `length`, and `percentage` https://github.com/tailwindlabs/tailwindcss/blob/b37a44a85d803ca353f299c1989c425874da9af5/src/util/dataTypes.js#L9

So we can now create some crazy stuff like
```css
/* bg-[rgb(11,22,33)] */
.bg-\[rgb\(11\2c 22\2c 33\)\] {
  --tw-bg-opacity: 1;
  background-color: rgb(11 22 33 / var(--tw-bg-opacity))
}

/* bg-[linear-gradient(45deg,_red_0_50%,_blue_50%_100%)] */
.bg-\[linear-gradient\(45deg\2c _red_0_50\%\2c _blue_50\%_100\%\)\] {
  background-image: linear-gradient(45deg, red 0 50%, blue 50% 100%)
}

/* bg-[auto_auto,cover,_contain,calc(10px_+_12%),max(20px,_10%)_10%] */
.bg-\[auto_auto\2c cover\2c _contain\2c calc\(10px_\+_12\%\)\2c max\(20px\2c _10\%\)_10\%\] {
  background-size: auto auto,cover, contain,calc(10px + 12%),max(20px, 10%) 10%
}
```
discussions: https://github.com/tailwindlabs/tailwindcss/discussions/9351

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -1481,7 +1481,7 @@ export let corePlugins = {
   },
 
   backgroundSize: createUtilityPlugin('backgroundSize', [['bg', ['background-size']]], {
-    type: ['lookup', 'length', 'percentage'],
+    type: ['lookup', 'length', 'percentage', 'size'],
   }),
 
   backgroundAttachment: ({ addUtilities }) => {
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -6,6 +6,10 @@ let cssFunctions = ['min', 'max', 'clamp', 'calc']
 
 // Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types
 
+function isCSSFunction(value) {
+  return cssFunctions.some((fn) => new RegExp(`^${fn}\\(.*\\)`).test(value))
+}
+
 // This is not a data type, but rather a function that can normalize the
 // correct values.
 export function normalize(value, isRoot = true) {
@@ -55,13 +59,11 @@ export function url(value) {
 }
 
 export function number(value) {
-  return !isNaN(Number(value)) || cssFunctions.some((fn) => new RegExp(`^${fn}\\(.+?`).test(value))
+  return !isNaN(Number(value)) || isCSSFunction(value)
 }
 
 export function percentage(value) {
-  return splitAtTopLevelOnly(value, '_').every((part) => {
-    return /%$/g.test(part) || cssFunctions.some((fn) => new RegExp(`^${fn}\\(.+?%`).test(part))
-  })
+  return (value.endsWith('%') && number(value.slice(0, -1))) || isCSSFunction(value)
 }
 
 let lengthUnits = [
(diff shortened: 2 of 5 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

We recently introduced another fix where we can define fallback plugins in case we have colliding plugins where we guess what the most likely plugin is that should generate the output instead of generating nothing which requires you to add a typehint.

Because of this a few changes have been made to your PR here: https://github.com/tailwindlabs/tailwindcss/pull/9448

Thanks for the PR!

PR #9393, opened 2022-09-22
✨ Add `word-break: keep-all` with `break-keep`

k-utsumi:
ref. https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

`whitespace-nowrap` and `word-break: keep-all` behave differently in different browsers.

Demo: https://jsfiddle.net/h1aj6nvy/

There is a difference between Firefox and Google chrome.


reinink:
@k-utsumi Hey thanks for this contribution! Just chatted with the team about this, and while we like the addition, we are going to call this utility `.break-keep` instead of `.keep-all` to keep all the `word-break` utilities consistent 👍

k-utsumi:
Thanks 😆

PR #9405, opened 2022-09-23
Add `@config` support

### LINKED ISSUE #6517: Resolve config file relative to the file being transformed instead of cwd

v3.0.2

Parcel 2 👋 

16

macOS

**Describe your issue**

A "standard" PostCSS config like

```js
const path = require("path");

module.exports = {
  plugins: [
      require('tailwindcss')
  ]
}
```

causes Tailwind to search for the config file relative to the current working directory because of this

https://github.com/tailwindlabs/tailwindcss/blob/a7263a8f6faf989cb98553491d5c456d0b86de9b/src/util/resolveConfigPath.js#L46-L48

I think it should instead be relative to the file that's being processed. This way you could also have a single project with multiple config files for different parts.

This might have performance downsides if you currently search just once for a config.

thecrypticace:
The new `@config` rule lets you specify a config file relative to your CSS. This, additionally, lets you have multiple config files for separate CSS entry points — oftentimes found in monorepos with multiple packages.

This means that you can now do this and it will look for the config file next to the CSS file:
```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```

This even works with `postcss-import` if you have a `@config` directive in an imported file. In which case the config will be relative to that imported CSS file. If you do not have an `@config` directive we will still use the config file specified by the postcss plugin for that CSS file.

> **Warning**
> This **does not** work the way you may expect with imports + preprocessors like Sass/LESS/etc… as those generally compile to one blob of CSS that's passed to PostCSS so it will likely be relative to the Sass entry file instead.

Fixes #6517


JamieGarcia123:
So would this be ideal if we have one Code repo for mulitple clients but need different style sheets based on that company?

zommerberg:
> So would this be ideal if we have monorepo for mulitple clients but need different style sheets based on that company? for instance the color themes will be different from client to client

I am also interested about that, especially if we dynamically fetch client specific data (fonts/colors) after we have built our app

eYinka:
> > So would this be ideal if we have monorepo for mulitple clients but need different style sheets based on that company? for instance the color themes will be different from client to client
> 
> I am also interested in that, especially if we dynamically fetch client specific data (fonts/colors) after we have built and deployed the app

This is my use case at the moment. Now I am thinking if it's possible to store the path to each client's custom config in an ENV file and then dynamically load it somehow. Wild thoughts maybe?

fabswt:
link to the docs: https://tailwindcss.com/docs/functions-and-directives#config

PR #9496, opened 2022-10-06
style violations fix

DawoodKMasood:

### DIFF
--- a/tests/arbitrary-variants.test.js
+++ b/tests/arbitrary-variants.test.js
@@ -107,7 +107,7 @@ test('variants without & or an at-rule are ignored', () => {
 
 test('arbitrary variants are sorted after other variants', () => {
   let config = {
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hey, appreciate the PR but this isn't correct. You're seeing this as a side-effect of the prettier plugin not able to load the git version of Tailwind CSS because this repo _is_ Tailwind itself. There's a difference in the version that's baked into the prettier plugin as a fallback which obviously can't match the git version. Technically, NPM can work around this for us in v16 but not earlier because a package requiring itself via `file:.` causes `npm install` to crash. We fix this ourselves by creating a symlink when running Github actions: https://github.com/tailwindlabs/tailwindcss/blob/5788a9753a6437f00697f62ecb64b36fc3ea3aa5/.github/workflows/nodejs.yml#L47

PR #9507, opened 2022-10-08
Adding support for `baseline` in `placeItems` utility

DawoodKMasood:
Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/place-items


thecrypticace:
Thanks! 🎉

PR #9557, opened 2022-10-14
Add `aria-*` variants

reinink:
This PR adds a new `aria-{state}` variant which makes it easier to style elements based on the states of ARIA attributes.

For example, you can update the background color of an element based on whether the `aria-checked` state is `true`:

```html
<div aria-checked="true" class="bg-gray-600 aria-checked:bg-blue-600">
  <!-- ... -->
</div>
```

## Included variants

By default Tailwind provides the variants for the most common boolean ARIA attributes:

| Variant  | CSS  |
|---|---|
| `aria-checked:`  | `[aria-checked="true"]` |
| `aria-disabled:`  | `[aria-disabled="true"]` |
| `aria-expanded:`  | `[aria-expanded="true"]` |
| `aria-hidden:`  | `[aria-hidden="true"]` |
| `aria-pressed:`  | `[aria-pressed="true"]` |
| `aria-readonly:`  | `[aria-readonly="true"]` |
| `aria-required:`  | `[aria-required="true"]` |
| `aria-selected:`  | `[aria-selected="true"]` |

## Using custom values

You can customize which ARIA variants are available by editing `theme.aria` or `theme.extend.aria` in your `tailwind.config.js` file:

```js
module.exports = {
  theme: {
    extend: {
      aria: {
        'asc': 'sort="ascending"',
        'desc': 'sort="descending"',
      }
    }
  }
}
```

Then you can use any custom ARIA variants you've configured in your project:

```html
<th
  aria-sort="ascending"
  class="
    aria-asc:bg-[url('/img/up-arrow.svg')]
    aria-desc:bg-[url('/img/down-arrow.svg')]
  "
>
  <!-- ... -->
</th>
```

## Arbitrary values

For more complex ARIA attributes that take specific values (like `aria-activedescendant`), or for situations where it just doesn't feel worth it to add an alias to your theme, you can use the arbitrary value/square bracket notation to create custom ARIA variants on the fly:

```html
<th
  aria-sort="ascending"
  class="
    aria-[sort=ascending]:bg-[url('/img/up-arrow.svg')]
    aria-[sort=descending]:bg-[url('/img/down-arrow.svg')]
  "
>
  <!-- ... -->
</th>
```

## Group and peer variants

The ARIA variants are also available as `group` and `peer` modifiers.

```html
<div>Message:</div>
<div role="textbox" contenteditable aria-required="true" class="peer"></div>
<div class="hidden peer-aria-required:block">*A message is required</div>
```

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -257,6 +257,13 @@ export let variantPlugins = {
     )
   },
 
+  ariaVariants: ({ matchVariant, theme }) => {
+    let options = { values: theme('aria') ?? {} }
+    matchVariant('aria', (value) => `&[aria-${value}]`, options)
+    matchVariant('group-aria', (value) => `:merge(.group)[aria-${value}] &`, options)
+    matchVariant('peer-aria', (value) => `:merge(.peer)[aria-${value}] ~ &`, options)
+  },
+
   orientationVariants: ({ addVariant }) => {
     addVariant('portrait', '@media (orientation: portrait)')
     addVariant('landscape', '@media (orientation: landscape)')
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -258,10 +258,23 @@ export let variantPlugins = {
   },
 
   ariaVariants: ({ matchVariant, theme }) => {
-    let options = { values: theme('aria') ?? {} }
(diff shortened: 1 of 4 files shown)


seb-jean:
I love this PR @reinink

PR #9610, opened 2022-10-19
Resolve broken relative logo sources on npmjs (relative → absolute)

brandonmcconnell:
@RobinMalfait Looks like that last PR caused a slight regression on npmjs where the logo is completely broken now, because it registers the `source` URLs as relative links instead of absolute links.

You can see that bug here: https://www.npmjs.com/package/tailwindcss

Before and after this PR (assuming this works 🤷🏻‍♂️… it should)

|  | Before | After |
| - | - | - |
| Light | Screen Shot 2022-10-19 at 18 40 02 | Screen Shot 2022-10-19 at 18 39 42 |
| Dark | Screen Shot 2022-10-19 at 18 40 02 | Screen Shot 2022-10-19 at 18 39 21 |

Unfortunately, npmjs doesn't yet have a dark mode of their own, so even after corrected, the logo will be mostly invisible on their site unless a stroke or some other sort of style is added to the logo to add contrast between the white text and white background.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 <p align="center">
   <a href="https://tailwindcss.com" target="_blank">
     <picture>
-      <source media="(prefers-color-scheme: dark)" srcset="./.github/logo-dark.svg">
-      <source media="(prefers-color-scheme: light)" srcset="./.github/logo-light.svg">
-      <img alt="Tailwind CSS" src="./.github/logo-light.svg" width="350" height="70" style="max-width: 100%;">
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Very odd, we used relative URLs before as well 🤔

brandonmcconnell:
@RobinMalfait It appears the relative links work on npmjs for `img` elements but not `source`.

Probably worth opening a ticket on `npmjs` for this as well. And… there: https://github.com/npm/feedback/discussions/792

brandonmcconnell:
Btw it appears that npmjs has since fixed that issue, so the correct logo displays for both light and dark mode, at least AFAICT when checking locally

PR #9773, opened 2022-11-08
Fix shadowed variable when using raw content

willcosgrove:
Using `{raw: ...}` in the content config no longer works. It looks like the reason why is the `content` variable was being shadowed. I just renamed the variable so `content` remains accessible inside the loop.


thecrypticace:
Whoops… thanks for finding / fixing this! 🎉 

It'll be available in our insiders build to use in a few minutes: `npm install tailwindcss@insiders`
Will look at tagging a release tomorrow morning.

PR #9787, opened 2022-11-09
Pick up changes from files that are both context and content deps

### LINKED ISSUE #9763: TailwindCSS > 3.1.8 in linux doesn't resolve content paths correctly
3.2.2

vue 3.2.41, vue-cli 5.0.8, webpack 5.74.0

tested on16, 18, it is irrelevant.

Irrelevant

Ubuntu, Windows 10

**Reproduction URL**

Can provide if requested. 

**Describe your issue**
When using latest tailwindcss, content resolution behavior is broken in Linux.

I have a project with the following tree structure. Notice there are two .ce.vue files where actual tailwind classes are present.
image
The tailwind.config.js is the following:
image

**When project is built with vue-cli under Windows 10, tailwind css classes are extracted correctly.**
**When doing the same under Ubuntu, no tailwind css classes are extracted.** 
Downgrading to tailwind 3.1.8 resolved this issue. Therefore, I believe it is an issue with tailwindcss.

thecrypticace:
We switched to absolute paths and it broke this revealing a lurking bug. Fun.

Fixes #9763

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed use of `raw` content in the CLI ([#9773](https://github.com/tailwindlabs/tailwindcss/pull/9773))
+- Pick up changes from files that are both context and content deps ([#9777](https://github.com/tailwindlabs/tailwindcss/pull/9777))
 
 ## [3.2.2] - 2022-11-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed use of `raw` content in the CLI ([#9773](https://github.com/tailwindlabs/tailwindcss/pull/9773))
-- Pick up changes from files that are both context and content deps ([#9777](https://github.com/tailwindlabs/tailwindcss/pull/9777))
+- Pick up changes from files that are both context and content deps ([#9787](https://github.com/tailwindlabs/tailwindcss/pull/9787))
 
 ## [3.2.2] - 2022-11-04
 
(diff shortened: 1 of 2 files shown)


zigomir:
This change negatively affects Vite HMR performance with Vue 3 on our project. File change updates used to be almost instant and now take several seconds.

We're on currently latest `vite@4.0.1` and `vue@3.2.45`. Besides main entry js file which does `import 'tailwindcss/tailwind.css'`, we also have two [SFC Custom Elements](https://vuejs.org/guide/extras/web-components.html#building-custom-elements-with-vue) with own Tailwind directives inside SFC's `style` element.

```
@tailwind base;
@tailwind components;
@tailwind utilities;
```

Removing these helps a bit, but of course breaks our styles.

Do you think there's a possible workaround for this, other than downgrading to `v3.2.2`? Thanks!

thecrypticace:
Can you please open an issue with a reproduction? It seems unlikely to me that this would cause such a significant slowdown but as noted in the comment it's not really the "right' fix for the problem.

zigomir:
@thecrypticace I prepared reproduction, see https://github.com/zigomir/vite-tailwind-regression/pull/1

Repo has been made with `npm create vite@latest` from https://vitejs.dev/guide/ + https://tailwindcss.com/docs/guides/vite

Instead of reverting from `3.2.4` to `3.2.2` (I'm aware 3.2.3 introduces regression), you can also directly change `content.js` in `node_modules` to observe sam perf regression.

```diff
-if (modified >= prevModified) {
+if (modified > prevModified) {
```

thecrypticace:
Okay, yeah there is definitely a regression here. However, at least for the reproduction, there's a fairly simple workaround for the situation.

Your content paths include all of `node_modules`:
```js
content: [
    "./index.html",
    "./**/*.{js,ts,jsx,tsx}",
],
```

This ends up scanning almost 1700 files totaling about 13.7MB.

I would recommend scanning only the files you want. Since there's no `src` folder or something similar you can _exclude_ node_modules from being scanned:

```js
content: [
    "./index.html",
    "./**/*.{js,ts,jsx,tsx}",
    "!./node_modules", // This is the new line -- note the ! in front
],
```

I'll look into implementing a more proper fix for the content + context deps change detection. Thanks!

thecrypticace:
@zigomir This has been fixed and will be available in our insiders build (should be just a few minutes). I've tested it against your project and the project from the original issue and both are fixed! :)

Thanks for the repro! It was helpful!

Issue #9759, opened 2022-11-07
safelist Regular expressions do not work


jipika:
![image](https://user-images.githubusercontent.com/59040157/200231707-887b8508-7a31-411c-b13f-4d1eaf5247e8.png)
![image](https://user-images.githubusercontent.com/59040157/200231723-056ccdce-1805-4275-b347-ac25abf2a0bc.png)

jipika:
> 如果这个问题没有被任何人接受！我可以解决这个问题，PR 将在 2 天内启动

thank

akbng:
Here you go the PR for this issue is open now!

thecrypticace:
This has been fixed by #9791 and will be available in our insiders build in a few minutes: `npm install tailwindcss@insiders`

PR #9791, opened 2022-11-09
Fix !important selectors not being classified as valid class inside safelist config

akbng:
## This PR resolves issue #9759 

# PR includes 
- add the '!' prefix as a valid class inside safelisting options
- include tests for the same


thecrypticace:
Hey, thanks for the PR! Could you add an additional check that the regex pattern contains the `!` character anywhere? We're good with adding this but don't want people to pay for the perf penalty of checking basically doubling the generated classes if they don't need them.

akbng:
> Hey, thanks for the PR! Could you add an additional check that the regex pattern contains the `!` character anywhere? We're good with adding this but don't want people to pay for the perf penalty of checking basically doubling the generated classes if they don't need them.

sure I can try that and also saw some failing tests as well - will try to fix them :+1:

thecrypticace:
If you do that you won't need to fix the failing tests :)

thecrypticace:
Actually, I've already got an implementation. Lemme push it

akbng:
Thanks for the help

thecrypticace:
Thank you! 🎉

akbng:
WOW! I didn't expected such fast action

PR #9944, opened 2022-11-28
Support renaming of output files by `PostCSS` plugin.

NakajimaTakuya:
If you use tailwindcss as a cli and set the postcss option, the output file will ignore changes to result.opts by the postcss plugin and respect the output option passed to the cli.

However, some postcss plugins, such as the cache buster plugins, work by rewriting [result.opts](https://postcss.org/api/#resultoptions).
Like this one, for example.
https://github.com/keithamus/postcss-hash/blob/v2.0.0/index.js#L55-L57

This change is to honor the end product, `result.opts.to`, as the output file during postcss adaptation.

Example of reproducing the current problem.
```
// commands.
tailwindcss -i /path/to/input.css -o /path/to/output.css --postcss postcss.config.js


// postcss.config.js
module.exports = {
  plugins: [
    require('postcss-hash')({manifest: './manifest.json'})
  ]
}

// expected: outputfile: /path/to/output.a1b2c3d4e5.css
// actual: outputfile: /path/to/output.css
```


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

Nice and simple, thanks!

NakajimaTakuya:
@RobinMalfait 
Thank you, I'm very happy❤️

PR #9953, opened 2022-11-28
Alphabetize `theme` keys in default config

chuntington:
This corrects the alphabetical position of several properties. (https://tailwindcss.com/docs/theme#configuration-reference)

Here is a full list of properties that were moved around:
- accentColor
- brightness
- caretColor
- columns
- contrast
- grayscale
- gridRowEnd/Start
- hueRotate
- invert
- supports
- textIndent/Opacity


reinink:
Hey thanks for this! Decided to go a step further and just alphabetize all the `theme` keys, including the `screens`, `colors` and `spacing`. Keeps things nice and simple, and in reality the order doesn't actually matter all that much.

PR #9972, opened 2022-11-30
Improve return value of `resolveConfig`, unwrap `ResolvableTo`

### LINKED ISSUE #9929: resolveConfig returns wrong TypeScript type for theme keys

3.2.4

Vite 3.2.4

16.18.1

N/A

Linux

**Reproduction URL**

[Repro on CodeSandbox](https://codesandbox.io/s/wizardly-chandrasekhar-sk6ln7?file=/src/App.vue) (may take a second or click in the editor for the IDE to pick up the type issue)

**Describe your issue**

Using the [recently exposed resolveConfig types](https://github.com/tailwindlabs/tailwindcss/pull/8924), many config keys (such as `theme.colors`) throw TS errors:

Unresolved config key type

`ResolvableTo<T>` is an internal type that equals `T | ((utils: PluginUtils) => T)` - presumably a type that gets resolved during `resolveConfig`. The expected colour key (gray, in this case) doesn't exist on the function type, causing the issue. I suspect the resolved config should return the expected `RecursiveKeyValuePair` type with `ResolvableTo` removed.

RobinMalfait:
This PR improves the return type fo the `resolveConfig` a little bit.

In the config, a lot of values can be written in various forms (and most of it is optional). We also allow for dynamic "runtime" presets, so generating the perfect TypeScript types are almost impossible.

However, some of the parts in the config accept a function that resolves to type `T`, this improvement gets rid of the callback function and results in `T` directly (because calling `resolveConfig` should resolve this).

Fixes: #9929

### DIFF
--- a/resolveConfig.d.ts
+++ b/resolveConfig.d.ts
@@ -1,3 +1,12 @@
-import type { Config } from './types/config'
-declare function resolveConfig(config: Config): Config
+import type { Config, ResolvableTo } from './types/config'
+
+type UnwrapResolvables<T> = {
+  [K in keyof T]: T[K] extends ResolvableTo<infer R> ? R : T[K]
+}
+
+type ResolvedConfig<T extends Config> = Omit<T, 'theme'> & {
+  theme: UnwrapResolvables<T['theme']>
+}
+
+declare function resolveConfig<T extends Config>(config: T): ResolvedConfig<T>
 export = resolveConfig
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `foo-[abc]/[def]` not being handled correctly ([#9866](https://github.com/tailwindlabs/tailwindcss/pull/9866))
 - Add container queries plugin to standalone CLI ([#9865](https://github.com/tailwindlabs/tailwindcss/pull/9865))
 - Support renaming of output files by `PostCSS` plugin. ([#9944](https://github.com/tailwindlabs/tailwindcss/pull/9944))
+- Improve return value of `resolveConfig`, unwrap `ResolvableTo` ([#9972](https://github.com/tailwindlabs/tailwindcss/pull/9972))
 
 ## [3.2.4] - 2022-11-11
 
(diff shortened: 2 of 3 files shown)


resolveConfig.d.ts
```diff
@@ -1,3 +1,12 @@
-import type { Config } from './types/config'
-declare function resolveConfig(config: Config): Config
+import type { Config, ResolvableTo } from './types/config'
+
+type UnwrapResolvables<T> = {
+  [K in keyof T]: T[K] extends ResolvableTo<infer R> ? R : T[K]
+}
+
+type ResolvedConfig<T extends Config> = Omit<T, 'theme'> & {
+  theme: UnwrapResolvables<T['theme']>
```
thecrypticace, line 8: You're gonna want to apply the same unwrapping to all the keys inside `theme.extend`

thecrypticace: Actually, let me re-word this. You need to unwrap all the keys inside theme.extend and then spread them into the top level object. Is that even possible??

thecrypticace (changes requested):
General approach looks good sans the change for handling `theme.extend` unwrapping and merging.

resolveConfig.d.ts
thecrypticace: Scratch this. It'll require some ridiculousness with conditional types and what not. Nevermind.

thecrypticace (approved):

justinh-kairos:
I'm still seeing this with tailwind 3.3.2:
```
Property 'colors' does not exist on type 'UnwrapResolvables<{ container: { center: true; padding: string; screens: { "2xl": string; }; }; extend: { fontFamily: { sans: [string, ...string[]]; display: [string, ...string[]]; }; colors: { input: string; ring: { DEFAULT: string; offset: string; }; ... 10 more ...; card: { ...; }; }; borderRadius: { ...; }; keyf...'.ts(2339)
```

stefusilviu:
I confirm the same on 3.4.1

image

theocerutti:
No it still doesn't work:

```
const config = resolveConfig(tailwindConfig);

const mycolor = config.theme.colors.neutral['750']; // Element implicitly has an 'any' type because expression of type '750' can't be used to index type
```

PR #9973, opened 2022-11-30
Clip unbalanced closing brackets in arbitrary values

### LINKED ISSUE #9911: Arbitrary values aren't recognised in Ruby percent arrays
Thank you for an awesome tool! I've been using it in all my projects recently and the productivity gains have been immense.

Tailwind v3.2.4

postcss: 8.4.19, postcss-cli 10.0.0

Node v16.18.1

Firefox

macOS

**Reproduction URL**

I have added a test case that fails, showing the incorrect behaviour. This is the first time I've popped the hood on Tailwind, so forgive me if the regexes I've linked are the incorrect ones.

https://github.com/developius/tailwindcss/commit/5fb86ab84603949212d8146b09d242b420c7df85

**Describe your issue**

We're using Ruby on Rails with Tailwind and use the Ruby percent string syntax to generate arrays.
Something like `%w[text-[#bada55]]` becomes `["text-[#bada55]"]` in Ruby.

The following ERB code doesn't correctly generate the arbitrary value Tailwind class:

```erb
<%= content_tag(:div, class: %w[text-[#bada55]]) {} %>
```

However, this does work:

```erb
<%= content_tag(:div, class: ["text-[#a55bad]"]) {} %>
```

As noted in my commit above, I think this is due to the extra closing square bracket tripping up the regex here[^1] but my knowledge of Tailwind's internals is very vague.

I'm not sure if there's a way to fix this or not, as I appreciate that handling extremely specific scenarios like this is a direction the project might not want to head in. But maybe there's a quick fix? Thanks for checking this issue out, anyway 😄 

[^1]: https://github.com/tailwindlabs/tailwindcss/blob/master/src/lib/defaultExtractor.js#L65

thecrypticace:
Fixes #9911

### DIFF
--- a/src/lib/defaultExtractor.js
+++ b/src/lib/defaultExtractor.js
@@ -184,7 +184,7 @@ function clipAtBalancedParens(input) {
     // This means that there was an extra closing `]`
     // We'll clip to just before it
     if (depth < 0) {
-      return input.substring(0, match.index)
+      return input.substring(0, match.index - 1)
     }
 
     // We've finished balancing the brackets but there still may be characters that can be included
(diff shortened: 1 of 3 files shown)


src/lib/defaultExtractor.js
```diff
@@ -184,7 +184,7 @@ function clipAtBalancedParens(input) {
     // This means that there was an extra closing `]`
     // We'll clip to just before it
     if (depth < 0) {
-      return input.substring(0, match.index)
+      return input.substring(0, match.index - 1)
```
RobinMalfait, line 187: Hah, yep that will do it!

RobinMalfait (approved):

developius:
Thank you @thecrypticace @RobinMalfait 🙏

PR #10023, opened 2022-12-08
Add "data" type in tailwind.config

Yovach:
This PR adds the missing [data attributes](https://tailwindcss.com/docs/hover-focus-and-other-states#data-attributes) type to the tailwind.config.js file.

This is small but can be useful for autocompletion.


thecrypticace:
Good catch, thanks! ✨

PR #10072, opened 2022-12-14
Fix missing `string[]` in the `dropShadow` type

ngrushkovsky:
This PR updates the `dropShadow` type to accurately reflect the existing support for both strings and arrays of strings.


thecrypticace:
Thanks!

PR #10374, opened 2023-01-20
Cache the `oxide` targets

RobinMalfait:
This will cache the oxide related build files to hopefully minimize the
amount of Rust compiling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,10 +63,10 @@ jobs:
         run: npm install
 
       - name: Build Tailwind CSS
-        run: npm run build
+        run: npx turbo run build --filter=//
 
       - name: Test
-        run: npm run test
+        run: npx turbo run test --filter=//
 
       - name: Lint
-        run: npm run style
+        run: npx turbo run style --filter=//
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -46,7 +46,19 @@ jobs:
...
-          key: ${{ runner.os }}-cargo-${{ hashFiles('./oxide/**/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      # Cache the `oxide` Rust build
+      - name: Cache oxide build
+        uses: actions/cache@v3
+        with:
+          path: |
+            ./oxide/crates/target
+            ./oxide/crates/node/target
```
thecrypticace, line 57: Aren't these two paths wrong? Afaik it's `./oxide/target`

RobinMalfait: Yep totally, was just playing with things and saw that I had a `target` folder in `./oxide/crates/node` so added them but that was wrong.

Reason it is wrong is because I use this on my personal machine:
```sh
export CARGO_TARGET_DIR=~/cargo-target
```
Which basically uses `~/cargo-target` as a shared folder for _all_ my Rust programs. Kinda nice because I can do this:
```sh
PATH=~/cargo-target/release:$PATH
```
And all my binaries are available from everywhere.

I think for now we just want `./oxide/target` indeed, but the moment we have more Rust binaries, this custom target might be nice to introduce.

PR #10399, opened 2023-01-23
Use `Lightning CSS` in the `PostCSS` Plugin

RobinMalfait:
This PR will add `Lightning CSS` to the `PostCSS` Plugin in the `oxide` engine. This also required a
little bit of fiddling around with tests because the `stable` version doesn't use that.

To solve this, we will run both version through `Lightning CSS` but purely in tests to get the same
output. It technically makes things a bit slower (but only in tests) and they are now also
consistent.

There is a confusing part though, if you use something like this:
```js
expect(output).toMatchFormattedCss(css`
  .example {
    color: rgb(0, 0, 0);
  }
`)
```

Then the test will pass, but logging the `output` could look like:
```css
.example {
  color: black;
}
```

Not sure if this is an _actual_ problem right now or not...


tests/arbitrary-variants.test.js
```diff
@@ -643,18 +643,18 @@ crosscheck(({ stable, oxide }) => {
...
-              <div class="group-aria-[labelledby=a_b]:underline"></div>
-              <div class="peer-aria-[labelledby=a_b]:underline"></div>
-              <div class="group-aria-[sort=ascending]/foo:underline"></div>
-              <div class="peer-aria-[sort=ascending]/foo:underline"></div>
+              <div class="group-aria-[sort="ascending"]:underline"></div>
+              <div class="peer-aria-[sort="ascending"]:underline"></div>
+              <div class="group-aria-[labelledby="a_b"]:underline"></div>
+              <div class="peer-aria-[labelledby="a_b"]:underline"></div>
+              <div class="group-aria-[sort=:"ascending"]/foo:underline"></div>
+              <div class="peer-aria-[sort="ascending"]/foo:underline"></div>
```
thecrypticace, line 657: While this technically works in the test it would be invalid HTML. Should you change to single quotes?

jest/customMatchers.js
```diff
@@ -1,85 +1,132 @@
...
+          },
+        })
+        .code.toString('utf8')
+
+      return formatPrettier(output)
+    } catch (err) {
+      try {
+        // Lightning CSS is pretty strict, so it will fail for `@media screen(md) {}` for example,
+        // in that case we can fallback to prettier since it doesn't really care. However if an
+        // actual sytnax is made, then we still want to show the proper error.
```
thecrypticace, line 33: actual **syntax error** is made

tests/arbitrary-variants.test.js
RobinMalfait: Yep thanks! I removed the quotes again in most cases, but added them for the once where it was required like `foo_bar` cases.

PR #10400, opened 2023-01-23
Handle group/peer variants with quoted strings

### LINKED ISSUE #10353: double escape happening in attribute value selectors

v3.2.4

postcss 8.4.19

v14.19.1

Chrome

Kubuntu Linux

**Reproduction URL**

https://play.tailwindcss.com/sRVqoiJfLV

**Describe your issue**

in the attribute value of a attribute selector the value is being double escaped and there appears to be no way to un-escape it.

thecrypticace:
Fixes #10353

This fixes two issues:
1. We were splitting on `/` in a quoted string in group/peer variants
2. We were splitting not-a-space but should've consumed spaces inside quoted string in group/peer variants.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -168,7 +168,32 @@ export let variantPlugins = {
           if (!result.includes('&')) result = '&' + result
 
           let [a, b] = fn('', extra)
-          return result.replace(/&(\S+)?/g, (_, pseudo = '') => a + pseudo + b)
+
+          let start = null;
+          let end = null;
+          let quotes = 0;
+
+          for (let i = 0; i < result.length; ++i) {
+            let c = result[i]
+            if (c === '&') {
+              start = i;
+            } else if (c === '\'' || c === '"') {
+              quotes += 1;
+            } else if (start !== null && c === ' ' && !quotes) {
+              end = i;
+            }
+          }
+
+          if (start !== null && end === null) {
+            end = result.length
+          }
+
+          // Basically this but can handle quotes:
+          // result.replace(/&(\S+)?/g, (_, pseudo = '') => a + pseudo + b)
+
+          result = result.slice(0, start) + a + result.slice(start+1, end) + b + result.slice(end)
+
+          return result
         },
         { values: Object.fromEntries(pseudoVariants) }
       )
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -169,18 +169,18 @@ export let variantPlugins = {
 
           let [a, b] = fn('', extra)
 
-          let start = null;
-          let end = null;
-          let quotes = 0;
+          let start = null
+          let end = null
+          let quotes = 0
 
           for (let i = 0; i < result.length; ++i) {
             let c = result[i]
             if (c === '&') {
-              start = i;
-            } else if (c === '\'' || c === '"') {
-              quotes += 1;
+              start = i
+            } else if (c === "'" || c === '"') {
+              quotes += 1
             } else if (start !== null && c === ' ' && !quotes) {
-              end = i;
+              end = i
             }
           }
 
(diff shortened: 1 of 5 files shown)


src/lib/generateRules.js
```diff
@@ -152,7 +152,27 @@ function applyVariant(variant, matches, context) {
 
   // Retrieve "modifier"
   {
-    let match = /(.*)\/(.*)$/g.exec(variant)
+    let start = null
+    let parens = 0
+    let brackets = 0
+
+    for (let i = 0; i < variant.length; i++) {
```
RobinMalfait, line 159: I was wondering if we could use the `splitAtTopLevelOnly` function that we already have 🤔

src/corePlugins.js
```diff
@@ -168,7 +168,32 @@ export let variantPlugins = {
...
+          if (start !== null && end === null) {
+            end = result.length
+          }
+
+          // Basically this but can handle quotes:
+          // result.replace(/&(\S+)?/g, (_, pseudo = '') => a + pseudo + b)
+
+          result = result.slice(0, start) + a + result.slice(start + 1, end) + b + result.slice(end)
+
+          return result
```
RobinMalfait, line 196: Nitpick:
```suggestion
          return result.slice(0, start) + a + result.slice(start + 1, end) + b + result.slice(end)
```

tests/basic-usage.test.js
```diff
@@ -949,4 +949,66 @@ crosscheck(({ stable, oxide }) => {
...
+      @tailwind utilities;
+    `
+
+    let result = await run(input, config)
+
+    oxide.expect(result.css).toMatchFormattedCss(css`
+      .group[href^='/'] .group-\[\[href\^\=\'\/\'\]\]\:hidden {
+        display: none;
+      }
+    `)
```
RobinMalfait, line 972: ❤️ the oxide parser

tests/util/run.js
```diff
@@ -60,6 +60,18 @@ let nullProxy = new Proxy(
...
+ * @typedef {object} CrossCheck
+ * @property {typeof import('@jest/globals')} oxide
+ * @property {typeof import('@jest/globals')} stable
+ * @property {object} engine
+ * @property {boolean} engine.oxide
+ * @property {boolean} engine.stable
+ */
+
+/**
+ * @param {(data: CrossCheck) => unknown} fn
```
RobinMalfait, line 73: Nitpick:
```suggestion
 * @param {(data: CrossCheck) => void} fn
```
(Is `void` a thing in these docblocks?)

RobinMalfait (approved):
Looking good!

src/lib/generateRules.js
thecrypticace: That's a good idea but it doesn't work because of this example: `ar-1/10/20:text-red-500`. I'll see if there's a way I can massage it into working though.

tests/util/run.js
thecrypticace: It is. This was just my mind picking the wrong word lol

tests/basic-usage.test.js
thecrypticace: ❤️

src/lib/generateRules.js
thecrypticace: Okay, I've tweaked the implementation to use `splitAtTopLevelOnly` _and_ account for it. It doesn't feel pretty.

It'd be nice if there was an option like a `limit` but where it could also accept negative numbers for N occurrences from the end kinda like `slice`. I'm not sure if that's feasible though. Or even the right API.

tests/util/run.js
RobinMalfait: Heh no worries!

src/lib/generateRules.js
RobinMalfait: Oh I see, yeah I noticed you had a `2` in the call but that's not used by splitAtTopLvelOnly afaik 🤔

thecrypticace: oh crap I left that idea in lol. It didn't pan out hahaha.

PR #10506, opened 2023-02-06
Add "object-inherit" util! 

zerodevelops:
Adds a object-inherit utility.

As with no object inherit, the typical use-case would be to control what happens when no value is specified. Specially when using multiple components in react or other areas we would not just need to inherit the objects and instead of remaking them. Understanding that typically tailwind inherits both, I think there could be some additional use cases here.

Thank you, 
Zack D.

### DIFF
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -628,7 +628,6 @@ module.exports = {
       fit: 'fit-content',
     },
     objectPosition: {
-      inherit: 'inherit',
       bottom: 'bottom',
       center: 'center',
       left: 'left',
--- a/stubs/defaultConfig.stub.js
+++ b/stubs/defaultConfig.stub.js
@@ -628,6 +628,7 @@ module.exports = {
       fit: 'fit-content',
     },
     objectPosition: {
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thank you for your PR!
Much appreciated! 🙏

I'm a bit hesitant to just add it to spacing alone and think if we want to start including `inherit` that we should go through the whole config and make sure it's included everywhere that makes sense. That sounds like a big project though and I'm not sure I'm going to have time to really dig in to that any time in the near future, so I'm going to say no to this PR for now, but not closed to revisiting it in the future.

In the mean time there is a super easy alternative, which is to use arbitrary values:

```html
<div class="object-[inherit]"></div>
```

Two extra characters but doesn't require any changes to Tailwind or your config at all 👍 Of course you can also add the `inherit` value to your own config if you'd like too.

Thanks regardless! ❤️

ZeroScriptz:
Very true now further thinking about that in retrospect of other configs! Very simple addition in near future to inherit others. Thanks for keeping it in mind. I just think it could save time on certain things. 

Thanks for the feedback!

PR #10645, opened 2023-02-21
Add `content-normal` and `content-stretch` utilities

dcastil:
I'd like to add a utility for the CSS `align-content: stretch`.

This is the default align-content value for flexbox and grid layouts if none other is specified. The reason to include this in Tailwind is that it is necessary to be able to reset the `align-content` value to its default on a breakpoint, e.g. `content-center lg:content-stretch`. Also it isn't possible to use arbitrary values for `align-content` since `content-[stretch]` would instead create the CSS `content: stretch`. So the only option to use that CSS currently is with arbitrary properties: `content-center lg:[align-content:stretch]`.

Let me know what you think. 😊

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -1196,6 +1196,7 @@ export let corePlugins = {
 
   alignContent: ({ addUtilities }) => {
     addUtilities({
+      '.content-normal': { 'align-content': 'normal' },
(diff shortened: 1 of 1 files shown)


reinink:
Hey thanks for this! I think this makes sense to add. I've also added `content-normal` with this update 👍

PR #10670, opened 2023-02-23
Gracefully handle missing config when Tailwind doesn't need to run

### LINKED ISSUE #10654: @config rule seems to break with scoped styles in vue

For example: v3.2.7

For example: vite 4.1.0

For example: v19.6.1

For example: Firefox

For example: Debian

**Reproduction URL**

https://github.com/bebetoalves/tailwind-config-bug

**Describe your issue**

Apparently the @config directive breaks compilation when using scoped styles in a vue component. Even if the class doesn't have any css present, the compilation will still break.

error

thecrypticace:
Right now we fail not-so-gracefully if we expect a config file and don't find one. This can become more apparent when using `@config` with a custom config file path _and_ Vue SFCs as style blocks are processed independently.

This proof-of-concept does two things:

1. Adds usable error messages when config files are missing.
2. Bails as early as possible if Tailwind doesn't need to run. We base this on the presence of `tailwind`, `layer`, `import`, and some other at-rules as well as uses of the `theme` and `screen` functions. However, we no longer deduplicate adjacent rules for all inputs where we did previously — even if Tailwind itself didn't otherwise produce the resulting CSS. It's possible we'll want to still do that no matter what.

Fixes #10654

### DIFF
--- a/src/lib/evaluateTailwindFunctions.js
+++ b/src/lib/evaluateTailwindFunctions.js
@@ -267,3 +267,26 @@ export default function (context) {
     })
   }
 }
+
+/**
+ *
+ * @param {import('postcss').Node} node
+ * @returns {boolean}
+ */
+export function hasTailwindFunctions(node) {
+  let property = nodeTypePropertyMap[node.type]
+  if (property === undefined) {
+    return false
+  }
+
+  let hasFunction = false
+
+  parseValue(node[property]).walk((vnode) => {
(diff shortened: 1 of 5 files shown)


thecrypticace:
We no longer error when a config is missing — it would still be a good optimization to figure out how to optimize thing such that Tailwind can skip setup and processing when unnecessary.

We still technically show a warning regarding empty content arrays — this isn't ideal but it's at the very least not breaking the build. So we're going to punt on this idea for now.

PR #10672, opened 2023-02-23
Sort class lists deterministically for Prettier plugin

thecrypticace:
Something we noticed in testing. With the latest release we made sure to sort generated rules deterministically rather than based on content-scanning order which can (and does) differ across environments. However, in testing, we noticed that the Prettier plugin didn't take this into account when ordering class names.

This PR enables that behavior in the API used by our prettier plugin so the next release of Tailwind will ensure prettier sorts the classes correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Disallow multiple selectors in arbitrary variants ([#10655](https://github.com/tailwindlabs/tailwindcss/pull/10655))
+- Sort class lists deterministically for Prettier plugin ([#10672](https://github.com/tailwindlabs/tailwindcss/pull/10672))
 
 ### Changed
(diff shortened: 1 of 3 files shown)


carlmjohnson:
This change results in a lot of churn on my site, and in particular it changes the order of margins. Before margins were trlb but now they aren't. Are we sure this didn't have a bug?

reinink:
@carlmjohnson So yeah, this is intended behavior, although you're totally right it does result in a bunch of new classes being sorted by our Prettier plugin that previously weren't. I actually just tweeted about this yesterday: https://twitter.com/reinink/status/1641222722179985408

When we added deterministic sorting to Tailwind CSS back in January (see #10382) we forgot to update the Prettier plugin to match this, so in v3.3 we've corrected this (#10672).

My recommendation is to run Prettier against your whole project to make all these changes at once.

I'm a fan of this change personally the Prettier plugin sorting is even more thorough now — meaning even less for us developers to have to think about 👍

PR #10785, opened 2023-03-13
Enable ESM and TS based config files

RobinMalfait:
This PR adds support for loading and scaffolding ESM and TypeScript config files.

### ESM and TS syntax support

Prior to this PR, your `tailwind.config.js` file had to be in CommonJS format. This was surprising to people using `"type": "module"` in their projects, and made it more difficult to do things like extract your theme tokens into a separate module that you could import into your `tailwind.config.js` as well as into your application code to maintain a single source of truth.

To make this work, we need to transpile ESM and TS config files on the fly (using [jiti](https://github.com/unjs/jiti)), since Tailwind itself isn't published in native ESM and of course there's no native TS support in Node.js. This means there is a performance penalty to using ESM and TS config files, which we're mitigating as much as possible by using [Sucrase](https://github.com/alangpierce/sucrase) as the parser instead of Babel. The penalty on my machine is about ~10ms (~50ms using Babel) for one-off builds and the initial build in watch mode, and basically not measurable for incremental builds.

### CLI improvements

This PR also adds `--esm` and `--ts` flags to the `tailwindcss init` CLI command to generate your config file using your preferred syntax:

```bash
npx tailwindcss init --esm
```

When running the `init` command with no flags, Tailwind will guess which syntax to use based on the `"type"` field in your `package.json`.

When using the `--postcss` flag, the generated `postcss.config.js` file will use the same syntax as your `tailwind.config.js` file, whether inferred or specified explicitly on the command line.

---

@natemoo-re and @innocenzi have been credited as co-authors for their efforts in earlier PRs:

- https://github.com/tailwindlabs/tailwindcss/pull/7327
- https://github.com/tailwindlabs/tailwindcss/pull/3544


loadConfig.d.ts
```diff
@@ -0,0 +1,4 @@
+import type { Config } from './types/config'
+
+declare function loadConfig(path: string): Config
+export = loadConfig
```
RobinMalfait, line 4: Does this seem correct to you, @thecrypticace?

Started form the `resolveConfig` file

src/lib/load-config.ts
```diff
@@ -0,0 +1,27 @@
...
+  )
+}
+
+export function loadConfig(path: string): Config {
+  try {
+    return path ? require(path) : {}
+  } catch {
+    return lazyJiti()(path)
+  }
+}
```
RobinMalfait, line 27: Simplified this to just a load config function that returns the config directly

loadConfig.d.ts
thecrypticace: for `require("tailwindcss/loadConfig")` yep looks right to me

integrations/io.js
```diff
@@ -97,6 +97,17 @@ module.exports = function ({
...
+        !(await fs
+          .access(filePath)
+          .then(() => true)
+          .catch(() => false))
+      ) {
+        return
+      }
+
       if (!fileCache[filePath]) {
         fileCache[filePath] = await fs.readFile(filePath, 'utf8')
```
thecrypticace, line 112: I think you can get rid of the .access.then.catch call if you do this:

```suggestion
        fileCache[filePath] = await fs.readFile(filePath, 'utf8').catch(() => null)
```

thecrypticace: Ah you'd need to check and early return still but might be worth it ¯\_(ツ)_/¯

innocenzi:
Super happy we finally have this, great job guys ❤️ 

Tested with two medium projects:
- Both with `@tailwindcss/forms` and a few other external plugins
- Both greatly customized
- One that uses two configs through `@config`

Everything works perfectly fine, migration is easy. Super nice.

I've noticed `zIndex` has type issues. I had the following:

```ts
zIndex: {
	'sticky-on-scroll': 20,
	'chat-bubble': 30,
	'navbar': 30,
	'navbar-dropdown': 25,
	'dialog-base': 35,
	'dialog-panel': 36,
	'notification': 50,
},
```

But I had to make the values strings instead of numbers. Using `@type {import('tailwindcss').Config}` did not catch that before.

The only thing I miss is a `defineConfig` util instead of having to add `satisfies Config` at the end of the exported object, which is subjectively less pretty. But it's minor.

woss:
hi, has this been released? i'm trying to get the TS and ESM js config to work but it doesn't.

RobinMalfait:
@woss no not yet, it will be soon though!

If you want to know what's released and what isn't you can always take a look at the [CHANGELOG.md](https://github.com/tailwindlabs/tailwindcss/blob/master/CHANGELOG.md) file.

That said, you can already play with it using the insiders version: `npm install tailwindcss@insiders`.

karlhorky:
This looks great, thanks for this @RobinMalfait 🙌

Since the current version number is `3.2.7`, I guess this will be probably released in a new minor (`tailwindcss@3.3.0`)?

joshmanders:
Thank you @RobinMalfait this is a long awaited feature I've been waiting for! Happy early birthday to me!

klib19:
Thank you all for this feature! This is exactly what I needed for my project, I appreciate all the hard work!

One thing that is unclear to me and isn't addressed in the documentation, is how do I import the Tailwind colors into my ESM config file? None of these work

`const colors = require('tailwindcss/colors')`
import colors from "tailwindcss/colors";
`import colors from "tailwindcss/colors.js";`

seb-jean:
Hello,

How to add `defaultTheme`, `plugin`, etc with `TS`.

Like this?

```ts
const defaultTheme = require('tailwindcss/defaultTheme');
const plugin = require('tailwindcss/plugin');
```

karlhorky:
The more common style is `import` in TypeScript:

```ts
import defaultTheme from 'tailwindcss/defaultTheme';
import plugin from 'tailwindcss/plugin';
```

Often your editor / IDE can do this conversion for you like [Convert 'require' to 'import' in VS Code](https://stackoverflow.com/questions/62714391/is-there-an-action-to-run-convert-require-to-import-in-vscode) (either use the lightbulb or hover over the three dots)

RobinMalfait:
@klib19 while I wouldn't recommend to mix everything. All three of those should just work:
image

If you are experiencing issues then I would recommend to open an issue with a minimal reproduction repo attached that shows the issue.

---

@seb-jean as @karlhorky mentioned, you would use the `import` syntax in this case 👍

seb-jean:
thanks @RobinMalfait

pi0:
Hi. Jiti maintainer here. Excellent work and glad that tailwind is using jiti for TS/ESM support 🔥 Please let me know if there were any issues or possible improvements.

genesiscz:
It would be really great if it was somehow documented in the official documentation on the website. We literally burnt a few hours trying to make this work. Here's my steps:

- change postcss.config.js to postcss.config.mjs (we're not using type: module in package.json)
- renamed tailwind.config.cjs to tailwind.config.ts
- added `import type { Config } from "tailwindcss";` and `satisfies Config` to tailwind.config.ts 
- I also needed to change all ResolvableTo from (theme) => { ...theme("colors") } to (plugin) => { ...plugin.theme("colors") }

Example:
```diff
-padding: (theme) => theme("spacing")
+padding: (plugin) => plugin.theme("spacing")
```
- vite.config.ts:
-- Deleted resolve.alias["tailwind.config.js"], optimizeDeps.include[path.resolve(__dirname, "src/tailwind.config.js")], build.commonJsOptions.include ("tailwind.config.cjs")
- Added css.postcss.plugins: [tailwindcss()] where tailwindcss is `import tailwindcss from "tailwindcss";`


Whole diff:
```diff
diff --git a/bun.lockb b/bun.lockb
index 94d93624..0db19a94 100644
Binary files a/bun.lockb and b/bun.lockb differ
diff --git a/package.json b/package.json
index 7fe57db1..c8cfb66c 100644
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "mui-color": "^2.0.0-beta.2",
     "notistack": "^2.0.5",
     "polished": "^4.2.2",
+    "postcss": "^8.4.30",
     "postcss-cli": "^10.0.0",
     "pretty-quick": "^3.1.3",
     "react": "^18.2.0",
@@ -108,7 +109,6 @@
     "resumablejs": "^1.1.0",
     "sass": "^1.55.0",
     "styled-components": "^5.3.5",
-    "tailwind-config": "^0.1.2",
     "tailwindcss": "^3.1.8",
     "text-mask-addons": "^3.8.0",
     "url-loader": "^4.1.1",
@@ -174,7 +174,6 @@
     "fs": "^0.0.1-security",
     "http-proxy-middleware": "^2.0.6",
     "husky": "^8.0.0",
-    "postcss": "^8.4.18",
     "postcss-import": "^15.0.0",
     "postcss-loader": "^7.0.1",
     "prettier": "^2.7.1",
diff --git a/postcss.config.js b/postcss.config.js
deleted file mode 100644
index 0a0fcca1..00000000
--- a/postcss.config.js
+++ /dev/null
@@ -1,7 +0,0 @@
-import tailwind from "tailwindcss";
-import autoprefixer from "autoprefixer";
-import tailwindConfig from "tailwind.config";
-
-export default {
-  plugins: [tailwind(tailwindConfig), autoprefixer],
-};
diff --git a/src/modules/auth/container/FindInstance.tsx b/src/modules/auth/container/FindInstance.tsx
index 0f19fe62..bdcd7cd2 100644
--- a/src/modules/auth/container/FindInstance.tsx
+++ b/src/modules/auth/container/FindInstance.tsx
@@ -14,14 +14,13 @@ import { LoadingButton } from "../../../core/components/LoadingButton";
 import { useAppDispatch } from "../../../redux/hooks/hooks";
 import { Logo } from "../../../core/components/_atoms/Logo/Logo";
 import { findInstancesAction } from "../store/action/findInstances.action";
-import myConfig from "tailwind.config";
+import { resolveTailwindConfig } from "tailwind.theme";
+import tw from "tailwind.theme";
 
 function FindInstance() {
   const dispatch = useAppDispatch();
-  console.log(myConfig);
-  const theme = myConfig;
-  console.log(theme);
-  //console.log(tailwindTheme?.width["mine"]);
+  console.log("tw", tw);
+  console.log(resolveTailwindConfig());
 
   return (
     <AuthWrapper>
diff --git a/src/tailwind.config.js b/src/tailwind.config.js
deleted file mode 100644
index cc038157..00000000
--- a/src/tailwind.config.js
+++ /dev/null
@@ -1,161 +0,0 @@
-/** @type {import('tailwindcss').Config} */
-export default {
-  content: ["./src/**/*.{js,jsx,ts,tsx}", "./public/index.html"],
-  // ### Safelisting these classnames because, they were not updating in Title component when changing size prop
-  safelist: ["text-xs", "text-sm", "text-md", "text-lg", "text-xlg"],
-  theme: {
-    borderColor: (theme) => ({
-      ...theme("colors"),
-      DEFAULT: theme("colors.gray.200", "currentColor"),
-    }),
-    padding: (theme) => theme("spacing"),
-    // extending
-    extend: {
-    },
-  },
-};
diff --git a/src/tailwind.theme.ts b/src/tailwind.theme.ts
index 7b344945..c496faa8 100644
--- a/src/tailwind.theme.ts
+++ b/src/tailwind.theme.ts
@@ -1,4 +1,9 @@
-import myConfig from "tailwind.config.cjs";
+import resolveConfig from "tailwindcss/resolveConfig";
+import tailwindConfig from "../tailwind.config";
 
-console.log(myConfig);
-export default myConfig;
+export const resolveTailwindConfig = () => {
+  return resolveConfig(tailwindConfig);
+};
+
+const twConfig = resolveConfig(tailwindConfig);
+export default twConfig;
diff --git a/vite.config.ts b/vite.config.ts
index 5f2371e3..c442d12c 100644
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -7,11 +7,11 @@ import macrosPlugin from "vite-plugin-babel-macros";
 import svgrPlugin from "vite-plugin-svgr";
 import mkcert from "vite-plugin-mkcert";
 import eslintPlugin from "@nabla/vite-plugin-eslint";
-import path from "path";
 import circleDependency from "vite-plugin-circular-dependency";
 import checker from "vite-plugin-checker";
 //import { visualizer } from "rollup-plugin-visualizer";
-import postcss from "./postcss.config";
+//import react from '@vitejs/plugin-react-swc'
+import tailwindcss from "tailwindcss";
 
 export default defineConfig({
   envPrefix: "REACT_APP_",
@@ -29,11 +29,11 @@ export default defineConfig({
       https: require.resolve("rollup-plugin-node-builtins"),
       http2: require.resolve("rollup-plugin-node-builtins"),
       process: require.resolve("rollup-plugin-node-builtins"),
-      "tailwind.config.js": path.resolve(__dirname, "src/tailwind.config.js"),
     },
   },
+
   optimizeDeps: {
-    include: [path.resolve(__dirname, "src/tailwind.config.js")],
+    include: [],
     esbuildOptions: {
       define: {
         global: "globalThis",
@@ -50,7 +50,7 @@ export default defineConfig({
   },
   build: {
     commonjsOptions: {
-      include: ["src/tailwind.config.js", "node_modules/**"],
+      include: ["node_modules/**"],
 
       // Rollup doesn't build CommonJS modules out of the box. That's why this
       // option needs to be enabled
@@ -146,7 +146,9 @@ export default defineConfig({
     },
   },
   css: {
-    postcss,
+    postcss: {
+      plugins: [tailwindcss()],
+    },
   },
   plugins: [
     react(),
```


Hope that helps somebody!

PR #10947, opened 2023-04-04
Order utilities clockwise where applicable

thecrypticace:
Fixes https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/147

We're gonna change things so the appropriate `inset`, `padding`, `margin`, etc… utilities are in a clockwise sort order.


src/corePlugins.js
```diff
@@ -631,14 +631,12 @@ export let corePlugins = {
...
-        ['right', ['right']],
-        ['bottom', ['bottom']],
-        ['left', ['left']],
-      ],
+      [['top', ['top']]],
+      [['right', ['right']]],
+      [['end', ['inset-inline-end']]],
+      [['bottom', ['bottom']]],
+      [['left', ['left']]],
+      [['start', ['inset-inline-start']]],
```
RobinMalfait, line 639: Do you think we should keep `start` and `end` as-is? This change would sort it `end` first then `start` which seems odd? 🤔

thecrypticace: Good question. Yo @adamwathan what do you think?

adamwathan: Yeah I put them there on purpose but doing some more research, it looks like the logical ones will be flipped when/if that’s ever supported:

image

So we should go top, right, inline-start, bottom, left, inline-end I think.

adamwathan:
Kinda second guessing some of this a bit, I think I would hate for `pr-*` to be sorted before `pl-*`. I think we should have an opinion about this in Tailwind but not convinced it should be TRBL 👀

adamwathan:
Kinda second guessing some of this a bit, I think I would hate for `pr-*` to be sorted before `pl-*`. I think we should have an opinion about this in Tailwind but not convinced it should be TRBL 👀

teddybradford:
A different approach could be to sort from
- inline-axis to block-axis
- inline-start to inline-end
- block-start to block-end
and
- x-axis to y-axis
- left to right
- top to bottom

But I think this sorting might be biased for Western locales?

adamwathan:
The people have spoken! Let's do:

- top
- bottom
- inline-start
- left
- inline-end
- right

image

Wolfgrammatikalazeredr:
Mo humanistic

Wolfgrammatikalazeredr:
Wowza

Wolfgrammatikalazeredr (approved):

Wolfgrammatikalazeredr:
Somwhere stream on through.

Wolfgrammatikalazeredr:
Anyone?

kachkaev:
@thecrypticac what is the current status of this PR? Happy to help with testing, if needed!

thecrypticace:
I'll re-open this PR once we've worked through the decisions we want to make here.

PR #10962, opened 2023-04-06
Don’t move unknown pseudo-elements to the end of selectors

### LINKED ISSUE #10960: 3.3.1 not handling pseudo-elements properly

`3.3.1`

tailwindcss

`v18.14.0`

#### Steps to Repro

```bash
mkdir repro
cd repro
yarn global add tailwindcss@3.3.1
tailwindcss init
```

Create `repro.css` with the following content:

```css
@config 'tailwind.config.js';
::deep [a] {
    @apply bg-white;
}
```

Run `tailwindcss --input .\repro.css --output .\result.css`

```css 
/* result.css / Incorrect */
[a]::deep {
  --tw-bg-opacity: 1;
  background-color: rgb(255 255 255 / var(--tw-bg-opacity))
}
```

You can fix this by downgrading tailwindcss to 3.3.0:

```bash
yarn global add tailwindcss@3.3.0
tailwindcss --input .\repro.css --output .\result.css
```` 

And the CORRECT output

```css
/* Correct! */
::deep [a] {
  --tw-bg-opacity: 1;
  background-color: rgb(255 255 255 / var(--tw-bg-opacity))
}
```

thecrypticace:
Fixes #10960

The `::deep` Shadow DOM piercing selector uses invalid selector syntax but needs to still work for now. Here we've add it as an exception to ensure it is not moved to the end of the selector (which is the only place that pseudo elements can normally be placed).

### DIFF
--- a/.github/workflows/ci-stable.yml
+++ b/.github/workflows/ci-stable.yml
@@ -50,7 +50,10 @@ jobs:
         run: npm run build
 
       - name: Test
-        run: npm run test
+        run: |
+          npm run test || \
+          npm run test || \
+          npm run test || exit 1
 
       - name: Lint
         run: npm run style
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,10 @@ jobs:
         run: npx turbo run build --filter=//
 
       - name: Test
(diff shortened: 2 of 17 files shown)


onyea (commented):
Open

PR #10982, opened 2023-04-11
Add first-class 3d transform utility support

brandonmcconnell:
Adds first-class support for 3d transformations using the existing TailwindCSS architecture.

This PR also sees GPU acceleration activated by default, but this can be easily switched to using CPU by default. It may also be useful to trigger some sort of warning if a user uses the `transform-cpu` utility but then also attempts to use one of the 3d transform utilities, which would naturally disable 3d transformations.

The new utilities added in this effort are:

| Utility | CSS Property/Transform Function | CSS Value |
|--------|--------|--------|
| `rotate-x` | `rotateX()` | same as `rotate` value(s) |
| `rotate-y` | `rotateY()` | same as `rotate` value(s) |
| `rotate-z` | `rotateZ()` | alias of `rotate` value(s) |
| `scale-z` | `scaleZ()` | same as `scale` value(s) |
| `perspective-self` | `perspective()` | incremental t-shirt sizing |
| `perspective` | `perspective` | incremental t-shirt sizing, same as `perspective-self` |
| `perspective-origin` | `perspective-origin` | same as `origin` (`transform-origin`) |
| `transform-flat` | `transform-style` | `flat` |
| `transform-3d` | `transform-style` | `preserve-3d` |
| `transform-content` | `transform-box` | `content-box` |
| `transform-border` | `transform-box` | `border-box` |
| `transform-fill` | `transform-box` | `fill-box` |
| `transform-stroke` | `transform-box` | `stroke-box` |
| `transform-view` | `transform-box` | `view-box` |
| `backface-visible` | `backface-visibility` | `visible` |
| `backface-hidden` | `backface-visibility` | `hidden` |


adamwathan:
Hey thanks!

Just to set expectations, 90% chance we will have to close this for now just because I can't drop everything and make this feature a top priority (especially the documentation effort). As I've mentioned in other PRs it just takes a lot of dedicated time and focus to ensure major new feature PRs are ready to be merged, because I have to go through the process of thinking about the feature from first principles and make sure the final solution matches what I think is the best solution for the framework, so it's always a _lot_ more work than simply looking at the changed files in a PR. So because we don't like to have PRs sit open for a year or more (like they have in the past), we generally close things that we aren't ready to merge basically right away.

That said the big blocker on this feature in the past has been designing a thoughtful scale for `perspective`. What is the justification/design story behind these specific values?

```
    perspective: {
      none: 'none',
      0: '0rem',
      xs: '6.25rem', // 100px
      sm: '18.75rem', // 300px
      md: '31.25rem', // 500px
      lg: '50rem', // 800px
      xl: '75rem', // 1200px
      '2xl': '106.25rem', // 1700px
      '3xl': '143.75rem', // 2300px
      '4xl': '187.5rem', // 3000px
      '5xl': '237.5rem', // 3800px
      '6xl': '300rem', // 4800px
      '7xl': '375rem', // 6000px
      '8xl': '468.75rem', // 7500px
      '9xl': '625rem', // 10000px
    },
```

brandonmcconnell:
@adamwathan No worries! Always a pleasure. It's too bad there isn't a way to archive these sorts of PRs for a later date, though I guess you can do that internally and then always re-open them once they surface as priorities.

I'm collecting a long laundry list of TailwindCSS plugins on my end and have no issue keeping them all local, but seeing them supported as first-class utilities is always great.

Re the `perspective` scale, that was one I was thinking through quite a bit as well. This was a very rough scale I came up with, and I 100% know it would need some work before anything is finalized. The idea was generally that the perspective values should allow for a lot of granular control ranging between a perspective as small as 100px to as large as 100x that (10,000px).

If we want to get something really mathematically uniform, we could parabolic regression to make a scale like this:

```
y = 50x^2 + 50x + 100

xs   sm   md   lg   xl    2xl   3xl   4xl   5xl   6xl   7xl   8xl    9xl
100, 250, 500, 900, 1450, 2200, 3150, 4300, 5650, 7200, 8950, 10900, 13050
```

or if you want more human-friendly values (more likely, presumably), something simpler like this:

```
sm   sm   md    lg    xl    2xl   3xl
250, 500, 1000, 2500, 5000, 7500, 10000
```

brandonmcconnell:
And to help visualize these two different perspective scales (NOT the one originally/currently included in the PR, which ould need to be updated):

## Parabolic progression

```c
// y = 50x^2 + 50x + 100

xs   sm   md   lg   xl    2xl   3xl   4xl   5xl   6xl   7xl   8xl    9xl
100, 250, 500, 900, 1450, 2200, 3150, 4300, 5650, 7200, 8950, 10900, 13050
```

<small>Below source: <a href="https://www.wolframalpha.com/input?i=%7B100%2C+250%2C+500%2C+900%2C+1450%2C+2200%2C+3150%2C+4300%2C+5650%2C+7200%2C+8950%2C+10900%2C+13050%7D" target="_blank">Wolframp Alpha</a></small>

image

## Human-predictable stop values

I added a few extra breakpoints to make the graph's curvature more naturally exponential. It's not as smooth as the parabolic form, though the stop values are more predictable.

```c
xs   sm   md    lg    xl    2xl   3xl   4xl   5xl
250, 500, 1000, 1500, 2500, 3750, 5000, 7500, 10000
```

<small>Below source: <a href="https://www.wolframalpha.com/input?i=%7B250%2C+500%2C+1000%2C+1500%2C+2500%2C+3750%2C+5000%2C+7500%2C+10000%7D" target="_blank">Wolframp Alpha</a></small>

image

brandonmcconnell:
Btw here's a little demo showcasing some of these new utilities:
https://play.tailwindcss.com/5Mp3tH8i40

Here is it how it looks:



Many of the classes used in that example are redundant and would be streamlined using a simple variable (e.g. `cubeFace` or similar) in a real-world scenario.

brandonmcconnell:
@adamwathan Also, if `perspective` is the main blocker here, a Twitter poll and getting community thoughts. However, a perspective scale isn't entirely necessary for the initial addition of 3d transforms anyway.

The example I showed in my previous comment was made with an intrinsic perspective (essentially infinity).

I think it would still help to include the `perspective`-related properties even without a scale at first, as this would allow easier usage of each of these:

```css
* {
  perspective: /* arbitrary value */;
  transform: perspective(/* arbitrary value */);
  perspective-origin: /* same config values as transform-origin */;
}
```

So this would still all be valid even without a `perspective` scale, and I think this would be justifiable over using an arbitrary property like `[perspective:1000px]` because it's **(a)** first-class supported in nature, and moreover **(b)** allows the use of `perspective()` which wouldn't otherwise be possible using an arbitrary property.

sambauers:
Some notes from my implementation of https://github.com/sambauers/tailwindcss-3d

I went for modern alternatives to `transform` by default (there is a legacy mode to use `transform`). Tailwind's use of `var` within `transform` functions is smart but limited (especially within `keyframe`s). I wonder if core would be better served by transitioning to the modern approach at this point in time, rather than extending use of `transform` into 3D.

There are some rough edges to browser support here as well. There are various triggers for GPU versus CPU usage that can kick in unexpectedly https://developer.mozilla.org/en-US/docs/Learn/Performance/CSS - there is also still a need to add `--webkit-transform` duplicate declarations in many/most cases to ensure broad support.

I do want to see 3D support in core, but maybe the demand can be gauged through stats on the plugin that exists (not huge usage number so far), I can also weed out various issues there. I'm also happy to have your input and support on development there @brandonmcconnell

Side note: @adamwathan sorry for not adhering to the brand/naming conventions for the 3D plugin. I only became aware of that guideline last night - I'll work through renaming soon.

sambauers:
On the question of `perspective` values. It generally takes big changes in values to see any visible difference in output. In the 3D plugin I used a very small set of defaults expecting that people would use arbitrary values if they want to get specific.

https://github.com/sambauers/tailwindcss-3d/blob/main/src/css-utilities/perspective.ts#L17

sambauers:
@brandonmcconnell what is the thinking behind having seperate `perspective` and `perspective-self` utilities? Is there a specific use case that would require one over the other? In the 3D plugin I implemented the `perspective` property as standard, or the `perspective()` transform function in legacy mode, but not both at once.

brandonmcconnell:
@sambauers Thanks for sharing your thoughts here!

* Re the GPU/CPU triggers:

  I purposely followed the current TailwindCSS as much as possible, so that should be handled near identically. The natural browser GPU/CPU triggers are intentional and should trigger GPU automatically when needed unless manually overridden using `transform-gpu`/`transform-cpu`.
* Re keyframes:

  Each of these properties are animatable using `@keyframes`. What limitations are you referring to?
* Re demand:

  There are have been at least 4 plugins created over the years for this precise purpose—
  * https://github.com/benface/tailwindcss-transforms
  * https://github.com/sambauers/tailwindcss-3d
  * https://github.com/Kamona-WD/tailwindcss-perspective
  * & my own

  …and probably others, among hundreds of people who have used these plugins. Having first-class support for 3d transforms would enable all CSS devs to make use of 3d transforms and related properties, a core features of the lang
* Re `perspective` values:

  Granularity also makes a great deal of difference on the smaller values like `100px` vs. `300px`, which is why I added the `100px` value, though I agree (re [my recent comment](https://github.com/tailwindlabs/tailwindcss/pull/10982#issuecomment-1511874536)), and I think those values can be all/mostly be left to arbitrary usage.
* Re your question on `perspective` vs. `perspective-self`:

  This PR also uses the `perspective` utility as the standard, but `perspective()` (`perspective-self`) serves a different purpose than the `perspective` property and is an ideal fit for situations where you don't want to enclose your 3d element in a wrapper simply to give it perspective, or where sibling elements require different `perspective` values.

  MDN makes this distinction [here](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/perspective): "The `perspective()` transform function is part of the [`transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) value applied on the element being transformed. This differs from the [`perspective`](https://developer.mozilla.org/en-US/docs/Web/CSS/perspective) and [`perspective-origin`](https://developer.mozilla.org/en-US/docs/Web/CSS/perspective-origin) properties which are attached to the parent of a child transformed in 3-dimensional space."
  
  

---

Re the naming conventions note, would you mind sharing the link to those so I can also update my plugins to adhere to the recommended naming conventions? Thanks!

brandonmcconnell:
@sambauers Personally, I am all for TailwindCSS switching to the more modern approach, but that's a bigger change which—as you pointed out—would produce breaking changes for those already building around the current approach.

This PR has no conflicts with the current methodology, making it a no-op, whereas introducing the newer transform properties would require _at least_ an opt-in option that defaults to the legacy format to prevent any breaking changes, essentially the inverse of how your package is currently configured.

In that way, this PR is actually the first essential step toward introducing the new transform properties.

adamwathan:
Circling back to this briefly just because found myself deep in the `perspective` research hole again the other day and don't want to lose my work — right now I think I like this for the perspective scale:

| Class | Value |
| --- | --- |
| `perspective-none` | `none` |
| `perspective-dramatic` | `100px` |
| `perspective-near` | `300px` |
| `perspective-standard` or `perspective-normal` | `500px` |
| `perspective-midrange` | `800px` |
| `perspective-distant` | `1200px` |

I like that this keeps the options really limited and easy to pick through for people who don't want to get lost nitpicking on this stuff, and that we have arbitrary values as an easy escape hatch for anyone who needs a specific value 👍

brandonmcconnell:
@adamwathan Yeah, I like that. I updated the PR to reflect that (using "normal", vs. "standard") and also added a `DEFAULT` value, which reflects the same normal value.

I think it could be intuitive that using `perspective` by itself without a value adds that normal value, but I could see a string case for either removing the `normal` value in favor of `DEFAULT` (no value needed), or removing `DEFAULT` in favor of `normal`. Let me know.

---

Other than that, I think this PR is in pretty good shape. It was a bit hard to test the CPU/GPU opt in/out logic without being able to run this exactly as is in a remote environment. StackBlitz and CodeSandbox don't seem to play nice with Tailwind CSS npm forks, but it builds fine for me locally.

It would just help to have someone on the team do a once-over on it to ensure that CPU/GPU bit works as expected. It might require a small patch.

ShingLi:
is there any result ?

KrisBraun:
Hey all! Thanks @brandonmcconnell for drafting this, and everyone else for the input. We're implementing this in Tailwind CSS v4 with #13248. As you'll see, we're closely following what was discussed here, with a few small changes:

1. All transforms are done using specific CSS properties, rather than `transform` functions. This prevents conflicts updating the `transform` property, and makes it easier to see in the inspector.
2. `rotate-x`, `rotate-y`, and `rotate-z` contribute to the axis of rotation, rather than specifying separate angles of rotation. See #13248 for more details.
3. Added `scale-3d` to scale in all three dimensions.
4. Skipped `perspective-self` since it doesn't correspond to a separate CSS property. Feel free to open a new discussion if it seems like something is missing without this.

Closing this PR in favour of #13248. Please follow along there!

PR #10988, opened 2023-04-11
Replace `__OXIDE__` at build time to prevent `@tailwindcss/oxide` leaks in the stable engine

RobinMalfait:
This PR will ensure that the oxide engine doesn't leak in the current stable version.

The oxide engine is currently behind a **runtime** feature flag:
```js
if (env.OXIDE) {
  require('@tailwindcss/oxide').doSomethingOxideSpecific()
}
```

But some build tools that use the commonjs plugin (often used in Vite / Rollup) will hoist that `require` part out of the guard... this typically results in a static import instead. But since the stable engine doesn't have any idea what `@tailwindcss/oxide` even is, it crashes.

If you would inspect the generated code, then it will look something like this:
```js
import moduleName from '@tailwindcss/oxide'

if (env.OXIDE) {
  moduleName.doSomethingOxideSpecific()
}
```

... which kind of defeats the whole purpose of the feature flag in the first place. In a project where `OXIDE` is defined this all works fine, but in other places it won't.

This PR will switch the **runtime** feature flag, to a **build time** feature flag. This will ensure that the code is not even there in the final built code but that should be fine since you can switch between the engines if you install a different version (`insiders` vs `oxide-insiders`).

We will replace `__OXIDE__` at build time. This also means that some parts of the code will now look something like `if (false) { require('@tailwindcss/oxide') }`. This means that the block will never be reachable, but luckily SWCs optimizer will get rid of the dead code entirely.


thecrypticace (approved):
LGTM

PR #11005, opened 2023-04-14
Formatting config.full.js

vjdevra:


thecrypticace:
Hey, thanks for the PR! The "full" stub is a file that gets copied into your project when using the `--full` flag in the CLI. The formatting is optimized for customization. That is the reason we're using the splat operator in the object. In case people want to add additional values to any of those keys.

Appreciate the effort on this but we're going to keep it as is. Thanks! ✨

PR #11089, opened 2023-04-25
Drop support for Node.js v12

### LINKED ISSUE #11081: TailwindCSS depends on an outdated version of postcss-load-config
TailwindCSS depends on `yaml` indirectly through `postcss-load-config`. [This CVE](https://huntr.dev/bounties/4b494e99-5a3e-40d9-8678-277f3060e96c/) was released earlier today, and was patched in `yaml@2.2.2`. It can be used by `postcss-load-config@4.0.1`, but tailwind depends on `postcss-load-config@^3.1.4`. This is causing our static analysis tools to yell at us.

RobinMalfait:
This PR drops support for Node.js v12 and updates any dependencies that were locked to old versions to their latest maintained versions. This makes Node.js v14 a minimum requirement for Tailwind CSS.

The primary motivation for this change is to update our transitive dependency on `yaml` to resolve [CVE-2023-2251](https://huntr.dev/bounties/4b494e99-5a3e-40d9-8678-277f3060e96c/). While this CVE is not really relevant to a build-time project like Tailwind, it causes warnings for lots of people who check for these sorts of things in CI (by running `npm audit` for example) and creates a lot of headache.

We were originally going to wait until Tailwind CSS v4.0 to drop support for Node.js v12, but since it is end-of-life and hasn't been receiving security updates since 2022-04-30, we've decided we're comfortable making this change sooner even though it will require upgrading to a maintained version of Node.js for people that are still using v12.

Resolves #11081.


capitalist:
For anyone wondering it's `CVE-2023-2251`  - might be worth mentioning in the changelog on the off chance that there's a second CVE in a short timeframe

levrik:
In an automated dependency upgrade PR I just saw the `@alloc/quick-lru` package being added via Tailwind. I couldn't really find the source of this package (it seems to be a fork of `quick-lru`?) so it looked a bit suspicious to me. This PR seems to have replaced it but I can't find a reason stated here for the switch.

PR #11102, opened 2023-04-26
Log error for default config file not found.

LorenzoBloedow:
So here's the story:
After spending probably about 2 hours stuck on a bug where Tailwind would only log:
```
warn - No utility classes were detected in your source files. If this is unexpected, double-check the `content` option in your Tailwind CSS configuration.
warn - https://tailwindcss.com/docs/content-configuration
```
I finally realized that it was because I was using `npx tailwindcss` in an npm script (thus executing from package.json) while my tailwind config was actually nested in a lower directory (a test server).
When I went to the code that resolved the default path, I noticed that it wasn't logging anything, so I thought it'd be really helpful to log what happened so that it, hopefully, will save someone else some time in the future!

### DIFF
--- a/src/util/resolveConfigPath.js
+++ b/src/util/resolveConfigPath.js
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import path from 'path'
+import log from './log'
 
 const defaultConfigFiles = [
   './tailwind.config.js',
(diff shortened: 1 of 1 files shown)


RobinMalfait:
We made some changes to `master` and had to force push which is why your branch looks a bit funny now. I don't have push rights on your branch, so if you want to fix it yourself then you can pull down the latest `master` branch and run `git rebase master -X ours`.

LorenzoBloedow:
> `you can pull down the latest master branch and run git rebase master -X ours`.
I'm not gonna lie, currently I'm not the best at git but I gave it my best shot, hopefully didn't mess anything up

LorenzoBloedow:
@RobinMalfait can we merge this?

PR #11122, opened 2023-04-28
Sort classes width then height

PixelJanitor:
Dimensions are pretty much universally documented as `{width} x {height}` but due to the current ordering here the prettier plugin sorts these classes as `{height} x {width}`.

This PR swaps those orderings to align with the universal size descriptions 📏


RobinMalfait:

It looks like the PR is incomplete and only updates some of the test files and not the real implementation. That said, we have an open PR where we updated all the plugins with the new order (not only the width & height order).

That PR can be found here: #10947

Going to close this PR as a duplicate of the other one 💪

PixelJanitor:
Ah yea - been meaning to ask you about the final push on this to see what I was missing. I didn't understand how things compiled to detect the sort order but knew you would. All good on the close since it's elsewhere!

PR #11319, opened 2023-05-30
Fix source maps issue resulting in a crash

RobinMalfait:
This PR fixes an issue where source maps weren't properly working. This is because we started using Lightning CSS in between our PostCSS plugins.

The implementation of Lightning CSS currently re-parses the CSS therefore losing the original information from the `root` tree. After that we reset the result.root based on the result of Lightning CSS.

This PR should resolve the source map issues by passing the previous source maps to Lightning CSS, and passing the resulting source maps from Lightning CSS back to PostCSS.


tests/source-maps.test.js
```diff
@@ -38,33 +38,34 @@ test.skip('apply generates source maps', async () => {
...
-    '9:8-33 -> 15:8-63',
-    '10:6 -> 16:6',
-    '13:8 -> 18:6',
-    '13:8-31 -> 19:8-20',
-    '13:8-31 -> 20:8-19',
-    '13:8-31 -> 21:8-26',
-    '13:8 -> 22:8',
-    '13:31 -> 23:0',
-  ])
+  expect(annotations).toMatchInlineSnapshot(`
```
RobinMalfait, line 43: Now that we don't use `crosscheck` anymore, we can use `toMatchInlineSnapshot` again!

tests/source-maps.test.js
```diff
@@ -38,33 +38,34 @@ test.skip('apply generates source maps', async () => {
   // All CSS generated by Tailwind CSS should be annotated with source maps
   // And always be able to point to the original source file
   expect(sources).not.toContain('<no source>')
-  expect(sources.length).toBe(1)
+  expect(sources.length).toBe(2)
```
RobinMalfait, line 41: All of these are increased by 1. If you (temporarily) set the `minify: false` to `minify: true` in the lightningCssPlugin then it goes down by 1 again.

tests/source-maps.test.js
thecrypticace: 🎉🎉🎉

tests/source-maps.test.js
thecrypticace: the number of sources should definitely just be `1`

Introducing additional sources that disappear during minification sounds like a lightningcss bug. I'll set up an isolated test case for that and file a bug report.

thecrypticace (approved):
aside from the source path issues which is probs a lightning bug this LGTM. We can merge it and take care of those later.

PR #11548, opened 2023-07-04
Make PostCSS plugin async to improve performance

RobinMalfait:
This PR improves the performance by making the main plugin async. This means that we can replace a bunch of `fs.readFileSync` calls for each template file with a `Promise.all(f => fs.promises.readFile(f))` instead, without blocking IO.


codemile:
FYI: This change broke support for NativeWind on React Native. Triggers a "Use process(css).then(cb) to work with async plugins" error when running 3.3.3.

Any insight from the Tailwind team on the reasons? I suspect the error is being thrown by Babel.

lahaxearnaud:
FYI: This change also break the postCss plugin postcss-multiple-tailwind.

robbevp:
> FYI: This change also break the postCss plugin postcss-multiple-tailwind.

I created postcss-multiple-tailwind, but since tailwindcss v2.3 there has been internal support for this functionality. You should use that instead of my plugin.

Jorge-Ortiz-Mata:
Error! 
This PR did something wrong! I'm getting an internal error server.
I'm not sure what's going on but it must be solved.

Grenish:
`Use process(css).then(cb) to work with async plugins` is what I'm getting.

PR #11670, opened 2023-07-24
Changed the border width

AnuragNishad12:
Just changed the border-width in preflight.css file

### DIFF
--- a/src/css/preflight.css
+++ b/src/css/preflight.css
@@ -7,7 +7,7 @@
 ::before,
 ::after {
   box-sizing: border-box; /* 1 */
-  border-width: 0; /* 2 */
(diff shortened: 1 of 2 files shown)


RobinMalfait:

This would be a breaking change, but I would suggest to add it to your own `@layer base` configuration if you want to have this change.

The goal of the `border-width: 0` is that every element already has a border and a border color. Adding the `border` class sets a border with and the corresponding border color will already be shown.

A `border-width: 2` is invalid syntax, you probably want `2px`, but if you do this, then the result will be that every element on the page has a 2px border. e.g.: https://play.tailwindcss.com/s1I3GhKTdh

PR #11692, opened 2023-07-26
Add `prefers-reduced-transparency` variants

lukewarlow:
Adds support for the [`prefers-reduced-transparency`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-transparency) media query.

I've implemented and filed a Chrome Intent to Ship for this feature: https://groups.google.com/a/chromium.org/g/blink-dev/c/EiZ2vlPy9xs


lukewarlow:
No worries if you'd rather wait for the 3 LGTM on the I2S. I'll comment here to let you know when I unflag the support.

lukewarlow:
Got the 3 approvals to ship so this media query will be coming to chrome 118

RobinMalfait:
Hey!

PR #11693, opened 2023-07-26
Add `inverted-colors` variant

lukewarlow:
Add a variant for the [`inverted-colors`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/inverted-colors) media query.

This has been supported in Safari for a while. I'm also implementing support in Chrome atm.

I've decided to only add the inverted-colors: inverted variant because the `none` state isn't really useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


RobinMalfait:
Hey!

philipp-spiess (approved):
Thanks for the PR! We're landing this in `main` behind a feature flag and plan to release this with `4.1` later 👍

PR #11694, opened 2023-07-26
Add `forced-colors` variant

lukewarlow:
Add a variant for the [`forced-colors`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) media query.

Also add a contrast-custom variant to match [custom contrast](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast#:~:text=level%20of%20contrast.-,custom,-Indicates%20that%20user) preferences

Like with #11693 i've only added a variant for `forced-colors: active` because the `forced-colors: none` case isn't overly useful.

This has widespread browser support across multiple platforms

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -459,7 +459,6 @@ export let variantPlugins = {
   prefersContrastVariants: ({ addVariant }) => {
     addVariant('contrast-more', '@media (prefers-contrast: more)')
     addVariant('contrast-less', '@media (prefers-contrast: less)')
-    addVariant('contrast-custom', '@media (prefers-contrast: custom)')
   },
 
   forcedColorsVariants: ({ addVariant }) => {
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -459,6 +459,11 @@ export let variantPlugins = {
   prefersContrastVariants: ({ addVariant }) => {
     addVariant('contrast-more', '@media (prefers-contrast: more)')
     addVariant('contrast-less', '@media (prefers-contrast: less)')
+    addVariant('contrast-custom', '@media (prefers-contrast: custom)')
+  },
+
+  forcedColorsVariants: ({ addVariant }) => {
+    addVariant('forced-colors', '@media (forced-colors: active)')
   },
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add utilities for CSS subgrid ([#12298](https://github.com/tailwindlabs/tailwindcss/pull/12298))
 - Add spacing scale to `min-w-*`, `min-h-*`, and `max-w-*` utilities ([#12300](https://github.com/tailwindlabs/tailwindcss/pull/12300))
 - Add `forced-color-adjust` utilities ([#11931](https://github.com/tailwindlabs/tailwindcss/pull/11931))
+- Add `forced-colors` variant ([#11694](https://github.com/tailwindlabs/tailwindcss/pull/11694))
 - [Oxide] New Rust template parsing engine ([#10252](https://github.com/tailwindlabs/tailwindcss/pull/10252))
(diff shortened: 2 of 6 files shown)


src/corePlugins.js
```diff
@@ -459,6 +459,11 @@ export let variantPlugins = {
   prefersContrastVariants: ({ addVariant }) => {
     addVariant('contrast-more', '@media (prefers-contrast: more)')
     addVariant('contrast-less', '@media (prefers-contrast: less)')
+    addVariant('contrast-custom', '@media (prefers-contrast: custom)')
```
lukewarlow, line 462: It would be useful to also have a variant for `@media (prefers-contrast)` in a boolean context. But I couldn't think of a good name for it? `contrast-specified`?

To quote the [spec](https://drafts.csswg.org/mediaqueries-5/#prefers-contrast):

>it is also common to reduce visual clutter and color complexity in response to both high and low contrast preferences. In that case, it is appropriate to use @media (prefers-contrast) { … } without specifying [more](https://drafts.csswg.org/mediaqueries-5/#valdef-media-prefers-contrast-more) or [less](https://drafts.csswg.org/mediaqueries-5/#valdef-media-prefers-contrast-less), to do things like replacing background images with plain colors, turning off decorative gradients, or replacing border images or box shadows with simple solid borders. As [prefers-contrast: custom](https://drafts.csswg.org/mediaqueries-5/#descdef-media-prefers-contrast)—like prefers-contrast: more or prefers-contrast: less—evaluates to true in a [boolean context](https://drafts.csswg.org/mediaqueries-5/#boolean-context), such simplifications would also benefit users of [forced colors mode](https://drafts.csswg.org/css-color-adjust-1/#forced-colors-mode), even when their colors of choice do not result in a particularly high or low contrast. This is desirable, as the reduced palette enforced by forced colors mode calls for some visual simplification of the page.

RobinMalfait:
Hey!

Thanks for the PR. I did drop the `contrast-custom` from this PR and changed the order of the variants plugin slightly. Apart from this everything looks good.

This will be available in the next release but you can already try it using the insiders version.

devongovett:
hey @RobinMalfait @lukewarlow! Thanks for adding this. I was trying it out, and wondered if the order of the media queries in the generated CSS should be changed. For example, currently if you have `bg-red-400 dark:bg-red-800 forced-colors:bg-black`, the dark mode variant wins over forced colors. This is the generated CSS:

```css
.bg-red-400 {
  --tw-bg-opacity: 1;
  background-color: rgb(248 113 113 / var(--tw-bg-opacity));
}

@media (forced-colors: active) {
  .forced-colors\:bg-black {
    --tw-bg-opacity: 1;
    background-color: rgb(0 0 0 / var(--tw-bg-opacity));
  }
}

  .dark\:bg-red-800 {
    --tw-bg-opacity: 1;
    background-color: rgb(153 27 27 / var(--tw-bg-opacity));
  }
}
```

I think forced colors should basically always win. Is there a way to have this automatically pushed to the end so we don't need to also add a redundant `dark:forced-colors:bg-black` class?

lukewarlow:
@devongovett Forced-colors mode could result in a light theme or a dark theme (and a high or low or custom contrast). So you could end up needing to set a separate class for forced-colors with dark mode than without. Having said that provided the dark:forced comes after normal forced, your change probably makes sense.

@RobinMalfait I think https://github.com/tailwindlabs/tailwindcss/pull/11694/commits/d620471d4d7c6d90f235a80c381f644523c162cf is what changed that ordering, was this deliberate or was it just for neatness?

lukewarlow:
I've made a PR to do this so it's ready to go if the team want to make the change. https://github.com/tailwindlabs/tailwindcss/pull/12582

Depends on whether we want to be consistent with contrast etc or not I guess.

RobinMalfait:
Yep you are totally correct. @lukewarlow we moved it because it made more sense to group the related variants together, but this does indeed break the dark mode case. Good catch! Will take a look at your PR tomorrow, thanks 🙏

PR #11709, opened 2023-07-28
Fix incorrectly generated CSS when using square brackets inside arbitrary properties

### LINKED ISSUE #11241: Arbitrary CSS variable with variant prefix
3.3.2

Not sure, simulated on Tailwind Play though.

See above.

Doesn't matter.

Doesn't matter.

**Reproduction URL**
https://play.tailwindcss.com/uGZbOX6jll

**Describe your issue**
When using arbitrary CSS variables as seen in docs https://tailwindcss.com/docs/adding-custom-styles#arbitrary-properties, there is an issue when:
- Variant prefix for media is used (may be a problem with other variants as well, not sure)
- and theme function is used using `[]` to retrieve value.

E.g. these classes does work properly
✔️ `[--bar:theme('spacing[1.5]')]` - even though `[]` is used, it still works because there is no variant.
✔️ `md:[--foo:theme('spacing.2')]` - variant is used, but dot notation for value.

But this one doesn't:
❌ `md:[--bar:theme('spacing[2.5]')]` - `[]` and variant is used together, resulting in malformed CSS output.

All three classes are used in demo above, where first two works as intended, but the last one does not.

RobinMalfait:
This PR fixes an issue where using square brackets inside arbitrary properties could result in incorrect CSS.

If you have the following class: `md:[--bar:theme('spacing[2.5]')]`
Then the output looked like this:
```css
@media (min-width: 768px) {
  {
    --bar: 0.625rem;
  }
}
```

With this fix in place, the output looks like this:
```css
@media (min-width: 768px) {
  .md\:\[--bar\:theme\(\'spacing\[2\.5\]\'\)\] {
    --bar: 0.625rem;
  }
}
```

Fixes: #11241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow variant to be an at-rule without a prelude ([#11589](https://github.com/tailwindlabs/tailwindcss/pull/11589))
 - Improve normalisation of `calc()`-like functions ([#11686](https://github.com/tailwindlabs/tailwindcss/pull/11686))
 - Skip `calc()` normalisation in nested `theme()` calls ([#11705](https://github.com/tailwindlabs/tailwindcss/pull/11705))
+- Fix incorrectly generated CSS when using square brackets inside arbitrary properties ([#11709](https://github.com/tailwindlabs/tailwindcss/pull/11709))
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


ktmn:
Should it work at play.tailwindcss.com if I set the version to `Insiders (b6b3f85)`?

Currently I get `Error: Invalid empty selector` with these classes:

```html
<div class="border p-[--var] [--var:1rem] md:[--var:8rem]">Works</div>
<div class="border p-[--var] [--var:theme(spacing[4])] md:[--var:theme(spacing[32])]">Doesn't</div>
```

PR #11730, opened 2023-08-01
Make `content` optional for presets in TypeScript types

### LINKED ISSUE #11725: Typescript type for `presets` should allow config without `content` property

3.3.2

postcss 8.3.1, Next.js 13.4.9, webpack 5.88.2

For example: v18.16.0

N/A

vscode devcontainer from `mcr.microsoft.com/vscode/devcontainers/typescript-node:0-18-bullseye`

**Reproduction URL**

https://codesandbox.io/s/upbeat-engelbart-srwsmw?file=/src/tailwind.preset.ts

**Describe your issue**

`content` is required in preset, but preset configs sometimes/often/generally should not override the `content` property.

[edit] Updated the code-sandbox slightly to better illustrate the point.

The (js) example on the tailwind website does not have a `content` property.
https://tailwindcss.com/docs/presets

Currently the relevant typing is:
```ts
interface OptionalConfig {
  // ...
  presets: Partial<PresetsConfig>;
  // ...
}

type PresetsConfig = Config[];
```

Instead, I think it should be:

```ts
interface OptionalConfig {
  // ...
  presets: PresetsConfig;
  // ...
}

type PresetsConfig = Partial<Config>[];
```

(I assume the intent was to make array members partial, instead of making the array itself partial; as I cannot come up with any reason why the array itself should be partial/would benefit from being partial)

wimbarelds:
Instead of `Partial<Array<Thing>>` have `Array<Partial<Thing>>`

Fixes: https://github.com/tailwindlabs/tailwindcss/issues/11725

Also applies the same fix to safelist and blocklist properties of Config type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve normalisation of `calc()`-like functions ([#11686](https://github.com/tailwindlabs/tailwindcss/pull/11686))
 - Skip `calc()` normalisation in nested `theme()` calls ([#11705](https://github.com/tailwindlabs/tailwindcss/pull/11705))
 - Fix incorrectly generated CSS when using square brackets inside arbitrary properties ([#11709](https://github.com/tailwindlabs/tailwindcss/pull/11709))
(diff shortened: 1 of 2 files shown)


RobinMalfait:

I made a few small adjustments, but these types make more sense now, thanks!

Issue #11090, opened 2023-04-25
jit mode compilet error with font-size


hellosanbao:
I  set the `font-size` like it: `text-[30rpx]`.
I expect get`font-size:30rpx`,but the result is :`color:30rpx`,
It must end width 'px',but i want set 'rpx'
version: 3.3.1

Mrlilili:
@hellosanbao How was it solved in the end?

Simon-He95:
> @hellosanbao How was it solved in the end?

you can use text-[length:123rpx], or try this vscode plugin when saving,auto transform text-16rpx to text-[length:16rpx]
https://github.com/Simon-He95/vscode-tailwind-magic

PR #11787, opened 2023-08-10
feat: Compatible with applet rpx

Simon-He95:
#11090

### DIFF
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -146,6 +146,7 @@ let lengthUnits = [
   'pc',
   'pt',
   'px',
+  'rpx',
   'em',
   'ex',
   'ch',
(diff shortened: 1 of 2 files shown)


xiBread:
`rpx` isn't a standard CSS unit. You can just do `text-[length:24rpx]` or `[font-size:24px]`.

RobinMalfait:

I'm going to say no for now because as @xiBread mentioned the `rpx` unit isn't a standard CSS unit. Luckily there is a workaround by forcing the current type for the arbitrary values:

```html
<div class="text-[length:123px]"></div>
```

We will revisit this if more people want this feature out of the box or if CSS makes it a recognized length unit.

Thanks!

Simon-He95:
All right, thanks for your reply. I will do it in [vscode-tailwind-magic](https://github.com/Simon-He95/vscode-tailwind-magic), text-123rpx -> text-[length:123rpx]

屏幕录制2023-08-11 00 10 44

Simon-He95:
Hi，@RobinMalfait, by the way, Do you think this plugin can be linked to the homepage of tailwindcss, I think it will help people who use tailwindcss

PR #11808, opened 2023-08-12
Obey EditorConfig

szepeviktor:
Now all our bytes pass [`eclint`](https://github.com/szepeviktor/byte-level-care/blob/db9d972a6c789a190ed3faa0de701b1b494b0b23/.github/workflows/reusable-integrity.yml#L78-L83).


.editorconfig
```diff
@@ -7,11 +7,17 @@ end_of_line = lf
...
+block_comment_end = */
 
 [*.rs]
-indent_style = space
 indent_size = 4
-end_of_line = lf
-charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
+#max_line_length = 120
```
szepeviktor, line 16: This is for the future. To limit our liberties.

oxide/crates/node/.gitignore
```diff
@@ -121,7 +121,7 @@ dist
 .AppleDouble
 .LSOverride
 
-# Icon must end with two
```
szepeviktor, line 124: There was a Windows line end (CR) here.

szepeviktor:
Do these changes make `tailwindcss-cli` tests fail?
https://github.com/tailwindlabs/tailwindcss/pull/11808/files#diff-f9f898fc2134d83b8c635c6c9be912c5dce99cda88e2f32c13576a600cab18f3

oxide/crates/node/.gitignore
thecrypticace: Yea, that one is on purpose. Though in the comment you could just say `\r`. Line 125 `Icon` should end with two windows line endings as well. You can see more about the file here: https://superuser.com/questions/298785/icon-file-on-os-x-desktop

thecrypticace: (two are required because of a quirk of gitignore — even though the filename is spelled `Icon\r`)

szepeviktor: Thank you @thecrypticace!
Added it name: Icon must end with two CR characters
https://en.wikipedia.org/wiki/ASCII#End_of_line

reinink:
Hey thanks for this contribution. I've just done a quick review of these changes, and a bunch of things are actually wrong. For example, our CLI indentation uses 3 spaces, not 2 spaces, so that change is incorrect. Similarly, the indentation change in Preflight is also wrong.

While we always very much appreciate PRs to improve the codebase, quite often sweeping code formatting changes like this simply create unnecessary work for us as we need to carefully review all the changes to make sure things are correct, and overall don't add a ton of value to the project.

Because of that I'm going to this for now, but thanks either way for the contribution 👍

PR #11833, opened 2023-08-15
Enhancing Text Clarity: The Impact of the `font-smooth` Property in `preflight.css` 📝

sanjaiyan-dev:
Adding the `font-smooth` property in `preflight.css` helps render text with grayscale anti-aliasing instead of subpixel rendering. Switching from subpixel rendering to anti-aliasing for light text on dark backgrounds gives it a lighter appearance. This property has currently been added in Material-UI. For reference, you can check the following link: https://github.com/mui/material-ui/blob/ddf840bf3cc72ff863d91dabf65408fed3f5d046/packages/mui-material/src/CssBaseline/CssBaseline.js#L8.

I apologize if I've made any mistakes :(

### DIFF
--- a/src/css/preflight.css
+++ b/src/css/preflight.css
@@ -39,11 +39,15 @@ html {
 /*
 1. Remove the margin in all browsers.
 2. Inherit line-height from `html` so users can set them as a class directly on the `html` element.
+3. Smooth the font on the level of the pixel, as opposed to the subpixel. Switching from subpixel rendering to anti-aliasing for light text on dark backgrounds makes it look lighter.
+4. Render text with grayscale anti-aliasing, as opposed to the subpixel. Switching from subpixel rendering to anti-aliasing for light text on dark backgrounds makes it look lighter.
 */
 
 body {
   margin: 0; /* 1 */
   line-height: inherit; /* 2 */
+  -webkit-font-smoothing: antialiased; /* 3 */
+  -moz-osx-font-smoothing: grayscale; /* 4 */
 }
 
 /*


xiBread:
There are already [font smoothing utilities](https://tailwindcss.com/docs/font-smoothing). I don't see the need to add them to the preflight.

sanjaiyan-dev:
> 

I appreciate your perspective on font smoothing utilities. I thought that incorporating font smoothing by default, similar to material UI design, could enhance the CSS touch. However, I understand your point and apologize if I may have misunderstood its necessity. I'm open to discussing this further. Thank you for the constructive input.

sanjaiyan-dev:
> There are already [font smoothing utilities](https://tailwindcss.com/docs/font-smoothing). I don't see the need to add them to the preflight.

I appreciate your perspective on font smoothing utilities. I thought that incorporating font smoothing by default, similar to material UI design, could enhance the CSS touch. However, I understand your point and apologize if I may have misunderstood its necessity. I'm open to discussing this further. Thank you for the constructive input.

reinink:
Hey! Thanks for this suggestion, and while I kind of like the idea (we basically always add the `antialiased` class to the body of our projects here at Tailwind Labs), I'm concerned that this is too opinionated of a change for Preflight given that not everyone agrees with doing this:

- https://github.com/google/fonts/issues/1170 (Google has actually stopped doing this on Google Fonts)
- https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

Given that, and the fact that this is trivial to enable on your own projects, I think it's best that we leave this out of Preflight at this time.

```html
<body class="antialiased">
```

Hope that makes sense! 🤙

PR #11917, opened 2023-08-28
[Issue] font-[inherit] is assigned to font-family

mathieutu:
Hey folks.

Today I tried to `font-[inherit]` (with prose), and saw that the generated style is font-family.
I'd think more relevant to have it in font-weight. 
WDYT?

More generaly, how have you though this kind of conflict?

Thanks.

In the commit, the failing test.

See you. 
👋
Mathieu.

### DIFF
--- a/tests/plugins/fontWeight.test.js
+++ b/tests/plugins/fontWeight.test.js
@@ -6,6 +6,7 @@ quickPluginTest('fontWeight', {
     'font-[bold]',
     'font-[650]',
     'font-[var(--my-value)]',
+    'font-[inherit],
   ],
 }).toMatchFormattedCss(css`
   .font-\[650\] {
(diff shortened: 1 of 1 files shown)


tests/plugins/fontWeight.test.js
```diff
@@ -6,6 +6,7 @@ quickPluginTest('fontWeight', {
     'font-[bold]',
     'font-[650]',
     'font-[var(--my-value)]',
+    'font-[inherit],
```
mathieutu, line 9:
```suggestion
    'font-[inherit]' ,
```

adamwathan:
Hey! Unfortunately this would be a breaking change so not something we can do 😕 The general solution to this stuff though is to use type-hints:

https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities

In this case you want to hint `number` since that's the data type the `font-weight` property accepts in CSS:

image

Sort of a weird/unintuitive one in this case admittedly.

PR #11931, opened 2023-08-30
Add `forced-color-adjust` utilities

lukewarlow:
Adds utilities for [`forced-colors-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/forced-color-adjust)


lukewarlow:
This is linked to the forced-colors functionality that https://github.com/tailwindlabs/tailwindcss/pull/11694 also helps to expose in Tailwind.

RobinMalfait:
Hey!

Thanks for the PR, I've updated the name to be `forced-color-adjust` (singular `color`) instead to match the property.

lukewarlow:
Ah good catch I must have got confused with the media query.

PR #11954, opened 2023-09-05
Remove matrix.setup from release-insiders.yml

szepeviktor:
`setup` was never part of the matrix.


thecrypticace:
It was part of it until the end of May — but yeah it can be removed now. Thanks.

PR #12051, opened 2023-09-20
Make `dark:` variant fully customizable.

MichaelAllenWarner:
This PR adds a `'custom'` strategy to the dark-mode config, in order to allow users to fully customize the `dark:` definition. It works like this:

```javascript
module.exports = {
  darkMode: [
    'custom',
    [
      '@media (prefers-color-scheme: dark) { &:not(:root[data-mode="light"] *) }',
      '&:is(:root[data-mode="dark"] *)',
    ],
  ],
}
```

Note that this is extremely similar to the custom-selector flavor of the `'class'` strategy that already exists (i.e., `darkMode: ['class', '[data-mode="dark"]']`). The only difference is that with the `'custom'` strategy, the second element of the `darkMode` array is used _verbatim_ as the second argument of the `addVariant('dark',)` call.

The [rationale](https://github.com/tailwindlabs/tailwindcss/discussions/11534) for this PR is that there's currently no way for users to make their own custom `dark:` variant. This is because the core `dark:` variant is registered _after_ user-provided variants, which means that a custom `addVariant('dark',)` does nothing. And while the custom-selector flavor of the `'class'` strategy is powerful, it's not enough to let you use the media-query as a fallback when JavaScript is disabled; for that, you need something like the selectors in my snippet above.

I've added a test for the new strategy and updated the `DarkModeConfig` type.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -213,7 +213,7 @@ export let variantPlugins = {
   },
 
   darkVariants: ({ config, addVariant }) => {
-    let [mode, className = '.dark'] = [].concat(config('darkMode', 'media'))
+    let [mode, classNameOrAddVariantDefinition = '.dark'] = [].concat(config('darkMode', 'media'))
 
     if (mode === false) {
       mode = 'media'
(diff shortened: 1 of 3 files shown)


MichaelAllenWarner:
Follow-up: @olets has [shown me ](https://github.com/tailwindlabs/tailwindcss/discussions/11534#discussioncomment-7869682) that setting `darkMode: null` successfully disables Tailwind's core `dark:` variant, allowing the user to make an entirely custom variant. This takes care of the problem for me, so I'm going to close this PR.

PR #12149, opened 2023-10-05
Add `transition-behavior` utilities

lukewarlow:
[`transition-behavior` ](https://drafts.csswg.org/css-transitions-2/#transition-behavior-property) is a new CSS property that allows you to transition on discrete properties. 

This is new in Chrome 117. See https://chromestatus.com/feature/5071230636392448

One major use case for this is styling dialog and popover transitions when showing. (See comment for exit transitions)

See https://github.com/tailwindlabs/tailwindcss/pull/12040 for a related PR to add a `@starting-style` variant.

### DIFF
--- a/packages/tailwindcss/src/__snapshots__/intellisense.test.ts.snap
+++ b/packages/tailwindcss/src/__snapshots__/intellisense.test.ts.snap
@@ -1717,6 +1717,8 @@ exports[`getClassList 1`] = `
   "transform-view",
   "transition",
   "transition-all",
+  "transition-allow-discrete",
+  "transition-behavior-normal",
   "transition-colors",
   "transition-none",
   "transition-opacity",
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -10475,6 +10475,19 @@ test('transition', () => {
   ).toEqual('')
 })
 
+test('transition-behavior', () => {
+  expect(run(['transition-allow-discrete', 'transition-behavior-normal'])).toMatchInlineSnapshot(`
+    ".transition-allow-discrete {
+      transition-behavior: allow-discrete;
+    }
+
+    .transition-behavior-normal {
+      transition-behavior: normal;
+    }"
+  `)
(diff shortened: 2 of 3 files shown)


src/corePlugins.js
```diff
@@ -2860,6 +2860,12 @@ export let corePlugins = {
     )
   },
 
+  transitionBehavior: ({ addUtilities }) => {
+    addUtilities({
+      '.transition-allow-discrete': { 'transition-behavior': 'allow-discrete' },
+      '.transition-behavior-normal': { 'transition-behavior': 'normal' },
+    })
+  },
```
lukewarlow, line 2868: Not sure on the exact best class names for this?

lukewarlow:
It would also be good to add a utility for `transition: overlay`. https://drafts.csswg.org/css-position-4/#overlay

The overlay property shipped in Chrome 117 See https://chromestatus.com/feature/5138724910792704

While transition-behavior allows styling entry animations for popover and dialog for exit animations you need transition-property to include `overlay`

See https://developer.chrome.com/en/blog/new-in-chrome-117/#exit-entry-animations for a good example.

lukewarlow:
See https://developer.chrome.com/blog/entry-exit-animations/ for a good overview on all these new capabilities in Chrome

brandonmcconnell:
@lukewarlow I've been wanting a CSS property like that for years! 🙌🏼

For a Tailwind CSS implementation, I think we may want to wait until at least all modern browsers implement the feature:
https://caniuse.com/mdn-css_properties_transition-behavior

lukewarlow:
https://github.com/WebKit/standards-positions/issues/148

WebKit have a positive standards position on this and look like they have it implemented as well.

lukewarlow:
This is part of Interop 2024 so should be in all 3 browsers by the end of the year

lukewarlow:
I have no clue what's happened to the branch and why there's so many conflicts and commits but I imagine it's a simple fix

RobinMalfait:
Hey!

brandonmcconnell:
Congrats, @lukewarlow!

philipp-spiess:
Looks like this was added to v4 in https://github.com/tailwindlabs/tailwindcss/pull/15051.

PR #12171, opened 2023-10-09
Add `font-stretch` utilities

Gregory-Gerard:
Hey there!

First-time contributor to Tailwind CSS, if there's anything specific you'd like me to adjust or if you have any questions, please don't hesitate to let me know. 

**Goal**

I've added the `.stretch-*` classes to manage the `font-stretch` property. This addition is approximately 96% compatible overall, as per the information available on [caniuse.com](https://caniuse.com/).

Stretch should allow percentage value or any config value. I've relied on the `fontWeight` property for this new feature. I hesitated between adding statics utilities or the version I sent, but I think that being able to use arbitrary values and extend configurations is interesting because this property accepts percentage values in addition to global values.

I found myself deliberating between using `font-stretch` and simply `stretch`. I would appreciate your guidance on which option is more suitable for this.

**Documentation**

Additional documentation of `font-stretch` property on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch)

If this feature is accepted, I could also take care of creating the corresponding documentation in https://github.com/tailwindlabs/tailwindcss.com.

**Related discussions**

https://github.com/tailwindlabs/tailwindcss/discussions/10156
https://github.com/tailwindlabs/tailwindcss/discussions/5686
https://github.com/tailwindlabs/tailwindcss/discussions/8324

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -2072,6 +2072,10 @@ export let corePlugins = {
     type: ['lookup', 'number', 'any'],
   }),
 
+  fontStretch: createUtilityPlugin('fontStretch', [['stretch', ['fontStretch']]], {
+    type: ['lookup', 'percentage', 'any'],
+  }),
+
   textTransform: ({ addUtilities }) => {
     addUtilities({
       '.uppercase': { 'text-transform': 'uppercase' },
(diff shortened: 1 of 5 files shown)


KrisBraun:
Hey @Gregory-Gerard,

Thanks for contributing! These utilities will be helpful. With Tailwind CSS v4 coming, we'll add new feature there. I've opened #13153 to add `font-stretch` utilities to v4.

Thanks again for contributing!

PR #12208, opened 2023-10-15
Allow plugins from a parent document to be used in an iframe

### LINKED ISSUE #11899: TailwindCDN breaks when loading a custom plugin within an iframe

v3.7.2

**What build tool are you using?**

parcel v2.0.0, but it seems unrelated (also tried on esbuild v0.17.19)

v14.18.1 (also tried on v18.14.2)

Chrome, Firefox

macOS

**Reproduction URL**

https://codesandbox.io/s/tailwindcdn-with-cusotm-plugins-in-iframe-gj9fkl

**Describe your issue**

TailwindCDN breaks when loading a custom plugin within an iframe.

[It works flawlessly with custom plugins outside iframes, as you can see at this CodeSandbox link](https://codesandbox.io/s/tailwindcdn-with-cusotm-plugins-6tqwgj?file=/index.mjs). Here, the logic is similar to  one of the bug reproduction – the assigned `tailwind.config` is the same.

However, when you place it within an iframe, it throws "raws is undefined" from `...t[1].raws.tailwind` in the producion code.

The issue seems to be adding plugins. If you uncomment the example custom plugin from the plugins array in the bug reproduction code (`/index.mjs`, line 23), it loads properly.

Am I doing something wrong? I'd be happy to contribute.

2hu12:
Fix https://github.com/tailwindlabs/tailwindcss/issues/11899, to make `isPlainObject` util could work across iframe.

Another [example](https://github.com/vercel/satori/pull/561) that uses the `cdn.tailwindcss.com` in iframe, we have to use  `Object.create(null)` as workaround.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve RegEx parser, reduce possibilities as the key for arbitrary properties ([#12121](https://github.com/tailwindlabs/tailwindcss/pull/12121))
 - Fix sorting of utilities that share multiple candidates ([#12173](https://github.com/tailwindlabs/tailwindcss/pull/12173))
 - Ensure variants with arbitrary values and a modifier are correctly matched in the RegEx based parser ([#12179](https://github.com/tailwindlabs/tailwindcss/pull/12179))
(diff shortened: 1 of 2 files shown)


thecrypticace:
Thanks — this was a much simpler fix for that problem than I expected. It'll be available in out insiders build in probably about 15 minutes or so:
```
npm install tailwindcss@insiders
```

PR #12236, opened 2023-10-18
Improve automatic `var` injection

### LINKED ISSUE #12205: `var()` is incorrectly wrapped around dashed idents with CSS scroll animations

For example: v3.3.3

Next.js 13.4.16

v20.6.1

Chrome

macOS

**Reproduction URL**

https://play.tailwindcss.com/mFflelRLYg

**Describe your issue**

I've been exploring using Tailwind with CSS scroll animations and I've noticed that Tailwind is incorrectly adding `var()` to specific keywords. For example, with this code:

```html
<div class="[timeline-scope:--tl]">
  <div class="[scroll-timeline:--tl]">Scroller</div>
  <div class="[animation-timeline:--tl]">Element</div>
</div>
```

It is incorrectly setting the css properties like so:

```css
timeline-scope: var(--tl);
scroll-timeline: var(--tl);
animation-timeline: var(--tl);
```

When the intended and correct output should be:

```css
timeline-scope: --tl;
scroll-timeline: --tl;
animation-timeline: --tl;
```

[MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/timeline-scope#dashed-ident)

RobinMalfait:
This PR improves the automatic `var` injection behaviour.

When using arbitrary properties such as:
```html
<div class="[timeline-scope:--foo]"></div>
```

Then the compiled CSS will be:
```css
.\[timeline-scope\:--foo\]{
  timeline-scope: var(--foo)
}
```

However, the `timeline-scope` property accepts a `<dashed-ident>` data type, which means that the value should stay as `--foo`, and not `var(--foo)`.

This PR has an exceptions list for these properties, and won't wrap the variable with `var()` for properties that accept `<dashed-ident>`.

More info:
- https://drafts.csswg.org/scroll-animations/#propdef-timeline-scope
- https://developer.mozilla.org/en-US/docs/Web/CSS/timeline-scope#dashed-ident

Fixes: #12205

### DIFF
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -11,10 +11,35 @@ function isCSSFunction(value) {
   return IS_CSS_FN.test(value)
 }
 
+// These properties accept a `<dashed-ident>` as one of the values. This means that you can use them
+// as: `timeline-scope: --tl;`
+//
+// Without the `var(--tl)`, in these cases we don't want to normalize the value, and you should add
+// the `var()` yourself.
+//
+// More info:
+// - https://drafts.csswg.org/scroll-animations/#propdef-timeline-scope
+// - https://developer.mozilla.org/en-US/docs/Web/CSS/timeline-scope#dashed-ident
+//
+const AUTO_VAR_INJECTION_EXCEPTIONS = new Set([
+  // Concrete properties
+  'scroll-timeline-name',
+  'timeline-scope',
+  'view-timeline-name',
+
+  // Shorthand properties
+  'scroll-timeline',
+  'animation-timeline',
+  'view-timeline',
+])
+
 // This is not a data type, but rather a function that can normalize the
 // correct values.
-export function normalize(value, isRoot = true) {
-  if (value.startsWith('--')) {
+export function normalize(value, context = null, isRoot = true) {
+  if (
+    (context ? !AUTO_VAR_INJECTION_EXCEPTIONS.has(context.property) : true) &&
+    value.startsWith('--')
+  ) {
     return `var(${value})`
   }
 
@@ -28,7 +53,7 @@ export function normalize(value, isRoot = true) {
           return part
         }
 
-        return normalize(part, false)
+        return normalize(part, context, false)
       })
       .join('')
   }
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -26,6 +26,7 @@ const AUTO_VAR_INJECTION_EXCEPTIONS = new Set([
   'scroll-timeline-name',
   'timeline-scope',
   'view-timeline-name',
+  'font-palette',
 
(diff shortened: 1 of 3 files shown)


RobinMalfait:
@thecrypticace what do you think about this solution?

This only solves it for arbitrary properties, and not when you use it inside of a plugin. Alternatively we can write a Lightning CSS plugin that strips the `var()` for properties that accept `<dashed-ident>`. That way _all_ properties regardless where they are coming from will be fixed.

But, not generating them in the first place seems better I think.

Also, it looks like you can't use `var(--foo)` at all, because if you then define `--foo` as `--tl` Chrome devtools don't seem to resolve these correctly, so a Lightning CSS plugin seems doable. However, if it _is_ allowed, then this solution won't work.

Example of what I mean with variables: https://play.tailwindcss.com/Y5fbjtUBHa

src/util/dataTypes.js
```diff
@@ -11,10 +11,35 @@ function isCSSFunction(value) {
...
+const AUTO_VAR_INJECTION_EXCEPTIONS = new Set([
+  // Concrete properties
+  'scroll-timeline-name',
+  'timeline-scope',
+  'view-timeline-name',
+
+  // Shorthand properties
+  'scroll-timeline',
+  'animation-timeline',
+  'view-timeline',
```
thecrypticace, line 33: `font-palette` also needs to be in this list

src/util/dataTypes.js
```diff
@@ -11,10 +11,35 @@ function isCSSFunction(value) {
...
+])
+
 // This is not a data type, but rather a function that can normalize the
 // correct values.
-export function normalize(value, isRoot = true) {
-  if (value.startsWith('--')) {
+export function normalize(value, context = null, isRoot = true) {
+  if (
+    (context ? !AUTO_VAR_INJECTION_EXCEPTIONS.has(context.property) : true) &&
```
thecrypticace, line 40: can we split this into it's own var or something. This is just kinda hard to scan.

thecrypticace: Maybe:

```js
let isVarException = context && AUTO_VAR_INJECTION_EXCEPTIONS.has(context.property)

if (value.startsWith('--') && !isVarException) { /* … */ }
```

thecrypticace (changes requested):
Looks good other than needing to add the `font-palette` property and cleanup that one if statement.

thecrypticace:
LGTM. The `rollup-sass` integration test has become super flaky with timing btw. Need to figure out how to fix that so you can ignore it.

thecrypticace (approved):

mrcsmcln:
+1 as I'm currently experiencing this issue while transitioning @property values. Currently using an underscore (e.g., `transition-[_--foo]`.

PR #12251, opened 2023-10-19
Refactor length function to utilize a pre-compiled RegExp

marco-carvalho:
This PR refactors the `length` function by extracting the regular expression used for validation into a separate, pre-compiled RegExp. This change aims to improve the performance of the function, especially in scenarios where it's called frequently, by avoiding the overhead of compiling the RegExp each time the function is called.

### DIFF
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -173,10 +173,11 @@ let lengthUnits = [
   'cqmax',
 ]
 let lengthUnitsPattern = `(?:${lengthUnits.join('|')})`
+let lengthRegExp = new RegExp(`^[+-]?[0-9]*\.?[0-9]+(?:[eE][+-]?[0-9]+)?${lengthUnitsPattern}$`);
 export function length(value) {
   return (
     value === '0' ||
-    new RegExp(`^[+-]?[0-9]*\.?[0-9]+(?:[eE][+-]?[0-9]+)?${lengthUnitsPattern}$`).test(value) ||
+    lengthRegExp.test(value) ||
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Thanks!

Made some synthetic micro performance benchmarks that at the end of the day don't matter in practice, but there is a speedup so that's useful. Thanks!

```console
❯ node bench.js
cpu: Apple M1 Max
runtime: node v20.1.0 (arm64-darwin)

benchmark                 time (avg)             (min … max)
------------------------------------------------------------
• RegExp performance
------------------------------------------------------------
baseline              327.74 ps/iter     (300 ps … 12.73 ns)
Pre-compiled          107.04 ns/iter (102.57 ns … 118.06 ns)
Compiled when needed    2.57 µs/iter     (2.53 µs … 2.67 µs)

summary for RegExp performance
  baseline
   326.59x faster than Pre-compiled
   7832.34x faster than Compiled when needed
```

```console
❯ bun bench.js
cpu: Apple M1 Max
runtime: bun 1.0.6 (arm64-darwin)

benchmark                 time (avg)             (min … max)
------------------------------------------------------------
• RegExp performance
------------------------------------------------------------
baseline              328.41 ps/iter      (300 ps … 5.42 ns)
Pre-compiled            3.45 ns/iter    (3.15 ns … 12.25 ns)
Compiled when needed  607.87 ns/iter (577.39 ns … 684.43 ns)

summary for RegExp performance
  baseline
   10.51x faster than Pre-compiled
   1850.95x faster than Compiled when needed
```

PR #12265, opened 2023-10-23
[PERFORMANCE IMPROVEMENT]

yogesh-aggarwal:
**Here's a summary:**

1. Emphasized the use of `const` declarations over `let` declarations when possible. This is a best practice that is recommended by the eslint itself.
2. Discouraged the use of the `var` keyword and replaced it with modern `let` keyword for clarity in the codebase.
3. Integrations aren't touched in this commit. They are as is, no changes so far.

**Result:**
1. Reduction in the build times and of course faster developer experience.
2. Clarity in the codebase while reading it to identify the mutability of the variable.

If there're any issues, feel free to reach out to me here!

---
**Yogesh Aggarwal**


RobinMalfait:
Hey!

Thanks for the PR, but this doesn't have any performance impact and we prefer `let` for block scope, and only use `const` on a module level scope.

Issue #12264, opened 2023-10-22
Typescript: resolveConfig return type does not represent theme merging


baffalop:
**What version of Tailwind CSS are you using?**

v3.3.3

**What version of Typescript are you using?**

v5.2.2

**What build tool (or framework if it abstracts the build tool) are you using?**

Vite 4.5.0

**What version of Node.js are you using?**

v18.13.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

https://github.com/baffalop/tailwind-resolveconfig-repro/blob/main/src/main.ts

### Describe your issue

The return type for `resolveConfig` does not accurately represent what will be returned for the `theme` config. It's typed as simply returning the config passed in (via `UnwrapResolvables`), but the resolved `theme` output potentially differs significantly from this.

I would expect all of the default theme keys to be present, unless overridden; and they should be merged with the contents of `extend`.

It's particularly problematic that `config.theme.extend` is preserved in the return type, but absent in the actual return value, causing typescript to allow a runtime error if this is accidentally accessed.

This issue was originally pointed out to me here: https://github.com/tailwindlabs/tailwindcss/issues/6422#issuecomment-1761163362

RobinMalfait:
Hey!

This should be fixed by #12272, and will be available in the next release. Once again, thanks for your contribution!

You can already try it by using the insiders build `npm install tailwindcss@insiders`.

PR #12272, opened 2023-10-23
Improve resolveConfig return type: merge themes

baffalop:
This PR addresses https://github.com/tailwindlabs/tailwindcss/issues/12264, adding some special handling of the `theme` key in `ResolvedConfig` to better reflect how `theme` and `theme.extend` are merged with the default theme values. Specifically:

* Include the generated type `DefaultTheme` in the return value
* If a theme key is set in `config.theme`, it overrides the default theme value
* If a theme key is extended in `config.theme.extend`, it's intersected with the resolved theme value
* Custom theme keys set in the user config will also still show up in the return type

The generated `DefaultTheme` type doesn't itself contain all possible theme config keys (those that are defined via a callback are omitted), so we fall back to values from `ThemeConfig`, which are of a less concrete type (eg. `KeyValuePair<string, string>`).

Basically, I think I've implemented the "ridiculousness with conditional types" mentioned in [the review for this PR](https://github.com/tailwindlabs/tailwindcss/pull/9972/files#r1035948691)

### Incidental changes

A couple of small changes were necessary elsewhere in the codebase to facilitate this:

**`generate-types.js`**: Do not intersect the generated `DefaultTheme` type with `Config['theme']`. We need to use the concrete value types in here to improve the `ResolvedConfig` type, and the `Config['theme']` types get in the way of this. As far as I can see, the only place where `DefaultTheme` is currently referenced is in `defaultTheme.d.ts` where it is once again intersected with `Config['theme']`, so this change shouldn't make a difference to that.

**`config.d.ts`**: I exported `ThemeConfig` to use in `ResolvedConfig`, and I also split out a separate `ThemeConfigCustomizable` interface which includes the final `[key: string]: any`. Otherwise there was no way to avoid `ResolvedConfig` itself including a fallback `[key: string]: any`, rendering any unknown property access valid in typescript's eye.

### Testing

Given that the codebase itself isn't written in typescript, I wasn't sure of a good way to test the changes. I resorted to creating a sample ts file defining a config, resolving it, and trying to access various keys on the output---using my IDE for typescript feedback.

<details>
<summary>Example test file</summary>

```ts
import { Config } from './types/config'
import resolveConfig = require('./resolveConfig')

const config = {
  content: ['**/*.js'],
  theme: {
    gap: {
      1: '1',
      2: '2',
    },
    myCustom: {
      foo: 'custom',
      bar: 'custom',
    },
    extend: {
      colors: {
        cerulean: {
          50: 'bla',
          100: 'yeah',
          200: 'another',
        }
      },
      gap: {
        3: '3'
      }
    }
  }
} satisfies Config

const theme = resolveConfig(config).theme

theme.colors.cerulean // ✅
theme.colors.zinc // ✅
theme.gap // { 1, 2, 3 }
theme.spacing // { 0, 1, 2, 3, ... }
theme.padding // KeyValuePair<string, string>
// @ts-expect-error
theme.extend
// @ts-expect-error
theme.bla
theme.myCustom // ✅
```

</details>

I'd gladly do more real-world testing or add tests to the codebase if an approach could be suggested.

This is my first PR to a large-scale OS project. I hope it's useful!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fallback to RegEx based parser when using custom transformers or extractors ([#11335](https://github.com/tailwindlabs/tailwindcss/pull/11335))
 - Bump `lightningcss` and reflect related improvements in tests ([#11550](https://github.com/tailwindlabs/tailwindcss/pull/11550))
 - Fix incorrect spaces around `-` in `calc()` expression ([#12283](https://github.com/tailwindlabs/tailwindcss/pull/12283))
+- Improve types for `resolveConfig` ([#12272](https://github.com/tailwindlabs/tailwindcss/pull/12272))
 
 ### Added
 
--- a/resolveConfig.d.ts
+++ b/resolveConfig.d.ts
@@ -1,4 +1,4 @@
-import { Config, ResolvableTo, ThemeConfig } from "./types/config";
+import { Config, ResolvableTo, ThemeConfig } from './types/config'
 import { DefaultTheme } from './types/generated/default-theme'
 import { DefaultColors } from './types/generated/colors'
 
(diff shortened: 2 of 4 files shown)


baffalop:
I'm used to using Gitlab which gives an option to squash commits when merging an MR. Is there no do the same in Gitlab? Please let me know if I should squash my commits on the branch itself.

RobinMalfait:
Hey!

While this isn't 100% correct yet, it is still more correct than it was before. GitHub isn't showing it properly, but I rebased this PR to have the latest changes from `master` as well. You are still the author on most of the commits:
image

I added 2 commits to:
- Fix the TypeScript warning when trying to use `T['theme']['extend']`
- To format the types using Prettier

Regarding your comment about squashing, that's something maintainers can choose when merging. Contributors can't choose this. We always "Squash and merge" so you don't have to do anything.

image

Thanks for the PR!

baffalop:
That's great, thank you @RobinMalfait ! I'm not particularly precious about getting the credit, but good to know—it was a bit weird that Github started showing all my commits as yours. I agree with your changes. I had forgotten prettier was to be used for formatting, I was expecting eslint to say something but it was falling over parsing typescript.

And yes, the return type isn't perfect but it's much better. I had actually meant to write up the caveats. Maybe I'll add them into the PR description for posterity.

PR #12299, opened 2023-10-26
Disable tap highlights on iOS

thecrypticace:
Here we're adding `-webkit-tap-highlight-color: transparent;` to disable tap highlights on iOS.

This means that the brief gray overlay that appears when tapping links or buttons will not anymore.

**Before**

https://github.com/tailwindlabs/tailwindcss/assets/614993/1b2a6d27-5f51-4a1d-a7e9-7382ba25a48e

**After**

https://github.com/tailwindlabs/tailwindcss/assets/614993/14c4b101-474a-4d6a-9de4-6d29e0620d80

If you want to re-enable the tap highlight in your own project you can do so by adding a rule to revert it to the original value:

```css
@layer base {
  html,
  :host {
    -webkit-tap-highlight-color: revert;
  }
}
```


src/css/preflight.css
```diff
@@ -24,6 +24,7 @@
 4. Use the user's configured `sans` font-family by default.
 5. Use the user's configured `sans` font-feature-settings by default.
 6. Use the user's configured `sans` font-variation-settings by default.
+7. Disable tap highlights on iOS
```
zepfietje, line 27: Missing `.` here, @thecrypticace. 😉

thecrypticace: Fixed, thanks!

oncet:
This also applies to Android, right?

thecrypticace:
It does (at least in Chrome)

babakfp:
Why is this added? That tap highlight is an accessibility/UX thing. This means by default all websites built with TailwindCSS are now less accessible!

@adamwathan, what changed? https://github.com/tailwindlabs/tailwindcss/pull/9050#issuecomment-1208468878

BenHerbst:
I hate this, i wondered why it got removed in my website and then i found this pr ...

mmikkel:
I agree with the above comments. Tap highlights being removed by default is an accessibility issue.

hichemfantar:
I agree that this makes tailwind less accessible by default.

PR #12342, opened 2023-11-02
Ensure configured `font-feature-settings` for `mono` are included in Preflight

tombl:
If a `fontFeatureSettings` value is configured for `mono`, elements that default to monospace, like `pre`, don't automatically inherit this value.

#9707 fixed this for `sans`, and this PR fixes it for `mono`.


thecrypticace:
Thanks!

PR #12506, opened 2023-11-29
Fix loading of built-in plugins when using an ESM or TypeScript config with the Standalone CLI

### LINKED ISSUE #12374: Standalone CLI build fails with first-party plugin in config

tailwindcss v3.3.5

Standalone CLI (https://tailwindcss.com/blog/standalone-cli)

N/A

Chrome

Windows

**Reproduction URL**

Not Needed

**Describe your issue**

Standalone CLI throws error after adding first party plugin to the config (@tailwindcss/forms). 

Config:

module.exports = {
  content: [
    "../resources/**/*.blade.php",
  ],
  plugins: [
    require('@tailwindcss/forms'),
  ]
}

Error:
Rebuilding...
Error: Cannot find module '@tailwindcss/forms'
Require /public/tailwind.config.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1955:46)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at _resolve (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:250334)
    at jiti (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:252917)
    at /home/eduquiz-5wyxc/staging.eduquiz.ca/public/tailwind.config.js:8:5
    at evalModule (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:255614)
    at jiti (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:253542)
    at /snapshot/tailwindcss/lib/lib/load-config.js:37:30
    at loadConfig (/snapshot/tailwindcss/lib/lib/load-config.js:39:6) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/[website]/public/tailwind.config.js'
  ]
}

thecrypticace:
In the case of a config that's not directly loadable by NodeJS — for example because it is TypeScript, ESM, or uses unsupported syntax, we invoke JITI to load the config file.

Normally there are no problems caused by this. However, in the standalone CLI we provide our plugins already built-in. This becomes a problem because JITI uses a version of require that we were not patching to load the built in plugins.

Making this work is a bit more complicated than it seems at first glance and relies on an internal implementation detail of JITI. However, it's _probably_ one that's not subject to change too much (… famous last words — I know)

Basically we now:
- Patch `require` and `require.resolve` that's available natively to every module loaded by Node
- Patch `createRequire` to return patched require + resolve functions that work the same way
- Overload the JITI instance used so we can provide additional options that effectively instruct it to always use the "native" require to load the module rather than trying to find it on the filesystem

Fixes #12374

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don't crash when given applying a variant to a negated version of a simple utility ([#12514](https://github.com/tailwindlabs/tailwindcss/pull/12514))
 - Fix support for slashes in arbitrary modifiers ([#12515](https://github.com/tailwindlabs/tailwindcss/pull/12515))
 - Fix source maps of variant utilities that come from an `@layer` rule ([#12508](https://github.com/tailwindlabs/tailwindcss/pull/12508))
+- Fix loading of built-in plugins when using an ESM or TypeScript config with the Standalone CLI ([#12506](https://github.com/tailwindlabs/tailwindcss/pull/12506))
 - [Oxide] Remove `autoprefixer` dependency ([#11315](https://github.com/tailwindlabs/tailwindcss/pull/11315))
 - [Oxide] Fix source maps issue resulting in a crash ([#11319](https://github.com/tailwindlabs/tailwindcss/pull/11319))
 - [Oxide] Fallback to RegEx based parser when using custom transformers or extractors ([#11335](https://github.com/tailwindlabs/tailwindcss/pull/11335))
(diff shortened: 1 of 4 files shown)


thecrypticace:
@RobinMalfait any objections to my `useCustomJiti` thingy?

I can't override loadConfig directly I'm pretty sure because it's compiled as a non-configurable property of the module's exports.

privatenumber:
If you're interested, tsx just released a `tsx.require()` and `register`/`unregister` functions if you want to load TypeScript/ESM files: https://tsx.is/node#only-commonjs-enhancement

PR #12711, opened 2024-01-05
Support theme values for has variant

crswll:
This allows you to add values for the `has` variant to your theme so we can do things like:
```
theme: {
  has: {
    headers: ':where(h1, h2, h3, h4, h5, h6)'
  }
}
```
```
<div class="has-headers:bg-green-500">...</div>
```
over using arbitrary values all the time like:
```
<div class="has-[:where(h1,h2,h3,h4,h5,h6)]:bg-green-500">...</div>
```

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -394,15 +394,15 @@ export let variantPlugins = {
     )
   },
 
-  hasVariants: ({ matchVariant }) => {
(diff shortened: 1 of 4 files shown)


crswll:
No trouble at all. Excited to see how it all works out.

PR #12718, opened 2024-01-05
Improve Windows support for escaped glob syntax

RobinMalfait:
This PR is a continuation of #12715 to apply the same improvements for Windows.

On Windows the path separator is `\\` and `\\` is also used for escaping the `[`, `]`, `(` and `)` characters. This will result in `\\\\[` for example. In this case it means that the first `\\` is the path separator and the second `\\` is the escape for the `[` or `(` characters.

Additionally, we also ensure that we only escape `[`, `]` and `(`, `)` when the balanced pairs are _both_ not escaped yet. Right now it was going over the individual characters and escaping them accordingly.

This is not enough, because in case of this path:

```js
let path = ".\\src\\**\\[foo]\\bar.html"
```

The first `\\[` looks like it is escaped, but on Windows this is just the path separator. The `foo]` part is _not_ escaped therefor we should escape the whole thing. As a first step, this will result in:

```js
let path = ".\\src\\**\\\\[foo\\]\\bar.html"
```

Now we do have the 4 backslashes `\\\\`, then in the normalization step, we normalize `\\` to `/`. If we do this _as is_, then it would look like this, which is also incorrect:

```js
let path = "./src/**/[foo/]/bar.html"
```

... it's incorrect because we have `/[foo/]` and we expected `/\\[foo\\]` instead. 

If we apply the logic I mentioned earlier where the first `\\` is the path separator and the second `\\` is the escape, then we can convert: 

```js
let path = ".\\src\\**\\\\[foo\\]\\bar.html"
``` 

... to this first: 

```js
let path = ".\\src\\**/\\[foo\\]\\bar.html"
```

Then we can apply the rest of the normalization, which looks like this:

```js
let path = "./src/**/\\[foo\\]/bar.html"
```

... and now the `[foo]` part is properly escaped as `\\[foo\\]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure max specificity of `0,0,1` for button and input Preflight rules  ([#12735](https://github.com/tailwindlabs/tailwindcss/pull/12735))
-- Improve glob handling for folders with `(`, `)`, `[` or `]` in the file path ([#12715](https://github.com/tailwindlabs/tailwindcss/pull/12715))
+- Improve glob handling for folders with `(`, `)`, `[` or `]` in the file path ([#12715](https://github.com/tailwindlabs/tailwindcss/pull/12715), [#12718](https://github.com/tailwindlabs/tailwindcss/pull/12718))
 - Split `:has` rules when using `experimental.optimizeUniversalDefaults` ([#12736](https://github.com/tailwindlabs/tailwindcss/pull/12736))
 
 ### Added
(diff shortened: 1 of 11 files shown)


mominshaikhdevs:
please merge this PR soon as its absolutely essential.

thecrypticace:
It is unfortunately not finished as it doesn't work 100% yet on Windows after I did some testing on Friday.

RobinMalfait:
Going to close this PR for now because the original issue that was reported was for macOS / Linux. This is probably still a good fix to do, but the current implementation is not ideal just because an escaped `\\` also happens to be a path separator in Windows which means that this is ambiguous.

The real fix should be to split the base path and the actual glob so that the ambiguity goes away. Because then the path can contain `\\` as the separator and the glob itself containing `\\` will be a guaranteed escape value.

PR #12747, opened 2024-01-11
feat: add contain utilities

TheAlexLichter:
This adds utilities for the [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) property, which is supported in all evergreen browsers by now 🙏👌

After that, contain-intrinsic-* (height/width/size) could be a nice addition too (but lacks browser support right now).

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -2946,5 +2946,5 @@ export let corePlugins = {
       '.contain-style': { contain: 'style' },
       '.contain-unset': { contain: 'unset' },
     })
-  }
+  },
 }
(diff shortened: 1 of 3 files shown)


BioPhoton:
The containe-intrinsic-* is for [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) which is a massive performance gainer but not supposed in most of the browsers 🥲.

BioPhoton (approved):
Looks good and is a nice performance feature!

KrisBraun:
Hey @manniL! Thanks for the contribution. We'll get this merged in soon via https://github.com/tailwindlabs/tailwindcss/pull/12993.

PR #12835, opened 2024-01-25
Remove Webkit Search Cancel Button

ljg2gb:
On Chrome and Safari webkit includes a cancel button that I think Preflight should hide. [MDN lists this cancel button as nonstandard][1] and Chrome and Sarafi both style this button differently. This appears to happen in macOS and linux.

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button

Safari:
Screenshot 2024-01-25 at 3 49 51 PM
Chrome:
Screenshot 2024-01-25 at 3 49 23 PM
Firefox (no cancel button with Preflight):
image (7)

Co-authored-by: Brooks Johnson <2008881+bkjohnson@users.noreply.github.com>

### DIFF
--- a/src/css/preflight.css
+++ b/src/css/preflight.css
@@ -255,6 +255,13 @@ Remove the inner padding in Chrome and Safari on macOS.
 }
 
 /*
+Remove x on Chrome and Safari.
+*/
+::-webkit-search-cancel-button {
+  display: none;
+}
+
+/*
 1. Correct the inability to style clickable types in iOS and Safari.
 2. Change font properties to `inherit` in Safari.
 */


reinink:
Hey! Thanks for this suggestion, but to be honest we already feel like some of our form related resets in Preflight are a little heavy-handed, and we're likely going to do less resetting of default browser styles in future versions of Tailwind.

For example, buttons and inputs are not even recognizable in Tailwind today without adding your own styles: https://play.tailwindcss.com/EYvrowvKOP

So given the direction we're heading with this, I don't think adding this to Preflight makes sense. You can, of course, still add this to your own project by adding this style to your CSS:

```css
@layer base {
  /*
  Remove x on Chrome and Safari.
  */
  ::-webkit-search-cancel-button {
    display: none;
  }
}
```

Hope that makes sense, and thanks either way for this suggestion!

PR #12911, opened 2024-02-09
Sort arbitrary properties alphabetically across multiple class lists

### LINKED ISSUE #246: Inconsistent sorting between formatting all files and a single file
**What version of `prettier-plugin-tailwindcss` are you using?**

v0.5.11

v3.4.1

v20.10.0

npm, pnpm

Windows (WSL)

**Reproduction URL**

https://github.com/lesha1201/tailwind-prettier-plugin-inconsistent-sorting-issue

**Describe your issue**

https://github.com/tailwindlabs/prettier-plugin-tailwindcss/assets/10157660/8590929d-0f29-4c0c-8752-a2be1496bd5a

`prettier-plugin-tailwindcss` formats a certain file differently when you format all files and only that file. In the reproduction example, there are two components: `ButtonSocialApple` and `ButtonSocialFacebook` which have similar tailwind classes.

```tsx
// src/app/button-social/button-social.apple.tsx
// ...
export function ButtonSocialApple({ children }: ButtonSocialAppleProps) {
  return (
    <div className="[--dr-social-button-color-bg-brand:#000] [--dr-social-button-color-bg-brand_hover:#000] [--dr-social-button-color-fg_on-brand:#fff]">
      {children}
    </div>
  );
}

// src/app/button-social/button-social.apple.tsx
// ...
export function ButtonSocialFacebook({ children }: ButtonSocialFacebookProps) {
  return (
    <div className="[--dr-social-button-color-bg-brand:#1877F2] [--dr-social-button-color-bg-brand_hover:#0C63D4] [--dr-social-button-color-fg_on-brand:#fff]">
      {children}
    </div>
  );
}
```

When you format all files by running:

```sh
npx prettier --write .
```

It produces one result for `ButtonSocialFacebook`:

```tsx
export function ButtonSocialFacebook({ children }: ButtonSocialFacebookProps) {
  return (
    <div className="[--dr-social-button-color-fg_on-brand:#fff] [--dr-social-button-color-bg-brand:#1877F2] [--dr-social-button-color-bg-brand_hover:#0C63D4]">
      {children}
    </div>
  );
}
```

When you format only `ButtonSocialFacebook`:

```sh
npx prettier --write ./src/app/button-social/button-social-facebook.tsx
```

It produces a different result:

```tsx
export function ButtonSocialFacebook({ children }: ButtonSocialFacebookProps) {
  return (
    <div className="[--dr-social-button-color-bg-brand:#1877F2] [--dr-social-button-color-bg-brand_hover:#0C63D4] [--dr-social-button-color-fg_on-brand:#fff]">
      {children}
    </div>
  );
}
```

However, when you remove `ButtonSocialApple` component, it formats `ButtonSocialFacebook` consistently.

thecrypticace:
Fixes https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/246

In Tailwind we sort classes alphabetically before processing to produce a deterministic order. And, when we encounter an arbitrary property we register it _at that time_ so that's the order it'll take when sorting. In most cases this is fine because we sort up front, however, in the Prettier plugin we don't know all classes up front — each individual class list is sorted separately but uses the same internal context for performance reasons. This means that we need to do some extra bookkeeping to ensure they're sorted consistently regardless of the order of individual class lists or files scanned.

This PR does this bookkeeping and will sort classes to ensure that the arbitrary properties are always alphabetical when handled by the prettier plugin even scross separate class lists or files.

Now given these two files, a.jsx and b.jsx:

```jsx
// a.jsx
export default function ComponentA() {
  return <div className="[--fg:#fff]" />
}
```

```jsx
// b.jsx
export default function ComponentB() {
  return <div className="[--bg:#111] [--bg_hover:#000] [--fg:#fff]" />
}
```

If you format each individual one with prettier or format all of them at one time the order of the properties will no longer flip-flop between these two lists:
```
[--bg:#111] [--bg_hover:#000] [--fg:#fff]
```

```
[--fg:#fff] [--bg:#111] [--bg_hover:#000]
```

### DIFF
--- a/src/lib/offsets.js
+++ b/src/lib/offsets.js
@@ -24,6 +24,8 @@ import { remapBitfield } from './remap-bitfield.js'
  * @property {bigint} variants Dynamic size. 1 bit per registered variant. 0n means no variants
  * @property {bigint} parallelIndex Rule index for the parallel variant. 0 if not applicable.
  * @property {bigint} index Index of the rule / utility in its given *parent* layer. Monotonically increasing.
+ * @property {bigint} propertyOffset Offset for the arbitrary property. Only valid after sorting.
+ * @property {string} property Name/Value of the arbitrary property.
  * @property {VariantOption[]} options Some information on how we can sort arbitrary variants
  */
 
@@ -88,17 +90,21 @@ export class Offsets {
       variants: 0n,
       parallelIndex: 0n,
       index: this.offsets[layer]++,
+      propertyOffset: 0n,
+      property: '',
       options: [],
     }
   }
 
   /**
+   * @param {string} name
    * @returns {RuleOffset}
    */
-  arbitraryProperty() {
+  arbitraryProperty(name) {
     return {
       ...this.create('utilities'),
       arbitrary: 1n,
+      property: name,
     }
   }
 
@@ -262,6 +268,11 @@ export class Offsets {
       return a.arbitrary - b.arbitrary
     }
 
+    // Always sort arbitrary properties alphabetically
+    if (a.propertyOffset !== b.propertyOffset) {
+      return a.propertyOffset - b.propertyOffset
+    }
+
     // Sort utilities, components, etc… in the order they were registered
     return a.index - b.index
   }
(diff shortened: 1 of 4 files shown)


src/lib/offsets.js
```diff
@@ -320,14 +331,62 @@ export class Offsets {
...
   /**
    * @template T
    * @param {[RuleOffset, T][]} list
    * @returns {[RuleOffset, T][]}
    */
   sort(list) {
+    // Sort arbitrary variants so they're in alphabetical order
     list = this.remapArbitraryVariantOffsets(list)
 
+    // Sort arbitrary properties so they're in alphabetical order
```
RobinMalfait, line 387: Nit: This comment doesn't reflect what is happening

RobinMalfait (approved):
Do you think we have to worry about this in v4? I think we handle this scenario already the way we built it. 🤔

thecrypticace:
I think we may but it entirely depends on how the APIs end up working with tools like the prettier plugin. As for right now it is already handled.

PR #12923, opened 2024-02-11
Add `mix-blend-plus-darker` utility

clementinelove:
This PR adds 'plus-darker' support to mix-blend-mode.

One PR already added 'mix-blend-plus-lighter' (#8288) , but there should also be 'mix-blend-plus-darker' since blend mode also has a 'plus-darker' value that could be added last time. 

[mix-blend-mode on MDN lists all possible values](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode)

From what I know, it seems only Safari supports `plus-darker` at the moment since this style is used heavily in macOS's vibrant visual style.


reinink:
Hey thanks! This seems like a reasonable addition given the Safari support 👍

PR #12938, opened 2024-02-13
Automatically quote invalid values in attribute selectors

rvanvelzen:
This should resolve https://github.com/tailwindlabs/tailwindcss/discussions/12928


rvanvelzen:
Thanks for the detailed explanation!

PR #12993, opened 2024-02-21
feat: add contain utilities

KrisBraun:
> Add utilities for the [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) property, which is supported in all evergreen browsers by now 🙏👌

Based on https://github.com/tailwindlabs/tailwindcss/pull/12747 by @manniL. Updated to support multiple `contain` keywords.


thecrypticace:
Only question I have is if the `contain` utility is placed where we want it?

thecrypticace (approved):

KrisBraun:
> Only question I have is if the `contain` utility is placed where we want it?

Yes, placed after `willChange` as both are related to performance.

PR #13087, opened 2024-03-05
[v4] Only output `@property foo` once instead of N times

RobinMalfait:
This PR improves the performance of the AST to CSS generation.

This now only outputs `@property foo` once, if it was already generated at the top-level then it will not be generated again.

On the tailwindcss.com codebase, this reduces the CSS generation (before the Lightning CSS step) time from `~6ms` to `~3ms`.

If we weren't using Lightning CSS at all, then it would reduce the generated CSS size from ~28k lines of code to ~14k lines of code (again, on the tailwindcss.com repo).

This means that we are sending fewer bytes to Lightning CSS to optimize, making the whole process a bit faster.

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -73,7 +73,6 @@ export function walk(
 
 export function toCss(ast: AstNode[]) {
   let atRoots: string = ''
-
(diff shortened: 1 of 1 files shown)


thecrypticace (approved):
😍

thecrypticace (approved):

PR #13096, opened 2024-03-05
[v4] Copy README file to every package

RobinMalfait:
This PR copies the root README file to every sub-package. 

Currently the contents is exactly the same, but we will update them in the future.

This will also automatically publish the README.md file, without listing it in the `files/` array.
image

Requires: #13095

### DIFF
--- a/packages/@tailwindcss-cli/README.md
+++ b/packages/@tailwindcss-cli/README.md
@@ -1,0 +1,40 @@
+<p align="center">
+  <a href="https://tailwindcss.com" target="_blank">
+    <picture>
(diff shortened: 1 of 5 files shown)


packages/@tailwindcss-cli/README.md
A7med3bdulBaset, line 1: It's better to reference it like https://github.com/vercel/next.js/blob/canary/readme.md

RobinMalfait: That's only true if you want the README's to always be in sync. This is a temporary workaround until we have dedicated README files per package. This way we don't have to worry about symlinks either.

PR #13097, opened 2024-03-06
CI: Add provenance to all published packages

saibotk:
This commit adds provenance for all published packages. See the NPM documentation [0].

Provenance will allow people to verify that the tailwindcss packages were actually built on GH Actions and with the content of the corresponding commit. This will help with supply chain security.

For this to work, the `id-token` permission was added only where necessary.
Additionally, all the oxide packages + cli package were adjusted to include their repository link in the `package.json`, so that the build works and the builds can be linked to this repository.

[0]: https://docs.npmjs.com/generating-provenance-statements


KrisBraun (approved):
Thanks, @saibotk. This looks good. Thanks for contributing!

PR #13130, opened 2024-03-07
[v4] Make CSS optimization and minification configurable

RobinMalfait:
This PR disables Lightning CSS by default in development to improve the performance of incremental builds.

This means that during development, the compiled CSS will still use native CSS nesting, modern color syntax without fallbacks, and be missing vendor prefixes for older browsers.

While incremental builds in v4 are already faster than they were in v3, this further improves the performance in projects we've tested. For example in the tailwindcss.com codebase, this reduces incremental build times from 40ms to 30ms.

In the CLI build, the following flags are available:

- `--minify` — this will optimize the CSS output and minify it.
- `--optimize` — this will optimize the CSS output but **not** minify it.

In the Vite plugin, the Lightning CSS optimizations are only enabled for production builds.

In the PostCSS plugin, we added a new option `optimize` that can be used to enable the optimizations.

```ts
module.exports = {
  plugins: {
    '@tailwindcss/postcss': {
      // This will optimize and minify the CSS output.
      optimize: true
    }
  }
}
```

```ts
module.exports = {
  plugins: {
    '@tailwindcss/postcss': {
      optimize: {
        // This will optimize but not minify the CSS output.
        minify: false
      } 
    }
  }
}
```

### DIFF
--- a/packages/@tailwindcss-cli/src/commands/build/index.ts
+++ b/packages/@tailwindcss-cli/src/commands/build/index.ts
@@ -46,7 +46,7 @@ export function options() {
     },
     '--minify': {
       type: 'boolean',
-      description: 'Minify the output',
+      description: 'Optimize and minify the output',
       alias: '-m',
     },
     '--optimize': {
--- a/packages/@tailwindcss-cli/src/commands/build/index.ts
+++ b/packages/@tailwindcss-cli/src/commands/build/index.ts
@@ -96,10 +96,15 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
   )
 
   // Compile the input
-  let result = optimizeCss(compile(input, candidates), {
-    file: args['--input'] ?? 'input.css',
-    minify: args['--minify'],
-  })
+  let result = compile(input, candidates)
+
+  // Optimize the output
+  if (args['--minify']) {
+    result = optimizeCss(result, {
+      file: args['--input'] ?? 'input.css',
+      minify: true,
+    })
+  }
 
   // Write the output
   if (args['--output']) {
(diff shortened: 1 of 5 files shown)


thecrypticace (approved):

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -49,6 +49,10 @@ export function options() {
       description: 'Minify the output',
```
KrisBraun, line 49: Should we say "Optimize and minify the output"? Now that the `--optimize` option exists, users may be confused that optimization isn't mentioned here.

KrisBraun (approved):

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -38,7 +42,13 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
 
         function replaceCss(css: string) {
           root.removeAll()
-          root.append(postcss.parse(optimizeCss(css), result.opts))
+          let output = css
+          if (optimize) {
+            output = optimizeCss(output, {
+              minify: typeof optimize === 'object' ? optimize.minify : false,
```
Facug03, line 48: But if opts.optimize is true, minify should be true right? But in this case it would be set to false

RobinMalfait: No there are 2 things happening:

1. "optimize" means, "optimize the output by handling nesting, merging adjacent at-rules, lowering CSS property values (`white` -> `#fff`) and applying vendor prefixes
2. Then there is also the `minify` part that minifies the CSS (mainly removing whitespacing). If minification happens, then optimization will also happen because that's done automatically by Lightning CSS.

Facug03: According to this example it says that `optimize: true` will minify too

```
module.exports = {
  plugins: {
    '@tailwindcss/postcss': {
      // This will optimize and minify the CSS output.
      optimize: true
    }
  }
}
```

RobinMalfait: Aha you are right, I was wrong. `{ optimize: true }` should also minify by default! Opened a PR here to fix this: https://github.com/tailwindlabs/tailwindcss/pull/13201

Thanks for catching this!

PR #13158, opened 2024-03-08
Include percent sign in font-stretch utilities

KrisBraun:
Include a percent sign in `font-stretch` utilities (e.g. `font-stretch-50%` rather than `font-stretch-50`) to make the meaning of the value clearer.


KrisBraun:
Updated the description on #13153 since it's linked from the changelog.

adamwathan (approved):

Issue #13180, opened 2024-03-10
[V4] after:scale-x-0 doesn't take effect


coolemur:
**What version of Tailwind CSS are you using?**

v4-alpha

**What build tool (or framework if it abstracts the build tool) are you using?**

Not relevant as in Tailwind 3 it was working. ("postcss": "^8.4.34")

**What version of Node.js are you using?**

Not relevant as in Tailwind 3 it was working. (v21.6.1)

**What browser are you using?**

Not relevant as in Tailwind 3 it was working. (Opera)

**What operating system are you using?**

Not relevant as in Tailwind 3 it was working. (MacOS)

**Reproduction**

```
<a class="relative first-letter:uppercase after:absolute after:bottom-0 after:left-0 after:h-[4px] after:w-full after:scale-x-0 after:bg-black after:transition-transform after:content-[''] hover:after:scale-x-100" href="#">hover effect</a>
```

Try it with v4 and compare it in https://play.tailwindcss.com .

**Describe your issue**

* after:scale-x-0 doesn't take effect.
* ~~first-letter:uppercase doesn't take effect.~~

wongjn:
`first-letter:uppercase` doesn't work in v3 either, as evidenced by the first element in [this Tailwind Play](https://play.tailwindcss.com/6FbmPfq5Oq). This is because, as per [the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter):

> The `::first-letter` [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) applies styles to the first letter of the first line of a [block container](https://developer.mozilla.org/en-US/docs/Web/CSS/Visual_formatting_model#block_containers), […]

`<a>` elements are **not** block containers by default, thus `::first-letter` would not match. However, applying `display: block` via the `block` class for example *would* make `::first-letter` match and then `first-letter:uppercase` to take effect, as evidenced by the second element in [this Tailwind Play](https://play.tailwindcss.com/6FbmPfq5Oq).

PR #13183, opened 2024-03-10
Fix allowed `--tw-scale-x/y` syntax

wongjn:
Fixes the issue where `scale-*` classes can't use percentages, even though they are permitted by the [CSS specification for the `scale` property](https://www.w3.org/TR/css-transforms-2/#propdef-scale) that the `scale-*` utilities use.

Related to #13180.

### Before

image

### After

image


wongjn:
Sniped by #13182 😅

PR #13218, opened 2024-03-12
Process Tailwind output with Vite CSS plugins

KrisBraun:
Transform Tailwind-generated CSS with Vite CSS plugins. `vite:css` does useful things like transforming `url()` paths and inlining images. `vite:css-post` generates bundle hashes. Before this change, the CSS bundle hash wasn't changing when the generated CSS changed.


KrisBraun:
@ArnaudBarre I'm interested in your review here. In particular, do you see any cases where this won't work with other plugins correctly?

ArnaudBarre:
Answered in Discord: this way of doing things doesn't make the CSS hash independent of the utils content. To workaround that I found that calling again the transform function of the `vite:css-post` plugin with the same id but the new content gives expected result, without the need for a fake id or placeholder. It will automatically apply all the transformation from Vite (but not from other plugins, which is fine IMO)

KrisBraun:
@ArnaudBarre The PR has been updated to apply the `renderChunk` technique.

playgrounds/vite/vite.config.js
thecrypticace, line 1: The `package.json` in The Vite playground needs to be updated otherwise it doesn't work.

thecrypticace (changes requested):

thecrypticace:
Some small changes but otherwise looks good and appears to work quite well. 💯

playgrounds/vite/vite.config.js
KrisBraun: Unintentional change (part of testing Storybook). Backing it out.

thecrypticace (approved):

Issue #13263, opened 2024-03-16
[v4] Doesn't detect Svelte class directives 


ryanylee:
**What version of Tailwind CSS are you using?**

v4.0.0-alpha.9

In Svelte, I have a div as follows
```
<script>
   let show = true;
</script>

<div class:opacity-10={show}>hi!</div>
```

Currently `.opacity-10` is not generated

RobinMalfait:
Hey!

This has been implemented in #13274 and will be available in the next v4 alpha release.

PR #13274, opened 2024-03-18
Parse candidates in `.svelte` files with `class:abc="condition"` syntax

RobinMalfait:
This PR introduces backwards compatibility for the `class:` syntax in Svelte files.

In Svelte there is a convention where you can use `<div class:px-4="condition" />` which allows you to apply the `px-4` class conditionally based on a condition. This also means that we extract `class:px-4` as a utility where `class` is a variant and `px-4` is the utility.

This is obviously incorrect, to solve this we can ignore the `class:` part before parsing the whole file.

This is also what we do in v3.

In a perfect world, we have more specific parsers depending on the file type before using the generic parser we use now. But that's a bigger project on its own way beyond the scope of this PR.

Fixes: #13263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Validate bare values ([#13245](https://github.com/tailwindlabs/tailwindcss/pull/13245))
+- Parse candidates in `.svelte` files with `class:abc="condition"` syntax ([#13274](https://github.com/tailwindlabs/tailwindcss/pull/13274))
 
 ### Changed
 
@@ -108,3 +109,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move the CLI into a separate `@tailwindcss/cli` package ([#13095](https://github.com/tailwindlabs/tailwindcss/pull/13095))
 
 ## [4.0.0-alpha.1] - 2024-03-06
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,4 +109,3 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move the CLI into a separate `@tailwindcss/cli` package ([#13095](https://github.com/tailwindlabs/tailwindcss/pull/13095))
 
 ## [4.0.0-alpha.1] - 2024-03-06
-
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -108,3 +109,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move the CLI into a separate `@tailwindcss/cli` package ([#13095](https://github.com/tailwindlabs/tailwindcss/pull/13095))
 
 ## [4.0.0-alpha.1] - 2024-03-06
+
```
thecrypticace, line 112: This is causing linting to fail. We probably _should_ have something here though. Maybe line `- Initial release` or something?

oxide/crates/core/src/lib.rs
```diff
@@ -399,6 +400,23 @@ pub fn scan_files(input: Vec<ChangedContent>, options: u8) -> Vec<String> {
...
+            },
+            Err(e) => {
+                event!(tracing::Level::ERROR, "Failed to read file: {:?}", e);
+                Default::default()
+            }
+        },
+        (None, Some(content)) => Some(content.into_bytes()),
+        _ => Default::default(),
+    }
+}
```
thecrypticace, line 418: The triple nested match feels a bit weird here. It's _shorter_ but harder to fallow. All but one of these conditions are early-returnable so maybe something like this instead:
```suggestion
fn read_changed_content(c: ChangedContent) -> Option<Vec<u8>> {
  if let Some(content) = c.content {
    return Some(content.into_bytes())
  }

  let Some(file) = c.file else {
    return Default::default()
  }
  
  let Ok(content) = std::fs::read(&file) else {
    event!(tracing::Level::ERROR, "Failed to read file: {:?}", e);
    return Default::default()
  }
  
   match file.extension().map(|x| x.to_str()) {
    Some(Some("svelte")) => Some(content.replace(" class:", " ")),
    _ => Some(content),
  }
}
```

thecrypticace (changes requested):

oxide/crates/core/src/lib.rs
RobinMalfait: I kept the original code and added the extension part to it. Splitting this up is a bit clearer indeed.

I did fix some syntax errors and changed it slightly.

thecrypticace (approved):

oxide/crates/core/tests/auto_content.rs
```diff
@@ -297,9 +297,14 @@ mod auto_content {
             ("foo.jpg", Some("xl:font-bold")),
             // A file that is ignored
             ("foo.html", Some("lg:font-bold")),
+            // A svelte file with `class:foo="bar"` syntax
+            ("index.svelte", Some("<div class:px-4='condition'></div>")),
```
ryanylee, line 301: In Svelte, expression would be in squiggly brackets and not quotes
`<div class:px-4={condition}>`

PR #13290, opened 2024-03-20
Improve `transform`, `transform-cpu` and `transform-gpu` utilities

RobinMalfait:
This PR improves the transform utilities by making sure that `transform-cpu` and `transform-gpu` also keep the `skewX` and `skewY` functions into account.

Additionally, `skewX` and `skewY` are dropped from arbitrary values when using `transform-[…]`. This way you always get the arbitrary values as-is without anything else.

We also move the `skewX` and `skewY` functions into the `--tw-skew-x` and `--tw-skew-y` variables themselves, because otherwise it could happen that the variable is not defined and then we use `skewX(var(--variable-that-is-not-defined))`.

By doing that, we also updated the `transform-cpu` and `transform-gpu` to use the `--tw-skew-x` and `--tw-skew-y` variables that already include the `skewX` and `skewY` functions.

Apart from that we also introduced `translate-none`, `scale-none` and `transform-none` which just sets the property to `none`.


packages/tailwindcss/src/utilities.ts
```diff
@@ -1488,20 +1488,15 @@ export function createUtilities(theme: Theme) {
...
-      // are applied right to left, so we put them at the end so they are
-      // applied before other transforms, similar to other specific transform
-      // properties.
-      //
-      // https://developer.mozilla.org/en-US/docs/Web/CSS/transform#values
-      value = `${candidate.value.value} ${value}`
+      value = candidate.value.value
     }
 
+    if (value === null) return
```
KrisBraun, line 1498: This could probably be an `else` with `null` dropped from type.

KrisBraun (approved):

packages/tailwindcss/src/utilities.ts
RobinMalfait: The candidate.value.value could still be `null`. The previous implementation hacked around that by (incorrectly) converting it to a string

packages/tailwindcss/src/utilities.ts
```diff
@@ -1510,14 +1511,8 @@ export function createUtilities(theme: Theme) {
...
 
-  staticUtility('transform-cpu', [['transform', 'translate(0,0)']])
-  staticUtility('transform-gpu', [['transform', 'translate(0,0,0)']])
-  staticUtility('transform-none', [
-    ['translate', 'none'],
-    ['rotate', 'none'],
-    ['scale', 'none'],
-    ['transform', 'none'],
-  ])
+  staticUtility('transform-gpu', [['transform', 'translateZ(0)']])
```
thecrypticace, line 1514: This needs the skew stuff present otherwise it'll get reset

thecrypticace (dismissed):

adamwathan (approved):

PR #13306, opened 2024-03-21
Fix crash showing completions in Intellisense when using a custom separator

### LINKED ISSUE #894: Config separator not work

For example: v1.85.1

For example: v0.10.5

For example: v3.4.0

For example: npm

 Windows 10

**Tailwind config**

```js
export default {
  content: ["**/*.html"],
  separator:'_', 
} 
```

**VS Code settings**

```json
 // all default config
```

**Reproduction URL**

A public GitHub repo that includes a minimal reproduction of the bug. **Please do not link to your actual project**, what we need instead is a _minimal_ reproduction in a fresh project without any unnecessary code. This means it doesn't matter if your real project is private/confidential, since we want a link to a separate, isolated reproduction anyways.

**Describe your issue**

Config separator not work ，hope work when  `separator:'_'`!

thecrypticace:
Intellisense uses the internal `getVariants` API and its use of `__TAILWIND_PLACEHOLDER__` crashes inside Tailwind's variant formatting when the separator is `_`. This crashes the entire completion routine in intellisense and prevents them from appearing.

I've changed it to use no "special" characters to prevent the crash. Instead, we use `TAILWINDPLACEHOLDER` with a unique, random ID attached to eliminate potential conflicts with registered utilities.

Fixes https://github.com/tailwindlabs/tailwindcss-intellisense/issues/894

### DIFF
--- a/src/lib/setupContextUtils.js
+++ b/src/lib/setupContextUtils.js
@@ -1044,6 +1044,11 @@ function registerPlugins(plugins, context) {
 
   // Generate a list of available variants with meta information of the type of variant.
   context.getVariants = function getVariants() {
+    // We use a unique, random ID for candidate names to avoid conflicts
+    // We can't use characters like `_`, `:`, `@` or `.` because they might
+    // be used as a separator
+    let id = Math.random().toString(36).substring(7).toUpperCase()
+
     let result = []
     for (let [name, options] of context.variantOptions.entries()) {
       if (options.variantInfo === VARIANT_INFO.Base) continue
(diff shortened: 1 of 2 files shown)


src/lib/setupContextUtils.js
```diff
@@ -1054,7 +1059,7 @@ function registerPlugins(plugins, context) {
         values: Object.keys(options.values ?? {}),
         hasDash: name !== '@',
         selectors({ modifier, value } = {}) {
-          let candidate = '__TAILWIND_PLACEHOLDER__'
+          let candidate = `TAILWINDPLACEHOLDER${id}`
```
RobinMalfait, line 1062: I don't love the placeholder, but I didn't love it before either haha. One potential improvement could be to use a UUID (without the `-`) but that's almost overkill for this.

```js
crypto.randomUUID().replaceAll('-','')
```

RobinMalfait (approved):

src/lib/setupContextUtils.js
thecrypticace: yeah I really don't like the placeholder either but it's necessary for the current implementation. UUID would definitely be overkill for this.

PR #13322, opened 2024-03-22
Transpile `import.meta.url` in config files

### LINKED ISSUE #909: Doesn't work with ES Module syntax
Latest

Latest 0.10.5

3.3.2

Npm

Linux, macOS

**Tailwind config**
import { fileURLToPath } from 'url';
const componentsDir = fileURLToPath(new URL('./components', import.meta.url));

export default {
  content: [`${componentsDir}/**/*.{js,ts,jsx,tsx}`],
  important: true,
  theme: {
    extend: {},
  },
  plugins: [],
};

**Describe your issue**

[Error - 10:50:21 AM] Tailwind CSS: Cannot use 'import.meta' outside a module
{directory}/tailwind.config.js:2
const componentsDir = _url.fileURLToPath.call(void 0, new URL('./components', import.meta.url));
                                                                                     ^^^^

SyntaxError: Cannot use 'import.meta' outside a module

Telling me that it doesn't support ES6 syntax. Converting the config to commonJS fixes this bug but the build pipeline forces me to use ES6.

thecrypticace:
Right now if you try to use `import.meta.url` in a `tailwind.config.js` file (or a dependency) Node will throw an error. This happens because config files are compiled to Common JS internally using Jiti and Sucrase. However, Sucrase doesn't rewrite `import.meta.url` so we must use Jiti's built-in Babel transform in this case.

Fixes https://github.com/tailwindlabs/tailwindcss-intellisense/issues/909

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `mix-blend-plus-darker` utility ([#12923](https://github.com/tailwindlabs/tailwindcss/pull/12923))
 - Ensure dashes are allowed in variant modifiers ([#13303](https://github.com/tailwindlabs/tailwindcss/pull/13303))
 - Fix crash showing completions in Intellisense when using a custom separator ([#13306](https://github.com/tailwindlabs/tailwindcss/pull/13306))
+- Transpile `import.meta.url` in config files ([#13322](https://github.com/tailwindlabs/tailwindcss/pull/13322))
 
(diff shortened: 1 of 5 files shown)


RobinMalfait (approved):
Looks good, only thing remaining is probably a changelog fixed entry.

thecrypticace:
Ah yeah will do! I'm testing the Standalone CLI to verify it works there too.

AnzeKop:
Hi @thecrypticace thanks for getting on the issue so quick, updated to latest tailwind version but I still get the same error in place.

PR #13324, opened 2024-03-22
Fix missing `xx-large` and remove double `x-large` absolute size

allain:
x-large appears twice in the set of absolute sizes.

I believe the second one is meant to be 'xx-large'.


thecrypticace:
@RobinMalfait make sure we fix this in v4 as well if it's not already been fixed there.

RobinMalfait:
@thecrypticace yep verified that it's correct there 👍

RobinMalfait (approved):

RobinMalfait:
Very nice catch, thanks for fixing!

allain:
Thank you. :)

PR #13327, opened 2024-03-22
Fix incorrect syntax for `translate-z`'s `@property` definition

RobinMalfait:
This PR is a continuation of #13321. This used to be `<length-percentage>`, but we dropped percentage support because it's not valid in #13321 and I forgot about this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


thecrypticace:
ah good point. Didn't think about needing to change these.

thecrypticace (approved):

RobinMalfait:
Yep totally missed it as well but @adamwathan caught it: https://github.com/tailwindlabs/tailwindcss/pull/13321#issuecomment-2015449074

PR #13328, opened 2024-03-22
Inherit letter spacing in form controls

thecrypticace:
(Same as #13150 but for v4)

Some browsers set `letter-spacing: normal` for form elements. This PR inherits `letter-spacing` styles for consistency since we also inherit other font attributes like family, style, line-height, etc…

### DIFF
--- a/packages/tailwindcss/preflight.css
+++ b/packages/tailwindcss/preflight.css
@@ -189,6 +189,7 @@ textarea,
   font: inherit; /* 1 */
   font-feature-settings: inherit; /* 1 */
   font-variation-settings: inherit; /* 1 */
+  letter-spacing: inherit; /* 1 */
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/preflight.css
```diff
@@ -189,6 +189,7 @@ textarea,
   font: inherit; /* 1 */
   font-feature-settings: inherit; /* 1 */
   font-variation-settings: inherit; /* 1 */
+  letter-spacing: inherit; /* 1 */
```
RobinMalfait, line 192: The big difference compared to the v3 implementation is that the selector includes `::file-selector-button`. But I tested this in various browsers and I don't see a difference so I think this good to merge 👍

RobinMalfait (approved):

PR #13334, opened 2024-03-23
Remove leading space in `box-shadow`'s `addDefaults` call

allain:
It's probably not an issue but it might cause some deep bugs when looking up the defaults by name. I tried to trace it down to see if there's some cleanup going on to remove it but got lost on the way to grandma's house.


thecrypticace (approved):

thecrypticace:
Nah, it doesn't cause issues because they're not directly looked up by name. `addDefaults` inserts fake at rules called `@defaults {name-here}` and the whitespace isn't significant.

Thanks for catching tho!

PR #13349, opened 2024-03-25
Collect user-defined CSS that can be used in `@apply`

### LINKED ISSUE #13342: [v4] @apply previous custom classes error
Just a remember that this works on v3.x but as of v4 alpha 10 it stills doesn't work, it gets the error: "Cannot apply unknown utility class: btn"

.btn {
 @apply py-2 px-3;
}
.btn-red {
 @apply btn bg-red-600;
 }

Appreciate any guidance

RobinMalfait:
This allows you to apply user-defined CSS inside of `@apply`. It works a tiny bit different compared to v3 because we only allow you to apply simple selectors. Only a single class can be applied such as a `.foo {}`.

If you want more complex selectors, then you can use nesting to enhance the definition of what `.foo` means exactly.

Implementing it this way allows us to keep the codebase simple and keep expectations simple. 

In v3 you can do crazy things such as:

```css
.foo .bar .baz {
  color: red;
}
```

And you can apply `@apply foo` or `@apply bar` or `@apply baz`, but that's not intuitive if you look at the final result.

This PR also ensures that you can't apply classes that would result in circular dependencies.

What this PR is not:

This PR does not make custom classes "utilities" or "components". This means that you won't be able to do this:

```css
.foo {
  color: red;
}

.example {
  @apply hover:foo;
  /*     ----- You will not be able to add variant modifiers to `.foo` */
}
```

Fixes: #13342

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -179,6 +179,13 @@ export function compile(css: string): {
     walk(ast, (root) => {
       if (root.kind !== 'rule') return WalkAction.Continue
 
+      // It's possible to `@apply` user-defined classes. We need to make sure
+      // that we never run into a situation where we are eventually applying
+      // the same class that we are currently processing otherwise we will end
+      // up in an infinite loop (circular dependency).
+      //
+      // This means that we need to track the current node as a candidate and
+      // error when we encounter it again.
       let rootAsCandidate = root.selector.slice(1)
 
       walk(root.nodes, (node, { replaceWith }) => {
@@ -199,6 +206,8 @@ export function compile(css: string): {
             // Collect all user-defined classes for the current candidates that
             // we need to apply.
             for (let candidate of candidates) {
+              // If the candidate is the same as the current node we are
+              // processing, we have a circular dependency.
               if (candidate === rootAsCandidate) {
                 throw new Error(`Detected circular \`@apply\`: ${candidate}`)
               }
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -176,70 +176,88 @@ export function compile(css: string): {
 
   // Replace `@apply` rules with the actual utility classes.
   if (containsAtApply) {
-    walk(ast, (node, { replaceWith }) => {
-      if (node.kind === 'rule' && node.selector[0] === '@' && node.selector.startsWith('@apply')) {
-        let candidates = node.selector
-          .slice(7 /* Ignore `@apply ` when parsing the selector */)
-          .trim()
-          .split(/\s+/g)
+    walk(ast, (root) => {
+      if (root.kind !== 'rule') return WalkAction.Continue
 
-        // Replace the `@apply` rule with the actual utility classes
-        {
-          let newNodes: AstNode[] = []
+      let rootAsCandidate = root.selector.slice(1)
 
-          // Collect all user-defined classes for the current candidates that we
-          // need to apply.
-          for (let candidate of candidates) {
-            let nodes = userDefinedApplyables.get(candidate)
-            if (!nodes) continue
+      walk(root.nodes, (node, { replaceWith }) => {
+        if (
+          node.kind === 'rule' &&
+          node.selector[0] === '@' &&
+          node.selector.startsWith('@apply')
+        ) {
+          let candidates = node.selector
+            .slice(7 /* Ignore `@apply ` when parsing the selector */)
+            .trim()
+            .split(/\s+/g)
 
-            for (let child of nodes) {
-              newNodes.push(structuredClone(child))
-            }
-          }
+          // Replace the `@apply` rule with the actual utility classes
+          {
+            let newNodes: AstNode[] = []
 
-          // Parse the candidates to an AST that we can replace the `@apply` rule with.
-          let candidateAst = compileCandidates(candidates, designSystem, {
-            onInvalidCandidate: (candidate) => {
-              // When a candidate is invalid, we want to first verify that the
-              // candidate is a user-defined class or not. If it is, then we can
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/index.ts
```diff
@@ -153,17 +185,48 @@ export function compile(css: string): {
...
+              // a known utility class. For example, the following CSS would be:
+              //
+              // ```css
+              // .flex {
+              //   --display-mode: flex;
+              // }
+              // ```
+              //
+              // When the user then uses `@apply flex`, we want to both apply
+              // the user-defined class and the utility class.
```
thecrypticace, line 220:
```suggestion
              // We must pass in user-defined classes and then filter them out
              // here because, while they are usually not known utilities, the
              // user can define a class that happens to *also* be a known
              // utility. For example, given the following, `flex` counts as
              // both a user-defined class and a known utility:
              //
              // ```css
              // .flex {
              //   --display-mode: flex;
              // }
              // ```
              //
              // When the user then uses `@apply flex`, we want to both apply
              // the user-defined class and the utility class.
```

thecrypticace (changes requested):

packages/tailwindcss/src/index.ts
```diff
@@ -143,40 +175,98 @@ export function compile(css: string): {
...
+      // This means that we need to track the current node as a candidate and
+      // error when we encounter it again.
+      let rootAsCandidate = root.selector.slice(1)
+
+      walk(root.nodes, (node, { replaceWith }) => {
+        if (
+          node.kind === 'rule' &&
+          node.selector[0] === '@' &&
+          node.selector.startsWith('@apply')
+        ) {
```
thecrypticace, line 196: I'd flip the condition and do an early return here.

packages/tailwindcss/src/index.ts
```diff
@@ -143,40 +175,98 @@ export function compile(css: string): {
...
+            let newNodes: AstNode[] = []
+
+            // Collect all user-defined classes for the current candidates that
+            // we need to apply.
+            for (let candidate of candidates) {
+              // If the candidate is the same as the current node we are
+              // processing, we have a circular dependency.
+              if (candidate === rootAsCandidate) {
+                throw new Error(
+                  `You cannot \`@apply\` the \`${candidate}\` utility here because it creates a circular dependency.`,
```
thecrypticace, line 213: I think it'd be worth printing the selector of `root` in the error message to help guide the user to the thing with the circular dependency.

thecrypticace (changes requested):

packages/tailwindcss/src/utils/is-simple-class-selector.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (selector[0] !== '.') return false
+
+  for (let i = 1; i < selector.length; i++) {
+    switch (selector[i]) {
+      // The character is escaped, skip the next character
+      case '\\':
+        i += 1
+        continue
+
+      case ' ': // Descendat combinator
```
francoismassart, line 19:
```suggestion
      case ' ': // Descendant combinator
```

packages/tailwindcss/src/index.ts
```diff
@@ -28,6 +33,33 @@ export function compile(css: string): {
...
     if (node.kind !== 'rule') return
 
+    // Track all user-defined classes for `@apply` support
+    if (
+      containsAtApply &&
+      // Verify that it is a valid applyable-class. An applyable class is a
+      // class that is a very simple selector, like `.foo` or `.bar`, but doesn't
+      // contain any spaces, combinators, pseudo-selectors, pseudo-elements, or
+      // attribute selectors.
+      node.selector[0] === '.' &&
```
francoismassart, line 43: I believe you don't need `node.selector[0] === '.'` ?

Because inside `isSimpleClassSelector`, you perform this test with an early return:
`if (selector[0] !== '.') return false`

I would suggest to remove this line
```suggestion
```

packages/tailwindcss/src/index.ts
```diff
@@ -143,40 +175,98 @@ export function compile(css: string): {
...
+      let rootAsCandidate = root.selector.slice(1)
+
+      walk(root.nodes, (node, { replaceWith }) => {
+        if (
+          node.kind === 'rule' &&
+          node.selector[0] === '@' &&
+          node.selector.startsWith('@apply')
+        ) {
+          let candidates = node.selector
+            .slice(7 /* Ignore `@apply ` when parsing the selector */)
```
francoismassart, line 198: I really enjoy the comments in your code, you could have use
```js
.slice('@apply '.length)
```
instead of
```js
.slice(7 /* Ignore `@apply ` when parsing the selector */)
```
Off course the latest is slightly faster & shorter (if you omit the comment).

In the end, I prefer your version.
Longer with the comment but once you read it once, it feels right 🤓.

RobinMalfait: Yep we focus on performance and re-allocating and re-computing the length is unnecessary work.

packages/tailwindcss/src/index.ts
RobinMalfait: This is still required, if you comment this out a test should fail. This is basically for the scenario where you have `.foo.bar` which contains 2 classes which is not allowed.

francoismassart: Good catch 😉

nam-co:
Good day, this amazing/magical feature seems to be forgotten, hopefully 🙏 this gets included in the next release, please @adamwathan

RobinMalfait:
To keep things simple, we currently only allow using `@apply` with utilities that were defined via `@utility`. Maybe we can revisit this in the future, but this keeps the mental model of what you can apply clean and simple.

Going to close it for now, _if_ we need it, it needs some work anyways.

PR #13380, opened 2024-03-27
Fix Vite serve issues

KrisBraun:
Fix Vite hanging in Remix builds by temporarily disabling use of the Vite `waitForReuestsIdle` (#13291) feature. This will cause multiple Tailwind CSS generations in some requests, especially the initial one. This might be visible as a FOUC, but the builds happen so quickly it may not be noticeable. This only affects dev server mode.

A second fix addresses an issue introduced in https://github.com/tailwindlabs/tailwindcss/pull/13218 (unreleased). The `vite:css-post` plugin was being double-applied in serve mode.


thecrypticace (approved):

thecrypticace:
Hopefully we'll get a fix in Vite soon and can re-enable this.

PR #13415, opened 2024-03-31
Speed up segmentation

ottomated:
Saw [this tweet](https://twitter.com/adamwathan/status/1767937096214257837) and wanted to pitch in. This speeds up the `segment` function from ~1.2x to ~1.5x, at the cost of some readability.

Benchmark is below. Could use some larger or smaller test cases!
image


ottomated:
#13393 also defines some char code constants - these should probably be moved out to a common file

thecrypticace (approved):

RobinMalfait (approved):

thecrypticace:
Thanks! 💯

PR #13424, opened 2024-04-01
Chore refactor repeated tests

ghost:
1. Refactor repeated tests
2. Add `rust-toolchain.toml`
3. Ignore .idea folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ playwright-report/
 blob-report/
 playwright/.cache/
 target/
+.idea/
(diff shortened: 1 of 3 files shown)


thecrypticace:
Hey, appreciate the PR but we're going to leave this as is. The repetition is not a problem right now. It might be different if we already had a few thousand such tests but there's only a handful — so the macro isn't buying us much.

Thanks anyway though!

0xqd:
@thecrypticace yeah, i think we hold off for now, macros for the win .. someday!

PR #13457, opened 2024-04-05
Ensure deterministic SSR builds in `@tailwindcss/vite`

### LINKED ISSUE #13441: [v4] Styles not being generated on project build
## Environment

```
Astro                    v4.5.15
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          bun
Output                   static
Adapter                  none
Integrations             @astrojs/svelte

tailwindcss             v4.0.0-alpha.11
@tailwindcss/vite       v4.0.0-alpha.11
```
Can also reproduce the issue with using npm instead of bun.

**Reproduction URL**

Repo: https://github.com/Brian-Pob/tailwind-v4-vite/tree/alpha-11-bug

Vercel deploy preview: https://tailwind-v4-vite-git-alpha-11-bug-brian-pobletes-projects.vercel.app/?_vercel_share=HBCiq51fIy1xjh2hdYtOnZrineQnVvSn

Vercel working deploy: https://tailwind-v4-vite.vercel.app/

**Describe your issue**

Outcome

Running `bun run build` does not generate Tailwind styles. The build process completes successfully but the styles are not applied to the output.

alpha 11 bug

Expected outcome

Running `bun run build` should generate Tailwind styles and apply them to the output.

Using Tailwind `4.0.0-alpha.10` with same versions for every other package produces the expected outcome.

## Steps to Reproduce
1. Clone this branch of the repository.
2. Ensure Tailwind version is `v4.0.0-alpha.11`. Same for `@tailwindcss/vite`.
3. Run `bun install`.
4. Run `bun run build` and `bun run preview`.
5. Open the preview URL in the browser.

RobinMalfait:
This PR fixes an issue where running `astro build` doesn't generate the correct CSS when using `@tailwindcss/vite`. This issue only seems to happen when running a build, and not when running in `dev` mode.

To fix this, we apply the same code (`waitForRequestsIdle`) as we do in the `apply: "server"` step.

Adding this to the `apply: "build"` step as well does generate the correct result.

While this does fix the issue, I'm not 100% convinced that this is the correct solution /cc @thecrypticace.

Fixes: #13441

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -187,6 +187,9 @@ export default function tailwindcss(): Plugin[] {
         cssModules[id] = src
 
         if (!options?.ssr) {
+          // Wait until all other files have been processed, so we can extract
+          // all candidates before generating CSS. This must not be called
+          // during SSR or it will block the server.
           await server?.waitForRequestsIdle?.(id)
         }
 
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -182,9 +182,16 @@ export default function tailwindcss(): Plugin[] {
       name: '@tailwindcss/vite:generate:build',
       apply: 'build',
 
-      transform(src, id) {
+      async transform(src, id, options) {
         if (!isTailwindCssFile(id, src)) return
         cssModules[id] = src
+
+        if (!options?.ssr) {
+          await server?.waitForRequestsIdle?.(id)
+        }
+
+        let code = await transformWithPlugins(this, id, generateCss(src))
+        return { code }
       },
 
       // renderChunk runs in the bundle generation stage after all transforms.
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -85,7 +85,7 @@ export default function tailwindcss(): Plugin[] {
       try {
         // Directly call the plugin's transform function to process the
         // generated CSS. In build mode, this updates the chunks later used to
-        // generate the bundle. In serve mode, the transformed souce should be
+        // generate the bundle. In serve mode, the transformed source should be
         // applied in transform.
         let result = await transformHandler.call(transformPluginContext, css, id)
         if (!result) continue
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -9,7 +9,13 @@ export default function tailwindcss(): Plugin[] {
   let candidates = new Set<string>()
   // In serve mode, we treat this as a set, storing storing empty strings.
   // In build mode, we store file contents to use them in renderChunk.
-  let cssModules: Record<string, string> = {}
+  let cssModules: Record<
+    string,
+    {
+      content: string
+      handled: boolean
+    }
+  > = {}
   let minify = false
   let cssPlugins: readonly Plugin[] = []
 
@@ -163,7 +169,7 @@ export default function tailwindcss(): Plugin[] {
         if (!isTailwindCssFile(id, src)) return
 
         // In serve mode, we treat cssModules as a set, ignoring the value.
-        cssModules[id] = ''
+        cssModules[id] = { content: '', handled: true }
 
         if (!options?.ssr) {
           // Wait until all other files have been processed, so we can extract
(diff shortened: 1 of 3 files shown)


thecrypticace:
I don't think this is the right fix.
- I'm not certain if `waitForRequestsIdle` is valid to call during a normal build — rather than just for the dev server.
- As I understand it, we have to use `renderChunk` during the build because that's the only way we'll have the guarantee that all files have already been seen.

Even if `waitForRequestsIdle()` works during a normal build it won't during an SSR build which means we'd be back to needing to use `renderChunk` for SSR builds and we'd be left with similar problems.

Our setup works in a normal Vite project, and even a basic Astro project. The problem in the project in the issue is caused by a few things:
1. Importing **and** using the `Image` component from `astro:assets` also breaks
2. An Astro component containing `import.meta.glob(…)` in the front matter block causes our implementation renderChunk to not work. Dynamic imports in general seem to have this issue.
3. However, an eager glob also causes renderChunk to break.

I'm not sure yet if this is a problem specific to Astro or is in Vite itself. I need to do some more investigating.

Brian-Pob:
Hi. I'm the person who filed the issue. I did a bit more testing with my project and removed the `ProjectCard.astro` component that uses astro's `Image` and the usage of `import.meta.glob(...)`.

The issue still seems to persist when using the latest version of `@tailwindcss/vite`.

The screenshot below shows that I commented out the component usage but I also completely deleted the file.

https://github.com/Brian-Pob/tailwind-v4-vite/tree/alpha-11-bug-2

CleanShot 2024-04-05 at 13 34 49

Brian-Pob:
And please feel free to reach out to me in the Tailwind Discord if you would like me to try anything or run any tests.

thecrypticace:
@Brian-Pob Importing a Svelte component _also_ seems to do it.

Even then, it appears that when a component statically imports images, the build's CSS is non-deterministic. Sometimes our plugin appears to run at the appropriate time and sometimes it does not.

fwiw, I stress-tested this over ~500 runs and this did *not* happen for a vanilla Vite installation. Only when astro components are involved.

thecrypticace:
Notes for me later:

I stripped down the astro build to just enable the following plugins:

- `@astro/plugin-middleware`
- `@astrojs/vite-plugin-astro-ssr-manifest`
- `astro:build`

My guess is that `astro:build` is possibly doing something with the module graph that is causing this non-determinism possibly _because_ our `renderChunk` method re-runs the `vite:css` and `vite:css-post` plugins on our output. I'm not certain of this but it would be the next thing to test.

thecrypticace:
Something interesting I've discovered in my stripped down test is that anytime `transform()` for **an image** get's called AFTER the `transform()` for the CSS containing `@tailwind` the output is broken.

Which means if we delay resolution to make sure the CSS file is transformed _after_ all images things start working. In my stripped down reproduction a delay of ~100ms is sufficient. In the original reproduction that was provided the delay has to be a good bit longer.

So, in the world of "how and why does this work", this plugin solves the issue:

```ts
{
  name: "delay-the-css",
  enforce: "pre",

  async resolveId(id) {
    if (!id.includes(".css")) return;
    await new Promise((resolve) => setTimeout(resolve, 3000));
  },
},
```

I'm still investigating the root of the problem. I also tested this in a Vite + Vue project and could not reproduce the issue. Even when manually delaying images such that they were transformed after CSS files.

thecrypticace:
Okay @RobinMalfait I figured this one out. During SSR there is a server with a module graph (when normally there isn't during a build) and we're deleting references to modules that don't exist. Normally this is fine but during an SSR build the module won't get reloaded unless its transform is handled after everything else.

This is what's causing the problem — and the "after everything else" is the cause of the non-deterministic behavior. With this the problem should be resolved entirely.

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -113,16 +125,21 @@ export default function tailwindcss(): Plugin[] {
...
-        // URL path rewriting and image inlining.
-        //
-        // In build mode, since renderChunk runs after all transformations, we
-        // need to also apply vite:css-post.
-        cssPlugins = config.plugins.filter((plugin) =>
-          ['vite:css', ...(config.command === 'build' ? ['vite:css-post'] : [])].includes(
-            plugin.name,
-          ),
-        )
+        isSSR = config.build.ssr !== false && config.build.ssr !== undefined
```
RobinMalfait, line 128: I was confused why the second undefined check was there, but didn't realise `ssr` could be a string.

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -85,7 +97,7 @@ export default function tailwindcss(): Plugin[] {
       try {
         // Directly call the plugin's transform function to process the
         // generated CSS. In build mode, this updates the chunks later used to
-        // generate the bundle. In serve mode, the transformed souce should be
+        // generate the bundle. In serve mode, the transformed source should be
```
RobinMalfait, line 100: Nice

RobinMalfait (commented):
Makes sense what you did here @thecrypticace, nice debugging job on the SSR stuff!

packages/@tailwindcss-vite/src/index.ts
thecrypticace: heh yeah I looked at the types and I was like… oh okay

thecrypticace (approved):

PR #13478, opened 2024-04-08
Invert variant order application

RobinMalfait:
This PR can be a breaking change, but it will make the order of variants more intuitive. Instead of applying the variants "inside-out", they will be applied "left-to-right".

What is affected:

- Utilities with multiple variants, where the order matters.

  E.g.: In combination with pseudo elements:

  - `before:hover:flex` — This means: "when the before pseudo element is hovered, apply flex". This doesn't work in CSS.
  - `hover:before:flex` — This means: "when the element is hovered, apply flex to the before pseudo element".

  E.g.: Variants that rely on another class in the selector. This will now change the order of the classes.

  - `*:first:underline`
    ```
    *:first:underline
      ^^^^^ first child `&:first-child`
    ^       Child selector `& > *`
    ```

    - Previous behavior:  "when the current element is the first child, add an underline to all its direct children"
      E.g.: `.foo:first-child > *`
    - New behavior: "add an underline to the first child of the current element"
      E.g.: `.foo > *:first-child`

What's not affected:

- Utilities used without variants. E.g.: `flex`
- Utilities used with a single variant. E.g.: `hover:flex`
- Utilities used with multiple variants, but where the order doesn't matter. E.g.: `hover:focus:flex`

### DIFF
--- a/packages/tailwindcss/src/index.test.ts
+++ b/packages/tailwindcss/src/index.test.ts
@@ -333,7 +333,7 @@ describe('arbitrary variants', () => {
 describe('variant stacking', () => {
   it('should stack simple variants', () => {
     expect(run(['focus:hover:flex'])).toMatchInlineSnapshot(`
-      ".focus\\:hover\\:flex:hover:focus {
+      ".focus\\:hover\\:flex:focus:hover {
         display: flex;
       }"
     `)
@@ -341,7 +341,7 @@ describe('variant stacking', () => {
 
   it('should stack arbitrary variants and simple variants', () => {
     expect(run(['[&_p]:hover:flex'])).toMatchInlineSnapshot(`
-      ".\\[\\&_p\\]\\:hover\\:flex:hover p {
+      ".\\[\\&_p\\]\\:hover\\:flex p:hover {
         display: flex;
       }"
     `)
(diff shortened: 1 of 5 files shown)


packages/tailwindcss/src/index.test.ts
```diff
@@ -359,8 +359,11 @@ describe('variant stacking', () => {
...
   it('pseudo element variants are re-ordered', () => {
     expect(run(['before:hover:flex', 'hover:before:flex'])).toMatchInlineSnapshot(`
-      ".before\\:hover\\:flex:hover:before {
+      ".before\\:hover\\:flex:before {
         content: var(--tw-content);
+      }
+
+      .before\\:hover\\:flex:before:hover {
         display: flex;
       }
```
RobinMalfait, line 368: This is now added because we don't do any "fixups" to correct this behavior.

packages/tailwindcss/src/index.test.ts
```diff
@@ -359,8 +359,11 @@ describe('variant stacking', () => {
 
   it('pseudo element variants are re-ordered', () => {
     expect(run(['before:hover:flex', 'hover:before:flex'])).toMatchInlineSnapshot(`
-      ".before\\:hover\\:flex:hover:before {
+      ".before\\:hover\\:flex:before {
         content: var(--tw-content);
+      }
+
+      .before\\:hover\\:flex:before:hover {
```
RobinMalfait, line 366: This is now added because we don't do any "fixups" to correct this behavior.

adamwathan (approved):

PR #13529, opened 2024-04-16
fix: doesn't work spaces inside calc

bae-sh:
This PR fix a uncomfortable thing that spaces inside calc doesn't work.

When i use arbitrary calc values with spaces `<div class="w-[calc(100px + 200px)]></div>`, tailwind extract separate words.
Finally, It doesn't work.

So I've added a function that remove spaces inside calc. It can remove spaces inside calc before using extractor.

#### before

`w-[calc(100px + 200px)]` -> `w-[calc(100px` , `+` , `200px)]`

#### after

`w-[calc(100px + 200px)]` -> `w-[calc(100px+200px)]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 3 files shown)


bae-sh:
Thank you for the review!

The intent of this PR was not only to avoid spaces inside calc, but also to be recognized when spaces are used. 
However, i was missing the fact that the html itself uses spaces to separate them. 

Thanks for pointing that out! I hope I can contribute again in the future. 😀

PR #13608, opened 2024-04-29
Ensure strings are consumed as-is when using internal `segment()`

### LINKED ISSUE #13607: v4: Bug with how the parser handles nested `quote_stack` and `bracket_stack` characters when in a string context
<details><summary> Technical Configuration</summary>
<br>
<table>
    <tr></tr>
    <tr>
        <th align="left">What version of Tailwind CSS are you using?</th>
    </tr>
    <tr>
        <td>Next (v4, latest)</td>
    </tr>
    <tr>
        <th align="left">What build tool (or framework if it abstracts the build tool) are you using?</th>
    </tr>
    <tr>
        <td>`pnpm build && pnpm test`</td>
    </tr>
    <tr>
        <th align="left">What version of Node.js are you using?</th>
    </tr>
    <tr>
        <td>v21.6.1</td>
    </tr>
    <tr>
        <th align="left">What browser are you using?</th>
    </tr>
    <tr>
        <th align="left">N/A</td>
    </tr>
    <tr>
        <th align="left">What operating system are you using?</th>
    </tr>
    <tr>
        <td>macOS</td>
    </tr>
</table>

</details>

---

**Reproduction URL**

The accepted test results from #13596

**Describe your issue**

> attn @adamwathan @RobinMalfait

This issue delves into some of the issues surrounding why the parser doesn't properly handle `supports-[content:"("]:grid`, first explored in #13596. This specific case turned out to be quite the rabbit hole.

From what I can tell, Tailwind's Rust parser runs into issues related to its `quote_stack` and `bracket_stack` when encountering a construct like `'('`.

None of these process correctly, apparently for this same reason:
* `supports-[content:'(']`
* `supports-[content:'[']`
* `supports-[content:'"']`
* `supports-[content:"'"]`

This Tailwind Play example demonstrates the issue I'm describing: https://play.tailwindcss.com/EAhSfQFfTa

I can think of a few potential solutions, all of which come with trade-offs:

1. We could interpret everything after an open quote as a string until we hit the closing quote. This would break some things, but we'd be able to handle brackets or quotes within strings.

2. We could always ignore `bracket_stack` characters within `quote_stack` characters, but not ignore `quote_stack` characters within `bracket_stack` characters.

3. Specifically enforce escaping characters related to `quote_stack` and `bracket_stack` when they're in a string context. This would mean that `supports-[content:"("]` would need to be written as `supports-[content:"\("]` to work.

   * `supports-[content:'(']` -> `supports-[content:'\(']`
   * `supports-[content:'[']` -> `supports-[content:'\[']`
   * `supports-[content:'"']` -> `supports-[content:'\"']`
   * `supports-[content:\"'\"]` -> `supports-[content:\"\'\"]`
   ☝🏼 this last one might be problematic for more reasons than one, related to escaping the same quotes used when opening the `class` attribute as well as escaping the inner `quote_stack` character

4. We could strictly enforce escaping special characters like these in string contexts.

   This might be most consistent and with common escape patterns, but it would also be one of the most disruptive solutions.
   
   In the case of my Multi plugin, for example, I was surprised—when chatting with @RobinMalfait ([see discussion](https://github.com/tailwindlabs/tailwindcss/issues/13473#issuecomment-2059785107))—to discover that nesting quotes in arbitrary values is not only supported but that they appear to be infinitely nestable.
   
   In other words, this works:
   
   ```
   multi-['multi-['multi-['font-bold']'];text-green-700;before:content-['$']']
   ```

   If escapes were required to resolve this issue, each level of nesting might require a greater level of escape characters, which could get out of hand quickly. For example, the above would need to be written like this:
   
   ```
   multi-['multi-[\'multi-[\\\'font-bold\\\']\'];text-green-700;before:content-[\'$\']']
   ```
   
   Of course, it would be convenient for me not to need to update my plugin(s) with new syntax and safety measures again, and such escaping would probably render my plugin too cumbersome to justify using. However, greater parity with CSS is probably a higher and more evergreen standard here.
   
   Simply wrapping the arbitrary value in quotes was an easy fix, but the magic around it allowing nested quotes may be the culprit behind `"("` not working, to some degree.
   
   <table><tr><td>
   
   [conversation between Brandon McConnell and Robin Malfait about the nature of quoted strings in Tailwind CSS](https://github.com/tailwindlabs/tailwindcss/issues/13473#issuecomment-2059785107)
   
   </td></tr></table>

---

I'm new to Rust, so all of this is "to the best of my current knowledge" from what I could grok reading through the parser. I also consulted a colleague who is more experienced with Rust to corroborate my findings and theories, and they came to roughly the same conclusions.

RobinMalfait:
This PR fixes an issue where parens, bracket and curlies needed to be balanced inside of strings instead of consuming the string until the end.

When encountering strings when using `segment` we didn't really treat them as actual strings. This means that if you used any parens, brackets, or curlies then we wanted them to be properly balanced.

This should not be the case, whenever we encounter a string, we want to consume it as-is and don't want to worry about bracket balancing. We will now consume it until the end of the string (and make sure that escaped closing quotes are not seen as real closing quotes).

Fixes: #13607

### DIFF
--- a/packages/tailwindcss/src/utils/segment.test.ts
+++ b/packages/tailwindcss/src/utils/segment.test.ts
@@ -45,10 +45,6 @@ it('should skip escaped single quotes', () => {
   expect(segment(String.raw`a:'b:c\':d':e`, ':')).toEqual(['a', String.raw`'b:c\':d'`, 'e'])
 })
 
-it('should not crash when single quotes are unbalanced', () => {
-  expect(segment("a:'b:c:d", ':')).toEqual(['a', "'b:c:d"])
-})
-
 it('should split by the escape sequence which is escape as well', () => {
   expect(segment('a\\b\\c\\d', '\\')).toEqual(['a', 'b', 'c', 'd'])
   expect(segment('a\\(b\\c)\\d', '\\')).toEqual(['a', '(b\\c)', 'd'])
--- a/packages/tailwindcss/src/utils/segment.test.ts
+++ b/packages/tailwindcss/src/utils/segment.test.ts
@@ -21,6 +21,34 @@ it('should not split inside of curlies', () => {
   expect(segment('a:{b:c}:d', ':')).toEqual(['a', '{b:c}', 'd'])
 })
 
+it('should not split inside of double quotes', () => {
+  expect(segment('a:"b:c":d', ':')).toEqual(['a', '"b:c"', 'd'])
+})
+
+it('should not split inside of single quotes', () => {
+  expect(segment("a:'b:c':d", ':')).toEqual(['a', "'b:c'", 'd'])
+})
+
+it('should not crash when double quotes are unbalanced', () => {
+  expect(segment('a:"b:c:d', ':')).toEqual(['a', '"b:c:d'])
+})
+
+it('should not crash when single quotes are unbalanced', () => {
+  expect(segment("a:'b:c:d", ':')).toEqual(['a', "'b:c:d"])
+})
+
+it('should skip escaped double quotes', () => {
+  expect(segment(String.raw`a:"b:c\":d":e`, ':')).toEqual(['a', String.raw`"b:c\":d"`, 'e'])
+})
+
+it('should skip escaped single quotes', () => {
+  expect(segment(String.raw`a:'b:c\':d':e`, ':')).toEqual(['a', String.raw`'b:c\':d'`, 'e'])
+})
+
+it('should not crash when single quotes are unbalanced', () => {
+  expect(segment("a:'b:c:d", ':')).toEqual(['a', "'b:c:d"])
+})
+
 it('should split by the escape sequence which is escape as well', () => {
   expect(segment('a\\b\\c\\d', '\\')).toEqual(['a', 'b', 'c', 'd'])
   expect(segment('a\\(b\\c)\\d', '\\')).toEqual(['a', '(b\\c)', 'd'])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Make sure `contain-*` utility variables resolve to a valid value ([#13521](https://github.com/tailwindlabs/tailwindcss/pull/13521))
-- Ensure strings are consumed as-is when using internal `segment()` ([#13608](https://github.com/tailwindlabs/tailwindcss/pull/13608))
+- Support unbalanced parentheses and braces in quotes in arbitrary values and variants ([#13608](https://github.com/tailwindlabs/tailwindcss/pull/13608))
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


packages/tailwindcss/src/utils/segment.test.ts
```diff
@@ -21,6 +21,34 @@ it('should not split inside of curlies', () => {
...
+  expect(segment(String.raw`a:"b:c\":d":e`, ':')).toEqual(['a', String.raw`"b:c\":d"`, 'e'])
+})
+
+it('should skip escaped single quotes', () => {
+  expect(segment(String.raw`a:'b:c\':d':e`, ':')).toEqual(['a', String.raw`'b:c\':d'`, 'e'])
+})
+
+it('should not crash when single quotes are unbalanced', () => {
+  expect(segment("a:'b:c:d", ':')).toEqual(['a', "'b:c:d"])
+})
```
thecrypticace, line 50: this test is listed twice so this last one can be dropped

thecrypticace (changes requested):
LGTM — just one tweak in the tests plz

thecrypticace (approved):

brandonmcconnell:
@RobinMalfait @thecrypticace It may be useful to include tests for bracket and quote chars in arbitrary values, specifically, to include tests similar to the original issue: e.g. `supports-[content:"("]:[content:"("]`

We don't need to include all of these, but these should all pass with the proper fix, perhaps with some minor alterations: https://play.tailwindcss.com/EAhSfQFfTa

adamwathan (commented):
Nice! This looks good to me.

Couple things:

- Should we add a test _somewhere_ for the higher level issue that made us notice this, or is there not really a good/obvious place to put it? It's kinda weird to put it in the `supports` tests because it applies to other stuff like the `data` variant too.
- Did you benchmark this against the previous implementation (even though it had a bug)? Hoping it's basically negligible.

brandonmcconnell:
@RobinMalfait Thanks for the quick fix here!

RobinMalfait:
Yeah good ideas to add another, more high level test.

My thinking is to add a generic test for candidate parsing so that we know that we at least parse it into the correct data structure. If we want, we can add a specific one for `supports-[...]` but as you mentioned, this is a more generic fix anyway.

Benchmarking wise, it's actually even better when we encounter strings because we can skip a lot of work. The checks to know that we are in a string or a single integer check.

Current PR:
```
 ✓ src/utils/segment.bench.ts (1) 792ms
     name               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · segment  1,087,386.63  0.0008  0.9782  0.0009  0.0009  0.0010  0.0012  0.0033  ±0.53%   543694   fastest
```

Next branch:
```
 ✓ src/utils/segment.bench.ts (1) 789ms
     name               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · segment  1,042,913.63  0.0008  0.9799  0.0010  0.0010  0.0011  0.0014  0.0028  ±0.55%   521457   fastest
```

PR #13660, opened 2024-05-08
Use long form `<length> | <percentage>` syntax for properties

thecrypticace:
In Firefox Nightly, when using properties that declare the shorthand syntax `<length-percentage>`, there is a bug where two properties side-by-side e.g. `var(—foo)var(—bar)` invalidate the property value when it should not. Switching the `@property` definition to use the long form syntax `<length> | <percentage>` fixes this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use `length` data type for `background-size` instead of `background-position` ([#13771](https://github.com/tailwindlabs/tailwindcss/pull/13771))
 - Support negative values for `{col,row}-{start,end}` utilities ([#13780](https://github.com/tailwindlabs/tailwindcss/pull/13780))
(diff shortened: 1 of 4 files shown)


RobinMalfait (commented):
Should we update the `<length-percentage>` for gradient stops as well for completeness sake? 🤔 

It looks like it's working, even though it's also generating concatenated vars:
```css
.from-red-500 {
  --tw-gradient-from: var(--color-red-500, #ef4444);
  --tw-gradient-stops: var(--tw-gradient-via-stops, var(--tw-gradient-from)var(--tw-gradient-from-position), var(--tw-gradient-to)var(--tw-gradient-to-position))
}
```

https://jsfiddle.net/jyqn70w1/1/

thecrypticace:
Not sure what you mean? The variable `--tw-gradient-stops` is `*` not `<length-percentage>`. I'd already updated the position properties in the above commit:
https://github.com/tailwindlabs/tailwindcss/pull/13660/commits/6a283170677e6f8d6de63eb0c39f74169077618d#diff-036c40df8e1498a5c399fb971238a19fcdab257001b9703661a46a110984b56aR2574-R2576

RobinMalfait (approved):

PR #13661, opened 2024-05-08
Add `nth-*` variants

thecrypticace:
This adds variants for the `:nth-child()`, `:nth-last-child()`, `:nth-of-type()`, and `:nth-last-of-type()` pseudo classes. These variants support arbitrary values as well as numeric bare values. Refer to the table below for example usasges on what selectors these variants produce:

| Variant | Selector |
| -- | -- |
| `nth-3:*` | `&:nth-child(3)` |
| `nth-[2n+1]:*` | `&:nth-child(2n+1)` |
| `nth-[2n+1_of_.item]:*` | `&:nth-child(2n+1 of .item)` |
| `nth-last-3:*` | `&:nth-last-child(3)` |
| `nth-last-[2n+1]:*` | `&:nth-last-child(2n+1)` |
| `nth-last-[2n+1_of_.item]:*` | `&:nth-last-child(2n+1 of .item)` |
| `nth-of-type-3:*` | `&:nth-of-type(3)` |
| `nth-of-type-[2n+1]:*` | `&:nth-of-type(2n+1)` |
| `nth-last-of-type-3:*` | `&:nth-last-of-type(3)` |
| `nth-last-of-type-[2n+1]:*` | `&:nth-last-of-type(2n+1)` |


packages/tailwindcss/src/variants.test.ts
```diff
@@ -1594,6 +1594,67 @@ test('forced-colors', () => {
...
+    }
+
+    .nth-last-of-type-\\[2n\\+1\\]\\:flex:nth-last-of-type(odd) {
+      display: flex;
+    }"
+  `)
+
+  expect(
+    run(['nth-foo:flex', 'nth-of-type-foo:flex', 'nth-last-foo:flex', 'nth-last-of-type-foo:flex']),
+  ).toMatchInlineSnapshot(`""`)
```
RobinMalfait, line 1655: Nit: let's use `.toEqual('')` instead, that way it's similar to the other cases and that way we guarantee that we never accidentally update a snapshot (if it starts producing results all of a sudden)

RobinMalfait (approved):
Only thing it doesn't really do is read from a config, but I think think that even makes sense so this is looking good to me 👍

packages/tailwindcss/src/variants.test.ts
thecrypticace: good call. updated.

SH20RAJ (commented):
Nice to see this ✅

adamwathan (approved):
Looks good 👍

PR #13693, opened 2024-05-16
Add ESM build of the `@tailwindcss/postcss` package

wesbos:
This PR adds the ability to import the postcss plugin into esm files.

the package is currently published as CJS, but does not provide an export for ESM files to consume it. I ran into this when  trying to use the postcss plugin inside my esm `vite.config.ts` file.

See: https://nodejs.org/api/packages.html#conditional-exports

"import" - matches when the package is loaded via import or import(), or via any top-level import or resolve operation by the ECMAScript module loader. Applies **regardless of the module format of the target file**.

### DIFF
--- a/packages/@tailwindcss-postcss/package.json
+++ b/packages/@tailwindcss-postcss/package.json
@@ -12,7 +12,7 @@
   "homepage": "https://tailwindcss.com",
   "scripts": {
     "lint": "tsc --noEmit",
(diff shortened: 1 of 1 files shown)


thecrypticace:
heh I didn't do this because it broke nextjs. I wonder if that's fixed now…

thecrypticace:
looks like it is 💯  — i think we can actually ship a proper ESM build with this now too.

wesbos:
👌 Sweet, feel free to close this then

RobinMalfait (approved):

PR #13780, opened 2024-06-03
[v4] Support negative values for `{col,row}-{start,end}` utilities

### LINKED ISSUE #13754: Col-end does not support negative values
This is a very small issue, col-end does not support negative values out of the box, but negative values makes a lot of sense there. This happens on **v3.4.3**.

https://play.tailwindcss.com/8L8o0VhDfr

The code is here:

~~~ html
<div class="grid grid-cols-3">
  <div class="-col-end-1">BROKEN</div>
  <div class="col-end-[-1]">WORKS</div>
</div>
~~~

I'm not sure if this is exactly a bug, because it is not explicitly described in the documentation, but it is a broken expectation (I've expected that to work for consistency).

thecrypticace:
Fixes #13754

Not including support for this seems to have been an oversight. We'll be backporting this to v3 as well.


RobinMalfait (approved):

RobinMalfait:
We should also add this for rows

thecrypticace:
Added

RobinMalfait (approved):

PR #13781, opened 2024-06-03
[v3] Support negative values for `{col,row}-{start,end}` utilities

### LINKED ISSUE #13754: Col-end does not support negative values

thecrypticace:
Fixes #13754


RobinMalfait (approved):

RobinMalfait:
Just a changelog item 👍

PR #13815, opened 2024-06-09
Prevent Preflight from affecting list semantics

benface:
The comment should be self-explanatory. I learned this trick thanks to Sara Soueidan's talk at CSS Day. 🙃


thecrypticace:
Looks like using `list-style: ""` works too given that the semantics are lost when the type is set to `none` and since it's set to `""` they stay around. Interestingly some screen readers don't seem to care if `none` is used but some other ones do so this seems like a good change. 👍

thecrypticace (approved):

sntran:
How would we go about re-adding it? For example, we want the `<ul>` to have bullet on desktop but not on tablet, while wanting to keep its sematics?

benface:
Huh, good point @sntran. I guess you can do `md:list-['']` or customize your theme to have `{ listStyleType: { none: "''" } }`. But I wonder if that should be changed in the default theme @thecrypticace?

PR #13826, opened 2024-06-12
Disable automatic `var()` injection for anchor properties

### LINKED ISSUE #13818: Arbitrary value for `anchor-name` outputs as custom variable instead of identifier value

`v3.4.4`

`"next": "^14.2.3"`

`v18.15.0`

N/A

macOS

**Reproduction URL**

https://play.tailwindcss.com/fagqAhZl3j

```html
<div class="[anchor-name:--foo]"></div>

<div class="hover:[anchor-name:--foo]"></div>
```

**Describe your issue**

Arbitrary value for `anchor-name` outputs as custom variable (`var(--foo)`) instead of the desired identifier value (`--foo`).

Per specs by Chrome: https://developer.chrome.com/blog/anchor-positioning-api

[Current issue] the above outputs as:
```css
.\[anchor-name\:--foo\] {
  anchor-name: var(--foo);
}

.hover\:\[anchor-name\:--foo\]:hover {
  anchor-name: var(--foo);
}
```

[Desired] `--foo` instead of `var(--foo)`
```css
.\[anchor-name\:--foo\] {
  anchor-name: --foo;
}

.hover\:\[anchor-name\:--foo\]:hover {
  anchor-name: --foo;
}
```

thecrypticace:
This PR disables automatic var() injsection for the following properties related to CSS Anchor Positioning as they all take dashed identifiers as values:

- `anchor-name`
- `anchor-scope`
- `position-anchor`
- `position-try-options`
- `position-try`

Before this PR when using these properties with custom properties you'd get output like this — notice how we wrap the identifier in `var()`:

```css
.\[anchor-name\:--tooltip\] {
  anchor-name: var(--tooltip);
}

.\[position-anchor\:--tooltip\] {
  position-anchor: var(--tooltip);
}
```

Normally, this behavior is reasonable for most properties like colors, sizes, spacing, etc… but these properties take a dashed ident directly as a custom name and are not referencing a custom property. So, after this PR you'll get the following output:

```css
.\[anchor-name\:--tooltip\] {
  anchor-name: --tooltip;
}

.\[position-anchor\:--tooltip\] {
  position-anchor: --tooltip;
}
```

Fixes #13818

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Disable automatic `var()` injection for anchor properties ([#13826](https://github.com/tailwindlabs/tailwindcss/pull/13826))
 
 ## [3.4.4] - 2024-06-05
 
--- a/src/util/dataTypes.js
+++ b/src/util/dataTypes.js
@@ -19,6 +19,7 @@ function isCSSFunction(value) {
 // More info:
 // - https://drafts.csswg.org/scroll-animations/#propdef-timeline-scope
 // - https://developer.mozilla.org/en-US/docs/Web/CSS/timeline-scope#dashed-ident
+// - https://www.w3.org/TR/css-anchor-position-1
 //
 const AUTO_VAR_INJECTION_EXCEPTIONS = new Set([
   // Concrete properties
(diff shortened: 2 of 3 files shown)


adamwathan (commented):
Do we have tests for the other exceptions that we should update for these new ones?

thecrypticace:
We have tests but they only contain two cases:
```js
  let config = {
    content: [
      // Automatic var injection
      { raw: '[color:--foo]' },

      // Automatic var injection is skipped
      { raw: '[timeline-scope:--foo]' },
    ],
  }
```

I can add all of them though if you think it's warranted

adamwathan (approved):
Approved

RobinMalfait (approved):

PR #13900, opened 2024-06-26
Loosen :is() wrapping rules in applyImportantSelector for more readable output

joshwilsonvu:
Hi, this is just a small quality of life improvement to the changes made in #10835, intended for a v3.x.

I'm using the `important` selector strategy, and noticed that for most cases, output like

```css
.tw :is(.text-citrine-500) {
  --tw-text-opacity: 1;
  color: rgb(246 220 82 / var(--tw-text-opacity)) /* #f6dc52 */;
}
```

can be reduced to simply

```css
.tw .text-citrine-500 {
  --tw-text-opacity: 1;
  color: rgb(246 220 82 / var(--tw-text-opacity)) /* #f6dc52 */;
}
```

because the nested selector doesn't contain a descendant selector like `.dark .text-citrine-500`. I'm sure the unnecessary `:is()` would be removed in minification, but having the "cleanest" possible output for Intellisense would be nice. (Trying to convince my coworkers to get on board, and making the previews feel more "hand-written" might help! 😁)

As far as I can tell, this change can't affect the behavior of any possible selector passed in. Feel free to close if you can find a counterexample. Thank you!

**Edit:** we're switching from the important selector strategy to prefixing, so we won't actually need this, if that affects whether you'll consider it.

### DIFF
--- a/src/util/applyImportantSelector.js
+++ b/src/util/applyImportantSelector.js
@@ -5,13 +5,12 @@ export function applyImportantSelector(selector, important) {
   let sel = parser().astSync(selector)
 
   sel.each((sel) => {
-    // Wrap with :is if it's not already wrapped
-    let isWrapped =
-      sel.nodes[0].type === 'pseudo' &&
-      sel.nodes[0].value === ':is' &&
-      sel.nodes.every((node) => node.type !== 'combinator')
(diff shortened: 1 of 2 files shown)


thecrypticace (approved):

dickeylth:
This solves a critical performance issue in our long list scroll render scene, thanks! Great job!

PR #13912, opened 2024-06-28
Parent state modifier without explicit parent (in-*)

eloyesp:
I'm doing a follow up on https://github.com/tailwindlabs/tailwindcss/discussions/13751 where I propose a new modifier, that is a simplified "group-*" without an explicit parent.

There are many times when the styling is based on the parent state, but it is not important which parent, there are many examples of that on the tests with the form of: `[.checked &]:underline`, but there are many other cases, for example:

- `in-aria-busy`: If any parent is busy is enough, we don't care which parent it is.
- `in-data-active`: This is part of an active element.

The syntax is quite simple, the implementation is clean and the generated css is shorter than using `group-*` or `[.checked_&]:underline`, and it also allows using variants like data- and aria-.

Hope you like the idea.

### DIFF
--- a/src/corePlugins.js
+++ b/src/corePlugins.js
@@ -494,6 +494,9 @@ export let variantPlugins = {
 
   dataVariants: ({ matchVariant, theme }) => {
     matchVariant('data', (value) => `&[data-${normalize(value)}]`, { values: theme('data') ?? {} })
+    matchVariant('in-data', (value) => `[data-${normalize(value)}] &`, {
+      values: theme('data') ?? {},
+    })
     matchVariant(
       'group-data',
       (value, { modifier }) =>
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey!

Now that Tailwind CSS v4 is out, we won't be adding more features to v3. That said, we did implement this feature in v4 so I would highly recommend to upgrade to v4 if you can!

We wrote an upgrade guide and included some tooling to upgrade your projects to minimize the amount of work you have to do yourself. More info: https://tailwindcss.com/docs/upgrade-guide

Docs related to the `in-*` variant: https://tailwindcss.com/docs/hover-focus-and-other-states#implicit-groups

Thanks again for this feature!

eloyesp:
Yes, I already saw it and I'm already using this! Thanks a lot!

Issue #13927, opened 2024-06-30
Add support for tailwind.config.mjs and tailwind.config.mts


jscarle:
**What version of Tailwind CSS are you using?**

tailwindcss 3.4.4

**What build tool (or framework if it abstracts the build tool) are you using?**

vue 3.4.31, typescript 5.5.2, vue-tsc 2.0.24, vite 5.3.2

**What version of Node.js are you using?**

node 22.3.0

**What browser are you using?**

N/A

**What operating system are you using?**

Windows

**Reproduction URL**

https://github.com/jscarle/ReproTailwindCSS

**Describe your issue**

In preparation for Node 22 and ESM module requirements, I have started converting my projects to use `.mts` files so that vite transforms them to `.mjs` files. When building the project with vite using `tailwind.config.ts`, node generates the following error:

<img width="344" alt="image" src="https://github.com/tailwindlabs/tailwindcss/assets/29762210/4e1dce44-8c44-4603-92ff-4ae9297c3693" alt="transforming (1) index.html(node:43860) Warning: To load an ES module, set 'type': 'module' in the package.json or use the .mjs extension. (Use `node --trace-warnings ...` to show where the warning was created) ">

If I change the filename to `tailwind.config.mts`, it no longer finds the configuration.

<img width="380" alt="image" src="https://github.com/tailwindlabs/tailwindcss/assets/29762210/15237c1b-4d84-4139-846b-01ace5e8b9ed" alt="transforming (1) index.html warn - The `content` option in your Tailwind CSS configuration is missing or empty.  warn - Configure your content sources or your generated CSS will be missing styles.  warn - https://tailwindcss.com/docs/content-configuration">

### Node 22 documentation

https://nodejs.org/api/esm.html

RobinMalfait:
Hey!


While we can add support for the `.mts` file extension, the `.ts` file extension should just work for your use case. The big difference is the authoring experience, which allows you to use types in `.ts`, `export default` in `.mjs` and `module.exports` in `.cjs`.

Vite doesn't have to transpile the `tailwind.config.ts` file at all, it's only used by Tailwind CSS itself to eventually generate some CSS. The generated CSS will be used by Vite later on to put it in the correct location.

Are you running into other issues while using the `.ts` file extension for `tailwind.config.ts`?

Hope this helps!

jscarle:
[...]

PR #13940, opened 2024-07-03
Add `.mts` and `.cts` config file detection

RobinMalfait:
This PR will add support for detecting `tailwind.config.cts` and `tailwind.config.mts` files by default.

This PR will also remove some warnings that Node v22 was showing in the terminal when using ESM files (or TS files). If you happen to use an ESM file (`.mjs`) or any TS file (`.ts`, `.cts` or `.mts`), then we will immediately use `jiti` instead of first trying to `require` the file. Requiring these files triggers Node in logging the warning.

Fixes: #13927


thecrypticace:
We should probably add tests for these proving .mts and .cts config files work. e.g. https://github.com/tailwindlabs/tailwindcss/blob/0de1f0cce404c2291e7673e60d7cf3d0347f168c/integrations/tailwindcss-cli/tests/integration.test.js#L137

thecrypticace (changes requested):
add integration tests plz

thecrypticace (approved):

PR #13977, opened 2024-07-10
Discard invalid `variants` and `utilities` with modifiers

RobinMalfait:
This PR discards candidates that contain modifiers for variants and utilities that don't need it.

For example, right now you can use `hover/foo:flex` or even `flex/foo`. These generate the exact same contents as `hover:flex` and `flex` respectively. However, the `foo` modifier isn't used but is still included in the class name, which means that this currently just generates additional unnecessary CSS.

For static utilities and static variants, we can bail early when we detect a modifier and we don't have to continue parsing the candidate. We also don't have to run the known invalid candidate through the system. (In theory if you have a lot of invalid candidates like this it will result in better performance and smaller CSS)

Some utilities and variants accept modifiers, e.g.: `bg-red-500/50` or `group-hover/foo`. In these cases we can't discard the `modifier` automatically, which is why the PR is relatively big because we have to discard on a case by case basis.

Requires:
- [x] #13970

### DIFF
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -317,12 +317,15 @@ export function createUtilities(theme: Theme) {
       let value: string | null = null
 
       if (!candidate.value) {
+        if (candidate.modifier) return
+
         // If the candidate has no value segment (like `rounded`), use the
         // `defaultValue` (for candidates like `grow` that have no theme values)
         // or a bare theme value (like `--radius` for `rounded`). No utility
         // will ever support both of these.
         value = desc.defaultValue ?? theme.resolve(null, desc.themeKeys ?? [])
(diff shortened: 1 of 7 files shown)


packages/tailwindcss/src/utilities.ts
```diff
@@ -310,12 +310,15 @@ export function createUtilities(theme: Theme) {
       let value: string | null = null
 
       if (!candidate.value) {
+        if (candidate.modifier) return
```
thecrypticace, line 313: I feel like this is wrong but I can't remember which "default" utility was/is supposed to take a modifier (this was a bug in v3 iirc).

But if we can't come up with any then maybe it's fine.

thecrypticace (approved):

Issue #14005, opened 2024-07-16
Slim template regression with v3.4.5


ledermann:
**What version of Tailwind CSS are you using?**

v3.4.5

**What build tool (or framework if it abstracts the build tool) are you using?**

postcss-cli 8.4.39 (but IMHO not relevant here)

**What version of Node.js are you using?**

v20.9.0

**What browser are you using?**

Not relevant

**What operating system are you using?**

macOS (but not relevant)

**Reproduction URL**

Because the bug belongs to Pug or slim, `play.tailwindcss.com` cannot be used to build an example. I will send a PR to fix, please don't close this issue.

**Describe your issue**

After upgrading my Rails application (using Slim templates) to Tailwind CSS [v3.4.5](https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.4.5), I found a regression that some classes are not generated anymore. Example:

```slim
.bg-green-300.2xl:bg-red-300
  '(Look mom, no closing tag!)
```

The class `2xl:bg-red-300` is missing. It seems this happens because #13959 uses an invalid number check.

Stay tuned, I'm working on a PR to fix this.

thecrypticace:
fix released in v3.4.6 -- thanks!

Skulli:
I have a similar problem (in haml) with the class `xl:col-span-2` in the line `.xl:col-span-2.xl:pr-8`. With 3.4.5 and 3.4.6 the `col-span-2` class is not extracted. Can open a new issue, if required.

ledermann:
@Skulli I just tried your example in the test suite of tailwindcss `v3.4.6` and it works fine. Both classes are extracted. So it seems to be a different thing, maybe your example is not complete. I suggest creating a new issue with a minimal failing example.

PR #14006, opened 2024-07-16
Fix extracting utilities from slim/pug templates

ledermann:
This pull request addresses an issue with extracting utility classes from slim/pug templates. Specifically, it corrects the handling of segments starting with a number to avoid the loss of relevant class names.

This fixes a regression (described in #14005) of yesterday's release `v3.4.5`.

It uses `Number()` instead of `parseInt()` to check whether a string is a number. This is better because:

```js
const segment = "2xl:bg-red-500"
parseInt(segment) => 2
Number(segment) => NaN
```

I've added a test to ensure my change works as expected.


thecrypticace (approved):

thecrypticace:
I'll get a release out today. Thanks!

PR #14037, opened 2024-07-23
Normalize attribute selector for `data-*` and `aria-*` modifiers

### LINKED ISSUE #14026: Data attribute modifier doesn't work with numeric values

3.4.6

Vite 5.3.3 (no Vue/React/Angular)

20.12.1

Any browser

MacOS

**Reproduction URL**

https://play.tailwindcss.com/urhLO2C0Zj

**Describe your issue**

When the `data-*` modifier is used and the `data` attribute has a numeric value, the class isn't applied.

philipp-spiess:
Fixes #14026 
See #14040 for the v4 fix

When translating `data-` and `aria-` modifiers with attribute selectors, we currently do not wrap the target attribute in quotes. This only works for keywords (purely alphabetic words) but breaks as soon as there are numbers or things like spaces in the attribute:

```html
<div data-id="foo" class="data-[id=foo]:underline">underlined</div>
<div data-id="f1" class="data-[id=1]:underline">not underlined</div>
<div data-id="foo bar" class="data-[id=foo_bar]:underline">not underlined</div>
```

Since it's fairly common to have attribute selectors with `data-` and `aria-` modifiers, this PR will now wrap the attribute in quotes unless these are already wrapped. 

| Tailwind Modifier  | CSS Selector |
| `.data-[id=foo]`  | `[data-id='foo']`  |
| `.data-[id='foo']`  | `[data-id='foo']`  |
| `.data-[id=foo_i]`  | `[data-id='foo i']`  |
| `.data-[id='foo'_i]`  | `[data-id='foo' i]`  |
| `.data-[id=123]`  | `[data-id='123']`  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Fix class detection in Slim templates with attached attributes and ID ([#14019](https://github.com/tailwindlabs/tailwindcss/pull/14019))
-- Attribute selectors in `data-` and `aria-` modifiers are now wrapped in quotation marks by default, allowing numbers and spaces in them ([#14037])(https://github.com/tailwindlabs/tailwindcss/pull/14037)
+- Attribute selectors in `data-*` and `aria-*` modifiers are now wrapped in quotation marks by default, allowing numbers and spaces in them ([#14037])(https://github.com/tailwindlabs/tailwindcss/pull/14037)
 
 ## [3.4.6] - 2024-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Fix class detection in Slim templates with attached attributes and ID ([#14019](https://github.com/tailwindlabs/tailwindcss/pull/14019))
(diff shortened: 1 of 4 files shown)


thecrypticace:
fyi this is technically breaking. Though in practice it may not actually matter. I kinda doubt people are using this feature but today this candidate causes the div to be underlined:
```
<div class="data-[id=foo_i]:underline" data-id="fOO">test</div>
```

However, after this PR it won't. I think it's probably okay to ship as-is but we'll want to at least make note of this in case we need to follow it up with a fix.

philipp-spiess:
@thecrypticace Hmm yes indeed. As you say, it's easy to catch this case and check for the valid flags if we want to. Let's for sure make a not though!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Fix class detection in Slim templates with attached attributes and ID ([#14019](https://github.com/tailwindlabs/tailwindcss/pull/14019))
+- Attribute selectors in `data-` and `aria-` modifiers are now wrapped in quotation marks by default, allowing numbers and spaces in them ([#14037])(https://github.com/tailwindlabs/tailwindcss/pull/14037)
```
RobinMalfait, line 11:
```suggestion
- Attribute selectors in `data-*` and `aria-*` modifiers are now wrapped in quotation marks by default, allowing numbers and spaces in them ([#14037])(https://github.com/tailwindlabs/tailwindcss/pull/14037)
```

RobinMalfait (approved):

PR #14074, opened 2024-07-29
Prefix internal modules with `internal-*`

philipp-spiess:
Last week we discussed bringing in some consistency for our non-public npm packages in the repo. We discussed using custom namespaces (e.g. `@tailwindcss-internal`) vs. simple prefixes but it does not matter too much if we are both consistent with our pattern and it's easy for us to see whether a plugin is public or not. 

Since we have a mixture of public namespaced (`@tailwindcss/*`) and non-namespaced (`tailwindcss`) packages, I think it would be best if we use a prefix to signal internal dependencies. This PR proposes we use `internal-*` as the prefix and renames `test-utils` to `internal-example-plugin` (since, really, this package is just an example for the Tailwind plugin system).


RobinMalfait (approved):
I like this one, but let's wait for somebody else to approve as well before merging 💪

In the future if we have a JS and a CSS plugin, then we can rename it to `internal-example-js-plugin` or something. But for now this is fine.

thecrypticace:
I think I'm okay with this. The only odd thing about this is it's really just a test fixture so not sure if the name should be different _because_ of that.

But can always be changed later. I think this is at least better than what is is currently.

thecrypticace (approved):

adamwathan (approved):

PR #14076, opened 2024-07-29
Add `@content` support to `@tailwindcss/vite`

RobinMalfait:
This PR makes sure that when you use `@content`, that these files are also being watched if you are using Vite.

In Vite we rely on the module graph instead, but sometimes it can be useful to watch files outside of the module graph. E.g.: A Laravel project, where you are watching `*.blade.php` files.

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -75,19 +75,27 @@ export default function tailwindcss(): Plugin[] {
   }
 
   function generateCss(css: string, inputPath: string, addWatchFile: (file: string) => void) {
-    let basePath = path.dirname(path.resolve(inputPath))
+    let inputBasePath = path.dirname(path.resolve(inputPath))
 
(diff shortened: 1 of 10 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -95,10 +95,18 @@ export default function tailwindcss(): Plugin[] {
...
+    // Watch individual files
     for (let file of result.files) {
       addWatchFile(file)
     }
 
+    // Watch globs
+    for (let glob of result.globs) {
+      if (glob.glob[0] === '!') continue
+
+      addWatchFile(`${glob.base}/${glob.glob}`)
```
philipp-spiess, line 107: Note from our call: We need to be careful about this on Windows since we want glob to always use posix style separators but base can use windows-style. We will need to fix windows path issues anyways and will take care of this as part of https://github.com/tailwindlabs/tailwindcss/pull/14065

philipp-spiess (approved):

PR #14081, opened 2024-07-29
Automate checksum generation for standalone CLI

thecrypticace:
We've done this manually until now which is _fine_ but not ideal so this PR automates it so it cannot be forgotten.

@RobinMalfait do you know of a good way to test this w/o creating a tag? Should I add a `workflow_dispatch` trigger to the prepare release workflow?

### DIFF
--- a/standalone-cli/scripts/checksum.mjs
+++ b/standalone-cli/scripts/checksum.mjs
@@ -1,0 +1,27 @@
+import { createHash } from 'node:crypto'
+import { readFile, writeFile } from 'node:fs/promises'
+import * as path from 'node:path'
+
+const files = [
+  './tailwindcss-linux-arm64',
+  './tailwindcss-linux-armv7',
+  './tailwindcss-linux-x64',
+  './tailwindcss-macos-arm64',
+  './tailwindcss-macos-x64',
+  './tailwindcss-windows-arm64.exe',
+  './tailwindcss-windows-x64.exe',
+]
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname)
+
+const lines = await Promise.all(
+  files.map(async (file) => {
+    let sum = createHash('sha256')
+      .update(await readFile(path.resolve(__dirname, '../dist', file)))
+      .digest('hex')
+
+    return `${sum}  ${file}`
+  })
+)
+
+await writeFile(path.resolve(__dirname, '../dist', 'sha256sums.txt'), lines.join('\n') + '\n')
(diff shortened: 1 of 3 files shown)


standalone-cli/scripts/checksum.mjs
```diff
@@ -0,0 +1,27 @@
...
+  './tailwindcss-linux-arm64',
+  './tailwindcss-linux-armv7',
+  './tailwindcss-linux-x64',
+  './tailwindcss-macos-arm64',
+  './tailwindcss-macos-x64',
+  './tailwindcss-windows-arm64.exe',
+  './tailwindcss-windows-x64.exe',
+]
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname)
```
philipp-spiess, line 15: Learned today that this will fail on Windows and you're supposed to use the `fileURLToPath` helper 🙈  https://github.com/tailwindlabs/tailwindcss/pull/14065/files#diff-8cdc2dcef376d2bbc6b8c9029a227c928f2aaa0f460776ba24d9d3eba110aab1

philipp-spiess (approved):
> @RobinMalfait do you know of a good way to test this w/o creating a tag? Should I add a workflow_dispatch trigger to the prepare release workflow?

Maybe it's fine to run `npm run build` locally to validate this. Otherwise the `workflow_dispatch` trigger makes sense IMO but if we want to run a full release we might need to make it a `--dry-run` _somewhere_.

The changes look good to me, though :)

standalone-cli/scripts/checksum.mjs
RobinMalfait: Luckily, we don't have to worry about this for this particular case since it will be executed on a linux machine

RobinMalfait:
Yeah I would go for a manual `workflow_dispatch` or even a temporary change in an action that is already running to verify that the script works.

It looks good to me, but my only fear is that the `import` and `node:` prefix might not work if we are running on an older node version when preparing releases 🤔

Another alternative is to try and setup act (https://github.com/nektos/act) but that might not be as straightforward because of the generated executables.

philipp-spiess:
> It looks good to me, but my only fear is that the import and node: prefix might not work if we are running on an older node version when preparing releases 🤔

Seems like we use Node 20 for the releases as we also do for the CI: https://github.com/tailwindlabs/tailwindcss/blob/next/.github/workflows/release.yml#L17

thecrypticace:
Wrong branch: https://github.com/tailwindlabs/tailwindcss/blob/master/.github/workflows/release.yml#L21

We're using node 14 so… we need to update that if we can. 😱

thecrypticace:
Gonna merge this and give it a test. I think I need to create a dummy tag first though b/c if I don't it'll overwrite the existing release (which would be bad)

PR #14085, opened 2024-07-30
Fix Windows tests for new postcss plugin

philipp-spiess:
This PR fixes the new `postcss-fix-relative-paths` plugin for Windows paths. The issue was that we mixed Windows-style path separators from the absolute file paths with the Posix-style separators from globs. This caused the `dirname` functions to resolve to the wrong files.

To solve this, we now make the difference very clear by calling the content a `glob`. For globs, we always expect Posix-style path separators and for the case of making a glob absolute (by prefixing the directory), we now convert them into Posix-style explicitly.

This PR also fixes an issue where negative rules (e.g. `!./**/*.ts`) were not properly rewritten.

### DIFF
--- a/packages/internal-postcss-fix-relative-paths/src/index.ts
+++ b/packages/internal-postcss-fix-relative-paths/src/index.ts
@@ -44,6 +44,8 @@ export default function fixRelativePathsPlugin(): Plugin {
     let rulePath = path.posix.join(path.dirname(inputFilePath), content)
     let relative = path.posix.relative(path.posix.dirname(rootPath), rulePath)
 
+    // TODO: If we fix paths like this, ensure we have tests that cover
+    // POSIX style absolute globs on windows in the rust codebase
     console.log({
       content,
       inputFilePath,
--- a/packages/internal-postcss-fix-relative-paths/src/index.ts
+++ b/packages/internal-postcss-fix-relative-paths/src/index.ts
@@ -41,13 +41,21 @@ export default function fixRelativePathsPlugin(): Plugin {
       return
     }
 
-    let rulePath = path.posix.join(path.posix.dirname(inputFilePath), content)
+    let rulePath = path.posix.join(path.dirname(inputFilePath), content)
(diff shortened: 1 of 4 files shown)


RobinMalfait (approved):
All the Windows paths make me nervous, but was on a call with @philipp-spiess so going to approve already.

I think it's good to get at least @thecrypticace eyes on it before merging.

packages/internal-postcss-fix-relative-paths/src/index.ts
```diff
@@ -9,12 +9,12 @@ export default function fixRelativePathsPlugin(): Plugin {
   let touched: WeakSet<AtRule> = new WeakSet()
 
   function fixRelativePath(atRule: AtRule) {
-    let rootPath = getRoot(atRule)?.source?.input.file
+    let rootPath = atRule.root().source?.input.file
```
thecrypticace, line 13: I simplified this. `Node` has a `root()` method that walks up the tree for you.

It also stops at Root instead of a Document (an optional parent node of Root)

packages/internal-postcss-fix-relative-paths/src/index.ts
```diff
@@ -9,12 +9,12 @@ export default function fixRelativePathsPlugin(): Plugin {
...
     if (!rootPath) {
       return
     }
 
-    let inputFilePath = atRule?.source?.input.file
+    let inputFilePath = atRule.source?.input.file
```
thecrypticace, line 18: `atRule` isn't nullable so one of the `?` could be removed

packages/internal-postcss-fix-relative-paths/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import path from 'node:path'
 import type { AtRule, Plugin } from 'postcss'
+import { normalizePath } from './normalize-path'
```
thecrypticace, line 3: I inlined a corrected version of the `normalize-path` NPM package (specifically regarding network shares) and used that method to cleanup the code slightly.

thecrypticace (approved):

PR #14089, opened 2024-07-31
Add integration test setup and tests for the Vite integration

philipp-spiess:
This PR adds a new root `/integrations` folder that will be the home of integration tests. The idea of these tests is to use Tailwind in various setups just like our users would (by only using the publishable npm builds). 

To avoid issues with concurrent tests making changes to the file system, to make it very easy to test through a range of versions, and to avoid changing configuration objects over and over in test runs, we decided to inline the scaffolding completely into the test file and have no examples checked into the repo. 

Here's an example of how this can look like for a simple Vite test:

```ts
test('works with production builds', {
    fs: {
      'package.json': json`
        {
          "type": "module",
          "dependencies": {
            "@tailwindcss/vite": "workspace:^",
            "tailwindcss": "workspace:^"
          },
          "devDependencies": {
            "vite": "^5.3.5"
          }
        }
      `,
      'vite.config.ts': ts`
        import tailwindcss from '@tailwindcss/vite'
        import { defineConfig } from 'vite'

        export default defineConfig({
          build: { cssMinify: false },
          plugins: [tailwindcss()],
        })
      `,
      'index.html': html`
        <head>
          <link rel="stylesheet" href="./src/index.css">
        </head>
        <body>
          <div class="underline m-2">Hello, world!</div>
        </body>
      `,
      'src/index.css': css`
        @import 'tailwindcss/theme' reference;
      `,
    },
  },
  async ({ fs, exec }) => {
    await exec('pnpm vite build')

    expect.assertions(2)
    for (let [path, content] of await fs.glob('dist/**/*.css')) {
      expect(path).toMatch(/\.css$/)
      expect(stripTailwindComment(content)).toMatchInlineSnapshot(
        `
        ".m-2 {
          margin: var(--spacing-2, .5rem);
        }

        .underline {
        }"
      `,
      )
    }
  },
)
```

By defining all dependencies this way, we never have to worry about which fixtures are checked in and can more easily describe changes to the setup. 

For ergonomics, we've also added the [`embed` prettier plugin](https://github.com/Sec-ant/prettier-plugin-embed). This will mean that files inlined in the `fs` setup are properly indented. No extra work needed! 

If you're using VS Code, I can also recommend the [Language Literals](https://marketplace.visualstudio.com/items?itemName=sissel.language-literals) extension so that syntax highlighting also _just works_.

A neat feature of inlining the scaffolding like this is to make it very simple to test through a variety of versions. For example, here's how we can set up a test against Vite 5 and Vite 4:

```js
;['^4.5.3', '^5.3.5'].forEach(viteVersion => {
    test(`works with production builds for Vite ${viteVersion}`, {
      fs: {
        'package.json': json`
          {
            "type": "module",
            "devDependencies": {
              "vite": "${viteVersion}"
            }
          }
        `,
    async () => {
      // Do something
    },
  )
})
```

## Philosophy

Before we dive into the specifics, I want to clearly state the design considerations we have chosen for this new test suite:

- All file mutations should be done in temp folders, nothing should ever mess with your working directory
- Windows as a first-class citizen
- Have a clean and simple API that describes the test setup only using public APIs
- Focus on reliability (make sure cleanup scripts work and are tolerant to various error scenarios)
- If a user reports an issue with a specific configuration, we want to be able to reproduce them with integration tests, no matter how obscure the setup (this means the test need to be in control of most of the variables)
- Tests should be reasonably fast (obviously this depends on the integration. If we use a slow build tool, we can't magically speed it up, but our overhead should be minimal).

## How it works

The current implementation provides a custom `test` helper function that, when used, sets up the environment according to the configuration. It'll create a new temporary directory and create all files, ensuring things like proper `\r\n` line endings on Windows. 

We do have to patch the `package.json` specifically, since we can not use public versions of the tailwindcss packages as we want to be able to test against a development build. To make this happen, every `pnpm build` run now creates tarballs of the npm modules (that contain only the files that would also in the published build). We then patch the `package.json` to rewrite `workspace:^` versions to link to those tarballs. We found this to work reliably on Windows and macOS as well as being fast enough to not cause any issues. Furthermore we also decided to use `pnpm` as the version manager for integration tests because of it's global module cache (so installing `vite` is fast as soon as you installed it once).

The test function will receive a few utilities that it can use to more easily interact with the temp dir. One example is a `fs.glob` function that you can use to easily find files in eventual `dist/` directories or helpers around `spawn` and `exec` that make sure that processes are cleaned up correctly.

Because we use tarballs from our build dependencies, working on changes requires a workflow where you run `pnpm build` before running `pnpm test:integrations`. However it also means we can run clients like our CLI client with no additional overhead—just install the dependency like any user would and set up your test cases this way.

## Test plan

This PR also includes two Vite specific integration tests: One testing a static build (`pnpm vite build`) and one a dev mode build (`pnpm vite dev`) that also makes changes to the file system and asserts that the resources properly update.

### DIFF
--- a/integrations/utils.ts
+++ b/integrations/utils.ts
@@ -6,6 +6,7 @@ import fs from 'node:fs/promises'
 import net from 'node:net'
 import { homedir, platform, tmpdir } from 'node:os'
 import path from 'node:path'
+import tcpPortUsed from 'tcp-port-used'
 import { test as defaultTest } from 'vitest'
 
 export let css = dedent
@@ -13,17 +14,24 @@ export let html = dedent
 export let ts = dedent
 export let json = dedent
 
+interface SpawnedProcess {
+  dispose: () => void
+  onStdout: (predicate: (message: string) => boolean) => void
+  onStderr: (predicate: (message: string) => boolean) => void
+}
+
 interface TestConfig {
   fs: {
     [filePath: string]: string
   }
 }
 interface TestContext {
-  exec: (command: string) => Promise<string>
-  spawn: (command: string) => { dispose: () => void }
+  exec(command: string): Promise<string>
+  spawn(command: string): Promise<SpawnedProcess>
   getFreePort(): Promise<number>
   fs: {
-    glob: (pattern: string) => Promise<[string, string][]>
+    write(filePath: string, content: string): Promise<void>
+    glob(pattern: string): Promise<[string, string][]>
   }
 }
 type TestCallback = (context: TestContext) => Promise<void> | void
@@ -48,7 +56,7 @@ export function test(
       ),
     )
 
-    for (let [filename, content] of Object.entries(config.fs)) {
+    async function write(filename, content): Promise<void> {
       let full = path.join(root, filename)
 
       if (filename.endsWith('package.json')) {
@@ -63,6 +71,10 @@ export function test(
       let dir = path.dirname(full)
       await fs.mkdir(dir, { recursive: true })
       await fs.writeFile(full, content)
+    }
+
+    for (let [filename, content] of Object.entries(config.fs)) {
+      await write(filename, content)
     }
 
     try {
@@ -85,7 +97,7 @@ export function test(
       async exec(command: string) {
         return execSync(command, { cwd: root }).toString()
       },
-      spawn(command: string) {
+      async spawn(command: string) {
         let resolveDisposal: (() => void) | undefined
         let rejectDisposal: ((error: Error) => void) | undefined
         let disposePromise = new Promise<void>((resolve, reject) => {
@@ -113,7 +125,6 @@ export function test(
         }
         disposables.push(dispose)
         function onExit() {
-          console.log(`spawned proccess (${command}) exited`)
           resolveDisposal?.()
         }
 
@@ -185,7 +196,18 @@ export function test(
               if (port === null) {
                 reject(new Error(`Failed to get a free port: address is ${address}`))
               } else {
-                disposables.push(() => killPort(port))
+                disposables.push(async () => {
+                  // kill-port uses `lsof` on macOS which is expensive and can
+                  // block for multiple seconds. In order to avoid that for a
+                  // server that is no longer running, we check if the port is
+                  // still in use first.
+                  const isRunning = await tcpPortUsed.check(port, '127.0.0.1')
+                  if (!isRunning) {
+                    return
+                  }
+
+                  await killPort(port)
+                })
                 resolve(port)
               }
             })
@@ -193,6 +215,7 @@ export function test(
         })
       },
       fs: {
+        write,
         async glob(pattern: string) {
           let files = await fastGlob(pattern, { cwd: root })
           return Promise.all(
--- a/integrations/utils.ts
+++ b/integrations/utils.ts
@@ -28,6 +28,8 @@ type TestCallback = (context: TestContext) => Promise<void> | void
 
 const REPO_ROOT = path.join(__dirname, '..')
 
+type SpawnActor = { predicate: (message: string) => boolean; resolve: () => void }
+
 export function test(
   name: string,
   config: TestConfig,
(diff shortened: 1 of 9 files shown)


integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+                disposables.push(async () => {
+                  // Wait for 10ms in case the process was just killed
+                  await new Promise((resolve) => setTimeout(resolve, 10))
+
+                  // kill-port uses `lsof` on macOS which is expensive and can
+                  // block for multiple seconds. In order to avoid that for a
+                  // server that is no longer running, we check if the port is
+                  // still in use first.
+                  let isPortTaken = await testIfPortTaken(port)
+                  console.log({ isPortTaken })
```
thecrypticace, line 206: can be removed

integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+export let html = dedent
+export let ts = dedent
+export let json = dedent
+
+const REPO_ROOT = path.join(__dirname, '..')
+
+interface SpawnedProcess {
+  dispose: () => void
+  onStdout: (predicate: (message: string) => boolean) => void
+  onStderr: (predicate: (message: string) => boolean) => void
```
RobinMalfait, line 21:
```suggestion
  onStdout: (predicate: (message: string) => boolean) => Promise<void>
  onStderr: (predicate: (message: string) => boolean) => Promise<void>
```

integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+    let root = await fs.mkdtemp(
+      // On Windows CI, tmpdir returns a path containing a weird RUNNER~1 folder
+      // that apparently causes the vite builds to not work.
+      path.join(
+        process.env.CI && platform() === 'win32' ? homedir() : tmpdir(),
+        'tailwind-integrations',
+      ),
+    )
+
+    async function write(filename, content): Promise<void> {
```
RobinMalfait, line 58:
```suggestion
    async function write(filename: string, content: string): Promise<void> {
```

integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+          env: {
+            ...process.env,
+          },
+        })
+
+        function dispose() {
+          child.kill()
+
+          let timer = setTimeout(
+            () => rejectDisposal?.(new Error(`spawned proccess (${command}) did not exit in time`)),
```
RobinMalfait, line 118:
```suggestion
            () => rejectDisposal?.(new Error(`spawned process (${command}) did not exit in time`)),
```

integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+// Maps package names to their tarball filenames. See scripts/pack-packages.ts
+// for more details.
+function pkgToFilename(name: string) {
+  return `${name.replace('@', '').replace('/', '-')}.tgz`
+}
+
+function overwriteVersionsInPackageJson(content: string): string {
+  let json = JSON.parse(content)
+
+  // Resolve all worksplace:^ versions to local tarballs
```
RobinMalfait, line 249:
```suggestion
  // Resolve all workspace:^ versions to local tarballs
```

integrations/utils.ts
```diff
@@ -0,0 +1,296 @@
...
+  ;['dependencies', 'devDependencies', 'peerDependencies'].forEach((key) => {
+    let desp = json[key] || {}
+    for (let dependecy in desp) {
+      if (desp[dependecy] === 'workspace:^') {
+        desp[dependecy] = resolveVersion(dependecy)
+      }
+    }
```
RobinMalfait, line 256:
```suggestion
    let dependencies = json[key] || {}
    for (let dependency in dependencies) {
      if (dependencies[dependency] === 'workspace:^') {
        dependencies[dependency] = resolveVersion(dependency)
      }
    }
```

thecrypticace (approved):

PR #14110, opened 2024-08-02
Ensure `require` is possible in `esm` file

RobinMalfait:
When using the `@plugin "…";` in CSS, then the `@tailwindcss/cli` will currently crash with `Dynamic require` error:

```
Error: Dynamic require of "./plugin.js" is not supported
    at file:///private/var/folders/…/T/tailwind-integrations…/node_modules/@tailwindcss/cli/dist/index.mjs:2:195
    at loadPlugin (file:///private/var/folders/…/T/tailwind-integrations…/node_modules/@tailwindcss/cli/dist/index.mjs:6:439)
    at file:///private/var/folders/…/T/tailwind-integrations…/node_modules/tailwindcss/dist/lib.mjs:8:4328
    at S (file:///private/var/folders/…/T/tailwind-integrations…/node_modules/tailwindcss/dist/lib.mjs:1:445)
    at Module._r (file:///private/var/folders/…/T/tailwind-integrations…/node_modules/tailwindcss/dist/lib.mjs:8:4222)
    at p (file:///private/var/folders/…/T/tailwind-integrations…/node_modules/@tailwindcss/cli/dist/index.mjs:6:414)
    at le (file:///private/var/folders/…/T/tailwind-integrations…/node_modules/@tailwindcss/cli/dist/index.mjs:6:452)
```

To solve this, we will make sure that we create a `require` based on the `import.meta.url`.

### DIFF
--- a/packages/@tailwindcss-cli/src/commands/build/index.ts
+++ b/packages/@tailwindcss-cli/src/commands/build/index.ts
@@ -3,6 +3,7 @@ import { IO, Parsing, scanDir, scanFiles, type ChangedContent } from '@tailwindc
 import { Features, transform } from 'lightningcss'
 import { existsSync } from 'node:fs'
 import fs from 'node:fs/promises'
+import { createRequire } from 'node:module'
 import path from 'node:path'
 import postcss from 'postcss'
 import atImport from 'postcss-import'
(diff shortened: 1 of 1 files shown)


philipp-spiess:
From what we discussed before we headed to lunch: We probably want to do this not only for the cli client but all other clients as well as they do also build `mjs` variants.

thecrypticace:
We really should use dynamic imports instead.

philipp-spiess:
@thecrypticace Yeah that's probably even better, just need to make the loadPlugin API async for this first 👍

philipp-spiess:
Here's the async version 😬 https://github.com/tailwindlabs/tailwindcss/pull/14132

PR #14116, opened 2024-08-03
Only read from stdin when using `—input -`

thecrypticace:
Using `process.stdin.resume()` causes stdin to be consumed and buffered in Node immediately — even when using a file as input. It’d be better for our process to not read data from stdin at all in this case.


RobinMalfait:
Going to close this PR for now. If people are running into this on the v3 codebase we could re-introduce this, but until somebody runs into problems let's not worry about it right now.

PR #14132, opened 2024-08-06
Use `import` to load plugins

philipp-spiess:
Alternative to #14110

This PR changes the way how we load plugins to be compatible with ES6 async `import`s. This allows us to load plugins even inside the browser but it comes at a downside: We now have to change the `compile` API to return a `Promise`...

So most of this PR is rewriting all of the call sites of `compile` to expect a promise instead of the object.


adamwathan (approved):
Thank you for your sacrifice 😄 Looks good!

thecrypticace (approved):

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -73,26 +74,30 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
             hasTailwind = true
           }
         },
-        OnceExit(root, { result }) {
+        async OnceExit(root, { result }) {
           let inputFile = result.opts.from ?? ''
+          console.log({ inputFile })
           let context = cache.get(inputFile)
           let inputBasePath = path.dirname(path.resolve(inputFile))
+          console.log({ inputBasePath })
```
RobinMalfait, line 82: We merged it with these `console.log` calls

thecrypticace: removed them

PR #14133, opened 2024-08-06
Add `test.debug` to integration tests

philipp-spiess:
While working on #14078, there were a couple of debugging  techniques that we were using quite frequently:

- Being able to `cd` into the test setup
- Seeing the stdio and stdout data in real-time (this currently requires us to mark a test as failing)
- Checking the exact commands that are being run

Since we naturally worked around this quite often, I decided to make this a first-level API with the introduction of a new `test.debug` flag. When set, it will:

- Create the test setup in the project dir within a new `.debug` folder and won't delete it after the run. Having it in an explicit folder allows us to easily delete it manually when we need to.
- Logs all run commands to the console (`>` for a sync call, `>&` for a spawned process)
- Logs stdio and stderr to the console in real time.
- Run the test as `.only`

Screenshot 2024-08-06 at 13 19 49

### DIFF
--- a/integrations/utils.ts
+++ b/integrations/utils.ts
@@ -102,9 +102,15 @@ export function test(
             rejectDisposal = reject
           })
 
+          let cwd = childProcessOptions.cwd ?? root
+          if (debug && cwd !== root) {
+            let relative = path.relative(root, cwd)
+            if (relative[0] !== '.') relative = `./${relative}`
+            console.log(`> cd ${relative}`)
+          }
           if (debug) console.log(`>& ${command}`)
           let child = spawn(command, {
-            cwd: root,
+            cwd,
             shell: true,
             env: {
               ...process.env,
--- a/integrations/utils.ts
+++ b/integrations/utils.ts
@@ -41,6 +41,10 @@ interface TestContext {
   }
 }
 type TestCallback = (context: TestContext) => Promise<void> | void
+interface TestFlags {
+  only?: boolean
+  debug?: boolean
+}
 
 type SpawnActor = { predicate: (message: string) => boolean; resolve: () => void }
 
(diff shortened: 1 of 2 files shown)


integrations/utils.ts
```diff
@@ -51,68 +55,33 @@ export function test(
...
+
+      if (debug) {
+        console.log('Running test in debug mode. File system will be written to:')
+        console.log(root)
+        console.log()
       }
 
-      options.onTestFinished(dispose)
-
       let context = {
```
philipp-spiess, line 81: This was moved up so that `pnpm install` can use the same `exec` function that we also export

integrations/utils.ts
```diff
@@ -177,12 +147,14 @@ export function test(
 
           child.stdout.on('data', (result) => {
             let content = result.toString()
+            if (debug) console.log(content)
```
thecrypticace, line 150: Should this have like a `[stdout] ` prefix or something? Just so it's more easily distinguishable in the test output?

integrations/utils.ts
```diff
@@ -177,12 +147,14 @@ export function test(
...
             combined.push(['stdout', content])
             stdoutMessages.push(content)
             notifyNext(stdoutActors, stdoutMessages)
           })
           child.stderr.on('data', (result) => {
             let content = result.toString()
+            if (debug) console.error(content)
```
thecrypticace, line 157: Ditto but with `[stderr] `

integrations/utils.ts
thecrypticace: thinking out loud: should probably prefix every line of the output

philipp-spiess: Hmm, maybe? It could help with the `onStderr` helper we have but I wonder why we need to be so diligent about this in the first place. In normal terminal apps you don't see the difference between `stdout` and `stderr` in the console either unless you pipe line stream to another location. I decided to replicate this here by using `console.log` for `stdout` and `console.error` for `stderr`

integrations/utils.ts
```diff
@@ -51,68 +55,33 @@ export function test(
...
       }
 
-      options.onTestFinished(dispose)
-
       let context = {
         root,
         async exec(command: string, childProcessOptions: ChildProcessOptions = {}) {
+          if (debug) console.log(`> ${command}`)
           return execSync(command, {
             cwd: root,
```
RobinMalfait, line 86: While testing I ran into an issue that I was executing the commands in the wrong folder. If we pass in a custom `cwd`, I think we should show that as a step as well:
```suggestion
          let cwd = childProcessOptions.cwd ?? root
          if (debug && cwd !== root) {
            let relative = path.relative(root, cwd)
            if (relative[0] !== '.') relative = `./${relative}`
            console.log(`> cd ${relative}`)
          }
          if (debug) console.log(`> ${command}`)
          return execSync(command, {
            cwd,
```

This results in:
image

integrations/utils.ts
RobinMalfait: I personally don't think the `[stdout]` or `[stderr]` prefix helps here because Vitest already shows that for us:

image

integrations/utils.ts
```diff
@@ -127,6 +96,7 @@ export function test(
             rejectDisposal = reject
           })
 
+          if (debug) console.log(`>& ${command}`)
```
RobinMalfait, line 99: We could do the same here as above, but this now results in the same command log, but just with a different prefix:
image


I would maybe just drop this? 🤔 

```suggestion
```

philipp-spiess: @RobinMalfait I’m confused, how did you get to the above log? Were you doing both `exec()` and `spawn()`? (and if that is the case, would you not expect both to log? 🤔)

RobinMalfait:
I think this PR should be merged in `next` instead of the `@source` support PR.

philipp-spiess:
@RobinMalfait yeah I need to rebase it I just wanted it to be based on the latest changes :)

RobinMalfait (approved):

PR #14135, opened 2024-08-06
Apply general feedback for `@sources`

RobinMalfait:
Part of #14078

This PR applies general feedback to the `@sources` umbrella PR.

This PR includes the following changes:
- Used real classes instead of class-like objects.
- Only handle PostCSS imports when `@import` is used.
- Improve some comments
- Improve naming of `GlobEntry` (`glob.glob` -> `glob.pattern`)
- Use `scanFiles` on the `scanDirResult` as-if it's an incremental scan (under
  the hood we still do a full scan right now).
- Renamed `content_paths` to `sources`

This is a separate PR just so that we can merge the umbrella PR un-squashed for each sub-PR. This way all the commits here won't show up in the `next` branch as-is.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -24,6 +24,12 @@ impl From<ChangedContent> for tailwindcss_oxide::ChangedContent {
 #[derive(Debug, Clone)]
 #[napi]
 pub struct ScanResult {
+  // Private information necessary for incremental rebuilds. Note: these fields are not exposed
+  // to JS
+  base: String,
+  sources: Vec<GlobEntry>,
+
+  // Public API:
   pub globs: Vec<GlobEntry>,
   pub files: Vec<String>,
   pub candidates: Vec<String>,
@@ -32,11 +38,13 @@ pub struct ScanResult {
 #[napi]
 impl ScanResult {
   #[napi]
-  pub fn scan_files(&self, input: Vec<ChangedContent>) -> Vec<String> {
-    tailwindcss_oxide::scan_files_with_globs(
-      input.into_iter().map(Into::into).collect(),
-      self.globs.clone().into_iter().map(Into::into).collect(),
-    )
+  pub fn scan_files(&self, _input: Vec<ChangedContent>) -> Vec<String> {
+    let result = tailwindcss_oxide::scan_dir(tailwindcss_oxide::ScanOptions {
+      base: self.base.clone(),
+      sources: self.sources.clone().into_iter().map(Into::into).collect(),
+    });
+
+    result.candidates
   }
 }
 
@@ -82,9 +90,10 @@ pub fn clear_cache() {
 #[napi]
 pub fn scan_dir(args: ScanOptions) -> ScanResult {
   let result = tailwindcss_oxide::scan_dir(tailwindcss_oxide::ScanOptions {
-    base: args.base,
+    base: args.base.clone(),
     sources: args
       .sources
+      .clone()
       .unwrap_or_default()
       .into_iter()
       .map(Into::into)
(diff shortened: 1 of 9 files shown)


crates/node/src/lib.rs
```diff
@@ -32,35 +38,37 @@ pub struct ScanResult {
...
-      self.globs.clone().into_iter().map(Into::into).collect(),
-    )
+  pub fn scan_files(&self, _input: Vec<ChangedContent>) -> Vec<String> {
+    let result = tailwindcss_oxide::scan_dir(tailwindcss_oxide::ScanOptions {
+      base: self.base.clone(),
+      sources: self.sources.clone().into_iter().map(Into::into).collect(),
+    });
+
+    result.candidates
   }
```
RobinMalfait, line 48: Public API uses `scanDirResult.scanFiles()` for incremental rebuilds. Right now we will do a full scan behind the scenes.

Good idea, @philipp-spiess!

crates/node/src/lib.rs
```diff
@@ -24,6 +24,12 @@ impl From<ChangedContent> for tailwindcss_oxide::ChangedContent {
 #[derive(Debug, Clone)]
 #[napi]
 pub struct ScanResult {
+  // Private information necessary for incremental rebuilds. Note: these fields are not exposed
+  // to JS
+  base: String,
+  sources: Vec<GlobEntry>,
```
RobinMalfait, line 30: Mind = blown

Didn't know this was possible, but it keeps the information private. It's not exposed to node land.

crates/node/src/lib.rs
```diff
@@ -32,35 +38,37 @@ pub struct ScanResult {
...
+    result.candidates
   }
 }
 
 #[derive(Debug, Clone)]
 #[napi(object)]
 pub struct GlobEntry {
   pub base: String,
-  pub glob: String,
+  pub pattern: String,
```
RobinMalfait, line 55: We had to do some renames already, so made this changes as well.

crates/oxide/src/lib.rs
```diff
@@ -441,26 +443,6 @@ pub fn scan_files(input: Vec<ChangedContent>, options: u8) -> Vec<String> {
     }
 }
 
-#[tracing::instrument(skip(input, globs))]
-pub fn scan_files_with_globs(input: Vec<ChangedContent>, globs: Vec<GlobEntry>) -> Vec<String> {
```
RobinMalfait, line 445: Not used right now, we can add it back when we need it.

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -228,23 +228,13 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
...
-
-            // Re-scan the directory to get the new `candidates`
-            scanDirResult = scanDir({
-              base, // Root directory, mainly used for auto content detection
-              contentPaths: compiler.globs.map((glob) => ({
-                base: inputBasePath, // Globs are relative to the input.css file
-                glob,
-              })),
-            })
+            let candidates = scanDirResult.scanFiles(changedFiles)
```
philipp-spiess, line 231: Nice! This properly feels like "something we need to fix on the rust side" now :)

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -87,28 +87,28 @@ export default function tailwindcss(): Plugin[] {
...
     })
 
-    let result = scanDir({
+    scanDirResult = scanDir({
       // TODO: This might not be necessary if we enable/disabled auto content
-      // detection
-      base: inputBasePath, // Root directory, mainly used for auto content detection
-      contentPaths: globs.map((glob) => ({
+      // detection.
+      base: config!.root, // Root directory, mainly used for auto content detection
```
philipp-spiess, line 93: Maybe we can for now just disable auto-content when base is set to an empty string? We only need that API for this caller anyways.

philipp-spiess (approved):
Happy Patton Oswalt

packages/@tailwindcss-vite/src/index.ts
RobinMalfait: Good idea, instead of an empty string I made it an `Option<…>` in Rust and just removed it from here.

This will now disable auto content detection.

thecrypticace (approved):

PR #14150, opened 2024-08-08
Unify implementations of compile and `__unstable__loadDesignSystem`

thecrypticace:
Right now there's some minor duplication and a ton of missing stuff in `__unstable__loadDesignSystem` — this PR makes sure the functionality is unified between it and `compile()`

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -35,24 +35,8 @@ function parseThemeOptions(selector: string) {
   return { isReference, isInline }
 }
 
-export async function compile(
-  css: string,
-  { loadPlugin = throwOnPlugin }: CompileOptions = {},
-): Promise<{
-  globs: string[]
-  build(candidates: string[]): string
-}> {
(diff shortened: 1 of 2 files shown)


packages/tailwindcss/src/index.ts
```diff
@@ -48,25 +48,9 @@ function parseThemeOptions(selector: string) {
...
 }
 
-export function compile(
-  css: string,
-  { loadPlugin = throwOnPlugin }: CompileOptions = {},
-): {
-  globs: string[]
-  build(candidates: string[]): string
-} {
+function parseCss(css: string, { loadPlugin = throwOnPlugin }: CompileOptions = {}) {
```
thecrypticace, line 51: This is a bad name. Suggestions plz?

thecrypticace:
I suspect this whole design will change completely once we rethink the exported APIs and what not but it seems reasonable to go ahead and unify them for now.

RobinMalfait (approved):
The `parseCss` name is not perfect, but it is internal API. We already have an internal card where we will tackle the public API and probably come up with something completely different.

The reason we still want to merge this one already is so that we can unblock some work in intellisense where we need this information.

PR #14160, opened 2024-08-09
Bump dependencies

RobinMalfait:
This PR bumps dependencies

We also make some dependencies `catalog:` dependencies, which allows us to keep
the version in sync. E.g.: `lightningcss` and `@types/node`.

Bumped `turbo` to the latest version + enabled the new UI

Fixed a bug in the tests now that `lightningcss` outputs the correct value.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "pipeline": {
+  "tasks": {
     "@tailwindcss/oxide#build": {
       "dependsOn": ["^build"],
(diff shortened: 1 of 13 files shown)


turbo.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "pipeline": {
+  "ui": "tui",
```
philipp-spiess, line 3: this will require some getting used to hah

philipp-spiess (approved):

PR #14163, opened 2024-08-09
Add Next.js integration test

philipp-spiess:
Using the [new integration test setup](https://github.com/tailwindlabs/tailwindcss/pull/14089), this PR adds a test for a V4 Next.js setup using the Postcss plugin. It's testing both a full build and the dev mode (non-turbo for now).

Because of webpack, tests are quite slow which is worrisome since we probably need to add many more integrations in the future. One idea I have is that we separate tests in two buckets: _essential_ tests that run often and are fast and advanced suites that we only run on CI via custom, non-blocking, jobs.

### DIFF
--- a/integrations/postcss/next.test.ts
+++ b/integrations/postcss/next.test.ts
@@ -1,5 +1,5 @@
 import { expect } from 'vitest'
-import { candidate, css, fetchStylesFromIndex, js, json, test } from '../utils'
+import { candidate, css, fetchStylesFromIndex, js, json, retryAssertion, test } from '../utils'
 
 test(
   'production build',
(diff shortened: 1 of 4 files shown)


integrations/postcss/next.test.ts
```diff
@@ -0,0 +1,167 @@
...
+test(
+  'production build',
+  {
+    fs: {
+      'package.json': json`
+        {
+          "dependencies": {
+            "react": "^18",
+            "react-dom": "^18",
+            "next": "14.2.5"
```
RobinMalfait, line 13: Do we want to pin this version?

RobinMalfait (approved):

integrations/postcss/next.test.ts
philipp-spiess: Not really, we can do a major pin like we do with react :)

integrations/postcss/next.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    expect(stylesheets).toHaveLength(1)
+    ;[, css] = stylesheets[0]
+
+    expect(css).toContain(candidate`underline`)
+    expect(css).toContain(candidate`text-red-500`)
+  },
+)
+
+test(
+  'dev mode (turbo)',
```
philipp-spiess, line 161: @RobinMalfait added a turbo dev mode test as well

integrations/vite/index.test.ts
```diff
@@ -140,7 +125,10 @@ test(
 
     await process.onStdout((message) => message.includes('ready in'))
 
-    let css = await fetchCSS('/src/index.css', port)
+    let stylesheets = await fetchStylesheetsFromIndex(port)
+    expect(stylesheets).toHaveLength(1)
+    let [, css] = stylesheets[0]
```
philipp-spiess, line 130: Not super ergonomic. I am wondering if i should just concatenate all css files into one string instead 🤔

RobinMalfait (approved):
Looks good with the turbo bundler as well.
CI failed (I retried already and now it's working) but the stability of CI is not 100% solved yet (https://github.com/tailwindlabs/tailwindcss/actions/runs/10367654334/job/28699575402).

Should not be part of this PR though.

philipp-spiess:
So the tests here are stable now (re-run without issues 4 times in a row). I reverted all changes that were supposed to speed up the test on Windows (shenanigans like keeping `node_module` alive between tests etc.) but since Windows seems to keep some file handles open, all of these were erroneous in some way or another.

We will eventually speed up these tests by creating a ReFS disk for the Windows test.

RobinMalfait (approved):

RobinMalfait (approved):

PR #14165, opened 2024-08-09
Retry UI test assertions

philipp-spiess:
We noticed that some of the tests depending on the `hover:` variant were flaky. After some investigation, we found that injected elements had the `:hover` state without us explicitly hovering over the element.

To avoid this, we now set up an explicit placeholder element to move the mouse to before running the tests.

### DIFF
--- a/packages/tailwindcss/tests/ui.spec.ts
+++ b/packages/tailwindcss/tests/ui.spec.ts
@@ -305,12 +305,22 @@ async function render(page: Page, content: string) {
     }
   `)
 
+  // We noticed that some of the tests depending on the `hover:` variant were
+  // flaky. After some investigation, we found that injected elements had the
+  // `:hover` state without us explicitly hovering over the element.
+  //
+  // To avoid this, we now set up an explicit placeholder element to move the
+  // mouse to before running the tests.
+  content = `<div id="mouse-park" class="size-12"></div>${content}`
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hm, it looks like the flaky part is not the computed style collection (good, this would be confusing) but the `.hover()` state modification. I moved all of that into the retry block as well now, lets see 🤞

philipp-spiess:
Screenshot 2024-08-12 at 12 24 26

🥳

philipp-spiess:
No retries for 3 attempts in a row! Yay!

RobinMalfait (approved):

packages/tailwindcss/tests/ui.spec.ts
```diff
@@ -305,13 +305,23 @@ async function render(page: Page, content: string) {
     }
   `)
 
+  // We noticed that some of the tests depending on the `hover:` variant were
+  // flaky. After some investigation, we found that injected elements had the
+  // `:hover` state without us explicitly hovering over the element.
+  //
+  // To avoid this, we now set up an explicit placeholder element to move the
+  // mouse to before running the tests.
+  content = `<div id="mouse-park" class="size-12"></div>${content}`
```
thecrypticace, line 314: 😱

PR #14173, opened 2024-08-12
Add support for the `tailwindcss/plugin` export

thecrypticace:
This PR adds support for the `tailwindcss/plugin` import which has historically been used to define custom plugins:

```js
import plugin from "tailwindcss/plugin";

export default plugin(function ({ addBase }) {
  addBase({
    // ...
  });
});
```

This also adds support for `plugin.withOptions` which was used to define plugins that took optional initilization options when they were registered in your `tailwind.config.js` file:

```js
import plugin from "tailwindcss/plugin";

export default plugin.withOptions((options = {}) => {
  return function ({ addBase }) {
    addBase({
      // ...
    });
  };
});
```

We've stubbed out support for the `config` argument but we're not actually doing anything with it at the time of this PR. The scope of this PR is just to allow people to create plugins that currently work using the raw function syntax but using the `plugin` and `plugin.withOptions` APIs. Support for `config` will land separately.


packages/tailwindcss/src/plugin.cts
```diff
@@ -0,0 +1,5 @@
+// This file exists so that `plugin.ts` can be written one time but compatible with both CJS and ESM
+// without it we get a `.default` export when using `require` in CJS
+
+// @ts-ignore
+module.exports = require('./plugin.ts').default
```
thecrypticace, line 5: There doesn't appear to be a way for tsup to compile a default export to `module.exports` in my testing so we came up with this + two separate entry points in the tsup config. I like this because it feels like it's more clear that it's a workaround. We _might_ be able to do something like this for the postcss plugin too but would need to test.

RobinMalfait (approved):

philipp-spiess (approved):

adamwathan (approved):

PR #14183, opened 2024-08-13
Add `scrollbar-width` utilities

lukewarlow:
# What?

This PR adds 3 new utilities for styling the width of scrollbars.

`scrollbar-width-auto` - For resetting a scrollbar to full-width, useful for overriding the other two*.

`scrollbar-thin` - For setting the [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) property to `thin`, which normally results in a thinner scrollbar.

`scrollbar-none` - For setting the [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) property to `none`, which hides the scrollbar. This utility also includes a fallback to the legacy `::-webkit-scrollbar { display: none; }`. Read about `::-webkit-scrollbar` [here](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar).

* - I think `scrollbar-none` might take precendence over `scrollbar-width-auto` in WebKit (and chromium) browsers because the `::-webkit-scrollbar` rule beats the `scrollbar-width: auto` rule.

# Why?

Styling of scrollbars is a long requested feature. See https://github.com/tailwindlabs/tailwindcss/discussions?discussions_q=is%3Aopen+scrollbar for a list of discussions requesting utilities for scrollbar styling.

This PR would be an initial step in that direction allowing authors to hide scrollbars (while maintaining the ability to scroll) in UIs where that is needed.

The naming of `scrollbar-width-auto` is different because the auto value is used across 3 separate scrollbar properties ([`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width), [`scrollbar-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color) and [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter). We could keep this inconsistency, change all values to be `scrollbar-width-X`, or we could drop the auto utility?

`scrollbar-width` has 80% support according to [caniuse](https://caniuse.com/mdn-css_properties_scrollbar-width). Though this and [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) are part of Interop 2024 and both the `scrollbar-width` property and `scrollbar-gutter` shipped in Safari 18.2.

Due to the currently limited support mentioned above the `scrollbar-none` utility includes the `::-webkit-scrollbar` fallback to maximise Browser compatability including the fallback brings that utility up to 100% of the supported Tailwind browsers. 

See https://github.com/tailwindlabs/tailwindcss/pull/5732 for a previous attempt at contributing this. No worries if the Tailwind team still aren't ready for scrollbar styling, but given the change in browser support status of this API, I thought I'd give it another go.

If the team are open to these contributions I'd happily also open one for `scrollbar-gutter` and even `scrollbar-color` (might need some API discussions but see https://github.com/tailwindlabs/tailwindcss/discussions/5751 for an initial idea)


benface:
Why `scrollbar-width-auto` and not `scrollbar-auto`?

RobinMalfait:
Hey! Appreciate the PR, this was shipped as of Tailwind CSS [v4.3.0](https://github.com/tailwindlabs/tailwindcss/releases/tag/v4.3.0) with a slightly different API: https://tailwindcss.com/docs/scrollbar-width

Thanks again!

PR #14187, opened 2024-08-13
Improve Oxide scanner API

RobinMalfait:
This PR updates the API for interacting with the Oxide API. Until now, we used the name `scanDir(…)` which is fine, but we do way more work right now.

We now have features such as:

1. Auto source detection (can be turned off, e.g.: `@tailwindcss/vite` doesn't need it)
2. Scan based on `@source`s found in CSS files
3. Do "incremental" rebuilds (which means that the `scanDir(…)` result was stateful).

To solve these issues, this PR introduces a new `Scanner` class where you can pass in the `detectSources` and `sources` options. E.g.:

```ts
let scanner = new Scanner({
  // Optional, omitting `detectSources` field disables automatic source detection
  detectSources: { base: __dirname }, 

  // List of glob entries to scan. These come from `@source` directives in CSS.
  sources: [
    { base: __dirname, pattern: "src/**/*.css" },
    // …
  ],
});
```

The scanner object has the following API:

```ts
export interface ChangedContent {
  /** File path to the changed file */
  file?: string
  /** Contents of the changed file */
  content?: string
  /** File extension */
  extension: string
}
export interface DetectSources {
  /** Base path to start scanning from */
  base: string
}
export interface GlobEntry {
  /** Base path of the glob */
  base: string
  /** Glob pattern */
  pattern: string
}
export interface ScannerOptions {
  /** Automatically detect sources in the base path */
  detectSources?: DetectSources
  /** Glob sources */
  sources?: Array<GlobEntry>
}
export declare class Scanner {
  constructor(opts: ScannerOptions)
  scan(): Array<string>
  scanFiles(input: Array<ChangedContent>): Array<string>
  get files(): Array<string>
  get globs(): Array<GlobEntry>
}
```

The `scanFiles(…)` method is used for incremental rebuilds. It takes the `ChangedContent` array for all the new/changes files. It returns whether we scanned any new candidates or not.

Note that the `scanner` object is stateful, this means that we don't have to track candidates in a `Set` anymore. We can just call `getCandidates()` when we need it.

This PR also removed some unused code that we had in the `scanDir(…)` function to allow for sequential or parallel `IO`, and sequential or parallel `Parsing`. We only used the same `IO` and `Parsing` strategies for all files, so I just got rid of it.


thecrypticace:
I'll give this a more thorough look tomorrow but I have a few comments on naming:

1. I'd drop `get` from the method names. They're nouns anyway so it's self explanatory and the scanner is effectively "read-only" from the outside so the prefixes feel extra unnecessary. Also could / should we turn these into real JS getters? e.g. `get candidates()`? If not that's fine just wondering if we want to (don't feel strongly about it so 🤷‍♂️).

- `getCandidates()` -> `candidates()`
- `getFiles()` -> `files()`
- `getGlobs()` -> `globs()`

2. `autoContent` feels like the wrong name now that we've settled on `@source`. So maybe:
- `detectSources` (i think i like this one the best)
- `discoverSources`
- `autoSource`
- `sourceDetection`
- `sourceDiscovery`

3. Is there a future where the `base` dir for automatic source / content detection would be useful elsewhere? Seems maybe reasonable that `base` could be a top-level option? Don't feel strongly about this — just thinking out loud.

Overall though I'm a fan of this API. Feels more designed and more open to future expansion than the current one.

crates/node/src/lib.rs
```diff
@@ -77,67 +56,85 @@ impl From<GlobEntry> for tailwindcss_oxide::GlobEntry {
...
+    // Return whether we scanned new candidates or not. We can do this because:
+    // 1. It's a set, so we can't have duplicates
+    // 2. We're only adding new candidates, not removing them
+    candidate_count_before != candidate_count_after
+  }
+
+  #[napi]
+  pub fn get_candidates(&self) -> Vec<String> {
+    self.scanner.get_candidates()
+  }
```
philipp-spiess, line 124: What is the behavior if a vec is returned here and exported to Node via napi? My current thinking is that this needs to copy the whole array every time we call getCandidates which might be quite a bit of I/O for large projects.

Maybe alternatively we could only return the new candidates (so moving some of the bookeeping into rust by keeping it in a set that we can discriminate between new and old).

I understand that it is annoying to collect the results in JS land in another data structure but I imagine that we would otherwise call `getCandidates()` quite often in dev mode so maybe this is a reasonable optimization?

philipp-spiess (commented):
@thecrypticace 
> `autoContent` feels like the wrong name now that we've settled on @source. So maybe:

Yeah I agree, should be called something with source now 👍

crates/node/src/lib.rs
RobinMalfait: Yeah that's right. Thinking about it, I think we even track all known candidates in tailwind's core already. So this can definitely be simplified.

thecrypticace: If we do that I think it needs a different name too — maybe something like "newCandidates" or w/e.

On the flip side — even with this stuff being copied — I'd see if we can measure the overhead to see if it's significant or not.

crates/oxide/src/scanner/allowed_paths.rs
```diff
@@ -0,0 +1,59 @@
+use ignore::{DirEntry, WalkBuilder};
+use lazy_static::lazy_static;
+use std::path::Path;
+
+lazy_static! {
```
thecrypticace, line 5: fyi now that we're on Rust 1.80 I think `lazy_static` can be replaced with `std::sync::LazyLock`

crates/oxide/src/lib.rs
```diff
@@ -65,24 +74,89 @@ pub fn clear_cache() {
...
+        init_tracing();
 
-            (files, globs)
+        let mut scanner = Self {
+            auto_content,
+            sources,
+            ..Default::default()
+        };
+
+        scanner.scan();
```
thecrypticace, line 105: It feels super weird to me that `::new()` has the side effect of touching the filesystem. I think I'd prefer that the users for `Scanner` call `scan` instead of relying on `::new()` basically not returning until its done.

crates/oxide/src/lib.rs
```diff
@@ -65,24 +74,89 @@ pub fn clear_cache() {
...
 
-    let (mut files, mut globs) = match opts.base {
-        Some(base) => {
-            // Only enable auto content detection when `base` is provided.
-            let base = Path::new(&base);
-            let (files, dirs) = resolve_files(base);
-            let globs = resolve_globs(base, dirs);
+impl Scanner {
+    pub fn new(auto_content: Option<AutoContent>, sources: Option<Vec<GlobEntry>>) -> Self {
+        init_tracing();
```
thecrypticace, line 97: also thinking maybe this should be not in ::new() either but can't think of a good place to put it. Maybe it should be in `scan`?

thecrypticace:
Since `Scanner` is stateful I think the cache should live in / be owned by the `Scanner` object. Removes our need for a `clearCache()` top-level fn which always felt like a hack.

crates/oxide/src/lib.rs
RobinMalfait: Yep I agree, reasons I still did it is that when using it, we basically always did this:

```ts
let scanner = new Scanner(…);
scanner.scan();
```

I think in a perfect world you can do something like this:
```ts
let scanner = new Scanner(…);

// Setup watchers, so that they are already "listening" for changes before we perform the first scan. Prevents a potential race condition where a file changed between setting up the scanner and creating the watchers.
await setupWatchers();

// Perform initial scan
scanner.scan();
```

However, in this case, in order to setup the watchers, we need to resolve the globs (which touches the file system).

packages/tailwindcss/tests/ui.spec.ts
```diff
@@ -317,6 +316,7 @@ async function render(page: Page, content: string) {
 
   await page.setContent(content)
 
+  let scanner = new Scanner({})
```
thecrypticace, line 319: fyi @RobinMalfait I fixed the UI tests — scanner being stateful caused problems because scanFiles only returns new candidates now. I think using a new scanner every time here is probably a good idea regardless.

thecrypticace:
Something else to think about:

Intellisense does *not* need candidates when scanning for thing using Oxide. Just a list of files and globs is sufficient. Scanning the files themselves would be unnecessary work.

RobinMalfait:
> Something else to think about:
> 
> 
> 
> Intellisense does *not* need candidates when scanning for things using Oxide. Just a list of files and globs is sufficient. Scanning the files themselves would be unnecessary work.

Thats should already be the case now. You just have to use `scanner.files` and `scanner.globs`. Whenever the first one is used, the files/globs will be calculated.

`scanner.scan()` should not be required anymore.

thecrypticace (approved):

PR #14199, opened 2024-08-15
Add Glimmer template extensions to Oxide content detection

mfeckie:
I recently watched the talk at [Tailwind Connect 2023](https://youtu.be/CLkxRnRQtDE) and was very interested in the idea that we'd no longer have to specify "content" and so was looking through the implementation.

I noticed that currently it would not be able to support Ember.js next gen [templates](https://guides.emberjs.com/release/components/template-tag-format/) as the file extensions aren't listed (`gts` and `gjs`).

I added them and then went to update the tests.

When I did that, I noticed that it was hard to understand what was happening with the extensions as they weren't in alphabetical order (I suspect this is because of the FxHashMap algorithm), so I took the liberty of sorting the vec.

Something I may also have misunderstood is that the found extensions:

```rust
    let mut found_extensions = FxHashSet::from_iter(
        include_str!("fixtures/template-extensions.txt")
            .trim()
            .lines()
            .filter(|x| !x.starts_with('#')) // Drop commented lines
            .filter(|x| !x.is_empty()) // Drop empty lines
            .map(|x| x.to_string()),
    );
```

Seems to run each time the function is called. Perhaps this isn't a big performance penalty, but it seems like this could actually be done at compile time or via a `lazy_static` as it doesn't depend on any of the function's inputs, but wanted to check before considering updating that too.


thecrypticace (approved):

thecrypticace:
Thanks!

As for the note about `found_extensions` — I'll make some changes regarding that in #14187 which is refactoring this code.

mfeckie:
No worries 😄

PR #14221, opened 2024-08-19
Improve compatibility with `@tailwindcss/typography` and `@tailwindcss/forms`

thecrypticace:
This PR enables compatibility for the `@tailwindcss/typography` and `@tailwindcss/forms` plugins. This required the addition of new Plugin APIs and new package exports.

## New Plugin APIs and compatibility improvements

We added support for `addComponents`, `matchComponents`, and `prefix`. The component APIs are an alias for the utilities APIs because the sorting in V4 is different and emitting components in a custom `@layer` is not necessary. Since `prefix` is not supported in V4, the `prefix()` API is currently an identity function.

```js
 addComponents({
  '.btn': {
    padding: '.5rem 1rem',
    borderRadius: '.25rem',
    fontWeight: '600',
  },
  '.btn-blue': {
    backgroundColor: '#3490dc',
    color: '#fff',
    '&:hover': {
      backgroundColor: '#2779bd',
    },
  },
  '.btn-red': {
    backgroundColor: '#e3342f',
    color: '#fff',
    '&:hover': {
      backgroundColor: '#cc1f1a',
    },
  },
})
```

The behavioral changes effect the `addUtilities` and `matchUtilities` functions, we now:

- Allow arrays of CSS property objects to be emitted:
  ```js
  addUtilities({
    '.text-trim': [
      {'text-box-trim': 'both'},
      {'text-box-edge': 'cap alphabetic'},
    ],
  })
  ```
- Allow arrays of utilities
  ```js
  addUtilities([
    {
      '.text-trim':{
        'text-box-trim': 'both',
        'text-box-edge': 'cap alphabetic',
      },
    }
  ])
  ```
- Allow more complicated selector names
  ```js
  addUtilities({
    '.form-input, .form-select, .form-radio': {
      /* styles here */
    },
    '.form-input::placeholder': {
      /* styles here */
    },
    '.form-checkbox:indeterminate:checked': {
      /* styles here */
    }
  })
  ```

## New `tailwindcss/color` and `tailwindcss/defaultTheme` export

To be compatible to v3, we're adding two new exports to the tailwindcss package. These match the default theme values as defined in v3:

```ts
import colors from 'tailwindcss/colors'

console.log(colors.red[600])
```

```ts
import theme from 'tailwindcss/defaultTheme'

console.log(theme.spacing[4])
```


packages/tailwindcss/src/compat/colors.ts
philipp-spiess, line 1: I wonder if we can/should compute this from the `theme.css` file. 🤔

thecrypticace: We could but that'd be a build-step and I'd rather not do that for now at least. Can come back later and do it if we need to.

philipp-spiess: I guess we _could_ do it at runtime too but yeah it seems like a waste of CPU cycles lol

thecrypticace: doing it at runtime would necessitate being able to use `node:fs` in core which we don't want to do.

philipp-spiess: Can we maybe do something like a file content import during bundling so that we can read the `css` file contents into an import? 🤔

packages/tailwindcss/src/compat/default-theme.ts
```diff
@@ -0,0 +1,23 @@
...
+  ...createCompatConfig(theme).theme,
+  fontSize: {
+    base: ['1rem', { lineHeight: '1.5rem' }],
+  },
+  spacing: {
+    2: '0.5rem',
+    3: '0.75rem',
+    4: '1rem',
+    10: '2.5rem',
+  },
```
philipp-spiess, line 16: Should these move into the `createCompatConfig` function? I understand this won't have _values_ then which maybe is the point but perhaps we can do something like this: https://github.com/tailwindlabs/tailwindcss/pull/14221/files#r1726670761

packages/tailwindcss/src/compat/colors.ts
philipp-spiess: We can make it work (esbuild has loaders for CSS that we can use). The issue is that we made `compileCss` async so we'd have to extract the theme parsing part (into a stand alone function) but if we do that we can compute both this color file and the default theme. We can even just spread `createCompatTheme()` or `createCompatTheme().colors` in there I think 🤔

philipp-spiess: What's good is that the file imports can be only added to the `default-theme` and `colors` build, not the normal `tailwindcss` import 🥺

philipp-spiess: Been hacking on this a bit and the hardest part is getting vitest to work because I want to write a test to make sure it indeed creates the same output object 😭 🙈

packages/tailwindcss/src/compat/default-theme.ts
philipp-spiess: Hmm in `createCompatConfig` this would be of type bare so we would need to "excute" the bare fn handler for the known spaces which is awkward again. 

maybe copy pasting is fine for now lol but we should make sure we catch all values not just `2`, `3`, `4`, and `10`

thecrypticace: I'm not sure if hardcoding the values is the way to go but the list needs to at least be expanded on. I thought about using createCompatConfig + resolveConfig here but I think that'll introduce some complexity. Given that this is really for backwards compatibility I think it's *probably* fine to hardcode them.

packages/tailwindcss/src/plugin-api.test.ts
```diff
@@ -1011,6 +1011,83 @@ describe('addUtilities()', () => {
...
+
+        @theme reference {
+          --breakpoint-lg: 1024px;
+        }
+      `,
+      {
+        async loadPlugin() {
+          return ({ addUtilities }: PluginAPI) => {
+            addUtilities({
+              '.form-input, .form-textarea:hover:focus': {
```
thecrypticace, line 1069: can we add `.form-input::placeholder` to this list too just to show that the pseudo-element works?

philipp-spiess (approved):
seal of approval

PR #14223, opened 2024-08-20
Change line height of the `lg` font size to `1.625rem`

MartijnCuppens:
The default line height of the `lg` font size looks a bit large compared to the rest of the line heights. You can see this in these demo's (third paragraph):

| https://play.tailwindcss.com/0B3D7HVsj0 | https://play.tailwindcss.com/rZkSJPczKR |
| - | - |
| PixelSnap 2024-08-20 at 11 06 02 AM@2x | PixelSnap 2024-08-20 at 11 05 24 AM@2x |

Also, if you plot the values of the relative line heights, you see the gap between `lg` and `xl` (blue dots 4&5):

Relative line-height

This is a breaking change, so I changed this in `v4`.


philipp-spiess:
Unfortunately this PR has undergone the same fate as your other PR so I'm linking my response here too: https://github.com/tailwindlabs/tailwindcss/pull/14503#issuecomment-2653790041

Here's the copy-pastebale userland implementation of this change for anyone stumbling upon this:

```css
@theme {
  --font-size-lg--line-height: 1.625rem;
}
```

PR #14231, opened 2024-08-21
Add support for matching multiple utility definitions for one candidate

thecrypticace:
Currently if a plugin adds a utility called `duration` it will take precedence over the built-in utilities — or any utilities with the same name in previously included plugins. However, in v3, we emitted matches from _all_ plugins where possible.

Take this plugin for example which adds utilities for `animation-duration` via the `duration-*` class:

```ts
import plugin from 'tailwindcss/plugin'

export default plugin(
  function ({ matchUtilities, theme }) {
    matchUtilities(
      { duration: (value) => ({ animationDuration: value }) },
      { values: theme("animationDuration") },
    )
  },
  {
    theme: {
      extend: {
        animationDuration: ({ theme }) => ({
          ...theme("transitionDuration"),
        }),
      }
    },
  }
)
```

Before this PR this plugin's `duration` utility would override the built-in `duration` utility so you'd get this for a class like `duration-3500`:
```css
.duration-3000 {
  animation-duration: 3500ms;
}
```

Now, after this PR, we'll emit rules for `transition-duration` (Tailwind's built-in `duration-*` utility) and `animation-duration` (from the above plugin) and you'll get this instead:
```css
.duration-3000 {
  transition-duration: 3500ms;
}

.duration-3000 {
  animation-duration: 3500ms;
}
```

These are output as separate rules to ensure that they can all be sorted appropriately against other utilities.


philipp-spiess (approved):
Japanese Thumb

PR #14257, opened 2024-08-26
Spreading `tailwindcss/defaultTheme` exports keeps bare values

philipp-spiess:
In #14221 we added a new export to the `tailwindcss` package: `tailwindcss/defaultTheme`. This is build on top of the full config from V3 and will allow plugins to keep being compatible.

However, spreading in from this package has overwritten the bare value callback handler. This PR fixes it by sharing the bare value callbacks with the compat config.

### DIFF
--- a/packages/tailwindcss/src/compat/default-theme.ts
+++ b/packages/tailwindcss/src/compat/default-theme.ts
@@ -1,5 +1,14 @@
 import { Theme } from '../theme'
-import { createCompatConfig } from './config/create-compat-config'
+import {
+  bareAspectRatio,
+  bareDegrees,
+  bareIntegers,
+  bareMilliseconds,
+  barePercentages,
+  barePixels,
+  bareRepeatValues,
+  createCompatConfig,
+} from './config/create-compat-config'
 
 let theme = new Theme()
 
@@ -24,6 +33,7 @@ export default {
     selected: 'selected="true"',
   },
   aspectRatio: {
+    ...bareAspectRatio,
     auto: 'auto',
     square: '1 / 1',
     video: '16 / 9',
@@ -95,6 +105,7 @@ export default {
     none: 'none',
   },
   brightness: {
+    ...barePercentages,
     0: '0',
     50: '.5',
     75: '.75',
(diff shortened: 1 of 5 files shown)


packages/tailwindcss/src/compat/default-theme.ts
```diff
@@ -1,10 +1,62 @@
...
+  return value.fraction
+})
+
+let bareRepeatValues = bareValues((value) => {
+  if (!Number.isNaN(Number(value.value))) {
+    return `repeat(${value.value}, minmax(0, 1fr))`
+  }
+})
 
 export default {
```
philipp-spiess, line 58: This is now basically the same as https://github.com/tailwindlabs/tailwindcss/blob/main/stubs/config.full.js with a few bare value helpers spread in to match the V4 behavior

packages/tailwindcss/src/plugin-api.test.ts
```diff
@@ -510,7 +511,7 @@ describe('theme', async () => {
           utility('my-backdrop-opacity', 'backdropOpacity')
           utility('my-backdrop-saturate', 'backdropSaturate')
           utility('my-backdrop-sepia', 'backdropSepia')
-          utility('my-border', 'border')
```
philipp-spiess, line 513: We suspect this was a typo

packages/tailwindcss/src/functions.test.ts
```diff
@@ -285,6 +285,20 @@ describe('theme function', () => {
...
+                font-family: theme(fontFamily.sans);
+              }
+            `),
+          ).toMatchInlineSnapshot(`
+          ".fam {
+            font-family: ui-sans-serif, system-ui, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
+          }"
+        `)
+        })
       })
```
philipp-spiess, line 302: This test is for #14262 but it requires the changes in here (the default theme changes) to be reproducible so we didn't cherry pick it into the other PR. I can make yet-another-PR if request but I think it's reasonable to add this here since it does also test that the default theme works for the CSS `theme()` function

thecrypticace (approved):

PR #14265, opened 2024-08-26
Fix support for declaration fallbacks in plugins

thecrypticace:
This PR fixes support for "fallback" values for declarations in plugins. 

A plugin using `addUtilities`, `matchUtilities`, `addComponents`, etc… should be able to specify "fallback" values for declarations by passing an array as the value of a declaration however this does not currently work in v4 (but it does in v3):

```js
export default {
  plugins: [
    function ({ addUtilities }) {
      addUtilities({
        '.outlined': {
          outline: ['1px solid ButtonText', '1px auto -webkit-focus-ring-color'],
        },
      })
    },
  ],
};
```

After this PR the candidate `outlined` will now produce the following CSS — like it does in v3:

```css
.outlined {
  outline: 1px solid ButtonText;
  outline: 1px auto -webkit-focus-ring-color;
}
```


philipp-spiess (approved):
Ooof

cossssmin:
Super useful for progressive enhancement in HTML emails too - we can keep using custom utilities that produce stuff like this in order to provide fallback for email clients with poor CSS support:

```css
.bg-custom {
  background-color: #000;
  background-color: rgba(0,0,0,0.5);
}
```

Thank you! :)

PR #14269, opened 2024-08-27
Rework Vite plugin to support lightningcss pre processor and fast rebuilds

### LINKED ISSUE #14106: [v4] vite build --watch only compile html classes right on the init run

v4

Plain vite Projekt with css/html Input files in rollup config

v20.11.0

Chrome

Linux Debian 

**Describe your issue**

I have some css files ans HTML files. The HTML contains some tailwindcss classes like p-2 etc. If i run vite build --watch the first build is correct. And the p-2 css Definition is in the Output css. But if i change the classes in the HTML it recompiles but dont get the new classes. If i use @apply in my source css file the Compiler works every time. So must be a bug in the HTML stuff.

### LINKED ISSUE #14205: [v4] Vite plugin breaks when setting `css.transformer` to `"lightningcss"`

4.0.0-alpha.19

Vite 5.4.0

20.16.0

N/A

Ubuntu 22.04

**Reproduction URL**

Let me know if you need this and how you'd like it. [It works fine on Tailwind Play](https://play.tailwindcss.com/ViBCaGZhLm?file=css) but fails in Vite.

**Describe your issue**

With a Vite config specifies `"lightningcss"` as the CSS transformer:

```ts
import tailwindcss from "@tailwindcss/vite";
import react from "@vitejs/plugin-react";
import { defineConfig } from "vite";

export default defineConfig({
  css: {
    transformer: "lightningcss",
  },
  plugins: [react(), tailwindcss()],
});
```

The behaviour of the Vite plugin is undefined and buggy.

It crashes when using `@apply` with class names containing characters `*` or `/`:

```css
@import "tailwindcss";

@utility foo {
  @apply text-3xl/tight;
}
```

```
12:44:49 PM [vite] Pre-transform error: Cannot apply unknown utility class: /
12:44:49 PM [vite] Internal server error: Cannot apply unknown utility class: /
  Plugin: @tailwindcss/vite:generate:serve
  File: /workspace/packages/ui/src/tailwind.css
      at onInvalidCandidate (file:///workspace/node_modules/.pnpm/tailwindcss@4.0.0-alpha.19/node_modules/tailwindcss/dist/lib.mjs:6:109810)
...
```

It pollutes the console with warnings:

```
Error running vite:css on Tailwind CSS output. Skipping.
Error running vite:css on Tailwind CSS output. Skipping.
Error running vite:css on Tailwind CSS output. Skipping.
```

And it seems to ignore `@layer base` classes.

You can work around the problem by removing the offending part of the Vite config.

philipp-spiess:
Fixes #14205
Fixes #14106

This PR reworks the Vite extension in order to supprt `lightningcss` as the pre-processor, enable faster rebuilds, and adds support for `vite build --watch` mode. To make this change possible, we've done two major changes to the extension that have caused the other changes.

## 1. Tailwind CSS is a preprocessor

We now run all of our modifications in `enforce: 'pre'`. This means that Tailwind CSS now gets the untransformed CSS files rather than the one already going through postcss or lightningcss. We do this because Tailwind CSS _is_ a preprocessor at the same level as those tools and we do sometimes use the language in ways that [creates problems when it's the input for other bundlers](https://github.com/tailwindlabs/tailwindcss/pull/14269). 

The correct solution here is to make Tailwind not depend on any other transformations. The main reason we were not using the `enforce: 'pre'` phase in Vite before was becuase we relied on the `@import` flattening of postcss so we now have to do this manually. `@import` flattening is now a concern that every Tailwind V4 client has to deal with so this might actually be something we want to inline into tailwindcss in the future. 

## 2. A Vite config can have multiple Tailwind roots 

This is something that we have not made very explicit in the previous iteration of the Vite plugin but we have to support multiple Tailwind roots in a single Vite workspace. A Tailwind root is a CSS file that is used to configure Tailwind. Technically, any CSS file can be the input for `tailwindcss` but you have to add certain rules (e.g. `@import "tailwindcss";`) to make the compiler do something.

A workspace can have multiple of these rules (e.g. by having different Tailwind configures for different sub-pages). With the addition of [support for `@source` rules](https://github.com/tailwindlabs/tailwindcss/pull/14078) and [JS config files](https://github.com/tailwindlabs/tailwindcss/pull/14239), Tailwind roots become more complex and can have a custom list of _dependencies_ (that is other JavaScript modules the compiler includes as part of these new rules). In order to _only rebuild the roots we need to_, we have to make this separation very clear.


packages/@tailwindcss-vite/package.json
```diff
@@ -31,13 +31,15 @@
     "@tailwindcss/node": "workspace:^",
     "@tailwindcss/oxide": "workspace:^",
     "lightningcss": "catalog:",
-    "postcss-load-config": "^6.0.1",
+    "postcss": "^8.4.41",
```
philipp-spiess, line 34: This is temporary until we start to resolve `@import` ourselves (at this point every client need this so we might as well)

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -16,23 +18,19 @@ export default function tailwindcss(): Plugin[] {
   let scanner: Scanner | null = null
   let changedContent: { content: string; extension: string }[] = []
   let candidates = new Set<string>()
-  let fullRebuildPaths: string[] = []
+  let fullRebuildPaths = new Set<string>()
+  let rebuildStrategy: 'full' | 'incremental' = 'full'
+  let compiler: null | Awaited<ReturnType<typeof compile>> = null
```
philipp-spiess, line 23: The compiler is now persisted between rebuilds. We pipe every change through a `onFileChange` function that will compare it with the list of `fullRebuildPaths`. If a change is detected, it will set the `rebuildStrategy` for the next emitted build to `full`.

Todo for myself: The compiler and the `rebuildStrategy` should probably be keyed by the input css file ID in case there are multiple in one Vite project.

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -239,106 +266,57 @@ export default function tailwindcss(): Plugin[] {
...
 
+        console.log('plugins')
+        console.log(config.plugins)
         cssPlugins = config.plugins.filter((plugin) => {
           return allowedPlugins.includes(plugin.name)
         })
       },
 
-      // Append the postcss-fix-relative-paths plugin
-      async config(config) {
```
philipp-spiess, line 248: We no longer need this since we can just run our own at-import plugins

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -239,106 +266,57 @@ export default function tailwindcss(): Plugin[] {
           ...(config.command === 'build' ? ['vite:css-post'] : []),
         ]
 
+        console.log('plugins')
+        console.log(config.plugins)
         cssPlugins = config.plugins.filter((plugin) => {
           return allowedPlugins.includes(plugin.name)
         })
```
philipp-spiess, line 273: This is something I don't fully understand yet and needs more thought. Maybe this is no longer necessary since we now run in `pre` and downstream plugins would _just work_?

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -239,106 +266,57 @@ export default function tailwindcss(): Plugin[] {
...
         if (id.includes('/.vite/')) return
         let extension = getExtension(id)
         if (extension === '' || extension === 'css') return
 
-        scan(src, extension)
-        updateCssModules(options?.ssr ?? false)
+        let updated = scan(src, extension)
+        if (updated) {
+          queueRebuild(options?.ssr ?? false)
+        }
```
philipp-spiess, line 297: Unclear why this was always causing css files to be rebuild. Changed it to only do that when new candidates are detected.

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -239,106 +266,57 @@ export default function tailwindcss(): Plugin[] {
...
-          // all candidates before generating CSS. This must not be called
-          // during SSR or it will block the server.
-          await server?.waitForRequestsIdle?.(id)
-        }
+        // if (!options?.ssr) {
+        //   // Wait until all other files have been processed, so we can extract
+        //   // all candidates before generating CSS. This must not be called
+        //   // during SSR or it will block the server.
+        //   await server?.waitForRequestsIdle?.(id)
+        // }
```
philipp-spiess, line 319: Unclear why this is necessary. If any candidate change is detected, we already mark these files as updated anyways which will (if I’m not mistaken?) cause a retransform for this file anyways.

packages/@tailwindcss-vite/src/index.ts
philipp-spiess: I thought about it more and it should _definitely_ be keyed by the CSS input ID. In fact I think this should be called `root` or so and should also replace the current `cssModules` var. 

I need to find out exactly why we have the `handled` flag though.

thecrypticace (approved):

PR #14270, opened 2024-08-27
Add standalone CLI

philipp-spiess:
This PR adds a new standalone client: A single-binary file that you can use to run Tailwind v4 without having a node setup. To make this work we use Bun's single-binary build which can properly package up native modules and the bun runtime for us so we do not have to rely on any expand-into-tmp-folder-at-runtime workarounds.

When running locally, `pnpm build` will now standalone artifacts inside `packages/@tailwindcss-standalone/dist`. Note that since we do not build Oxide for other environments in the local setup, you won't be able to use the standalone artifacts for other platforms in local dev mode.

Unfortunately Bun does not have support for Windows ARM builds yet but we found that using the `bun-baseline` runtime for Windows x64 would make the builds work fine in ARM emulation mode:

Screenshot windows

Some Bun related issues we faced and worked around:

- We found that the regular Windows x64 build of  `bun` does not run on Windows ARM via emulation. Instead, we have to use the `bun-baseline` builds which emulate correctly.

- When we tried to bundle artifacts with [embed directories](https://bun.sh/docs/bundler/executables#embed-directories), node binary dependencies were no longer resolved correctly even though they would still be bundled and accessible within the [`embeddedFiles` list](https://bun.sh/docs/bundler/executables#listing-embedded-files). We worked around this by using the `import * as from ... with { type: "file" };` and patching the resolver we use in our CLI.
  
  
- If you have an import to a module that is used as a regular import _and_ a `with { type: "file" }`, it will either return the module in both cases _or_ the file path when we would expect only the `with { type: "file" }` import to return the path. We do read the Tailwind CSS version via the file system and `require.resolve()` in the CLI and via `import * from './package.json'` in core and had to work around this by patching the version resolution in our CLI. 
 
  ```ts
  import packageJson from "./package.json"
  import packageJsonPath from "./package.json" with {type: "file"}
  
  // We do not expect these to be equal
  packageJson === packageJsonPath 
  ```
- We can not customize the app icon used for Windows `.exe` builds without decompiling the binary. For now we will leave the default but one workaround is to [use tools like ResourceHacker](https://github.com/tailwindlabs/tailwindcss/pull/14270/commits/698d9c4bd182391adb4b236f19a539aff13f78f9) to decompile the binary first.


patches/@parcel__watcher@2.4.1.patch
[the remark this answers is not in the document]
thecrypticace: Yep, that's exactly why.

.github/workflows/release.yml
```diff
@@ -210,6 +207,9 @@ jobs:
           cp bindings-x86_64-unknown-linux-gnu/* ./npm/linux-x64-gnu/
           cp bindings-x86_64-unknown-linux-musl/* ./npm/linux-x64-musl/
 
+      - name: Build Tailwind CSS
+        run: pnpm run build
+
```
philipp-spiess, line 212: @thecrypticace Do you know if there was an inherent reason to doing the build step before we downloaded the artifacts? I move it further down so that when we build the standalone artifact we have access to everything.

.github/workflows/release.yml
```diff
@@ -220,3 +220,9 @@ jobs:
         run: pnpm --recursive publish --tag ${{ inputs.release_channel }} --no-git-checks
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Upload Standalone Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: tailwindcss-standalone
+          path: packages/@tailwindcss-standalone/dist/
```
philipp-spiess, line 228: This will only be saved to the workflow run for now but that's probably fine and should be easy to change to be added to the GitHub release once we have it.

scripts/pack-packages.mjs
```diff
@@ -44,7 +44,7 @@ Promise.all(
     function pack() {
       return new Promise((resolve) => {
         exec(
-          `pnpm pack --pack-destination="${path.join(root, 'dist').replace(/\\/g, '\\\\')}"`,
+          `pnpm pack --pack-gzip-level=0 --pack-destination="${path.join(root, 'dist').replace(/\\/g, '\\\\')}"`,
```
philipp-spiess, line 47: `pnpm pack` got very slow for the ~100MB standalone package. Since we don't care about compression here (it's only needed for running our integration tests) we decided to disable gzip compression to significantly speed up the build.

thecrypticace (approved):

packages/tailwindcss/src/index.ts
```diff
@@ -406,3 +408,11 @@ export async function __unstable__loadDesignSystem(css: string, opts: CompileOpt
   let result = await parseCss(css, opts)
   return result.designSystem
 }
+
+async function getVersion() {
```
thecrypticace, line 412: this doesn't need to be async

packages/@tailwindcss-standalone/scripts/build.ts
```diff
@@ -0,0 +1,71 @@
...
+await mkdir(path.resolve(__dirname, '../dist'), { recursive: true })
+
+// Build platform binaries and checksum them
+let results = await Promise.all([
+  build('bun-linux-arm64', './tailwindcss-linux-arm64'),
+  build('bun-linux-x64', './tailwindcss-linux-x64'),
+  // build('linux-armv7', 'tailwindcss-linux-armv7'),
+  build('bun-darwin-arm64', './tailwindcss-macos-arm64'),
+  build('bun-darwin-x64', './tailwindcss-macos-x64'),
+  // The Windows x65 build uses `bun-baseline` instead of the regular bun build/
```
thecrypticace, line 53:
```suggestion
  // The Windows x65 build uses `bun-baseline` instead of the regular bun build
```

philipp-spiess: LOL also 65 :D

Jarred-Sumner:
Very exciting - and thanks for the feedback.

> We found that the regular Windows x64 build of bun does not run on Windows ARM via emulation. Instead, we have to use the bun-baseline builds which emulate correctly.

This is unfortunately expected, and we should just document it. arm64 emulation (including Rosetta) tends not to support SIMD instructions


> If you have an import to a module that is used as a regular import and a with { type: "file" }, it will either return the module in both cases or the file path when we would expect only the with { type: "file" } import to return the path. We do read the Tailwind CSS version via the file system and require.resolve() in the CLI and via import * from './package.json' in core and had to work around this by patching the version resolution in our CLI.

Currently, we treat file paths as the key in some hash tables, but it really should be file path and loader so that this works as expected

Internally, `bun build --compile` does try to make `fs.readFile` and `Bun.file(path).{text,arrayBuffer,json,bytes}` support embedded files without extra code changes, but I don't think we're doing a good enough job of this yet

> When we tried to bundle artifacts with [embed directories](https://bun.sh/docs/bundler/executables#embed-directories), node binary dependencies were no longer resolved correctly even though they would still be bundled and accessible within the [embeddedFiles list](https://bun.sh/docs/bundler/executables#listing-embedded-files). We worked around this by using the import * as from ... with { type: "file" }; and patching the resolver we use in our CLI.

This is a bug. It's probably using the `"file"` loader instead of our `"napi"` loader. Would you be able to file an issue? 

> We can not customize the app icon used for Windows .exe builds without decompiling the binary. For now we will leave the default but one workaround is to [use tools like ResourceHacker](https://github.com/tailwindlabs/tailwindcss/pull/14270/commits/698d9c4bd182391adb4b236f19a539aff13f78f9) to decompile the binary first.

We can add a `--icon` CLI flag or similar.

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -372,7 +372,24 @@ function handleImports(
...
+            throw new Error(`Could not resolve ${id} from ${basedir}`)
+          }
+          return resolved
+        },
+        load(id) {
+          // We need to synchronously read the file here because when bundled
+          // with bun, some of the ids might resolve to files inside the bun
+          // embedded files root which can only be read by `node:fs` and not
+          // `node:fs/promises`.
+          return readFileSync(id, 'utf-8')
```
Jarred-Sumner, line 389: This seems like a bug in Bun.

packages/@tailwindcss-standalone/scripts/build.ts
```diff
@@ -0,0 +1,71 @@
...
+import { createHash } from 'node:crypto'
+import { mkdir, readFile, writeFile } from 'node:fs/promises'
+import * as path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const __dirname = fileURLToPath(new URL('.', import.meta.url))
+
+async function buildForPlatform(triple: string, outfile: string) {
+  // We wrap this in a retry because occasionally the atomic rename fails for some reason
+  for (let i = 0; i < 5; ++i) {
```
Jarred-Sumner, line 11: Definitely a bug in Bun

Is this happening on all platforms or on a specific platform?

packages/@tailwindcss-standalone/scripts/build.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+async function build(triple: string, file: string) {
+  let start = process.hrtime.bigint()
+
+  let outfile = path.resolve(__dirname, `../dist/${file}`)
+
+  await buildForPlatform(triple, outfile)
+
+  await new Promise((resolve) => setTimeout(resolve, 100))
```
Jarred-Sumner, line 29: Was this delay a workaround of a bug?

packages/@tailwindcss-cli/src/utils/resolve.ts
```diff
@@ -1,3 +1,32 @@
...
+  if (typeof globalThis.__tw_resolve === 'function') {
+    let resolved = globalThis.__tw_resolve(id)
+    if (resolved) {
+      return resolved
+    }
+  }
+  return localResolve(id)
+}
+
+const resolver = EnhancedResolve.ResolverFactory.createResolver({
```
Jarred-Sumner, line 16: It's probably not feature-ful enough, but you might be able to instead use `Bun.resolveSync` here. `Bun.resolveSync` supports a 2nd argument for the path.

You can see some tests for it here:
https://github.com/oven-sh/bun/blob/6603133732892227f0a32e3f19b343fd1e59a604/test/js/bun/resolve/import-meta.test.js#L133-L154

packages/@tailwindcss-standalone/scripts/build.ts
thecrypticace: When I ran into this it was on macOS. Not sure if it applies to other platforms. It happened enough that I added a loop for it but it was **very** inconsistently reproducible.

packages/@tailwindcss-standalone/scripts/build.ts
thecrypticace: It was because the `readFile` that follows would occasionally fail unless we waited a little bit. Perhaps it's like a filesystem buffer syncing thing?

I need to give it another test to see if its still a problem.

packages/@tailwindcss-cli/src/utils/resolve.ts
thecrypticace: Un fortunately I don't think we can because we need support for the `style` field and `style` conditions in `exports`. Good to know about this though — perhaps it could be usable elsewhere.

thecrypticace:
> This is unfortunately expected, and we should just document it. arm64 emulation (including Rosetta) tends not to support SIMD instructions

FWIW, Rosetta 2 translates SSE2 instructions in Sonoma at least. In Sequoia it'll also support translation of AVX2 instructions (I assume that includes normal AVX as well). Translation of AVX512 is not supported tho.

philipp-spiess:
Thanks a bunch @Jarred-Sumner, this is super helpful! I created issues with repros for the bugs and one issue for the `--icon` feature request :)

- https://github.com/oven-sh/bun/issues/13684
- https://github.com/oven-sh/bun/issues/13685
- https://github.com/oven-sh/bun/issues/13686

RobinMalfait (commented):
This seems to work on my machine as expected. Only thing I want to figure out is to extend the integration tests to also test using the standalone CLI.

We currently only test on Linux and Windows, so we don't fully test every single OS that we generate binaries for. But if those two are tested via integration tests already that would be sick.

RobinMalfait (approved):

henrikbjorn:
This does not work for me. 

If I download the standalone from the latest release and run the binary I get this output:

```
/app # /bin/tailwindcss
Bun is a fast JavaScript runtime, package manager, bundler, and test runner. (1.1.26+0a37423ba)

Usage: bun <command> [...flags] [...args]

Commands:
  run       ./my-script.ts       Execute a file with Bun
            lint                 Run a package.json script
  test                           Run unit tests with Bun
  x         bun-repl             Execute a package binary (CLI), installing if needed (bunx)
  repl                           Start a REPL session with Bun
  exec                           Run a shell script directly with Bun

  install                        Install dependencies for a package.json (bun i)
  add       @shumai/shumai       Add a dependency to package.json (bun a)
  remove    babel-core           Remove a dependency from package.json (bun rm)
  update    hono                 Update outdated dependencies
  outdated                       Display latest versions of outdated dependencies
  link      [<package>]          Register or link a local npm package
  unlink                         Unregister a local npm package
  patch <pkg>                    Prepare a package for patching
  pm <subcommand>                Additional package management utilities

  build     ./a.ts ./b.jsx       Bundle TypeScript & JavaScript into a single file

  init                           Start an empty Bun project from a blank template
  create    svelte               Create a new project from a template (bun c)
  upgrade                        Upgrade to latest version of Bun.
  <command> --help               Print help text for command.

Learn more about Bun:            https://bun.sh/docs
Join our Discord community:      https://bun.sh/discord
```

Which is just bun. The above is done in an Alpine container on a M2 MacBook Pro.

Getting the same binary but for macos gives the correct and following output:


```
~/Downloads ❯❯❯ xattr -d com.apple.quarantine  tailwindcss-macos-arm64                                                                                                                                                                                                                                             ✘
~/Downloads ❯❯❯ ./tailwindcss-macos-arm64
≈ tailwindcss v4.0.0-alpha.25

Usage:
  tailwindcss [--input input.css] [--output output.css] [--watch] [options…]

Options:
  -i, --input ··········· Input file
  -o, --output ·········· Output file
  -w, --watch ··········· Watch for changes and rebuild as needed
  -m, --minify ·········· Optimize and minify the output
      --optimize ········ Optimize the output without minifying
      --cwd ············· The current working directory [default: `.`]
  -h, --help ············ Display usage information
~/Downloads ❯❯❯
```

@philipp-spiess @RobinMalfait

PR #14278, opened 2024-08-28
Remove named opacity support for color opacity modifiers

philipp-spiess:
Right now if you have a custom opacity theme value configured like this…

```css
@theme {
  --opacity-disabled: 50%;
}
```

…then you can use that named opacity value as a color opacity modifier like this:

```html
<div class="bg-red-500/disabled">
```

We think this behavior is confusing. The color opacity modifier is not setting opacity but the alpha value of a color. Opacity is for setting the opacity of an element, so the custom names you'd come up with are named after those situations, which makes them not seem appropriate for color modifiers.


RobinMalfait (approved):

adamwathan (approved):

emmbm:
Is there any chance we could see support for named alpha values, maybe under a different theme key like `--alpha-foo`?

PR #14319, opened 2024-09-03
Fix Nuxt integration

philipp-spiess:
We noticed that Nuxt projects were not working with the tailwindcss project. The issue was traced down to the fact that Nuxt starts multiple Vite dev servers and calling the experimental `waitForRequestsIdle()` on one of the test servers would never resolve.

This was fixed upstream and is part of the latest Vite/Nuxt release: https://github.com/vitejs/vite/issues/17980.

We still need to handle the fact that Vite can spawn multiple dev servers. This is necessary because when we invalidate all roots, we need to find that module inside all of the spawned servers. If we only look at the _last server_ (what we have done before), we would not find the module and thus could not invalidate it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure content globs defined in `@config` files are relative to that file ([#14314](https://github.com/tailwindlabs/tailwindcss/pull/14314))
 - Ensure CSS `theme()` functions are evaluated in media query ranges with collapsed whitespace ((#14321)[https://github.com/tailwindlabs/tailwindcss/pull/14321])
+- Fix support for Nuxt projects in the Vite plugin (requires Nuxt 3.13.1+) ([#14319](https://github.com/tailwindlabs/tailwindcss/pull/14319))
 
 ## [4.0.0-alpha.21] - 2024-09-02
 
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -11,7 +11,7 @@ import postcssImport from 'postcss-import'
 import type { Plugin, ResolvedConfig, Rollup, Update, ViteDevServer } from 'vite'
 
 export default function tailwindcss(): Plugin[] {
-  let server: ViteDevServer | null = null
+  let servers: ViteDevServer[] = []
   let config: ResolvedConfig | null = null
 
   let isSSR = false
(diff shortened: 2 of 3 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -169,7 +168,7 @@ export default function tailwindcss(): Plugin[] {
       },
       transform(src, id, options) {
         let extension = getExtension(id)
-        if (extension === '' || extension === 'css') return
```
philipp-spiess, line 172: This is an optimization that makes sure vite `<style>` blocks aren't scanned for candidates in addition to `css` files.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure content globs defined in `@config` files are relative to that file ([#14314](https://github.com/tailwindlabs/tailwindcss/pull/14314))
 - Ensure CSS `theme()` functions are evaluated in media query ranges with collapsed whitespace ((#14321)[https://github.com/tailwindlabs/tailwindcss/pull/14321])
+- Fix support for Nuxt projects in the Vite plugin (requires Nuxt 3.13.1+) ([#14319](https://github.com/tailwindlabs/tailwindcss/pull/14319))
```
philipp-spiess, line 18: Not happy about this line but if we want to mention the improved Nuxt support we should mention the version number, otherwise people try it and it won't work 😢

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -193,7 +192,9 @@ export default function tailwindcss(): Plugin[] {
           // The reason why we can not rely on the invalidation here is that the
           // users would otherwise see a flicker in the styles as the CSS might
           // be loaded with an invalid set of candidates first.
-          await server?.waitForRequestsIdle?.(id)
+          for (let server of servers) {
+            await server.waitForRequestsIdle(id)
+          }
```
thecrypticace, line 197: Should we do a `Promise.all` here?

CHANGELOG.md
thecrypticace: Seems fine to me tbh

packages/@tailwindcss-vite/src/index.ts
philipp-spiess: Yeah probably better to start waiting immediately 👍

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -193,7 +192,7 @@ export default function tailwindcss(): Plugin[] {
           // The reason why we can not rely on the invalidation here is that the
           // users would otherwise see a flicker in the styles as the CSS might
           // be loaded with an invalid set of candidates first.
-          await server?.waitForRequestsIdle?.(id)
```
thecrypticace, line 196: We need to bump the minimum Vite version to the one where `waitForRequestsIdle` was introduced if it's not already btw.

thecrypticace (approved):

PR #14337, opened 2024-09-04
Use a bit flags for theme options instead of an object with booleans

RobinMalfait:
This PR is a small improvement to the theme options where it's just a simple number now instead of an object with booleans.

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -295,7 +295,7 @@ async function parseCss(
 
     // Keep a reference to the first `@theme` rule to update with the full theme
     // later, and delete any other `@theme` rules.
-    if (!firstThemeRule && (themeOptions & ThemeOptions.REFERENCE) !== ThemeOptions.REFERENCE) {
+    if (!firstThemeRule && !(themeOptions & ThemeOptions.REFERENCE)) {
       firstThemeRule = node
     } else {
       replaceWith([])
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/index.ts
```diff
@@ -26,18 +26,17 @@ function throwOnConfig(): never {
 }
 
 function parseThemeOptions(selector: string) {
-  let isReference = false
-  let isInline = false
+  let options = ThemeOptions.NONE
```
philipp-spiess, line 29: Does the enum access compile away or does each of these now do a key lookup on an object?

RobinMalfait: They compile away because the enum is defined as a `const enum`:

https://github.com/tailwindlabs/tailwindcss/blob/13a03930853a13190f4c8b2c30a9939c263d1abf/packages/tailwindcss/src/theme.ts#L3-L8

Example in the TypeScript playground: 
https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwgOwM72MgrgWzgFQAtgdgB5MGASxVTgG8AoOOAOTNYFE4BeOABgA0zOAElWAGXHc+ARjgAeBQOEsASpwBinDawDCMuPKVHVcACJaAggFUJ+XkcXKATMIC+jRgBtg8AIaORCTklDRoAHTiUlw+fnAARkHEpBTUtBEa2roGQA

```js
let a = 1 /* ThemeOptions.INLINE */;
let b = 2 /* ThemeOptions.REFERENCE */;
export {};
```

RobinMalfait: Was still curious to what we _actually_ produce. Looking at the `dist/lib.mjs` file. If I pretty print it, the code looks like this:
```js
function zr(r) {
  let i = 0
  for (let o of $(r.slice(6), ' '))
    o === 'reference' ? (i |= 2) : o === 'inline' ? (i |= 1) : o === 'default' && (i |= 4)
  return i
}
```

philipp-spiess: Oh, sweeet

thecrypticace (approved):

philipp-spiess (approved):

PR #14340, opened 2024-09-04
Vite: Add support for `<style>` tags in Astro files

philipp-spiess:
This works similar to the Vue setup. The styles that Astro will receive might still contain Tailwind CSS APIs but since it's not picky, we can pass that through to the regular Vite `transform` handlers for now.

This, however, will have issues like https://github.com/tailwindlabs/tailwindcss/issues/14205. We have to fix this together with Vue and other similar extensions later. For now, it will break when syntax is used that lightningcss rewrites (like `@apply text-3xl/tight;`)


integrations/utils.ts
```diff
@@ -536,32 +536,40 @@ export async function fetchStyles(port: number, path = '/'): Promise<string> {
...
-  let regex = /<link rel="stylesheet" href="([a-zA-Z0-9\/_\.\?=%-]+)"/g
+  let linkRegex = /<link rel="stylesheet" href="([a-zA-Z0-9\/_\.\?=%-]+)"/gi
+  let styleRegex = /<style\b[^>]*>([\s\S]*?)<\/style>/gi
+
+  let stylesheets: string[] = []
 
   let paths: string[] = []
   let match
-  while ((match = regex.exec(html)) !== null) {
+  while ((match = linkRegex.exec(html)) !== null) {
```
thecrypticace, line 546:
```suggestion
  for (let match of html.matchAll(linkRegex)) {
```

integrations/utils.ts
```diff
@@ -536,32 +536,40 @@ export async function fetchStyles(port: number, path = '/'): Promise<string> {
...
+          headers: {
+            Accept: 'text/css',
+          },
+        })
+        return await css.text()
+      }),
+    )),
   )
 
+  while ((match = styleRegex.exec(html)) !== null) {
```
thecrypticace, line 566:
```suggestion
  for (let match of html.matchAll(styleRegex)) {
```

thecrypticace (changes requested):

integrations/utils.ts
thecrypticace: Pretty sure matchAll is equivalent here and it reads better imo

thecrypticace (approved):

Issue #14054, opened 2024-07-25
addComponents Type doesn't accept nested properties in media queries


moshie:
**What version of Tailwind CSS are you using?**

v3.4.6

**What build tool (or framework if it abstracts the build tool) are you using?**

prejss-cli: 0.3.3
postcss-cli: 11.0.0

**What version of Node.js are you using?**

v20.15.1

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction:**

```tsx
import type { PluginCreator, Config } from 'tailwindcss/types/config';

const pluginCreator: PluginCreator = function pluginCreator({
  addBase,
  addComponents,
  config,
}) {

addComponents({
    "@media (min-width: 400px)": [
      {
        ".range::-webkit-slider-thumb": {
          "width":"var(--range-thumb-size)",
          "height":"var(--range-thumb-size)",
          "marginTop":"calc(((var(--range-thumb-size) / 2) - var(--range-track-height) / 2) * -1)",
          "background":"url(\"data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 512 512'%3E%3Cpath fill='white' d='M377.941 169.941V216H134.059v-46.059c0-21.382-25.851-32.09-40.971-16.971L7.029 239.029c-9.373 9.373-9.373 24.568 0 33.941l86.059 86.059c15.119 15.119 40.971 4.411 40.971-16.971V296h243.882v46.059c0 21.382 25.851 32.09 40.971 16.971l86.059-86.059c9.373-9.373 9.373-24.568 0-33.941l-86.059-86.059c-15.119-15.12-40.971-4.412-40.971 16.97z'/%3E%3C/svg%3E\")\n        no-repeat\n        center center / 20px 20px\n        var(--range-thumb-background)"
        }
      }
    ]
  });
}
```

<img width="819" alt="Screenshot 2024-07-25 at 10 40 36" src="https://github.com/user-attachments/assets/474c2187-6ad6-4843-95ed-741f3dd3be25">

```
Type '{ ".range::-webkit-slider-thumb": { width: string; height: string; marginTop: string; background: string; }; }' is not assignable to type 'string'
```

```json
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "compilerOptions": {
    "target": "ES2020",
    "lib": ["ES2020", "DOM", "DOM.Iterable"],
    "allowJs": true,
    "skipLibCheck": true,
[...]

PR #14347, opened 2024-09-04
Update Types to allow an array of CSSRuleObjects

moshie:
When generating tailwind css to js using prejss it combines sections in an array which tailwind's types are not happy with. 

```
"@supports (color: oklch(0% 0 0))": [
      {
        ".btn": {
          "backgroundColor": "var(--btn-color, var(--zui-primary-button))",
          "borderColor": "var(--btn-color, var(--zui-primary-button))"
        }
      },
      {
        ".btn-primary": {
          "-BtnColor": "var(--primary-button)"
        }
      }
    ],
```

I propose we update the type to allow for CSSRuleObjects[]

related to: #14054


RobinMalfait:
Hey!

While this works, I would recommend to use 2 sibling rules instead of the array syntax. We allow the array syntax at "runtime" for backwards compatibility reasons, but the proper way to do this is to use normal objects:
```ts
addComponents({
  '@supports (color: oklch(0% 0 0))': {
    '.btn': {
      backgroundColor: 'var(--btn-color, var(--zui-primary-button))',
      borderColor: 'var(--btn-color, var(--zui-primary-button))',
    },
    '.btn-primary': {
      '--btn-color': 'var(--primary-button)',
    },
  },
})
```

Example: https://play.tailwindcss.com/SYZAmHUI8p?file=config

Going to close this for now because there are better solutions, but I appreciate the PR!

Thanks!

PR #14349, opened 2024-09-05
Add integration test for Vite CSS module support

philipp-spiess:
Wanted to make sure stuff works with CSS modules for both the postcss and the lightningcss pipeline.

### DIFF
--- a/integrations/vite/css-modules.test.ts
+++ b/integrations/vite/css-modules.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect } from 'vitest'
 import { css, html, test, ts, txt } from '../utils'
-;['postcss', 'lightningcss'].forEach((transformer) => {
+
+for (let transformer of ['postcss', 'lightningcss']) {
   describe(transformer, () => {
     test(
       `dev mode`,
@@ -65,4 +66,4 @@ import { css, html, test, ts, txt } from '../utils'
       },
     )
   })
-})
+}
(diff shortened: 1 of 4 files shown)


integrations/vite/css-modules.test.ts
```diff
@@ -0,0 +1,68 @@
+import { describe, expect } from 'vitest'
+import { css, html, test, ts, txt } from '../utils'
+;['postcss', 'lightningcss'].forEach((transformer) => {
```
RobinMalfait, line 3: I know we do this in other places, but I think a normal for loop is a bit cleaner here (gets rid of the `;` and allows a new line again)

```suggestion
for (let transformer of ['postcss', 'lightningcss']) {
```

Or, if it gives us any benefits, we could use `describe.each` as well: https://vitest.dev/api/#describe-each

integrations/vite/css-modules.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+      async ({ exec, fs }) => {
+        await exec(`pnpm vite build`)
+
+        let files = await fs.glob('dist/**/*.css')
+        expect(files).toHaveLength(1)
+        let [filename] = files[0]
+
+        await fs.expectFileToContain(filename, [
+          /\.[^f]*_foo[^t]*text-decoration-line: underline;/gi,
```
RobinMalfait, line 63: Goodness gracious 😂

RobinMalfait (approved):
Looks good to me, just a suggestion

integrations/vite/css-modules.test.ts
thecrypticace: 😂😂😂

PR #14362, opened 2024-09-06
Make config resolution lazy

thecrypticace:
The internal `registerPlugins()` API is used to enable backwards compatibility with v3 plugins and configs and it is called on every build even when no v3 plugins or configs are used. This function has a non-trivial cost in that case — around 5ms.

So this PR does a few things:

## Implements a simpler, faster `theme(…)` function

We now have a much simpler `theme(…)` function that can be used when backwards compatibility is not necessary. It still supports many of the same features:
- The modern, v4 style CSS variable syntax `theme(--color-red-500)`
- The legacy, v3 path style `theme(colors.red.500)`
- And the v3-style alpha modifier `theme(colors.red.500 / 50%)`
- Path upgrades so things like `theme(accentColor.red.500)` pulls from `--color-red-500` when no `--accent-color-red-500` theme key exists

When you do have plugins or configs the more advanced `theme(…)` function is swapped in for more complete backwards compatibility.

## `registerPlugins` registers globs

Before `registerPlugins` passed the `ResolvedConfig` out so we could register globs in `compile()`. Since that one function is really the main driver for backwards compat we decided to move the content path registration into `registerPlugins` itself when it comes to paths provided by plugins and configs.

This is an internal implementation detail (well this entire PR is) but it's worth mentioning. This method is used to resolve a theme value from a theme key.

## `registerPlugins` is now only called when necessary

All of the above work made it so that `registerPlugins` can be called only as needed. This means that when no v3 plugins or configs are used, `registerPlugins` is never called thus elminating the performance impact of config resolution.

### DIFF
--- a/packages/tailwindcss/src/compat/plugin-functions.ts
+++ b/packages/tailwindcss/src/compat/plugin-functions.ts
@@ -11,7 +11,7 @@ export function createThemeFn(
   configTheme: () => UserConfig['theme'],
   resolveValue: (value: any) => any,
 ) {
-  return function theme(path: string, defaultValue?: any) {
+  return function theme(path: string, defaultValue?: unknown) {
     // Extract an eventual modifier from the path. e.g.:
     // - "colors.red.500 / 50%" -> "50%"
     // - "foo/bar/baz/50%"      -> "50%"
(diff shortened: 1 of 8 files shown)


adamwathan (approved):
Going to merge this as-is for now but have some ideas for cleaning up a bunch more of this stuff in general 👍

PR #14371, opened 2024-09-09
Add `matchVariant` API

philipp-spiess:
This PR adds support for the `matchVariant` plugin API. I've copied over all [V3 tests](https://github.com/tailwindlabs/tailwindcss/blob/f07dbff2a7f78fd75c53c6cfe01b58b6c0419f22/tests/match-variants.test.js) and made sure they still pass. 

## Sorted order of stacked arbitrary variants

The only difference in behavior is regarding the sort order of stacked arbitrary variants: Sorting in this case now works by the latest defined `matchVariant` taking precedence. 

So, if you define a plugin like this:

```ts
matchVariant('testmin', (value) => `@media (min-width: ${value})`, {
  sort(a, z) {
    return parseInt(a.value) - parseInt(z.value)
  },
})

matchVariant('testmax', (value) => `@media (max-width: ${value})`, {
  sort(a, z) {
    return parseInt(z.value) - parseInt(a.value)
  },
})
```

The resulting CSS is first sorted by the `testmax` values descending and then the `testmin` values ascending, so these candidates:

```txt
testmin-[150px]:testmax-[400px]:order-2
testmin-[100px]:testmax-[350px]:order-3
testmin-[100px]:testmax-[300px]:order-4
testmin-[100px]:testmax-[400px]:order-1
```

Will resolve to the order outlined by the `order-` utility.

## At-rules and placeholders support

Since we added support for at-rules and placeholders in the `matchVariant` syntax like this:

```ts
matchVariant(
  'potato',
  (flavor) => `@media (potato: ${flavor}) { @supports (font:bold) { &:large-potato } }`,
)
```

We also added support for the same syntax to the `addVariant` API:

```ts
addVariant(
  'potato',
  '@media (max-width: 400px) { @supports (font:bold) { &:large-potato } }',
)
```

The only change necessary in core was to call functional variants for when the variant value is set to `null`. This allows functional variants to define the un-parameterized implementation like `potato:underline` as opposed to `potato[big]:underline`.

### DIFF
--- a/packages/tailwindcss/src/plugin-api.ts
+++ b/packages/tailwindcss/src/plugin-api.ts
@@ -1,5 +1,5 @@
 import { substituteAtApply } from './apply'
-import { decl, rule, type AstNode } from './ast'
+import { decl, rule, walk, type AstNode } from './ast'
 import type { Candidate, NamedUtilityValue } from './candidate'
 import { applyConfigToTheme } from './compat/apply-config-to-theme'
 import { createCompatConfig } from './compat/config/create-compat-config'
@@ -8,6 +8,7 @@ import type { ResolvedConfig, UserConfig } from './compat/config/types'
 import { darkModePlugin } from './compat/dark-mode'
 import { createThemeFn } from './compat/plugin-functions'
 import { substituteFunctions } from './css-functions'
+import * as CSS from './css-parser'
 import type { DesignSystem } from './design-system'
 import type { Theme, ThemeKey } from './theme'
 import { withAlpha, withNegative } from './utilities'
@@ -27,7 +28,19 @@ export type Plugin = PluginFn | PluginWithConfig | PluginWithOptions<any>
 
 export type PluginAPI = {
   addBase(base: CssInJs): void
+
   addVariant(name: string, variant: string | string[] | CssInJs): void
+  matchVariant<T = string>(
+    name: string,
+    cb: (value: T | string, extra: { modifier: string | null }) => string | string[],
+    options?: {
+      values?: Record<string, T>
+      sort?(
+        a: { value: T | string; modifier: string | null },
+        b: { value: T | string; modifier: string | null },
+      ): number
+    },
+  ): void
 
   addUtilities(
     utilities: Record<string, CssInJs | CssInJs[]> | Record<string, CssInJs | CssInJs[]>[],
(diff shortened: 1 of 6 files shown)


packages/tailwindcss/src/plugin-api.ts
```diff
@@ -96,6 +103,71 @@ function buildPluginApi(
...
+          if (a.kind !== 'functional' || z.kind !== 'functional') {
+            return 0
+          }
+          if (!a.value || !z.value) {
+            return 0
+          }
+
+          if (options && typeof options.sort === 'function') {
+            let aValue = options?.values?.[a.value.value] ?? a.value.value
+            let zValue = options?.values?.[z.value.value] ?? z.value.value
```
RobinMalfait, line 156: We already checked if `options` exist, so this should be safe to do:
```suggestion
            let aValue = options.values?.[a.value.value] ?? a.value.value
            let zValue = options.values?.[z.value.value] ?? z.value.value
```

packages/tailwindcss/src/plugin-api.ts
```diff
@@ -96,6 +103,71 @@ function buildPluginApi(
...
+
+            return options.sort(
+              { value: aValue, modifier: a.modifier?.value ?? null },
+              { value: zValue, modifier: z.modifier?.value ?? null },
+            )
+          }
+
+          let aOrder = defaultOptionKeys.indexOf(a.value.value) ?? 0
+          let zOrder = defaultOptionKeys.indexOf(z.value.value) ?? 0
```
RobinMalfait, line 165: Don't think we will ever hit the `?? 0` branch since `indexOf(…)` would return `-1`. Man, I wish we had proper enums in JS 😂

```suggestion
          let aOrder = defaultOptionKeys.indexOf(a.value.value)
          let zOrder = defaultOptionKeys.indexOf(z.value.value)
```

packages/tailwindcss/src/plugin-api.ts
```diff
@@ -346,6 +418,20 @@ function objectToAst(rules: CssInJs | CssInJs[]): AstNode[] {
   return ast
 }
 
+function parseVariantValue(resolved: string | string[], nodes: AstNode[]): AstNode[] {
+  let resolvedArray = typeof resolved === 'string' ? [resolved] : resolved
+  return resolvedArray.flatMap((r) => {
+    if (r.trim().endsWith('}')) {
+      let updatedCSS = r.replace('}', '{@slot}}')
```
RobinMalfait, line 425: I don't think this is enough, the moment you have more nesting then only a single `}` would be replaced. I wonder if we can/should use the CSS parser directly on this value 🤔

RobinMalfait: Actually, maybe it is because if you do more nesting like:
```css
.foo {
  .bar {
    .baz {
      &:hover
    } /* <-- this is the one that will be replaced with `{@slot}}` */
  }
} /* <-- this is the main check to determine there is nesting */
```

Then the only invalid one that needs replacing is `&:hover`. The others `.foo`, `.bar` and `.baz` will already be proper CSS rules.

RobinMalfait: The case I was thinking of looks like this: `.foo { .bar { &:hover } .baz { &:hover } }`. Where you have multiple `&:hover` things. But that doesn't even work in  v3

packages/tailwindcss/src/plugin-api.ts
```diff
@@ -96,6 +103,71 @@ function buildPluginApi(
...
+            }
+          })
+        },
+        (a, z) => {
+          // Since we only define a functional variant in the group, the `kind`
+          // has to be `functional`.
+          if (a.kind !== 'functional' || z.kind !== 'functional') {
+            return 0
+          }
+          if (!a.value || !z.value) {
```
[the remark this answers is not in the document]
RobinMalfait: I _think_ our types in v3 only allow `'0'` instead of `0`, but pretty sure `0` is a value that's being used right now.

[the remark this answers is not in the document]
RobinMalfait: nvm, `a.value` and `z.value` are objects or `null`

RobinMalfait (approved):

packages/tailwindcss/src/plugin-api.ts
philipp-spiess: Yeah I worked on this with @adamwathan and we decided that this approach, while being quite limited, is enough for V3 parity. I'll add a comment with our explanation but we had something that was more powerful and it felt like total overkill especially since it isn't working in v3 either.

PR #14373, opened 2024-09-09
Add classes output switch to Tailwind CLI

bopm:
Introduce a new switch `--classes-output` to the Tailwind CLI build command to output only detected classes.

* **src/cli/index.js**
  - Add a new switch `--classes-output` to the build command.
* **src/lib/expandTailwindAtRules.js**
  - Update the `expandTailwindAtRules` function to handle the `--classes-output` switch.
  - Add logic to output only the detected classes when the switch is enabled.
* **tests/variants.test.js**
  - Add tests to verify the functionality of the `--classes-output` switch.
  - Include test cases for different scenarios with the new switch.


philipp-spiess:
Hey @bopm! Thanks for the PR and sorry for not getting back to you earlier (we're currently crunching on Tailwind v4 work to get it ready for a beta release)

Do you have some context on why you need this feature? Are you able to get a similar output by parsing the generated CSS file for class names?

bopm:
@philipp-spiess parsing generated CSS is not as straightforward as saving it while we have it. [Here is](https://github.com/full-stack-biz/turbo_material/blob/develop/lib/generators/turbo_material/install_generator.rb#L95-L105) how much I need to do completely unrelated to Tailwind or CSS project. 

So generally speaking, my problem is that as an author of Rails gem which uses Tailwind, I have to find a way to export my project classes (obviously generated with specific configuration that may not be the same as in the parent project) and then inject them into the parent application. And that may happen for the multiple libraries at the same time. So I discussed with @adamwathan how this can be achieved and this approach came up as a result of that discussion. I am still not sure if it's the best approach, but at least compute it's least wasteful. We save classes when we have them and reuse them on compilations of the host app.

bopm:
@RobinMalfait thanks for pointing me towards that, it allowed me to approach to a convention based solution for Rails. I just opened [PR](https://github.com/rails/tailwindcss-rails/pull/484) towards tailwindcss-rails with it. 

I am still a bit concerned about chicken v.s. egg flow of these includes (right now I am building list of includes starting from libs, and last one is host app CSS) in terms of theming and other stuff like that. But for now I rely on giving users ability to override library CSS in host application to resolve it. I happy to hear any thoughts on right approach with that.

PR #14407, opened 2024-09-12
Create variants for `aria`, `supports`, and `data` JS config theme keys

philipp-spiess:
This PR adds support for the `aria`, `supports`, and `data` properties found in JS config options. In v3, you could extend the theme to add more variants by using an object syntax like this:

```ts
{
   theme: {
    extend: {
      aria: {
        polite: 'live="polite"',
      },
      supports: {
        'child-combinator': 'h2 > p',
      },
      data: {
        checked: 'ui~="checked"',
      },
    },
  }
}
```

Since we no longer rely on theme variables for these variants, the way to make this work is by adding custom variants for each of these manually added variants.

### DIFF
--- a/packages/tailwindcss/src/compat/apply-config-to-theme.test.ts
+++ b/packages/tailwindcss/src/compat/apply-config-to-theme.test.ts
@@ -1,9 +1,9 @@
-import { test } from 'vitest'
+import { expect, test } from 'vitest'
 import { buildDesignSystem } from '../design-system'
 import { Theme } from '../theme'
 import { applyConfigToTheme } from './apply-config-to-theme'
 
-test('Config values can be merged into the theme', ({ expect }) => {
+test('Config values can be merged into the theme', () => {
   let theme = new Theme()
   let design = buildDesignSystem(theme)
 
--- a/packages/tailwindcss/src/compat/theme-variants.ts
+++ b/packages/tailwindcss/src/compat/theme-variants.ts
@@ -1,27 +1,11 @@
 import { rule } from '../ast'
 import type { DesignSystem } from '../design-system'
 import type { ResolvedConfig } from './config/types'
-import DefaultTheme from './default-theme'
 
 export function registerThemeVariantOverrides(config: ResolvedConfig, designSystem: DesignSystem) {
   let ariaVariants = config.theme.aria || {}
   let supportsVariants = config.theme.supports || {}
   let dataVariants = config.theme.data || {}
-
-  for (let [name, rule] of Object.entries(DefaultTheme.aria)) {
-    // `theme.aria` contains values from the default theme. We don't
-    // want to create variants for these values as these are already
-    // handled by the core utility.
-    if (ariaVariants[name] === rule) {
-      delete ariaVariants[name]
-    }
-  }
-
-  for (let [name, rule] of Object.entries(DefaultTheme.aria)) {
-    if (ariaVariants[name] === rule) {
-      delete ariaVariants[name]
-    }
-  }
 
   if (Object.keys(ariaVariants).length > 0) {
     let coreAria = designSystem.variants.get('aria')
--- a/packages/tailwindcss/src/compat/theme-variants.ts
+++ b/packages/tailwindcss/src/compat/theme-variants.ts
@@ -1,9 +1,23 @@
 import type { PluginAPI } from '../plugin-api'
+import DefaultTheme from './default-theme'
 
 export function themeVariantsPlugin({ addVariant, config }: PluginAPI) {
   let ariaVariants = config('theme.aria', {})
   let supportsVariants = config('theme.supports', {})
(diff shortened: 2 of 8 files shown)


adamwathan:
Should we actually replace the root data/supports/aria variants here maybe so all the custom ones stay grouped with the arbitrary ones in the generated CSS?

philipp-spiess:
> Should we actually replace the root data/supports/aria variants here maybe so all the custom ones stay grouped with the arbitrary ones in the generated CSS?

Can you think of ways where the grouping would be important here? I feel like it would be weird to replace core utilities that are perfectly functioning. Makes it feel brittle because we could create a wrong implementation at any of the two places now or we forget to update one, etc.

packages/tailwindcss/src/compat/apply-config-to-theme.test.ts
```diff
@@ -1,9 +1,9 @@
-import { test } from 'vitest'
+import { expect, test } from 'vitest'
 import { buildDesignSystem } from '../design-system'
 import { Theme } from '../theme'
 import { applyConfigToTheme } from './apply-config-to-theme'
 
-test('Config values can be merged into the theme', ({ expect }) => {
```
philipp-spiess, line 6: I noticed that we are inconsistent here with wether we capture `expect` from the `vi` module or from the test callback so I updated (hopefully) all places to match the more-commonly-used style

philipp-spiess:
@adamwathan Check out this tech: https://github.com/tailwindlabs/tailwindcss/pull/14407/commits/5dcb3018b2c0305c781ff7e6d5e97b0589cff36c

packages/tailwindcss/src/compat/apply-config-to-theme.test.ts
thecrypticace: I don't think we have any concurrent tests in this file but if any async tests in a given file run concurrently you have to use expect from the test function for error messages to be correct.

philipp-spiess: Hmmm maybe this is no longer needed (thanks to async local storage or so)? I got curious and checked but both cases have the same output: 

Screenshot 2024-09-13 at 11 15 01
Screenshot 2024-09-13 at 11 15 18

thecrypticace: Ah it looks like this only applies to `expect.assertions(…)` now: https://vitest.dev/api/expect.html#expect-assertions

So as long as that is not the thing being used then it's probably fine.

adamwathan:
Had a really old review I started and forgot to submit so just submit it now before I forget, probably had some outdated stuff and haven’t looked at everything yet either, on phone heh.

packages/tailwindcss/src/compat/theme-variants.ts
```diff
@@ -0,0 +1,63 @@
...
+    let coreCompounds = coreAria?.compounds
+    designSystem.variants.functional(
+      'aria',
+      (ruleNode, variant) => {
+        let value = variant.value
+        if (value && value.kind === 'named' && value.value in ariaVariants) {
+          ruleNode.nodes = [rule(`&[aria-${ariaVariants[value.value]}]`, ruleNode.nodes)]
+          return
+        }
+        return coreApplyFn?.(ruleNode, variant)
```
thecrypticace, line 22: I think this could be written like so:
```ts
let value = variant.value

if (value && value.kind === 'named' && value.value in ariaVariants) {
  return coreApplyFn?.(ruleNode, {
    ...variant,
    value: {
      kind: 'arbitrary',
      value: ariaVariants[value.value] as string,
    },
  })
}

return coreApplyFn?.(ruleNode, variant)
```

which would mean that the core variant is still the thing that decides how it's formatted which I think works the same way it did in v3.

Also means that the logic between all of these is identical _except_ for the list of values it checks.

packages/tailwindcss/src/compat/theme-variants.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (Object.keys(supportsVariants).length > 0) {
+    let coreSupports = designSystem.variants.get('supports')
+    let coreApplyFn = coreSupports?.applyFn
+    let coreCompounds = coreSupports?.compounds
+    designSystem.variants.functional(
+      'supports',
+      (ruleNode, variant) => {
+        let value = variant.value
+        if (value && value.kind === 'named' && value.value in supportsVariants) {
+          ruleNode.nodes = [rule(`@supports (${supportsVariants[value.value]})`, ruleNode.nodes)]
```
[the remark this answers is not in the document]
philipp-spiess: Good catch! I had an issue in the example for the attribute selector too. calling the core callback with an arbitrary value instead will make this work for free 👍

thecrypticace (approved):

PR #14408, opened 2024-09-12
Further clean up compatibility APIs

philipp-spiess:
This PR builds on top of #14365 and adds a few more changes we discussed during a sync on the latter PR:

- We now split `plugin-api.ts` into two files and moved it into `compat/`. One file is now defining the comat plugin API only where as another file deals with the compat hook.
- The walk for `@plugin` and `@config` is now happening inside the compat hook.

The one remaining work item is to change the `loadPlugin` and `loadConfig` APIs to a more unified `resolveModule` one that does not care on what we try to load it for. I suggest we should make this change at the same time we start working on finalizing the `tailwindcss` APIs, since a lot of things will have to be rethought then anyways.


packages/tailwindcss/src/compat/theme-variants.ts
philipp-spiess, line 1: ah that sneaked in from a wrong branch, mb

adamwathan (approved):
Looks good!

PR #14411, opened 2024-09-12
Add CSS codemods for migrating `@tailwind` directives

RobinMalfait:
This PR adds CSS codemods for migrating existing `@tailwind` directives to the new alternatives.

---

Typical default usage of `@tailwind` directives in v3.

Input:
```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```

Output:
```css
@import 'tailwindcss';
```

---

Similar as above, but always using `@import` instead of `@import` directly.

Input:
```css
@import 'tailwindcss/base';
@import 'tailwindcss/components';
```

Output:
```css
@import 'tailwindcss';
```

---

When you are _only_ using `@tailwind base`:

Input:
```css
@tailwind base;
```

Output:
```css
@import 'tailwindcss/theme' layer(theme);
```

---

When you are _only_ using `@tailwind utilities`:

Input:
```css
@tailwind utilities;
```

Output:
```css
```

---

If the default order changes (aka, `@tailwind utilities` was defined _before_ `@tailwind base`), then an additional `@layer` will be added to the top to re-define the default order. 

Input:
```css
@tailwind utilities;
@tailwind base;
```

Output:
```css
@layer theme, components, utilities, base;
@import 'tailwindcss';
```

---

When you are _only_ using `@tailwind base; @tailwind utilities;`:

Input:
```css
@tailwind base;
@tailwind utilities;
```

Output:
```css
@import 'tailwindcss';
```

We currently don't have a concept of `@tailwind components` in v4, so if you are not using `@tailwind components`, we can expand to the default `@import 'tailwindcss';` instead of the individual imports.

---

`@tailwind screens` and `@tailwind variants` are not supported/necessary in v4, so we can safely remove them.

Input:
```css
@tailwind screens;
@tailwind variants;
```

Output:
```css
```


packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      @tailwind utilities;
+      @tailwind base;
+      @tailwind components;
+    `,
+  ],
+  [
+    css`
+      @tailwind utilities;
+      @tailwind components;
+      @tailwind base;
```
thecrypticace, line 75: I'm not sure that this translation is correct — though maybe it's still fine?

In this case in v3 utilities would definitely appear _before_ base layer styles (though I can't think of any reason anyone would ever do that). `@import 'tailwindcss'` would definitely not operate that way.

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    `),
+  ).toEqual(css`
+    @import 'tailwindcss/utilities' layer(utilities);
+  `)
+})
+
+it('should migrate `@import "tailwindcss/utilities"` to an import', async () => {
+  expect(
+    await migrate(css`
+      @import 'tailwindcss/utilities';
```
thecrypticace, line 122: Did this even exist in v3? Feels weird to migrate this if it didn't.

RobinMalfait: Yep! it's for when you are using `postcss-import` where `@import` has to be at the top, so you would replace `@tailwind utilities` with `@import "tailwindcss/utilities"`. 

See: https://tailwindcss.com/docs/functions-and-directives#config

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
RobinMalfait: The order shouldn't matter anymore since we are using native layers and `@import 'tailwindcss'` defines the order at the top:
```css
/* tailwindcss/index.css */

@import './theme.css' layer(theme);
@import './preflight.css' layer(base);
@import './utilities.css' layer(utilities);
```

I think if we want to be 100% correct here, we should a modified version of `@layer theme, base, components, utilities;` at the top.

Thoughts?

RobinMalfait: If we have multiple `@layer foo, bar, baz` definitions, then the first one wins. E.g.:

https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40layer%20base%2C%20theme%2C%20components%2C%20utilities%3B%5Cn%40layer%20theme%2C%20components%2C%20base%2C%20utilities%3B%5Cn%40layer%20theme%2C%20base%2C%20components%2C%20utilities%3B%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
thecrypticace: oh right 🤦‍♂️

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
thecrypticace: I think that would be the most correct translation but uh maybe it's fine. Maybe @adamwathan has thoughts here? It could help us nudge people to order them in a more consistent way.

RobinMalfait: Added support for that in the latest commit: https://github.com/tailwindlabs/tailwindcss/pull/14411/commits/5796dadb9957d116caa19890e077f700c7fe98ca

thecrypticace (approved):

PR #14412, opened 2024-09-12
Add CSS codemods for migrating `@apply`

RobinMalfait:
This PR adds CSS codemods for migrating existing `@apply` directives to the new version. Note, the actual tooling to execute and apply the codemod will be available in another PR. This is just the codemod on its own.

---

In v4, the convention is to put the important modifier `!` at the end of the utility class instead of right before it. This makes it easier to reason about, especially when you are variants.

Input:
```css
.foo {
  @apply !flex flex-col! hover:!items-start items-center;
}
```

Output:
```css
.foo {
  @apply flex! flex-col! hover:items-start! items-center;
}
```

---

In v4 we don't support `!important` as a marker at the end of `@apply` directives. Instead, you can append the `!` to each utility class to make it `!important`.

Input:
```css
.foo {
  @apply flex flex-col !important;
}
```

Output:
```css
.foo {
  @apply flex! flex-col!;
}
```

### DIFF
--- a/packages/tailwindcss/package.json
+++ b/packages/tailwindcss/package.json
@@ -89,6 +89,7 @@
   "devDependencies": {
     "@tailwindcss/oxide": "workspace:^",
     "@types/node": "catalog:",
+    "dedent": "1.5.3",
     "lightningcss": "catalog:"
   }
 }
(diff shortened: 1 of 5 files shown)


RobinMalfait:
Closed in favor of #14434

PR #14415, opened 2024-09-13
Add support for setting `screens` in JS config

philipp-spiess:
This PR adds support for the _simple_ case of the `screens` option inside JS config paths. This allows JS configs to extend the responsive theme by adding custom breakpoints. Here's an example from our v3 docs:

```js
{
  theme: {
    screens: {
      'sm': '640px',
      // => @media (min-width: 640px) { ... }

      'md': '768px',
      // => @media (min-width: 768px) { ... }

      'lg': '1024px',
      // => @media (min-width: 1024px) { ... }

      'xl': '1280px',
      // => @media (min-width: 1280px) { ... }

      '2xl': '1536px',
      // => @media (min-width: 1536px) { ... }
    }
  }
}
```

For simple breakpoints, this will extend the core breakpoints and will work with the `min-*` and `max-*` utilities. However, we also support complex ways of setting up custom screens like this:

```js
{
  theme: {
    extend: {
      screens: {
        sm: { max: '639px' },
        md: [
          { min: '668px', max: '767px' },
          { min: '868px' },
        ],
        lg: { min: '868px' },
        xl: { min: '1024px', max: '1279px' },
        tall: { raw: '(min-height: 800px)' },
      },
    },
  },
}
```

For these complex setups, we _only_ generate the shorthand variant (e.g. `tall`) but those won't integrate within `min-*` and `max-*`. In v3, adding any of these complex configurations would omit any `min-*` and `max-*` variants.


packages/tailwindcss/src/compat/config.test.ts
```diff
@@ -978,3 +978,65 @@ describe('default font family compatibility', () => {
...
+        --breakpoint-md: 50rem;
+        --breakpoint-lg: 64rem;
+        --breakpoint-xl: 80rem;
+        --breakpoint-2xl: 96rem;
+      }
+      .sm\\:flex {
+        @media (width >= 44rem) {
+          display: flex;
+        }
+      }
```
philipp-spiess, line 1022: This currently does not work because the `sm` utility is static and is eagerly created from the Theme _before the compat layer runs_.

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,442 @@
...
+        display: flex;
+      }
+    }
+    .min-md\\:max-lg\\:underline {
+      @media (width >= 50rem) {
+        @media (width < 64rem) {
+          text-decoration-line: underline;
+        }
+      }
+    }
```
philipp-spiess, line 74: The order here will be fixed with https://github.com/tailwindlabs/tailwindcss/pull/14431

packages/tailwindcss/src/compat/screens-config.ts
```diff
@@ -0,0 +1,94 @@
...
+        if ('min' in value && !('max' in value)) {
+          insertOrder = coreOrder
+        }
+      }
+    } else {
+      continue
+    }
+
+    if (coreOrder && insertOrder === undefined) {
+      insertOrder = allocateOrderAfter(designSystem, coreOrder)
```
thecrypticace, line 51: Is it possible for us to do this all at one time at the end rather than  after every inserted variant?

We're definitely doing more traversals and object replacements than we need to but if it proves to be too complex then this is fine for now.

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+    .mini\\:flex {
+      @media (width >= 40rem) {
+        display: flex;
+      }
+    }
+    .sm\\:flex {
+      @media (width >= 40rem) {
+        display: flex;
+      }
+    }
```
thecrypticace, line 317: I think this is wrong. The `sm` variant shouldn't exist right — because the theme wasn't extended? (and as such the user is explicitly saying they don't want "default" screens values)

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+          },
+        },
+      },
+    }),
+  })
+
+  expect(
+    compiler.build([
+      'sm:flex',
+      'md:flex',
```
thecrypticace, line 393: Pretty sure these _should_ exist because we're using `theme.extend`

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+      .md\\:flex {
+        @media (min-width: 668px and max-width: 767px), (min-width: 868px) {
+          display: flex;
+        }
+      }
+      .sm\\:flex {
+        @media (max-width: 639px) {
+          display: flex;
+        }
+      }
```
thecrypticace, line 511: These feel like they're ordered backwards

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+      @config "./config.js";
+      @tailwind utilities;
+    `
+
+    let compiler = await compile(input, {
+      loadConfig: async () => ({
+        theme: {
+          extend: {
+            screens: {
+              sm: { max: '639px' },
```
thecrypticace, line 456: I feel like, if we can, we should make `max-sm` and `sm` the same when given a config like this.

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+    })
+
+    expect(
+      compiler.build([
+        'sm:flex',
+        'md:flex',
+        'portrait:flex',
+        'min-sm:flex',
+        'min-md:flex',
+        'min-portrait:flex',
```
thecrypticace, line 551: This should be a separate build + assertion that it equals `''`

packages/tailwindcss/src/compat/screens-config.test.ts
```diff
@@ -0,0 +1,592 @@
...
+        'sm:flex',
+        'md:flex',
+        'lg:flex',
+        'xl:flex',
+        'tall:flex',
+        'min-sm:flex',
+        'min-md:flex',
+        'min-lg:flex',
+        'min-xl:flex',
+        'min-tall:flex',
```
thecrypticace, line 482: This should be a separate build + assertion that it equals `''`

thecrypticace (dismissed):

packages/tailwindcss/src/compat/screens-config.ts
philipp-spiess: We have to do it once for every utility that is inserted this way. Each one will handed their own order. I do need a test for this since I think it might now be adding it in inverse insertion order.

packages/tailwindcss/src/compat/screens-config.test.ts
philipp-spiess: Yeah that something me and @adamwathan chatted about and we decided that we want this behavior instead since it’s also how `—color` extension work right now. We append the user resolved theme values without throwing anyway anything that is defined in the CSS.

packages/tailwindcss/src/compat/screens-config.test.ts
philipp-spiess: That’s not how v3 works though iirc so might not make sense to add support for it in v4?

thecrypticace: ah yeah right, okay then.

packages/tailwindcss/src/compat/screens-config.test.ts
thecrypticace: 👍

packages/tailwindcss/src/compat/screens-config.test.ts
philipp-spiess: They are lol we need to keep track of the last insert order and then insert after that instead 🫠 I guess we can also iterate over the list in reverse which is prob an easier fix

packages/tailwindcss/src/compat/screens-config.ts
thecrypticace: Yeah it would make sense that this is the reason some things are going being inserted in reverse order. I was thinking of something like this:

- Get a list of all the variants _after_ `min` (and their compare fns)
- Insert our appropriate screens into the `min` group
- Insert other screens after the `min` group
- Move the previously recorded variants past the last variant we just inserted (*only* if there is one)

I think this would work but I'm not sure if the separate compareFn map complicates that.

philipp-spiess: Hm I’m not sure I follow, I think we can keep doing what we do right now but only loop over the variants in reverse order, this way we:

- Still insert appropriate screens into the `min` group. For that, the order does not matter since it'll use the compare function
- Insert complex variants _after_ the `min` group. But since we loop in reverse order, we will will always move the previously inserted complex variant down. 

I pushed a commit that changes this and it fixes the order.

thecrypticace (approved):

PR #14416, opened 2024-09-13
Add ability to parse SCSS syntax

RobinMalfait:
This PR adds support for parsing SCSS syntax. This will help with codemods.

Note: this does not "run" or "execute" SCSS code, it only parses it. If you are using SCSS and any of its features that aren't native to CSS (e.g.: $variables, mixins, functions, …), you will need to use a SCSS compiler to transform the SCSS.

Note: while this parse CSS comments and SCSS single-line comments, they are not in the AST yet. This will be a a separate PR for when we need it.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@ crates/node/index.d.ts
 crates/node/index.js
 .next
 .fingerprint
+__snapshots__/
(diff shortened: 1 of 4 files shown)


philipp-spiess (dismissed):
Awesomesauce

PR #14431, opened 2024-09-16
Fix stacking variant order when variants inside a group are treated as equal

philipp-spiess:
This PR fixes an issue with the order of CSS when using stacked variants when two variants have the same order (as defined by the custom comperator function).

## The problem

Take, for example, our breakpoint variants. Those are split into `max-*` variants and a group containing all `min-*` variants as well as the unprefixed static ones (e.g. `lg`, `sm`).

We currently define a custom sort order for all breakpoints variants that will compare their order based on the resolved value provided. So if you define `--breakpoint-sm: 100px` and `--breakpoint-lg: 200px`, we first check if both breakpoints have the same unit and then we rank based on the numerical value, making `sm` appear before `lg`.

But since the `min-*` variant and the `sm` variant share the same group, this also means that `min-sm` and `sm` as well as `min-lg` and `lg` will always have the same order (which makes sense—they also have the exact same CSS they generate!)

The issue now arises when you use these together with variant stacking. So, say you want to stack the two variants `max-lg:min-sm`. We always want stacked variants to appear _after_ their non-stacked individual parts (since they are more specific). To do this right now, we generate a bitfield based on the variant order. If you have four variants like this:

| Order | Variant |
| 0  | `max-lg`  |
| 1  | `max-sm`  |
| 2  | `min-sm`  |
| 3  | `min-lg`  |

We will assign one bit for each used variant starting from the lowest bit, so for the stack `max-lg:min-sm` we will set the bitfield to `0101` and those for the individual variants would result in `0100` (for `min-sm`) and `0001` (for `max-lg`). We then convert this bitfield to a number and order based on that number. This ensures that the stack always sorts higher.

The issue now arises from the fact that the variant order also include the unprefixed variants for a breakpoint. So in our case of `lg` and `sm`, the full list would look like this:

| Order | Variant |
| 0  | `max-lg`  |
| 1  | `max-sm`  |
| 2  | `min-sm`  |
| 3  | `sm`  |
| 4  | `min-lg`  |
| 5  | `lg`  |

This logic now breaks when you start to compute a stack for something like `max-lg:min-lg` _while also using the `lg` utility:

| Stack | Bitmap | Integer Value |
| ------------- | ------------- | ------------- |
| `max-lg:min-lg` | `010001` | 17 |
| `lg` | `100000` | 18 |

As you can see here, the sole `lg` variant will now sort higher than the compound of `max-lg:min-lg`. That's not something we want!

## Proposed solution

To fix this, we need to encode the information of _same_ variant order somehow. A single array like the example above is not sufficient for this, since it will remove the information of the similar sort order. Instead, we now computed a list of nested arrays for the order lookup that will combine variants of similar values (while keeping the order the same). So from the 6 item array above, we now have the following nested array:

| Order | Variant |
| 0  | [`max-lg`]  |
| 1  | [`max-sm`]  |
| 2  | [`min-sm`, `sm`]  |
| 3  | [`min-lg`, `lg`]  |

When we use the first layer index for the bitfield, we can now see how this solves the issue:

| Stack | Bitmap | Integer Value |
| ------------- | ------------- | ------------- |
| `max-lg:min-lg` | `1001` | 9 |
| `lg` | `1000` | 8 |

That's pretty-much it! There are a few other changes in this PR that mostly handles with a small regression by this change where now, named `group` variants and unnamed `group` variants would now have the same order (something that was undefined behavior before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Support `borderRadius.*` as an alias for `--radius-*` when using dot notation inside the `theme()` function ([#14436](https://github.com/tailwindlabs/tailwindcss/pull/14436))
+### Fixed
+
+- Fix an issue with the CSS order when stacking variants. ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
 
 ## [4.0.0-alpha.24] - 2024-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix an issue with the CSS order when stacking variants. ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
+- Ensure individual variants from groups are always sorted earlier than stacked variants from the same groups ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
 
 ## [4.0.0-alpha.24] - 2024-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Support `borderRadius.*` as an alias for `--radius-*` when using dot notation inside the `theme()` function ([#14436](https://github.com/tailwindlabs/tailwindcss/pull/14436))
-### Fixed
-
-- Fix an issue with the CSS order when stacking variants. ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
 - Ensure individual variants from groups are always sorted earlier than stacked variants from the same groups ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
 
 ## [4.0.0-alpha.24] - 2024-09-11
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -30,10 +30,7 @@ export function compileCandidates(
     matches.set(rawCandidate, candidates)
   }
 
-  // Sort the variants
-  let variants = sortAndGroup(designSystem.getUsedVariants(), (a, z) =>
-    designSystem.variants.compare(a, z),
-  )
+  let variants = designSystem.getUsedVariantGroups()
 
   // Create the AST
   for (let [rawCandidate, candidates] of matches) {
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Fix an issue with the CSS order when stacking variants. ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
```
```suggestion
- Ensure individual variants from groups are always sorted earlier than stacked variants from the same groups ([#14431](https://github.com/tailwindlabs/tailwindcss/pull/14431))
```

packages/tailwindcss/src/compile.ts
```diff
@@ -321,3 +322,38 @@ function getPropertySort(nodes: AstNode[]) {
 
   return Array.from(propertySort).sort((a, z) => a - z)
 }
+
+/**
+ * Sort an array of entries into an array-of-sets. Similar entries (where the
+ * sort function returns 0) are grouped together. The sort function is stable
+ * and uses bubble-sort
```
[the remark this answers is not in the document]
adamwathan: Is the comment about bubble sort useful if we are just using `.sort()` and not implementing the algorithm ourselves?

Also wonder if we should use `toSorted()`/make a copy since this function has a return value — is it weird that it has the side effect of sorting one of the parameters in place?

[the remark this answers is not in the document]
thecrypticace: If we're using `sort` then it's *not* bubble sort — it's most likely quick sort but I think the comment about the sort function should just go 🤷 

The only use of this passes in `getUsedVariants(…)` which is a fresh array every time so sorting in place is more efficient.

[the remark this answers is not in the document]
philipp-spiess: > Is the comment about bubble sort useful if we are just using .sort() and not implementing the algorithm ourselves?

No that's a leftover, mb!

> Also wonder if we should use toSorted()/make a copy since this function has a return value — is it weird that it has the side effect of sorting one of the parameters in place?

Yeah Jordan and me were also thinking. I did have it operate on a clone first but given that this is pretty hot code we thought that the optimization might make sense. That said, since its a function now it feels like something that another callsite might want to use, at which point this will become confusing yeah

packages/tailwindcss/src/compile.ts
```diff
@@ -31,9 +31,9 @@ export function compileCandidates(
   }
 
   // Sort the variants
-  let variants = designSystem.getUsedVariants().sort((a, z) => {
-    return designSystem.variants.compare(a, z)
-  })
+  let variants = sortAndGroup(designSystem.getUsedVariants(), (a, z) =>
+    designSystem.variants.compare(a, z),
+  )
```
thecrypticace, line 36: I think we should move this sorting into `designSystem.getUsedVariants()`

Maybe the name should be something like `getUsedVariantGroups()` or something then as well?

packages/tailwindcss/src/design-system.ts
```diff
@@ -77,8 +79,29 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
...
+        if (prevVariant !== undefined && this.variants.compare(prevVariant, variant) !== 0) {
+          index++
+        }
+
+        order.set(variant, index)
+        prevVariant = variant
+      }
+
+      return order
     },
```
philipp-spiess, line 105: @thecrypticace came up with this nugget: We can use a`Map<Variant, number>` which helps us:

- Avoid creating _n_ sets
- Avoid doing a linear search in the compile function for constant lookups

packages/tailwindcss/src/design-system.ts
```diff
@@ -77,8 +79,29 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
...
+        // so it goes into a new group
+        if (prevVariant !== undefined && this.variants.compare(prevVariant, variant) !== 0) {
+          index++
+        }
+
+        order.set(variant, index)
+        prevVariant = variant
+      }
+
+      return order
```
philipp-spiess, line 104: @thecrypticace came up with this nugget: We can use a`Map<Variant, number>` which helps us:

- Avoid creating _n_ sets
- Avoid doing a linear search in the compile function for constant lookups

packages/tailwindcss/src/compile.ts
```diff
@@ -51,7 +48,9 @@ export function compileCandidates(
         // variants used.
         let variantOrder = 0n
         for (let variant of candidate.variants) {
-          variantOrder |= 1n << BigInt(variants.indexOf(variant))
+          let order = variantOrderMap.get(variant)
+          if (order === undefined) continue
```
thecrypticace, line 52: We can use `!` instead of the if check because the variants inside candidate.variants are guaranteed to be present in getVariantOrder().

thecrypticace (approved):

PR #14434, opened 2024-09-16
Add initial codemod tooling

RobinMalfait:
This PR adds some initial tooling for codemods. We are currently only interested in migrating CSS files, so we will be using PostCSS under the hood to do this. This PR also implements the "migrate `@apply`" codemod from #14412.

The usage will look like this:

```sh
npx @tailwindcss/upgrade
```

You can pass in CSS files to transform as arguments:

```sh
npx @tailwindcss/upgrade src/**/*.css
```

But, if none are provided, it will search for CSS files in the current directory and its subdirectories.

```
≈ tailwindcss v4.0.0-alpha.24

│ No files provided. Searching for CSS files in the current
│ directory and its subdirectories…

│ Migration complete. Verify the changes and commit them to
│ your repository.
```

The tooling also requires the Git repository to be in a clean state. This is a common convention to ensure that everything is undo-able. If we detect that the git repository is dirty, we will abort the migration.

```
≈ tailwindcss v4.0.0-alpha.24

│ Git directory is not clean. Please stash or commit your
│ changes before migrating.

│ You may use the `--force` flag to override this safety
│ check.
```

---

This PR alsoo adds CSS codemods for migrating existing `@apply` directives to the new version.

---

In v4, the convention is to put the important modifier `!` at the end of the utility class instead of right before it. This makes it easier to reason about, especially when you are variants.

Input:
```css
.foo {
  @apply !flex flex-col! hover:!items-start items-center;
}
```

Output:
```css
.foo {
  @apply flex! flex-col! hover:items-start! items-center;
}
```

---

In v4 we don't support `!important` as a marker at the end of `@apply` directives. Instead, you can append the `!` to each utility class to make it `!important`.

Input:
```css
.foo {
  @apply flex flex-col !important;
}
```

Output:
```css
.foo {
  @apply flex! flex-col!;
}
```


packages/@tailwindcss-upgrade/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@tailwindcss/upgrade",
```
philipp-spiess, line 2: Nit: The PR description mentions `@tailwindcss/codemod` and this here uses `@tailwindcss/upgrade` 😅

philipp-spiess (approved):

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  // Discover CSS files in case no files were provided
+  if (files.length === 0) {
+    info(
+      'No files provided. Searching for CSS files in the current directory and its subdirectories…',
+    )
+
+    files = await fastGlob(['**/*.css'], {
+      absolute: true,
+      ignore: ['**/node_modules', '**/vendor'],
```
philipp-spiess, line 51: Wonder if we should instead use the `.gitignore` info here, I remember that we had a few people who had `vendor` folders when we added the content-path-warning and we eventually had to revert it.

integrations/cli/codemods.test.ts
philipp-spiess, line 1: Should we call this file `upgrade.test.ts` so it's clear that this is for the upgrade CLI?

philipp-spiess (approved):

integrations/cli/codemods.test.ts
RobinMalfait: I think that might be better yeah, good catch!

packages/@tailwindcss-upgrade/src/index.ts
RobinMalfait: Hmm yeah, let's see what I can find to solve this!

RobinMalfait: Found globby https://github.com/sindresorhus/globby, based on fast-glob and has support for this out of the box

philipp-spiess: :chef-kiss:

PR #14438, opened 2024-09-16
Refactor some screen compatibility code

thecrypticace:
This adds support for the `screen(…)` CSS function which can be used to build the media queries for a given breakpoint:

Input:

```css
@theme {
  --breakpoint-sm: 640px;
}

@media screen(sm) {
  .foo {
    color: red;
  }
}
```

output:
  
```css
@media (width >= 640px) {
  .foo {
    color: red;
  }
}
```

We also support more advanced `screens` configurations from configuration files (mainly for backwards) compat:
```ts
export default {
  theme: {
    screens: {
      sm: '640px',
      md: { min: '768px' },
      lg: { min: '1024px', max: '1280px' },
      sidebar: [
        { min: '640px', max: '768px' },
        { min: '1024px' },
      ],
    }
  }
}
```

```css
@media screen(sm) {
  .foo {
    color: red;
  }
}

@media screen(md) {
  .foo {
    color: blue;
  }
}

@media screen(lg) {
  .foo {
    color: green;
  }
}

@media screen(sidebar) {
  .foo {
    color: yellow;
  }
}
```

Output:

```css
@media (width >= 640px) {
  .foo {
    color: red;
  }
}

@media (width >= 768px) {
  .foo {
    color: blue;
  }
}

@media (1280px >= width >= 1024px) {
  .foo {
    color: green;
  }
}

@media (768px >= width >= 640px), (width >= 1024px) {
  .foo {
    color: yellow;
  }
}
```

### DIFF
--- a/packages/tailwindcss/src/compat/config/resolve-config.ts
+++ b/packages/tailwindcss/src/compat/config/resolve-config.ts
@@ -203,4 +203,19 @@ function mergeTheme(ctx: ResolutionContext) {
   for (let key in ctx.theme) {
     ctx.theme[key] = resolveValue(ctx.theme[key])
   }
+
+  // Turn {min: '123px'} into '123px' in screens
+  if (ctx.theme.screens && typeof ctx.theme.screens === 'object') {
+    for (let key of Object.keys(ctx.theme.screens)) {
+      let screen = ctx.theme.screens[key]
+      if (!screen) continue
+      if (typeof screen !== 'object') continue
+
+      if ('raw' in screen) continue
+      if ('max' in screen) continue
+      if (!('min' in screen)) continue
+
+      ctx.theme.screens[key] = screen.min
+    }
+  }
 }
--- a/packages/tailwindcss/src/compat/screens-config.test.ts
+++ b/packages/tailwindcss/src/compat/screens-config.test.ts
@@ -457,7 +457,7 @@ describe('complex screen configs', () => {
               md: [
                 //
                 { min: '668px', max: '767px' },
-                { min: '868px' },
+                '868px',
               ],
               lg: { min: '868px' },
               xl: { min: '1024px', max: '1279px' },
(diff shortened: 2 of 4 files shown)


thecrypticace:
@philipp-spiess `[@media(screen(sm))]:flex` doesn't work however `[@media_screen(sm)]:flex` does work — presumably because of the space. We should take a look at the value parser.

packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -757,6 +757,101 @@ describe('theme function', () => {
...
+
+        @media (width >= 48rem) {
+          .foo {
+            display: flex;
+          }
+        }"
+      `)
+    })
+
+    test('@media screen(advanced)', async () => {
```
thecrypticace, line 813: None of these are going to pass at least until we do the complex screen config stuff in #14415. Probably will still need more work after that.

packages/tailwindcss/src/compat/apply-compat-hooks.ts
```diff
@@ -191,6 +191,13 @@ export async function applyCompatibilityHooks({
   registerThemeVariantOverrides(resolvedUserConfig, designSystem)
   registerScreensConfig(resolvedUserConfig, designSystem)
 
+  // Replace `resolveBreakpointQuery` with a version that can handle more
+  // complex breakpoint queries like objects, tuples, and custom media queries
+  // that are present in JS theme configurations.
+  designSystem.resolveBreakpointQuery = function resolveBreakpointQuery(name: string) {
+    return buildMediaQuery(pluginApi.theme(`screens.${name}`))
+  }
```
philipp-spiess, line 199: This is great!

philipp-spiess (commented):
It's still not great that we now have the `screen` function checks in core but this way the legacy implementation is at least contained.

Just brainstorming here, but one way to clean this up even further is by adding another walk for function replacement in the compat layer _and_ we'd need a way for the compat layer to hook into the build calls (since we do a function substitution there first).

Other idea is that we do the function replacement after the core layer and have a dictionary of functions that we can register (just like our `Utilities` and `Variants` maps). Then, the compat layer could push custom functions into it.

thecrypticace:
We've decided to do this as a code mod instead — I'll be landing some of the refactoring tweaks in a separate PR.

PR #14448, opened 2024-09-17
Add support for `important` in v4

thecrypticace:
This PR allows modifying utility output by wrapping all utility declarations in a custom selector or marking all utility declarations as `!important`. This is the v4 equivalent to the `important` option in the `tailwind.config.js` file.

## Mark all utility declarations as `!important`

To add `!important` to all utility declarations, you add an `important` flag after the `@import` statement for `tailwindcss` or `tailwindcss/utilities`:

```css
/** Importing `tailwindcss` */
@import "tailwindcss" important;

/** Importing the utilities directly */
@import "tailwindcss/utilities" important;
```

Example utility output:

```css
.mx-auto {
  margin-left: auto !important;
  margin-right: auto !important;
}

.text-center {
  text-align: center !important;
}
```

This is equivalent to adding `important: true` to the `tailwind.config.js` file — which is still supported for backwards compatibility.

## Wrap all utility declarations in a custom selector

For backwards compatibility, if you need to nest all utilities in an `#app` selector you add `important: "#app"` to your config file:

```js
export default {
  important: "#app",
  // …
}
```

**This _does not_ bring back support for the `respectImportant` flag in `addUtilities` / `matchUtilities`.**


packages/tailwindcss/src/index.ts
```diff
@@ -231,6 +232,19 @@ async function parseCss(
       return WalkAction.Skip
     }
 
+    // Drop instances of `@media selector(…)`
+    //
+    // We support `@import "tailwindcss" selector(ident)` as a way to
```
thecrypticace, line 238: not `ident`

packages/tailwindcss/src/design-system.ts
```diff
@@ -13,6 +13,8 @@ export type DesignSystem = {
   utilities: Utilities
   variants: Variants
 
+  important: string | boolean
```
philipp-spiess, line 16: Is this overloading of the `important` option a v3-ism that creeps in? Wondering if this should be two separate options instead.

Is there a world where you'd maybe want both (I guess the answer is no because we would rather have people not use these at all?).

philipp-spiess (commented):
This all seems easy enough, just one question inline. Happy to stamp it but I feel like this could benefit from someone else's 👀 too!

packages/tailwindcss/src/design-system.ts
thecrypticace: Yeah it is kinda a v3-ism. I'd be cool with them being separate options. Originally I wanted selector to operate like `#app { @tailwind utilities; }` but after talking with @RobinMalfait we figured it's probably useful for intellisense to actually see that the important selector is in use — like how it works in v3.

But yeah in an ideal world people don't use these at all unless they absolutely have to.

packages/tailwindcss/src/design-system.ts
```diff
@@ -45,6 +47,12 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
     utilities,
     variants,
 
+    // How to mark important utilities
+    // - wrap with a selector (any string)
+    // - add an !important (true)
+    // - do nothing (false)
```
RobinMalfait, line 53: I think we can improve this comment a bit. Took a stab at it, but not sure if it is much better:

```suggestion
    // Strategies:
    // - `string` — wrap each rule with the provided selector
    // - `true`   — add `!important` to every declaration
    // - `false`  — do nothing
```

packages/tailwindcss/src/important.test.ts
```diff
@@ -0,0 +1,64 @@
+import { expect, test } from 'vitest'
+import { compile } from '.'
+
+const css = String.raw
+
+test('Utilities can be wrapped in a selector', async () => {
+  // This is the v4 equivalent of `important: "#app"` from v3
+  let input = css`
+    @import 'tailwindcss/utilities' selector(#app);
```
RobinMalfait, line 9: Was this the final syntax we wanted? I'm fine with it, but it's not super clear what this does (could be a documentation problem).

But it's almost like you want:

If you are talking about the feature:
```suggestion
    @import 'tailwindcss/utilities' important(#app);
```

If you are talking about what it does:
```suggestion
    @import 'tailwindcss/utilities' wrapper(#app);
```

Don't like either option so think `selector` is fine, but still writing this down just in case.

RobinMalfait (approved):

packages/tailwindcss/src/design-system.ts
thecrypticace: I ended up removing this comment after splitting them

packages/tailwindcss/src/important.test.ts
thecrypticace: `selector(…)` was Adam's initial thought but he wasn't really set on it I don't think. We looked around for possible existing discussion and the closest thing we could find was scope(…) but that will: 1. be an actual thing eventually; 2. works a bit differently.

I don't think spelling this `important(#app)` is the right move — the fact that it's the equivalent of part of the v3 important feature is _fine_ but we'd probably name it differently had that feature not existed (and still wanted to add it).

I don't mind `wrapper(#app)` though.

thecrypticace: @adamwathan do you have an opinion on this ☝️

thecrypticace:
Recording decisions from our call:
- Eliminate the `selector(#app)` feature from the utilities import because we can't quite decide on the best naming or implementation for that API. A native CSS version will eventually be available called `scope(…)` anyway which would be a far better solution for end users.
- We'll require users to wrap manually `@tailwind utilities` with a selector
- We don't care of intellisense does _not_ show the wrapping selector
- Backwards compat will look for the `@tailwind utilities` node in the AST and
  wrap it with `#app` selector

RobinMalfait (approved):

thecrypticace:
@benface Yeah so we ended up doing it only for backwards compatibility purposes to hopefully be less breaking for the projects that are still using this feature. Ideally though users would switch entirely to relying on cascade layers and remove this from their projects.

benface:
> Ideally though users would switch entirely to relying on cascade layers and remove this from their projects.

Well, they _have_ to, as there's no way in Tailwind 4 to render utilities _outside_ the `utilities` layer, right? Nor do I think there should be... but I would love more control over sub-layers, to make some utilities more "important" than others.

thecrypticace:
You definitely can. You just have to inline the imports. We rely on CSS's native layer features including `layer(…)` on `@import` so to drop the layers entirely you can write this:

```
@import "tailwindcss/theme";
@import "tailwindcss/preflight";
@import "tailwindcss/utiltiies";
```

To keep the layers you do this (this is exactly what is in `index.css`):

```

@import './theme.css' layer(theme);
@import './preflight.css' layer(base);
@import './utilities.css' layer(utilities);
```

Juneezee:
@thecrypticace  I think it would be better if this change is included in the V4 upgrade guide (https://tailwindcss.com/docs/upgrade-guide) and in the section for the `@import` directive (https://tailwindcss.com/docs/functions-and-directives#import-directive).

merijnponzo:
Can anyone explain if it's still possible to use a important strategy?

I added '#app' to the config's 'important' property but it doesn't seem to generate #app .p-4 for example.

I can image this is for more specific usecases but our Wordpress plugin uses it for example to render tailwind css for specific blocks. 

It's also great to give a very specific priority, instead of just overrule '!' everything

Since id's have a higher prio then classes #block-1234 .tw-4 overruled the editor styles wrapper of wordpress, but you can still add !important on a higher level. 

So
```
@import "tailwindcss" important(#app);
```

Would be perfect, or even combine it with prefix

```
@import "tailwindcss" important(#app) prefix(tw);
```

That generates
#app .tw-pw-4 {

PR #14460, opened 2024-09-19
CI speed-up by splitting into different jobs

n-batalha:
Please ignore this PR, checking some quick improvements re: CI https://x.com/adamwathan/status/1836761788936798453

Hope it's ok, I can't push branches to your repo so I'm opening a draft PR to trigger your CI, not replicable locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "postbuild": "node ./scripts/pack-packages.mjs",
     "dev": "turbo dev --filter=!./playgrounds/*",
     "test": "cargo test && vitest run",
-    "test:integrations": "vitest --root=./integrations --no-file-parallelism",
+    "test:integrations": "vitest --root=./integrations --pool=threads",
     "test:ui": "pnpm run --filter=tailwindcss test:ui",
     "tdd": "vitest",
     "bench": "vitest bench",


philipp-spiess:
Haha, thanks for looking into it! Let me know if I can help somehow. We know the biggest issue is that our integration tests do `pnpm install` for a bunch of times but we haven't found a good way to speed this up, even though most examples are pretty similar. Another issue is with running test suites in parallel is that free port assignment for the dev server doesn't work reliably (two parallel tests could request the same port and it would become flaky).

n-batalha:
> We know the biggest issue is that our integration tests do `pnpm install` for a bunch of times but we haven't found a good way to speed this up, even though most examples are pretty similar. Another issue is with running test suites in parallel is that free port assignment for the dev server doesn't work reliably (two parallel tests could request the same port and it would become flaky).

@philipp-spiess is the dev server just for playwright? There is another type of parallelism, see below. As it means the whole job is as slow as the slowest "sub-job" (Windows integration tests it seems):

image

https://github.com/n-batalha/tailwindcss/blob/vitest-speed-improvement-test-2/.github/workflows/ci.yml
https://github.com/n-batalha/tailwindcss/actions/runs/10945213206/job/30388912690

Then next step being to improve that job specifically

philipp-spiess:
> @philipp-spiess is the dev server just for playwright? There is another type of parallelism, see below. As it means the whole workflow is as slow as the slowest "job" (Windows integration tests it seems):

No what I mean are the integration tests that start a `vite` process with a dev server etc. We access these with `fetch` directly right now!

PR #14465, opened 2024-09-19
Refactor some screen compatibility code

thecrypticace:
These changes were extracted from our work on a `screen()` function for CSS — we've decided to move those changes to a code mod instead of implementing support for `screen()` in the compiler — but the refactoring around the changes still makes sense to do so I'm landing that here separately.


philipp-spiess (approved):
Ah, yep, thanks!

PR #14477, opened 2024-09-20
Add missing theme props

lubomirblazekcz:
There are some default props missing in @theme, these are definitely handy for use as CSS variables in custom CSS too

* Added missing font-weights props
* Added missing z-index props
* Added missing radius full prop


lubomirblazekcz:
Ok, makes sense 👍

PR #14501, opened 2024-09-23
Add support for prefixes

thecrypticace:
This PR adds support for requiring a custom prefix on utility classes.

Prefixes work a bit differently in v4 than they did in v3:
- They look like a custom variant: `tw:bg-white`
- It is always first in a utility — even before other variants: `tw:hover:bg-white`
- It is required on **all** utility classes — even arbitrary properties: `tw:[color:red]`
- Prefixes also apply to generated CSS variables which will be separated by a dash: `--tw-color-white: #fff;`
- Only alpha (a-z) characters are allowed in a prefix — so no `#tw#` or `__` or similar prefixes are allowed

To configure a prefix you can use add `prefix(tw)` to your theme or when importing Tailwind CSS like so:

```css
/* when importing `tailwindcss` */

/* when importing the theme separately */
@import 'tailwindcss/theme' prefix(tw);

/* or when using an entirely custom theme */
@theme prefix(tw) {
  --color-white: #fff;
  --breakpoint-sm: 640px;
  /* … */
}
```

This will configure Tailwind CSS to require a prefix on all utility classes when used in HTML:

```html
<div class="tw:bg-white tw:p-4">
  This will have a white background and 4 units of padding.
</div>

<div class="bg-white p-4">
  This will not because the prefix is missing.
</div>
```

and when used in CSS via `@apply`:
```css
.my-class {
  @apply tw:bg-white tw:p-4;
}
```

Additionally, the prefix will be added to the generated CSS variables. You **do not** need to prefix the variables in the `@theme` block yourself — Tailwind CSS handles this automatically.

```css
:root {
  --tw-color-white: #fff;
  --tw-breakpoint-sm: 640px;
}
```

A prefix is not necessary when using the `theme(…)` function in your CSS or JS given that plugins will not know what the current prefix is and must work with or without a prefix:

```css
.my-class {
  color: theme(--color-white);
}
```

However, because the variables themselves are prefixed when outputting the CSS, you **do** need to prefix the variables when using `var(…)` in your CSS:

```css
.my-class {
  color: var(--tw-color-white);
}
```

If you want to customize the prefix itself change `tw` to something else:

```css
/* my:underline, my:hover:bg-red-500, etc… */
@import 'tailwindcss' prefix(my);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add support for prefixes ([#14501](https://github.com/tailwindlabs/tailwindcss/pull/14501))
+
 ### Fixed
 
 - _Experimental_: Improve codemod output, keep CSS after last Tailwind directive unlayered ([#14512](https://github.com/tailwindlabs/tailwindcss/pull/14512))
--- a/packages/tailwindcss/src/compat/prefix.test.ts
+++ b/packages/tailwindcss/src/compat/prefix.test.ts
@@ -17,9 +17,8 @@ test('utilities must be prefixed', async () => {
   let compiler = await compile(input)
 
   // Prefixed utilities are generated
-  expect(
-    compiler.build(['tw:underline', 'tw:hover:line-through', 'tw:custom']),
-  ).toMatchInlineSnapshot(`
+  expect(compiler.build(['tw:underline', 'tw:hover:line-through', 'tw:custom']))
+    .toMatchInlineSnapshot(`
     ".tw\\:custom {
       color: red;
     }
@@ -268,4 +267,14 @@ test('a prefix can be configured via @import prefix(…)', async () => {
   })
 
   expect(compiler.build(['underline', 'hover:line-through', 'custom'])).toEqual('')
+})
+
+test('a prefix must be letters only', async () => {
+  let input = css`
+    @theme reference prefix(__);
+  `
+
+  await expect(() => compile(input)).rejects.toThrowErrorMatchingInlineSnapshot(
+    `[Error: The prefix "__" is invalid. Prefixes must be lowercase ASCII letters (a-z) only.]`,
+  )
 })
--- a/packages/tailwindcss/src/compat/prefix.test.ts
+++ b/packages/tailwindcss/src/compat/prefix.test.ts
@@ -97,7 +97,7 @@ test('CSS variables output by the theme are prefixed', async () => {
   `)
 })
 
-test('CSS theme functions do not need to use the prefix', async () => {
+test('CSS theme functions do not use the prefix', async () => {
   let compiler = await compile(css`
     @theme prefix(tw) {
       --color-red: #f00;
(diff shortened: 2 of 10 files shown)


thecrypticace:
@RobinMalfait Should we switch to the new prefix syntax (where it is actually a prefix lol) in this PR and do the codemod later — or land the old prefix version and then switch to the new one when we do codemod?

RobinMalfait:
@thecrypticace since we didn't have prefix support before this in v4, I would say use the new syntax only and introduce a codemod later (@philipp-spiess is already making good progress on the infra for that). 

Don't think it's worth / necessary to ship support for the old prefix. Because, if we have a codemod, then we would remove support for the old prefix again which makes it a breaking change (it's still alpha so it's fine, but still.)

thecrypticace:
- [ ] Add one test to candidate.test for prefixes
- [ ] Make sure `var(--…)` is properly prefixed in generated utilities
- [ ] If `config.prefix` ends in dash we remove it and warn
- [ ] Make sure prefixes are validated — lowercase alpha characters only (a-z)

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Added
+
+- Add support for utility prefixes in v4 ([#14501](https://github.com/tailwindlabs/tailwindcss/pull/14501))
```
philipp-spiess, line 12: This is mega nit but now that the prefix is not longer only in front of the utility, maybe should maybe not call it that. Also I think the version identifier is already in the changelog so might as well drop that:

```suggestion
- Add support for prefixes ([#14501](https://github.com/tailwindlabs/tailwindcss/pull/14501))
```

packages/tailwindcss/src/compat/prefix.test.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+    .tw\\:\\[color\\:theme\\(--color-red\\)\\] {
+      color: #f00;
+    }
+    .tw\\:text-\\[theme\\(--color-red\\)\\] {
+      color: #f00;
+    }
+    "
+  `)
+})
```
philipp-spiess, line 126: Does "does not need" here mean that it still _can_? In either case, we should add an example for that in the test to document the current behavior (I think it doesn't from the code review but maybe I overlooked something?)

philipp-spiess (commented):
Looks good to me overall! One question that could need some clarification is what should the `theme()` function do when _it does get a prefix fed into it_, see the inline comment on the new test.

packages/tailwindcss/src/compat/prefix.test.ts
thecrypticace: Ah yeah I actually don't know if we should even allow it. Seems a bit weird but maybe it's fine?

thecrypticace: Given that in the theme you _don't_ write variables with the prefixes I'm kinda includes to say they should not.

philipp-spiess: Yeah I am inclined to agree. Maybe then the test name should remove the "not need to" part:

```suggestion
test('CSS theme functions do not use the prefix', async () => {
  let compiler = await compile(css`
    @theme prefix(tw) {
      --color-red: #f00;
      --color-green: #0f0;
      --breakpoint-sm: 640px;
    }

    @tailwind utilities;
  `)

  expect(compiler.build(['tw:[color:theme(--color-red)]', 'tw:text-[theme(--color-red)]']))
    .toMatchInlineSnapshot(`
    ":root {
      --tw-color-red: #f00;
      --tw-color-green: #0f0;
      --tw-breakpoint-sm: 640px;
    }
    .tw\\:\\[color\\:theme\\(--color-red\\)\\] {
      color: #f00;
    }
    .tw\\:text-\\[theme\\(--color-red\\)\\] {
      color: #f00;
    }
    "
  `)
})
```

(sorry for the giant diff but the conversation already had the whole block selected ahaha)

RobinMalfait (approved):
A bit annoying to configure, but I do like the new syntax for it. Can't wait for writing a codemod for this 😅

philipp-spiess (approved):

PR #14517, opened 2024-09-25
CSS codemod: do not wrap comment nodes in `@layer`

RobinMalfait:
When CSS exists between two tailwind directives, then the CSS will be wrapped in an `@layer` to ensure it all ends up in the correct location in the final CSS file.

However, if the only thing between two tailwind directives is a comment, then the comment will also be wrapped in an `@layer` directive.

E.g.:
```css
@tailwind base;
/* This is a comment */
@tailwind components;
/* This is another comment */
@tailwind utilities;
```

Results in:
```css
@import "tailwindcss";
@layer base {
  /* This is a comment */
}
@layer components {
  /* This is another comment */
}
```

In this case, the layers don't matter, so this can be simplified to:
```css
@import "tailwindcss";
/* This is a comment */
/* This is another comment */
```


philipp-spiess (approved):
Niceeee

PR #14519, opened 2024-09-25
CSS codemod: Add `prefix(…)` to necessary CSS imports when a prefix is used.

thecrypticace:
If your config file has the `prefix` option specified:

```ts
/* tailwind.config.js */
export default {
  prefix: 'tw-',
};
```

We'll migrate it to the new `prefix` option in CSS:

```css
/* app.css */
```

Note: we error on invalid prefixes currently, so if you have a prefix that is not valid, you'll need to update it to a valid one before running the migration.


philipp-spiess:
Superset by #14557

PR #14526, opened 2024-09-26
Template migrations: Add automatic var injection codemods

philipp-spiess:
In v4, we're [removing automatic var injection](https://github.com/tailwindlabs/tailwindcss/pull/13657) (please refer to this PR for more detail as to why). 

Automatic var injection made it so that if you have a candidate like `bg-[--my-color]`, v3 would automatically wrap the content of the arbitrary section with a `var(…)`, resulting in the same as typing `bg-[var(--my-color)]`.

This PR adds codemods that go over various arbitrary fields and does the `var(…)` injection for you. To be precise, we will add `var(…)` to:

- Modifiers, e.g.: `bg-red-500/[var(--my-opacity)]`
- Variants, e.g.:  `supports-[var(--test)]:flex`
- Arbitrary candidates, e.g.: `[color:var(--my-color)]`
- Arbitrary values for functional candidates, e.g.: `bg-[var(--my-color)]`


adamwathan:
In v3, some utilities didn’t automatically wrap things in var() (there’s a list in the codebase somewhere), so we should make sure not to mod those ones here 👍🏻

packages/@tailwindcss-upgrade/src/template/codemods/automatic-var-injection.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (
+      node.kind === 'declaration' &&
+      AUTO_VAR_INJECTION_EXCEPTIONS.has(node.property) &&
+      node.value?.startsWith('--')
+    ) {
+      isException = true
+    }
+  })
+  return isException
+}
```
philipp-spiess, line 122: This _kinda_ works, unless you use a utility that would takes an arbitrary value with `--` but uses it in a _legit_ declaration _while also adding one of the exceptions to the AST. E.g.:

```
matchUtilities({
  foo: (value) => ({ 
    'scrollbar-color': value,
    'scroll-timeline-name': --my-other-thing,
  }),
})
```

I’m not sure how common this usage is but maaybe we can pass the arbitrary value in to this function and ensure that it uses that exact value we're trying to mod in a declaration of that exception to be extra sure?

philipp-spiess: Ok yeah this is actually a simple fix, so, fixed ✅

packages/@tailwindcss-upgrade/src/template/codemods/automatic-var-injection.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  ['bg-[color:--my-color]/[--my-opacity]', 'bg-[color:var(--my-color)]/[var(--my-opacity)]'],
+
+  ['[color:--my-color]/[_--my-opacity]', '[color:var(--my-color)]/[_--my-opacity]'],
+  ['bg-red-500/[_--my-opacity]', null],
+  ['bg-[--my-color]/[_--my-opacity]', 'bg-[var(--my-color)]/[_--my-opacity]'],
+  ['bg-[color:--my-color]/[_--my-opacity]', 'bg-[color:var(--my-color)]/[_--my-opacity]'],
+
+  // Variants
+  ['supports-[--test]:flex', 'supports-[var(--test)]:flex'],
+  ['supports-[_--test]:flex', null],
```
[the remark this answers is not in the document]
philipp-spiess: Ah, I got this from your list here: https://github.com/tailwindlabs/tailwindcss/pull/13657/files#diff-5c09a7c68a4426650ba1209e3be9652052750e24131acb607e09e5efea477dacR1619-R1624


`supports-[var(--test)]` mapping to `@supports var(--test)` seems to be at least compatible with v3 though 🙈 

Screenshot 2024-09-26 at 18 19 27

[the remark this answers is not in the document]
RobinMalfait: Hmm I see. While this is compatible with v3, it's still wrong (aka will never match). So the question is:

1. Do we want to fix it (could be a breaking change since it's already broken right now)
2. Do we get rid of the supports variant in this case (or the candidate entirely because it won't do anything anyway)
3. Do we do what v4 expects, where it maps to `--test: var(--tw)` (which essentially is the fix from #1)
4. Don't do anything

packages/@tailwindcss-upgrade/src/template/codemods/automatic-var-injection.ts
```diff
@@ -0,0 +1,154 @@
...
+  let ast = designSystem.compileAstNodes(candidate).map((n) => n.node)
+
+  let isException = false
+  walk(ast, (node) => {
+    if (
+      node.kind === 'declaration' &&
+      AUTO_VAR_INJECTION_EXCEPTIONS.has(node.property) &&
+      node.value == value
+    ) {
+      isException = true
```
RobinMalfait, line 150: I think once we found an exception, we can stop walking over anything else:

```suggestion
      isException = true
      return WalkAction.Stop
```

packages/@tailwindcss-upgrade/src/template/codemods/automatic-var-injection.ts
```diff
@@ -0,0 +1,154 @@
...
+  return {
+    kind: 'arbitrary' as const,
+    property: 'color',
+    value: 'red',
+    modifier: null,
+    variants: [variant],
+    important: false,
+    raw: 'candidate',
+  }
+}
```
RobinMalfait, line 115: Do we want to make sure that this is always in sync with the `Candidate` type? If so, we can probably do something like:
```suggestion
function createEmptyCandidate(variant: Variant) {
  return {
    kind: 'arbitrary' as const,
    property: 'color',
    value: 'red',
    modifier: null,
    variants: [variant],
    important: false,
    raw: 'candidate',
  } satisfies Candidate
}
```

Or
```suggestion
function createEmptyCandidate(variant: Variant): Candidate {
  return {
    kind: 'arbitrary' as const,
    property: 'color',
    value: 'red',
    modifier: null,
    variants: [variant],
    important: false,
    raw: 'candidate',
  }
}
```

RobinMalfait (approved):
Looking nice! Just a few suggestions

RobinMalfait (approved):

PR #14530, opened 2024-09-26
Use official FxHash crate

PolyMeilex:
`fxhash` is an old (and abandoned?) crate that simply extracts the fxhash implementation from rustc. These days, rustc has an official fxhash crate, that is actively maintained by the rust-lang organization.

In practice this does not really matter, aside from the new crate having one less dependency (`byteorder`), but I think it's generally a good idea to use something straight from the rust team.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,12 +51,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "byteorder"
-version = "1.4.3"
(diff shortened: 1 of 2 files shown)


thecrypticace:
Oh cool — thanks!. I noticed `rustc-hash` recently when looking through the rust-analyzer codebase for something but didn't connect the dots.

thecrypticace (approved):

PR #14531, opened 2024-09-26
Remove `process.env.VERSION` check

### LINKED ISSUE #14523: Fix compiler in non-Node.js environments?

Since `v4.0.0-alpha.21`

**Reproduction URL**

I'm not sure it's necessary, plus let's first confirm you even want to support this scenario.

**Describe your issue**

Until recently I could `import { compile } from "tailwindcss"` in a worker (or, presumably, browser) runtime and build a Tailwind stylesheet from my own list of candidate classes. This was awesome because we could compile certain Tailwind styles at runtime instead of build time.

Starting with `v4.0.0-alpha.21` this is broken because of a new unguarded reference to `process.env`:

https://app.renovatebot.com/package-diff?name=tailwindcss&from=4.0.0-alpha.20&to=4.0.0-alpha.21

I'm not sure this is a bug _per se_, because I don't know if you officially support running the compiler directly in non-Node.js environments.

But… could you? Would you? Our use case doesn't require a scanner, because our classes are in pure HTML, where a regex works flawlessly; we just need the compiler to build the CSS (AFAIK the same is true for Tailwind Play).

Obviously we can polyfill `process` to work around this for now, but I wanted to get your POV here. It seems to me if the compiler could pass tests in both Node.js and browser environments, you'd have a pure ESM package that could run anywhere, including worker runtimes. I just don't know if it's something you want to support.

thecrypticace:
Fixes #14523

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -395,7 +395,7 @@ export async function compile(
   })
 
   if (process.env.NODE_ENV !== 'test') {
-    ast.unshift(comment(`! tailwindcss v${getVersion()} | MIT License | https://tailwindcss.com `))
+    ast.unshift(comment(`! tailwindcss v${version} | MIT License | https://tailwindcss.com `))
   }
 
   // Track all invalid candidates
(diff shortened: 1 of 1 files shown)


thecrypticace:
@philipp-spiess I couldn't find any references to `process.env.VERSION` anywhere else but I swear it was being used at one point. Was that a leftover from our Standalone CLI work? Tests seem to pass so maybe we ended up not actually needing it?

philipp-spiess (approved):
@thecrypticace Yeah this looks like it sneaked in by accident. The only place where we need to patch the version is for the CLI. e.g. grep for `__tw_version`. `VERSION` is used nowhere in the codebase.

PR #14539, opened 2024-09-27
Update template migration interface

philipp-spiess:
This PR lands a quick interface update for template migration with some lessons learned form our existing migrations. Specifically, this version attempts to:

- Allow migrations to access the raw candidate. This way we can migrate candidates that _would not parse as valid in v4_. This will help us migrate prefixes in candidates from v3 to v4.
- There is no more awkward "return null" if nothing has changed. The return `null` was necessary because we relied on mutating the Variant and since parsing/printing could remove some information, it was not easy to find out wether a candidate needed to be migrated at all. With a string though, we can do this cheaply by returning the `rawCandidate`.
- We previously asserted that if `parseCandidate` returns more than one candidate, we only picked the first one. This behavior is now moved into the migrations where we have more context. For now though, we still do not need to worry about this since in all cases, these duplicate candidates would serialize to the same `Candidate`. It is helpful if you only want to run a migration on a specific type of candidate (e.g. if there's a `static` one and a more generic `functional` one).  
- We need access to the `DesignSystem` inside migrations now to be able to `parseCandidate`s.

Opening this up as a separate PR since it can take some time to iron out the edge cases for the individual codemod PRs and I don't want to be rebasing all the time. 

## Before

```ts
type Migration = (candidate: Candidate) => Candidate | null
```

## After

```ts
type Migration = (designSystem: DesignSystem, rawCandidate: string) => string
```


packages/@tailwindcss-upgrade/src/template/codemods/migrate-important.test.ts
philipp-spiess, line 1: moved this since it doesn't make sense to prefix all files with `migrate-*`

packages/@tailwindcss-upgrade/src/template/codemods/important.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  // Does not change non-important candidates
+  ['bg-blue-500', 'bg-blue-500'],
+  ['min-[calc(1000px+12em)]:flex', 'min-[calc(1000px+12em)]:flex'],
+])('%s => %s', async (candidate, result) => {
+  let designSystem = await __unstable__loadDesignSystem('@import "tailwindcss";', {
+    base: __dirname,
+  })
+
+  expect(important(designSystem, candidate)).toEqual(result)
+})
```
philipp-spiess, line 22: Migrated this test to what I found more useful in other migrations. Basically it only tests the migration function and not the text replace logic (we have other tests for this incl. integration tests)

RobinMalfait (approved):

PR #14553, opened 2024-09-30
Expose timing information in debug mode

RobinMalfait:
This PR exposes when using the the `DEBUG` environment variable. This follows the `DEBUG` conventions where:

- `DEBUG=1`
- `DEBUG=true`
- `DEBUG=*`
- `DEBUG=tailwindcss`

Will enable the debug information, but when using:

- `DEBUG=0`
- `DEBUG=false`
- `DEBUG=-tailwindcss`

It will not.

This currently only exposes some timings related to:

1. Scanning for candidates
2. Building the CSS
3. Optimizing the CSS

We can implement a more advanced version of this where we also expose more fine grained information such as the files we scanned, the amount of candidates we found and so on. But I believe that this will be enough to start triaging performance related issues.

### DIFF
--- a/packages/tailwindcss/src/env.ts
+++ b/packages/tailwindcss/src/env.ts
@@ -1,0 +1,40 @@
+export const DEBUG = typeof process !== 'undefined' ? resolveDebug(process.env.DEBUG) : false
+
+function resolveDebug(debug: typeof process.env.DEBUG) {
+  if (debug === undefined) {
+    return false
+  }
+
+  // Environment variables are strings, so convert to boolean
+  if (debug === 'true' || debug === '1') {
+    return true
+  }
+
+  if (debug === 'false' || debug === '0') {
+    return false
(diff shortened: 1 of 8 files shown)


packages/tailwindcss/src/env.ts
philipp-spiess, line 1: What are your thoughts of moving this into the node specific library instead? This way we don't need to be too defensive with accessing `process.env` and we can use a proper export for the three clients instead of having to rely on private bundled imports:

```
import * as env from '../../tailwindcss/src/env'
```

Could also be maybe a `tailwindcss-shared` private package that we only use for this. It just seems strange that this file is in the `tailwindcss` package but not part of the released `tailwindcss` npm package at all

philipp-spiess (approved):
Good stuff! one thought inline

thecrypticace (approved):

PR #14556, opened 2024-09-30
Add `blocklist` support from v3 config files

thecrypticace:
This PR adds support for the `blocklist` config option when using a JS config file in v4. You can now block certain classes from being generated at all. This is useful in cases where scanning files sees things that look like classes but are actually not used. For example, in paragraphs in a markdown file:

  ```js
  // tailwind.config.js
  export default {
    blocklist: ['bg-red-500'],
  }
  ```
  
  ```html
  <!-- index.html -->
  <div class="bg-red-500 text-black/75"></div>
  ```

Output:
  
  ```css
  .text-black/75 {
    color: rgba(0, 0, 0, 0.75);
  }
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add support for prefixes ([#14501](https://github.com/tailwindlabs/tailwindcss/pull/14501))
 - Expose timing information in debug mode ([#14553](https://github.com/tailwindlabs/tailwindcss/pull/14553))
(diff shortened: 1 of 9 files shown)


RobinMalfait (approved):
Nice! Can we add a test to make sure that `@apply bg-white` also doesn't work?

This is what v3 does: https://play.tailwindcss.com/ewH6tmV5q4?file=css

thecrypticace:
Good call — I fixed that case and added a test for it.

thecrypticace:
I think there's some wiggle room in the future here such that blocklist _only_ applies to generated candidates. This is because the nature of content scanning means you're likely scanning content you can't change for the sake of Tailwind. But `@apply` you always write explicitly and the failure mode is always hard crash when something doesn't exist.

PR #14557, opened 2024-09-30
Template migrations: Migrate v3 prefixes to v4

philipp-spiess:
This PR adds a new migration that can migrate Tailwind CSS v3 style prefixes into Tailwind CSS v4.

The migration is split into three separate pieces of work:

1. Firstly, we need to read the full JavaScript config to get the _old_ prefix option. This is necessary because in v4, we will not allow things like custom-separators for the prefix. From this option we will then try and compute a new prefix (in 90% of the cases this is going to just remove the trailing `-` but it can also work in more complex cases).
2. Then we migrate all Candidates. The important thing here is that we need to operate on the raw candidate string because by relying on `parseCandidate` (which we do for all other migrations) would not work, as the candidates are not valid in v4 syntax. More on that in a bit.
3. Lastly we also make sure to update the CSS config to include the new prefix. This is done by prepending the prefix option like so:
    
    ```css
    @import "tailwindcss" prefix(tw);
    ```

### Migrating candidates

The main difference between v3 prefixes and v4 prefixes is that in v3, the prefix was _part of the utility_ where as in v4 it is _always in front of the CSS class. 

So, for example, this candidate in v3: 

```
hover:-tw-mr-4
```

Would be converted to the following in v4:

```
tw:hover:-mr-4
```

Since the first example _won't parse as a valid Candidate in v4, as the `tw-mr` utility does not exist, we have to operate on the raw candidate string first. To do this I created a fork of the `parseCandidate` function _without any validation of utilities or variants_. This is used to identify part of the candidate that is the `base` and then ensuring the `base` starts with the old prefix. We then remove this to create an "unprefixed" candidate that we validate against a version of the DesignSystem _with no prefixes configured_. If the variant is valid this way, we can then print it again with the `DesignSystem` that has the new prefix to get the migrated version.

Since we set up the `DesignSystem` to include the new prefix, we can also be certain that migrations that happen afterwards would still disqualify candidates that aren't valid according to the new prefix policy. This does mean we need to have the prefix fixup be the first step in our pipeline.

One interesting bit is that in v3, arbitrary properties did not require prefixes where as in v4 they do. So the following candidate:

```
[color:red]
```

Will be converted to:

```
tw:[color:red]
```


packages/@tailwindcss-upgrade/src/template/codemods/prefix.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    return {
+      base,
+      start,
+      end,
+    }
+  }
+}
+
+const VALID_PREFIX = /([a-z]+)/
```
philipp-spiess, line 106: The heuristics here is basically finding the first alphabetical pair that appears in the old prefix. e.g. `tw-` is going to be `tw`, so is`tw__` etc.

packages/@tailwindcss-upgrade/src/template/parseConfig.ts
```diff
@@ -24,11 +34,21 @@ export async function parseConfig(
     relative = './' + relative
   }
 
-  let input = `@import 'tailwindcss';\n@config './${relative}'`
+  let userConfig = (await loadModule(fullConfigPath, __dirname, () => {}).then(
+    (result) => result.module,
+  )) as Config
```
philipp-spiess, line 39: There is one caveat here: By not using core to read the config file to get the _old_ prefix (because if we use Core we'd throw since that prefix is no longer supported in v4), we now loose the evaluation of plugins which...might also contribute prefixes? Is that something that happens in reality?

thecrypticace: A plugin can contribute a prefix. In reality this probably doesn't happen. But I wouldn't be surprised if prefixes were specified in _presets_.

philipp-spiess: Hmmm yeah we should resolve presets probably. I wish there was a nicer way to run this through core but we really want it to error like it does right now. Maybe I can use the compat APIs to resolve _only_ the config file properly

packages/@tailwindcss-upgrade/src/template/codemods/prefix.ts
```diff
@@ -0,0 +1,113 @@
...
+export function migratePrefix(prefix: string): string {
+  let result = VALID_PREFIX.exec(prefix.toLocaleLowerCase())
+  if (!result) {
+    throw new Error(`The prefix "${prefix}" can not be converted to Tailwind CSS v4.`)
```
thecrypticace, line 110:
```suggestion
    throw new Error(`The prefix "${prefix}" can not be used with Tailwind CSS v4 and cannot be converted to a valid one automatically. Please change all your utilities to use a prefix with only letters and run the migration again.`)
```

Alternatively, we could issue a warning and return `tw`?

packages/@tailwindcss-upgrade/src/template/parseConfig.ts
thecrypticace: I really don't like that we're exposing `loadMoudle` just for this case. I wonder if there's some way we can load the config without validation? Only problem I see there is it'd require adding an API to core.

Maybe this is something that we can revise once we revisit the APIs — probably just add a card for it later?

packages/@tailwindcss-upgrade/src/template/codemods/prefix.ts
thecrypticace: yeah that seems good to me

packages/@tailwindcss-upgrade/src/template/parseConfig.ts
thecrypticace: Oh maybe we just import resolveConfig directly and use it?

thecrypticace: Given we can use internal API b/c it's going to get bundled into the upgrade package directly — rather than relying on exports.

philipp-spiess: Yeah we can probably import it directly, note that right now it only uses `loadModule` from the `node` package so we have jiti added, but yeah we can add the dependencies to the upgrade package as well and then import the files we need from these other packages. I think we'll need to do this anyways now since we also want to load the presets etc.

philipp-spiess: Okay so i changed this to use `resolveConfig` from core and to not export `loadModule` in the node package anymore and instead bundle the dependencies directly

adamwathan (approved):

PR #14561, opened 2024-10-01
Fix rebuilds when editing imported CSS files

RobinMalfait:
This PR fixes an issue where importing other CSS files, then changing those CSS files didn't trigger a rebuild. This PR fixes that by making sure that the imported CSS files are properly watched.

This PR also adds some integration tests for `@tailwindcss/postcss`, `@tailwindcss/vite` and `@tailwindcss/cli` to make sure this works as expected.

### DIFF
--- a/packages/@tailwindcss-node/src/compile.ts
+++ b/packages/@tailwindcss-node/src/compile.ts
@@ -74,6 +74,8 @@ async function loadStylesheet(id: string, base: string, onDependency: (path: str
   let resolvedPath = await resolveCssId(id, base)
   if (!resolvedPath) throw new Error(`Could not resolve '${id}' from '${base}'`)
 
+  onDependency(resolvedPath)
+
   if (typeof globalThis.__tw_readFile === 'function') {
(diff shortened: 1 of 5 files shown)


packages/@tailwindcss-node/src/compile.ts
```diff
@@ -74,6 +74,8 @@ async function loadStylesheet(id: string, base: string, onDependency: (path: str
   let resolvedPath = await resolveCssId(id, base)
   if (!resolvedPath) throw new Error(`Could not resolve '${id}' from '${base}'`)
 
+  onDependency(resolvedPath)
```
philipp-spiess, line 77: Oh no, I was so certain I did add this. Thank you for fixing this 💅

philipp-spiess (approved):
Nice re. test upgrades

packages/@tailwindcss-node/src/compile.ts
RobinMalfait: No worries! This made a very easy fix for me. Spend 10x more time on the integration tests 😂

PR #14565, opened 2024-10-01
Improve `@tailwindcss/postcss` performance for initial builds

RobinMalfait:
This PR improves the performance of the `@tailwindcss/postcss` plugin. Before this change we created 2 compiler instances instead of a single one. On a project where a `tailwindcss.config.ts` file is used, this means that the timings look like this:

```
[@tailwindcss/postcss] Setup compiler: 137.525ms
⋮
[@tailwindcss/postcss] Setup compiler: 43.95ms
```

This means that with this small change, we can easily shave of ~50ms for initial PostCSS builds.


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Experimental_: Do not wrap comment nodes in `@layer` when running codemods ([#14517](https://github.com/tailwindlabs/tailwindcss/pull/14517))
 - _Experimental_: Ensure we don't lose selectors when running codemods ([#14518](https://github.com/tailwindlabs/tailwindcss/pull/14518))
 - _Experimental_: inject `@import` in a more expected location when running codemods ([#14536](https://github.com/tailwindlabs/tailwindcss/pull/14536))
+- Improve `@tailwindcss/postcss` performance for initial builds ([#14565](https://github.com/tailwindlabs/tailwindcss/pull/14565))
```
adamwathan, line 30: Maybe this should be more specifc?

```suggestion
- Only setup a single compiler in `@tailwindcss/postcss` for initial builds ([#14565](https://github.com/tailwindlabs/tailwindcss/pull/14565))
```

Also should move it above the experimental stuff maybe to keep things organized.

RobinMalfait: Yeah that's a bit better. I move it locally already, but saw CI fail so wanted to fix that first haha. Thanks for this suggestion 💪

philipp-spiess (approved):

RobinMalfait:
Pushed a few more changes to make CI pass. Some things I noticed and that the new commits solve:

1. The cache inside of the PostCSS plugin seems to be re-created from scratch, which means that we essentially have no cache. I hoisted it to the module-level to keep the cache around.
2. Noticed that we call `onDependency` with the same file twice, so made sure it's only in the list once. (This happened because resolving module dependencies of a file `x`, also included the file `x`)

So these new changes need a second pair of eyes before we can merge this.

/cc @adamwathan @philipp-spiess

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -36,20 +36,20 @@ export type PluginOptions = {
...
 
+let cache = new DefaultMap(() => {
+  return {
+    mtimes: new Map<string, number>(),
+    compiler: null as null | Awaited<ReturnType<typeof compile>>,
+    css: '',
+    optimizedCss: '',
+    fullRebuildPaths: [] as string[],
+  }
+})
```
adamwathan, line 47: The only thing I can think of here is that previously this was scoped for the `opts` argument to the `tailwindcss` function and now it's not, so now the cache is shared for the same key even if the options are different.

Interestingly the [PostCSS docs explicitly say](https://postcss.org/docs/writing-a-postcss-plugin) to set up caches in the place we were setting it up before:

image

Do we know with absolute certainty it was getting re-created from scratch on every save? 🤔

RobinMalfait: If I apply this diff to our plugin:
```diff
diff --git a/packages/@tailwindcss-postcss/src/index.ts b/packages/@tailwindcss-postcss/src/index.ts
index 7d18ddc36..43b824d2f 100644
--- a/packages/@tailwindcss-postcss/src/index.ts
+++ b/packages/@tailwindcss-postcss/src/index.ts
@@ -50,6 +50,8 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
   let base = opts.base ?? process.cwd()
   let optimize = opts.optimize ?? process.env.NODE_ENV === 'production'
 
+  console.count('Setup caches here')
+
   return {
     postcssPlugin: '@tailwindcss/postcss',
     plugins: [
@@ -65,6 +67,7 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
           let inputFile = result.opts.from ?? ''
           let context = cache.get(inputFile)
           let inputBasePath = path.dirname(path.resolve(inputFile))
+          console.count('Run @tailwindcss/postcss')
 
           async function createCompiler() {
             env.DEBUG && console.time('[@tailwindcss/postcss] Setup compiler')
```

If I then use it, the output looks like this:
image

You have the initial build, then 2 builds because of watch mode and because I made 2 changes.

Some notes / observations:
1. Not only is the `Setup caches here` log called every time we save, it's called twice for an unknown reason.
2. This is using the `postcss-cli` package

Next steps are to verify if this behaviour also happens in other environments where the PostCSS plugin is being used.

RobinMalfait: Can confirm that this behaviour also exists in a `vite` project with the PostCSS plugin:
image

ZuBB:
I might be running ahead of the train, but can this change (theoretically) be ported to v3? If so, what do maintainers think about it?

packages/@tailwindcss-postcss/src/index.ts
RobinMalfait: Some updates for completeness:

If you structure your PostCSS config like this:
```js
module.exports = {
  plugins: {
    '@tailwindcss/postcss': {},
  },
}
```
or this:
```js
module.exports = {
  plugins: [
    require('@tailwindcss/postcss')
  ],
}
```
Then the above behavior can be observer. If you _call_ your function (even without options) then you see the correct behavior:
```js
module.exports = {
  plugins: [
    require('@tailwindcss/postcss')()
  ],
}
```

adamwathan:
@ZuBB This isn't really relevant in v3 — completely different codebase and all of the caching we do there is already at the module scope level, so this is just an improvement over how v4 was working prior to this.

ZuBB:
> @ZuBB This isn't really relevant in v3 — completely different codebase and all of the caching we do there is already at the module scope level, so this is just an improvement over how v4 was working prior to this.

okay, I get it. thanks for the response

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -123,18 +125,20 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
...
-            detectSources: { base },
-            sources: context.compiler.globs,
-          })
+          if (context.scanner === null) {
+            // Look for candidates used to generate the CSS
+            context.scanner = new Scanner({
+              detectSources: { base },
+              sources: context.compiler.globs,
+            })
+          }
```
thecrypticace, line 134: I think we need to recreate the scanner when doing a full rebuild too, right? Since the globs might've changed?

thecrypticace: basically after line 166 we also want to re-create the scanner

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use the right import base path when using the CLI to reading files from stdin ([#14522](https://github.com/tailwindlabs/tailwindcss/pull/14522))
 - Ensure that `@utility` is top-level and cannot be nested ([#14525](https://github.com/tailwindlabs/tailwindcss/pull/14525))
+- Editing imported CSS files should trigger a rebuild ([#14561](https://github.com/tailwindlabs/tailwindcss/pull/14561))
+- Only setup a single compiler in `@tailwindcss/postcss` for initial builds ([#14565](https://github.com/tailwindlabs/tailwindcss/pull/14565))
```
RobinMalfait, line 29:
```suggestion
- Only setup a single compiler in `@tailwindcss/postcss` for initial builds ([#14565](https://github.com/tailwindlabs/tailwindcss/pull/14565))
```

PR #14594, opened 2024-10-04
Support keyframes in JS theme config

philipp-spiess:
Adds support for defining `keyframes` via the v3 JS config file. E.g.:

```js
module.exports = {
  theme: {
    extends: {
      keyframes: {
        "fade-in": {
          "0%": { opacity: "0" },
          "100%": { opacity: "1" },
        },
        "fade-out": {
          "0%": { opacity: "1" },
          "100%": { opacity: "0" },
        },
      },
    },
  },
};

```

### DIFF
--- a/packages/tailwindcss/src/compat/plugin-api.test.ts
+++ b/packages/tailwindcss/src/compat/plugin-api.test.ts
@@ -67,6 +67,18 @@ describe('theme', async () => {
           }
         }
       }
+      @keyframes enter {
+        from {
+          opacity: var(--tw-enter-opacity, 1);
+          transform: translate3d(var(--tw-enter-translate-x, 0), var(--tw-enter-translate-y, 0), 0) scale3d(var(--tw-enter-scale, 1), var(--tw-enter-scale, 1), var(--tw-enter-scale, 1)) rotate(var(--tw-enter-rotate, 0));
+        }
+      }
+      @keyframes exit {
+        to {
+          opacity: var(--tw-exit-opacity, 1);
+          transform: translate3d(var(--tw-exit-translate-x, 0), var(--tw-exit-translate-y, 0), 0) scale3d(var(--tw-exit-scale, 1), var(--tw-exit-scale, 1), var(--tw-exit-scale, 1)) rotate(var(--tw-exit-rotate, 0));
+        }
+      }
       "
     `)
   })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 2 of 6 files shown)


packages/tailwindcss/src/compat/plugin-api.test.ts
```diff
@@ -67,6 +67,18 @@ describe('theme', async () => {
...
+          opacity: var(--tw-enter-opacity, 1);
+          transform: translate3d(var(--tw-enter-translate-x, 0), var(--tw-enter-translate-y, 0), 0) scale3d(var(--tw-enter-scale, 1), var(--tw-enter-scale, 1), var(--tw-enter-scale, 1)) rotate(var(--tw-enter-rotate, 0));
+        }
+      }
+      @keyframes exit {
+        to {
+          opacity: var(--tw-exit-opacity, 1);
+          transform: translate3d(var(--tw-exit-translate-x, 0), var(--tw-exit-translate-y, 0), 0) scale3d(var(--tw-exit-scale, 1), var(--tw-exit-scale, 1), var(--tw-exit-scale, 1)) rotate(var(--tw-exit-rotate, 0));
+        }
+      }
```
philipp-spiess, line 81: Uh oh wait what.

philipp-spiess: This is interesting, in v3, plugins require an `addBase` to add the keyframes but if it was part of the JS config theme, it would be added automatically. That's lovely:
https://play.tailwindcss.com/YQLQchsWWW?file=config

thecrypticace: In v3 the way this worked was keyframes were not emitted until used. So you had to do `@keyframes …`: theme('keyframes.whatever`) in your utility function. And we handle that already.

thecrypticace: Or we do in addUtilities at least. I don't think we do in matchUtilities which we should I think.

thecrypticace: Or at least I'm like 95% sure it works that way. Maybe we just de-duped them…

philipp-spiess: Oh gotch you're right they're only added when used! If you use `addBase` _and_ use it with an `animate-*` variable, it is emitted twice in v3 (at the end of the stylesheet).

Screenshot 2024-10-04 at 12 56 42

I guess it's fine if we always emit keyframes for compat now (we do this for v4 keyframes too!). However unsure if we should then guard against this case. Maybe it's ok if the keyframes are emitted twice here since it does no harm? WDYT?

thecrypticace: yeah it's probably fine — would be nice to dedupe them but that can be an optimization done later. They should still be de-duped when minifying with lightingcss.

thecrypticace (approved):

PR #14595, opened 2024-10-04
Add `*.js` export variants for compat files

philipp-spiess:
In v3 it was possible to import `*.js` variants of compat files, e.g. `tailwindcss/plugin.js` in addition to `tailwindcss/plugin`. We need to properly map the exports to support this.

### DIFF
--- a/packages/tailwindcss/package.json
+++ b/packages/tailwindcss/package.json
@@ -71,11 +71,23 @@
         "require": "./dist/plugin.js",
         "import": "./dist/plugin.mjs"
       },
+      "./plugin.js": {
+        "require": "./dist/plugin.js",
+        "import": "./dist/plugin.mjs"
+      },
       "./defaultTheme": {
         "require": "./dist/default-theme.js",
         "import": "./dist/default-theme.mjs"
       },
+      "./defaultTheme.js": {
+        "require": "./dist/default-theme.js",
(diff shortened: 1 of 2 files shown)


thecrypticace:
Are there projects / plugins / whatever actually doing this?

philipp-spiess:
@thecrypticace loads: https://sourcegraph.com/search?q=context:global+%22from+%27tailwindcss/plugin%22&patternType=keyword&sm=0

thecrypticace:
About 15–16% wow. alright 👍

packages/tailwindcss/package.json
```diff
@@ -27,14 +27,26 @@
...
+      "import": "./src/compat/default-theme.ts"
+    },
     "./plugin": {
       "require": "./src/plugin.cts",
       "import": "./src/plugin.ts"
     },
+    "./plugin.js": {
+      "require": "./src/plugin.cts",
+      "import": "./src/plugin.ts"
+    },
```
thecrypticace, line 49: These need to be updated in the publishConfig too

thecrypticace (dismissed):
small change

packages/tailwindcss/package.json
philipp-spiess: ah of course 🫠

thecrypticace (approved):

PR #14597, opened 2024-10-04
Ensure upgrade tool has access to a JS config

philipp-spiess:
In order to properly migrate your Tailwind CSS v3 project to v4, we need access to the JavaScript configuration object. This was previously only required for template migrations, but in this PR we're making it so that this is also a prerequisite of the CSS migrations. This is because some migrations, like `@apply`, also need to convert candidates that to the v4 syntax and we need the full config in order to properly validate them.

In addition to requiring a JS config, we also now attempt to automatically find the right configuration file inside the current working directory. This is now matching the behavior of the Tailwind CSS v3 CLI where it will find the config automatically if it's in the current directory and called `tailwind.conf.js`.


packages/@tailwindcss-upgrade/src/template/prepareConfig.ts
```diff
@@ -0,0 +1,101 @@
+import { __unstable__loadDesignSystem, compile } from '@tailwindcss/node'
```
RobinMalfait, line 1: Can we rename this whole file to `prepare-config.ts` instead of `prepareConfig.ts`?

packages/@tailwindcss-upgrade/src/template/prepareConfig.ts
```diff
@@ -0,0 +1,101 @@
...
+import { loadModule } from '../../../@tailwindcss-node/src/compile'
+import { resolveConfig } from '../../../tailwindcss/src/compat/config/resolve-config'
+import type { DesignSystem } from '../../../tailwindcss/src/design-system'
+import { error } from '../utils/renderer'
+import { migratePrefix } from './codemods/prefix'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+
+let css = String.raw
```
RobinMalfait, line 16:
```suggestion
const css = String.raw
```

packages/@tailwindcss-upgrade/src/template/prepareConfig.ts
```diff
@@ -0,0 +1,101 @@
...
+      configPath = await detectConfigPath(options.base)
+    }
+
+    // We create a relative path from the current file to the config file. This is
+    // required so that the base for Tailwind CSS can bet inside the
+    // @tailwindcss-upgrade package and we can require `tailwindcss` properly.
+    let fullConfigPath = path.resolve(options.base, configPath)
+    let fullFilePath = path.resolve(__dirname)
+    let relative = path.relative(fullFilePath, fullConfigPath)
+    // If the path points to a file in the same directory, `path.relative` will
```
RobinMalfait, line 39:
```suggestion

    // If the path points to a file in the same directory, `path.relative` will
```

packages/@tailwindcss-upgrade/src/codemods/migrate-at-apply.ts
```diff
@@ -27,20 +30,13 @@ export function migrateAtApply({
         utility += '!'
       }
 
-      // Migrate the important modifier to the end of the utility
-      if (utility[0] === '!') {
-        utility = `${utility.slice(1)}!`
-      }
-
```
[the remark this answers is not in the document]
philipp-spiess: Was on purpose, we don't need to move the important modified at the end inside this loop anymore, now that we _always_ have a config and _always_ run the candidate migrations => The candidate migration will take care of this.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -2,12 +2,10 @@
 
 import { globby } from 'globby'
 import path from 'node:path'
-import type { Config } from 'tailwindcss'
-import type { DesignSystem } from '../../tailwindcss/src/design-system'
 import { help } from './commands/help'
 import { migrate as migrateStylesheet } from './migrate'
 import { migrate as migrateTemplate } from './template/migrate'
-import { parseConfig } from './template/parseConfig'
+import { prepareConfig } from './template/prepareConfig'
```
RobinMalfait, line 8:
```suggestion
import { prepareConfig } from './template/prepare-config'
```

RobinMalfait (approved):
Looking good! While testing I did notice we have a bogus import still. But apart from this this seems to work as expected!

philipp-spiess:
@RobinMalfait haha whoops 🙈

PR #14603, opened 2024-10-04
Migrate `@media screen(…)`

RobinMalfait:
This PR adds a codemod that migrates the `@media screen(…)` to the properly expanded `@media (…)` syntax.

```css
@media screen(md) {
  .foo {
    color: red;
  }
}
```

Will be converted to:
```css
@media (width >= 48rem) {
  .foo {
    color: red;
  }
}
```

If you happen to have custom screens (even complex ones), the screen will be converted to a custom media query.

```css
@media screen(foo) {
  .foo {
    color: red;
  }
}
```
With a custom `tailwind.config.js` file with a config like this:
```js
module.exports = {
  // …
  theme: {
    screens: {
      foo: { min: '100px', max: '123px' },
    },
  }
}
```

Then the codemod will convert it to:
```css
@media (123px >= width >= 100px) {
  .foo {
    color: red;
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pass options when using `addComponents` and `matchComponents` ([#14590](https://github.com/tailwindlabs/tailwindcss/pull/14590))
 - _Upgrade (experimental)_: Ensure CSS before a layer stays unlayered when running codemods ([#14596](https://github.com/tailwindlabs/tailwindcss/pull/14596))
 - _Upgrade (experimental)_: Resolve issues where some prefixed candidates were not properly migrated ([#14600](https://github.com/tailwindlabs/tailwindcss/pull/14600))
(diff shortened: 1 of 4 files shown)


thecrypticace (approved):
💪

adamwathan (dismissed):
Nice! I think this should use the theme function instead of inlining the raw value though.

RobinMalfait:
Made some updates:

1. If we can use the `theme(--breakpoint-{name})` function, then we will use that:
   https://github.com/tailwindlabs/tailwindcss/blob/28b2c1a1e0afbe3e37aa3c3d17d35562e4564512/packages/%40tailwindcss-upgrade/src/codemods/migrate-media-screen.test.ts#L26-L42
3. If we cannot, then we will expand the full media query.
   https://github.com/tailwindlabs/tailwindcss/blob/28b2c1a1e0afbe3e37aa3c3d17d35562e4564512/packages/%40tailwindcss-upgrade/src/codemods/migrate-media-screen.test.ts#L152-L177

Also tried with an alternative using `@custom-media`, but since we can't use that I reverted it. You can see the changes here: https://github.com/tailwindlabs/tailwindcss/pull/14603/commits/a1113f837b738a360da5d31b145771ad329be72a

thecrypticace (approved):

PR #14607, opened 2024-10-05
Add safe alignment utilities

benface:
Fixes #7929 and #12916

This PR adds [safe alignment utilities](https://www.stefanjudis.com/today-i-learned/safe-unsafe-alignment-in-css-flexbox/) to Tailwind 4. I opted to include them for every justify/align content/items/self property, but only for the `end` and `center` values. I know that it doesn't make sense for `start` (as the point of safe alignment is to fall back to `start` when it overflows), but I'm not sure about `space-between`, `space-around`, or other values. I certainly never encountered a situation where I needed `safe` with those.


philipp-spiess (approved):
Hey! I've updated your PR and added the new utilities behind a feature flag. The idea is that we release these with 4.1 but we might change the exact names (although current thinking is that the `-safe` suffix is nice because we can educates users to use these new utilities when they start typing `items-en…` and we show IntelliSense suggestions). 

Thanks for the PR!

PR #14612, opened 2024-10-07
Migrate simple PostCSS setup

philipp-spiess:
This PR attempts to detect simple postcss setups: These are setups that do not load dynamic modules and are based off the examples we are [recommending in our docs](https://tailwindcss.com/docs/installation/using-postcss). We detect wether a config is appropriate by having it use the object plugin config and by not requiring any other modules:

```js
module.exports = {
  plugins: {
    tailwindcss: {},
    autoprefixer: {},
  },
}
```

When we find such a config file, we will go over it line-by-line and attempt to:

- Upgrade `tailwindcss:` to `'@tailwindcss/postcss':`
- Remove `autoprefixer` if used

We then attempt to install and remove the respective npm packages based on the package manger we detect.

And since we now have logic to upgrade packages, this also makes sure to install `tailwindcss@next` at the end of a sucessful migration.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/migrate-postcss.ts
+++ b/packages/@tailwindcss-upgrade/src/migrate-postcss.ts
@@ -8,10 +8,12 @@ import { info, success, warn } from './utils/renderer'
 //
 // ```js
 // module.exports = {
-//  plugins: {
-//    tailwindcss: {},
-//    autoprefixer: {},
-//  }
+//   plugins: {
+//     'postcss-import': {},
+//     'tailwindcss/nesting': 'postcss-nesting',
+//     tailwindcss: {},
+//     autoprefixer: {},
+//   }
 // }
 export async function migratePostCSSConfig(base: string) {
   let configPath = await detectConfigPath(base)
@@ -30,17 +32,49 @@ export async function migratePostCSSConfig(base: string) {
 
   let didAddPostcssClient = false
   let didRemoveAutoprefixer = false
+  let didRemovePostCSSImport = false
 
   let fullPath = path.resolve(base, configPath)
   let content = await fs.readFile(fullPath, 'utf-8')
   let lines = content.split('\n')
   let newLines: string[] = []
-  for (let line of lines) {
-    if (line.includes('tailwindcss:')) {
+  for (let i = 0; i < lines.length; i++) {
+    let line = lines[i]
+
+    if (isTailwindCSSPlugin(line)) {
       didAddPostcssClient = true
       newLines.push(line.replace('tailwindcss:', `'@tailwindcss/postcss':`))
-    } else if (line.includes('autoprefixer:')) {
+    } else if (isAutoprefixerPlugin(line)) {
       didRemoveAutoprefixer = true
+    } else if (isPostCSSImportPlugin(line)) {
+      // Check that there are no unknown plugins before the tailwindcss plugin
+      let hasUnknownPluginsBeforeTailwindCSS = false
+      for (let j = i + 1; j < lines.length; j++) {
+        let nextLine = lines[j]
+        if (isTailwindCSSPlugin(nextLine)) {
+          break
+        }
+        if (isTailwindCSSNestingPlugin(nextLine)) {
+          continue
+        }
+        hasUnknownPluginsBeforeTailwindCSS = true
+        break
+      }
+
+      if (!hasUnknownPluginsBeforeTailwindCSS) {
+        didRemovePostCSSImport = true
+      } else {
+        newLines.push(line)
+      }
+    } else if (isTailwindCSSNestingPlugin(line)) {
+      // Check if the following rule is the tailwindcss plugin
+      let nextLine = lines[i + 1]
+      if (isTailwindCSSPlugin(nextLine)) {
+        // Since this plugin is bundled with `tailwindcss`, we don't need to
+        // clean up a package when deleting this line.
+      } else {
+        newLines.push(line)
+      }
     } else {
       newLines.push(line)
     }
@@ -55,6 +89,11 @@ export async function migratePostCSSConfig(base: string) {
   if (didRemoveAutoprefixer) {
     try {
       await pkg('remove autoprefixer', base)
+    } catch {}
+  }
+  if (didRemovePostCSSImport) {
+    try {
+      await pkg('remove postcss-import', base)
     } catch {}
   }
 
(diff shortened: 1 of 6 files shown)


packages/@tailwindcss-upgrade/src/utils/packages.ts
```diff
@@ -0,0 +1,61 @@
...
+  do {
+    // 1. Check package.json for a `packageManager` field
+    let packageJsonPath = resolve(base, 'package.json')
+    try {
+      let packageJsonContent = await fs.readFile(packageJsonPath, 'utf-8')
+      let packageJson = JSON.parse(packageJsonContent)
+      if (packageJson.packageManager) {
+        if (packageJson.packageManager.includes('yarn')) {
+          return 'yarn'
+        }
```
RobinMalfait, line 32: Do we want bun support as well?

```suggestion
        if (packageJson.packageManager.includes('bun')) {
          return 'bun'
        }
        if (packageJson.packageManager.includes('yarn')) {
          return 'yarn'
        }
```

packages/@tailwindcss-upgrade/src/utils/packages.ts
```diff
@@ -0,0 +1,61 @@
...
+        if (packageJson.packageManager.includes('pnpm')) {
+          return 'pnpm'
+        }
+        if (packageJson.packageManager.includes('npm')) {
+          return 'npm'
+        }
+      }
+    } catch {}
+
+    // 2. Check for common lockfiles
```
RobinMalfait, line 42: We can probably do this in a cleaner way. But bun is switching from `bun.lockb` (binary) to a `bun.lock` (text) file.
```suggestion
    // 2. Check for common lockfiles
    try {
      await fs.access(resolve(base, 'bun.lockb'))
      return 'bun'
    } catch {}
    try {
      await fs.access(resolve(base, 'bun.lock'))
      return 'bun'
    } catch {}
```

RobinMalfait (commented):
We have some failing tests 🤔

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,95 @@
...
+  let didRemoveAutoprefixer = false
+
+  let fullPath = path.resolve(base, configPath)
+  let content = await fs.readFile(fullPath, 'utf-8')
+  let lines = content.split('\n')
+  let newLines: string[] = []
+  for (let line of lines) {
+    if (line.includes('tailwindcss:')) {
+      didAddPostcssClient = true
+      newLines.push(line.replace('tailwindcss:', `'@tailwindcss/postcss':`))
```
philipp-spiess, line 41: If the previous line here is the nesting plugin, we can remove it as well: 

- https://tailwindcss.com/docs/using-with-preprocessors#nesting

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,95 @@
...
+// module.exports = {
+//  plugins: {
+//    tailwindcss: {},
+//    autoprefixer: {},
+//  }
+// }
+export async function migratePostCSSConfig(base: string) {
+  let configPath = await detectConfigPath(base)
+  if (configPath === null) {
+    // TODO: We can look for an eventual config inside package.json
```
philipp-spiess, line 19: https://sourcegraph.com/search?q=context:global+file:package.json+%27%22postcss%22:+%7B%27+tailwindcss&patternType=keyword&sm=0 let's support it

packages/@tailwindcss-upgrade/src/utils/packages.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (!packageManager) {
+    if (!didWarnAboutPackageManager) {
+      didWarnAboutPackageManager = true
+      warn('Could not detect a package manager. Please manually update `tailwindcss` to v4.')
+    }
+    return
+  }
+  return execSync(`${packageManager} ${command}`, {
+    cwd: base,
+  })
```
philipp-spiess, line 19: If this fails, we should print a message that explains what we tried to do

philipp-spiess (commented):
Some comments from my call with @RobinMalfait that I will fix up 👍

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,95 @@
...
+    } catch {}
+  }
+  return null
+}
+
+async function isSimplePostCSSConfig(base: string, configPath: string): Promise<boolean> {
+  let fullPath = path.resolve(base, configPath)
+  let content = await fs.readFile(fullPath, 'utf-8')
+  return (
+    content.includes('tailwindcss:') && !(content.includes('require') || content.includes('import'))
```
philipp-spiess, line 93: what about postcss-import? heh

philipp-spiess, line 93:
```suggestion
    content.includes('tailwindcss:') && !(content.includes('require') || content.includes('import '))
```

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,95 @@
...
+  let newLines: string[] = []
+  for (let line of lines) {
+    if (line.includes('tailwindcss:')) {
+      didAddPostcssClient = true
+      newLines.push(line.replace('tailwindcss:', `'@tailwindcss/postcss':`))
+    } else if (line.includes('autoprefixer:')) {
+      didRemoveAutoprefixer = true
+    } else {
+      newLines.push(line)
+    }
```
philipp-spiess, line 46: Should check for `{}` before deleting a line

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
philipp-spiess: also remove `postcss-import` if nothing else is between it and `tailwindcss`

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,287 @@
...
+  if (didRemoveAutoprefixer) {
+    try {
+      await pkg('remove autoprefixer', base)
+    } catch {}
+  }
+  if (didRemovePostCSSImport) {
+    try {
+      await pkg('remove postcss-import', base)
+    } catch {}
+  }
```
thecrypticace, line 72: We should merge these two `remove` calls into one command given that npm add/remove can sometimes take a decent chunk of time.

As far as I know every package manager supports removing more than one package

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -0,0 +1,287 @@
...
+  '.postcssrc.cjs',
+  '.postcssrc.ts',
+  '.postcssrc.mts',
+  '.postcssrc.cts',
+  'postcss.config.js',
+  'postcss.config.mjs',
+  'postcss.config.cjs',
+  'postcss.config.ts',
+  'postcss.config.mts',
+  'postcss.config.cts',
```
thecrypticace, line 259: As far as I know `.postcssrc`, `.postcssrc.json`, and `postcss.config.json` can also exist and use JSON. I think this is more common when using Next. We'll need to double check the logic in `postcss-load-config` to verify.

We could probably reuse some of the package.json config stuff for these files.

thecrypticace (dismissed):
Just a couple things but otherwise looks good!

thecrypticace (approved):

PR #14618, opened 2024-10-07
Make custom `@at-root` private API

RobinMalfait:
This PR makes the internal `@at-root` API private. Before this PR you could use `@at-root` in your own CSS, which means that it was part of the public API. If you (accidentally) used it in variants, you could generate CSS that was completely broken.

This now introduces a new private `AtRoot` node (similar to the recently introduced `Context` node) and can only be constructed within the framework.

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -22,7 +22,12 @@ export type Context = {
   nodes: AstNode[]
 }
 
-export type AstNode = Rule | Declaration | Comment | Context
+export type AtRoot = {
+  kind: 'at-root'
+  nodes: AstNode[]
+}
+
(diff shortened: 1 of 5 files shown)


thecrypticace (approved):

packages/tailwindcss/src/ast.ts
```diff
@@ -22,7 +22,12 @@ export type Context = {
   nodes: AstNode[]
 }
 
-export type AstNode = Rule | Declaration | Comment | Context
+export type AtRoot = {
+  kind: 'at-root'
```
RobinMalfait, line 26: If we ever need more like this, we could have a more generic `marker` node where you attach an additional sub-kind of w/e. I think this is the best solution for now though.

philipp-spiess (approved):

PR #14635, opened 2024-10-09
Inject `@config "..."` when a `tailwind.config.{js,ts,...}` is detected

RobinMalfait:
This PR injects a `@config "…"` in the CSS file if a JS based config has been found.

We will try to inject the `@config` in a sensible place:
1. Above the very first `@theme`
2. If that doesn't work, below the last `@import`
3. If that doesn't work, at the top of the file (as a last resort)

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -39,7 +39,8 @@ test(
       <div class="flex! sm:block! bg-linear-to-t bg-[var(--my-red)]"></div>
 
       --- ./src/input.css ---
-      @import 'tailwindcss';"
+      @import 'tailwindcss';
+      @config "../tailwind.config.js";"
     `)
 
     let packageJsonContent = await fs.read('package.json')
@@ -95,6 +96,8 @@ test(
       --- ./src/input.css ---
       @import 'tailwindcss' prefix(tw);
 
+      @config "../tailwind.config.js";
+
       .btn {
         @apply tw:rounded-md! tw:px-2 tw:py-1 tw:bg-blue-500 tw:text-white;
       }"
@@ -139,6 +142,8 @@ test(
       "
       --- ./src/index.css ---
       @import 'tailwindcss';
+
+      @config "../tailwind.config.js";
 
       .a {
         @apply flex;
@@ -192,6 +197,8 @@ test(
       "
       --- ./src/index.css ---
       @import 'tailwindcss';
+
+      @config "../tailwind.config.js";
 
       @layer base {
         html {
@@ -250,6 +257,8 @@ test(
       "
       --- ./src/index.css ---
       @import 'tailwindcss';
+
+      @config "../tailwind.config.js";
 
       @utility btn {
         @apply rounded-md px-2 py-1 bg-blue-500 text-white;
@@ -615,6 +624,7 @@ test(
       --- ./src/index.css ---
       @import 'tailwindcss';
       @import './utilities.css';
+      @config "../tailwind.config.js";
 
       --- ./src/utilities.css ---
       @utility no-scrollbar {
@@ -730,6 +740,7 @@ test(
       @import './c.1.css' layer(utilities);
       @import './c.1.utilities.css';
       @import './d.1.css';
+      @config "../tailwind.config.js";
 
       --- ./src/a.1.css ---
       @import './a.1.utilities.css'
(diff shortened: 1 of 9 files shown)


packages/@tailwindcss-upgrade/src/codemods/migrate-at-config.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (!hasTailwindImport) return
+
+    // Figure out the path to the config file
+    let sheetPath = sheet.file
+    let configPath = configFilePath
+
+    let relativePath = path.relative(path.dirname(sheetPath), configPath)
+    if (relativePath[0] !== '.') {
+      relativePath = `./${relativePath}`
+    }
```
thecrypticace, line 45: We'll want to test this on Windows. I think this might be kinda broken there?

RobinMalfait: I think we decided that all these globs should be unix paths with `/`. But _maybe_ the `relativePath` itself might be wrong then. Running the tests as-is on Windows right now.

thecrypticace: yeah I meant `path.relative` is wrong. It uses Windows separators on windows.

packages/@tailwindcss-upgrade/src/codemods/migrate-at-config.ts
```diff
@@ -39,7 +38,7 @@ export function migrateAtConfig(
     let sheetPath = sheet.file
     let configPath = configFilePath
 
-    let relativePath = path.relative(path.dirname(sheetPath), configPath)
+    let relativePath = path.relative(path.dirname(sheetPath), configPath).replaceAll('\\', '/')
```
thecrypticace, line 41: I don't completely recall but we have a normalizePath function in `@tailwindcss/node` and I _think_ it can handle relative paths. Might be good to use that here.

packages/@tailwindcss-upgrade/src/codemods/migrate-at-config.ts
```diff
@@ -0,0 +1,109 @@
...
+      relative = `./${relative}`
+    }
+    // Ensure relative is a posix style path since we will merge it with the
+    // glob.
+    relative = normalizePath(relative)
+
+    // Inject the `@config` in a sensible place
+    // 1. Above the first `@theme`
+    // 2. Below the last `@import`
+    // 3. At the top of the file
```
philipp-spiess, line 42: This is going to be the place for adding the CSS config if we can migrate the JS config. Maybe we can make the "what is injected" here dynamic so we can decide in the JS config migrator what we need in the css?

philipp-spiess (commented):
Good stuff! Some thoughts on how we can reuse this for the CSS we generate from the JS config too. For the relative path rewrite: If you orient yourself on the other places where we do that, it should be good 👍

packages/@tailwindcss-upgrade/src/codemods/migrate-at-config.ts
RobinMalfait: Yep I think once this PR is merged that we can abstract this to a function if we want. Not sure if that's necessary for this PR yet. 

There might be some additional logic where we might want to control the order of `@config`, `@theme`, injected `@variant`s and so on.

philipp-spiess: Haha yep I would need it right away, let's pair on this afterwards 👍

integrations/upgrade/index.test.ts
```diff
@@ -32,15 +32,16 @@ test(
...
-    await fs.expectFileToContain('src/input.css', css`@import 'tailwindcss';`)
+    expect(await fs.dumpFiles('./src/**/*.{css,html}')).toMatchInlineSnapshot(`
+      "
+      --- ./src/index.html ---
+      <h1>🤠👋</h1>
+      <div class="flex! sm:block! bg-linear-to-t bg-[var(--my-red)]"></div>
+
+      --- ./src/input.css ---
+      @import 'tailwindcss';
+      @config "../tailwind.config.js";"
```
philipp-spiess, line 43:
```suggestion
      @config "../tailwind.config.js";
```

thecrypticace (approved):

PR #14637, opened 2024-10-10
Ensure custom variants using the JS API have access to modifiers

philipp-spiess:
Fixes https://github.com/tailwindlabs/tailwindcss/discussions/14623

We didn't properly pass through modifiers for all custom variant calls. This PR fixes that.


RobinMalfait (approved):
Aha good catch

PR #14644, opened 2024-10-10
Migrate `aria-*`, `data-*` and `supports-*` variants from arbitrary values to bare values

RobinMalfait:
This PR adds a new codemod that can migrate `data-*` and `aria-*` variants using arbitrary values to bare values.

In Tailwind CSS v3, if you want to conditionally apply a class using data attributes, then you can write `data-[selected]:flex`. This requires the DOM element to have a `data-selected=""` attribute. In Tailwind CSS v4 we can simplify this, by dropping the brackets and by using `data-selected:flex` directly.

This migration operates on the internal AST, which means that this also just works for compound variants such as `group-has-data-[selected]:flex` (which turns into `group-has-data-selected:flex`).

Additionally, this codemod is also applicable to `aria-*` variants. The biggest difference is that in v4 `aria-selected` maps to an attribute of `aria-selected="true"`. This means that we can only migrate `aria=[selected="true"]:flex` to `aria-selected:flex`.

Last but not least, we also migrate `supports-[gap]` to `supports-gap` if the passed in value looks like a property. If not, e.g.: `supports-[display:grid]` then it stays as-is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Upgrade (experimental)_: Fully convert simple JS configs to CSS ([#14639](https://github.com/tailwindlabs/tailwindcss/pull/14639))
 - _Upgrade (experimental)_: Migrate `@media screen(…)` when running codemods ([#14603](https://github.com/tailwindlabs/tailwindcss/pull/14603))
 - _Upgrade (experimental)_: Inject `@config "…"` when a `tailwind.config.{js,ts,…}` is detected ([#14635](https://github.com/tailwindlabs/tailwindcss/pull/14635))
-- _Upgrade (experimental)_: Migrate `data-*` and `aria-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
+- _Upgrade (experimental)_: Migrate `aria-*`, `data-*`, and `supports-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Upgrade (experimental)_: Fully convert simple JS configs to CSS ([#14639](https://github.com/tailwindlabs/tailwindcss/pull/14639))
 - _Upgrade (experimental)_: Migrate `@media screen(…)` when running codemods ([#14603](https://github.com/tailwindlabs/tailwindcss/pull/14603))
 - _Upgrade (experimental)_: Inject `@config "…"` when a `tailwind.config.{js,ts,…}` is detected ([#14635](https://github.com/tailwindlabs/tailwindcss/pull/14635))
+- _Upgrade (experimental)_: Migrate `data-*` and `aria-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
 
 ### Fixed
 
--- a/packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
+++ b/packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
@@ -1,0 +1,61 @@
+import type { Config } from 'tailwindcss'
+import { parseCandidate, type Candidate, type Variant } from '../../../../tailwindcss/src/candidate'
+import type { DesignSystem } from '../../../../tailwindcss/src/design-system'
+import { printCandidate } from '../candidates'
+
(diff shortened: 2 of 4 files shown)


thecrypticace:
This brings me such joy 🥳

I think we can migrate `supports-*` too! (if the arbitrary value matches `/[a-z0-9-]+/` at least)

packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
```diff
@@ -0,0 +1,61 @@
...
+          value: variant.value.value,
+        }
+      }
+
+      // Convert `aria-[selected="true"]` to `aria-selected`
+      else if (
+        variant.kind === 'functional' &&
+        variant.root === 'aria' &&
+        variant.value?.kind === 'arbitrary' &&
+        variant.value.value.endsWith('="true"')
```
philipp-spiess, line 34: Is it valid to pass all CSS attribute selectors here? e.g. :`!=` or `*=` etc. This would match more often than we want it in that case

thecrypticace: The goal here is to match what the user would've typed in the variant in v3 that is _equivalent_ to the variant in v4 so anything other than `="true"` would be wrong because `="true"` is what v4 uses.

RobinMalfait: Ah but I see what @philipp-spiess means. If we have `aria-[foo*="true"]` then this could will match, but it should _not_ match because we can't safely migrate that. Will adjust and add a test.

philipp-spiess: Haha yeah that's what I meant! This will then migrate to `aria-foo*` which seems unexpected

packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
```diff
@@ -0,0 +1,94 @@
...
+          // aria-[foo~="true"]
+          key[key.length - 1] === '~' ||
+          // aria-[foo|="true"]
+          key[key.length - 1] === '|' ||
+          // aria-[foo^="true"]
+          key[key.length - 1] === '^' ||
+          // aria-[foo$="true"]
+          key[key.length - 1] === '$' ||
+          // aria-[foo*="true"]
+          key[key.length - 1] === '*'
```
RobinMalfait, line 50: I had a regex that covered all of these cases but it was horrible so made it more explicit 🙈

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Upgrade (experimental)_: The upgrade tool now automatically discovers your JavaScript config ([#14597](https://github.com/tailwindlabs/tailwindcss/pull/14597))
 - _Upgrade (experimental)_: Migrate legacy classes to the v4 alternative ([#14643](https://github.com/tailwindlabs/tailwindcss/pull/14643))
 - _Upgrade (experimental)_: Fully convert simple JS configs to CSS ([#14639](https://github.com/tailwindlabs/tailwindcss/pull/14639))
+- _Upgrade (experimental)_: Migrate `aria-*`, `data-*` and `supports-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
```
thecrypticace, line 18:
```suggestion
- _Upgrade (experimental)_: Migrate `aria-*`, `data-*`, and `supports-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
```

Oxford comma plz

packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
thecrypticace: lol its almost all special regex chars 😂  — I don't think any of them need to be escaped in a character class tho right? (unless `^` was listed first iirc). I think its this right `/[~|^$*]$/.test(key)`?

thecrypticace: explicit list of cases is fine though. It definitely reads better.

thecrypticace (approved):

packages/@tailwindcss-upgrade/src/template/codemods/arbitrary-value-to-bare-value.ts
RobinMalfait: This is the one I used originally `/[^~|^$*]=(['"]?)true\1/`

thecrypticace: ahhh yeah that is kinda awful lol

adamwathan (approved):

PR #14648, opened 2024-10-11
Detect and migrate static plugin usages

philipp-spiess:
This PR builds on top of the new [JS config to CSS migration](https://github.com/tailwindlabs/tailwindcss/pull/14651) and extends it to support migrating _static_ plugins.

What are _static_ plugins you might ask? Static plugins are plugins where we can statically determine that these are coming from a different file (so there is nothing inside the JS config that creates them). An example for this is this config file:

```js
import typographyPlugin from '@tailwindcss/typography'

export default {
  content: ['./src/**/*.{js,jsx,ts,tsx}'],
  darkMode: 'selector',
  plugins: [typographyPlugin],
} satisfies Config
```

Here, the `plugins` array only has one element and it is a static import from the `@tailwindcss/typography` module. In this PR we attempt to parse the config file via Tree-sitter to extract the following information from this file:

- What are the contents of the `plugins` array
- What are statically imported resources from the file

We then check if _all_ entries in the `plugins` array are either static resources or _strings_ (something I saw working in some tests but I’m not sure it still does). We migrate the JS config file to CSS if all plugins are static and we can migrate them to CSS `@plugin` calls.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/utils/extract-static-plugins.ts
+++ b/packages/@tailwindcss-upgrade/src/utils/extract-static-plugins.ts
@@ -8,7 +8,7 @@ const treesitter = String.raw
 const PLUGINS_QUERY = new Parser.Query(
   TS.typescript,
   treesitter`
-    ; export default = {}
+    ; export default {}
     (export_statement
       value: (satisfies_expression (object
         (pair
--- a/packages/@tailwindcss-upgrade/src/utils/extract-static-plugins.ts
+++ b/packages/@tailwindcss-upgrade/src/utils/extract-static-plugins.ts
@@ -54,10 +54,9 @@ const PLUGINS_QUERY = new Parser.Query(
         )?
       )
     )
-
   `,
 )
-export function findSimplePlugins(source: string): string[] | null {
+export function findStaticPlugins(source: string): string[] | null {
   try {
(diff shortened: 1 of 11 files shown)


philipp-spiess:
(Purposely not adding a change log since it's still covered in the initial JS config => CSS change log entry)

packages/@tailwindcss-upgrade/src/utils/extract-static-plugins.ts
```diff
@@ -0,0 +1,199 @@
...
+            pluginDefinition.type === '[' ||
+            pluginDefinition.type === ']' ||
+            pluginDefinition.type === ','
+          )
+            continue
+
+          switch (pluginDefinition.type) {
+            case 'identifier':
+              let source = imports[pluginDefinition.text]
+              if (!source || (source.export !== null && source.export !== '*')) {
```
philipp-spiess, line 83: Can also be `default`

philipp-spiess:
Should add support for:

```ts
import typography from '@tailwindcss/typography'

plugins: [
  require('@tailwindcss/typography')({
      className: 'wysiwyg',
    }),
   typography({
      className: 'wysiwyg',
    })
]
```

and migrate that to the CSS api

adamwathan (approved):

PR #14653, opened 2024-10-11
Fix `var(…)` as the opacity value inside the `theme(…)` function

RobinMalfait:
Inside the `theme(…)` function, we can use the `/` character for applying an opacity. For example `theme(colors.red.500 / 50%)` will apply a 50% opacity to the `colors.red.500` value.

However, if you used a variable instead of the hardcoded `50%` value, then this was not parsed correctly. E.g.: `theme(colors.red.500 / var(--opacity))`

_If_ we have this exact syntax (with the spaces), then it parses, but some information is lost:

```html
<div class="bg-[theme(colors.red.500_/_var(--opacity))]"></div>
```

Results in:
```css
.bg-\[theme\(colors\.red\.500_\/_var\(--opacity\)\)\] {
  background-color: color-mix(in srgb, #ef4444 calc(var * 100%), transparent);
}
```
Notice that the `var(--opacity)` is just parsed as `var`, and the `--opacity` is lost.

Additionally, if we drop the spaces, then it doesn't parse at all:

```html
<div class="bg-[theme(colors.red.500/var(--opacity))]"></div>
```

Results in:
```css
```

This means that we have to handle 2 issues to make this work:
1. We have to properly handle the `/` character as a proper separator.
2. If we have sub-functions, we have to make sure to print them in full (instead of only the very first node (`var` in this case)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pass options when using `addComponents` and `matchComponents` ([#14590](https://github.com/tailwindlabs/tailwindcss/pull/14590))
 - Ensure `boxShadow` and `animation` theme keys in JS config files are accessible under `--shadow-*` and `--animate-*` using the `theme()` function ([#14642](https://github.com/tailwindlabs/tailwindcss/pull/14642))
 - Ensure all theme keys with new names are also accessible under their old names when using the `theme()` function with the legacy dot notation syntax ([#14642](https://github.com/tailwindlabs/tailwindcss/pull/14642))
-- Ensure `var(…)` can be used as the opacity value inside the `theme(…)` function ([#14653](https://github.com/tailwindlabs/tailwindcss/pull/14653))
+- Ensure `var(…)` can be used as the opacity value inside the `theme([path] / [modifier])` function ([#14653](https://github.com/tailwindlabs/tailwindcss/pull/14653))
 - _Upgrade (experimental)_: Ensure CSS before a layer stays unlayered when running codemods ([#14596](https://github.com/tailwindlabs/tailwindcss/pull/14596))
 - _Upgrade (experimental)_: Resolve issues where some prefixed candidates were not properly migrated ([#14600](https://github.com/tailwindlabs/tailwindcss/pull/14600))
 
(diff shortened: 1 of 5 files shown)


packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -136,6 +136,27 @@ describe('theme function', () => {
         `)
       })
 
+      test('theme(colors.red.500/75%)', async () => {
```
RobinMalfait, line 139: Same test as below, but without spaces. Already worked, but added for completeness sake.

packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -178,6 +199,49 @@ describe('theme function', () => {
...
+
+      test('theme(colors.red.500/var(--opacity,50%))', async () => {
+        expect(
+          await compileCss(css`
+            @theme {
+              --color-red-500: #f00;
+            }
+            .red {
+              /* prettier-ignore */
+              color: theme(colors.red.500/var(--opacity,50%));
```
RobinMalfait, line 231: Prettier injected a space `var(--opacity, 50%)`, but explicitly didn't want it so I added the `prettier-ignore` comment.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pass options when using `addComponents` and `matchComponents` ([#14590](https://github.com/tailwindlabs/tailwindcss/pull/14590))
 - Ensure `boxShadow` and `animation` theme keys in JS config files are accessible under `--shadow-*` and `--animate-*` using the `theme()` function ([#14642](https://github.com/tailwindlabs/tailwindcss/pull/14642))
 - Ensure all theme keys with new names are also accessible under their old names when using the `theme()` function with the legacy dot notation syntax ([#14642](https://github.com/tailwindlabs/tailwindcss/pull/14642))
+- Fix `var(…)` as the opacity value inside the `theme(…)` function ([#14653](https://github.com/tailwindlabs/tailwindcss/pull/14653))
```
philipp-spiess, line 35:
```suggestion
- Allow `var(…)` for the opacity modifier inside `theme([path] / [modifier])` function calls ([#14653](https://github.com/tailwindlabs/tailwindcss/pull/14653))
```

Maybe if we word it like this it's also an addition and not a fix? I’m not sure about the theme function arguments, maybe it's fine to abbreviate with `…` but I was wondering if this makes it clearer what the modifier position is in this case.

philipp-spiess (approved):

CHANGELOG.md
RobinMalfait: I think it should be a fix because without it we would generate invalid CSS

thecrypticace (approved):

CHANGELOG.md
RobinMalfait: How about something like this? 

`Ensure `var(…)` can be used as the opacity value inside the `theme(…)` function`

philipp-spiess: Sounds good!

PR #14659, opened 2024-10-14
Callback in theme properties is also the theme function

philipp-spiess:
Something we noticed while testing the codemods on one of our codebases is that the callback passed to the `theme` function properties doesn't only expose some properties like `colors`, but it's also a function itself.

```ts
export default {
  theme: {
    extend: {
      colors: (theme) => {
        // The `theme` is a theme function _and_ the object...
        console.log(theme('spacing.2'), theme.colors.red['500'])
        return {}
      },
    },
  },
  plugins: [],
}
```

E.g.: https://play.tailwindcss.com/eV7Jgv17X1?file=config

---
h/t to @RobinMalfait for the issue description

### DIFF
--- a/packages/tailwindcss/src/compat/config/resolve-config.ts
+++ b/packages/tailwindcss/src/compat/config/resolve-config.ts
@@ -177,12 +177,11 @@ function extractConfigs(ctx: ResolutionContext, { config, base, path }: ConfigFi
 }
 
 function mergeTheme(ctx: ResolutionContext) {
-  let theme = createThemeFn(ctx.design, () => ctx.theme, resolveValue)
-  let api: PluginUtils = {
-    theme,
+  let themeFn = createThemeFn(ctx.design, () => ctx.theme, resolveValue)
+  let theme = Object.assign(themeFn, {
+    theme: themeFn,
     colors,
-  }
-  Object.assign(theme, api)
+  })
 
   function resolveValue(value: ThemeValue | null | undefined): ResolvedThemeValue {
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
Nice!

RobinMalfait:
Looks like we do have some type issues but otherwise this looks good

philipp-spiess:
@RobinMalfait Lol this passes locally, what 🤔

packages/tailwindcss/src/compat/config/resolve-config.ts
```diff
@@ -176,14 +177,15 @@ function extractConfigs(ctx: ResolutionContext, { config, base, path }: ConfigFi
...
 
 function mergeTheme(ctx: ResolutionContext) {
-  let api: PluginUtils = {
-    theme: createThemeFn(ctx.design, () => ctx.theme, resolveValue),
+  let themeFn = createThemeFn(ctx.design, () => ctx.theme, resolveValue)
+  let theme = Object.assign(themeFn, {
+    theme: themeFn,
     colors,
-  }
+  })
```
thecrypticace, line 184: I'm like 95% sure this is not documented anywhere. Pretty sure the types for v3 consider this usage invalid too. We'll still need to support it for B/C reasons but I think we should introduce a wrapper function that warns the first time its used.

cc @philipp-spiess @RobinMalfait thoughts?

philipp-spiess: See #14661

PR #14661, opened 2024-10-14
Warn on use of plugin parameters as function

philipp-spiess:
Quick follow-up to #14659 base don @thecrypticace's idea:

- This behavior is no longer added to the types of the Plugin API to be consistent with v3
- When the plugin argument is used as a function, we now warn the first time

### DIFF
--- a/packages/tailwindcss/src/compat/config/resolve-config.test.ts
+++ b/packages/tailwindcss/src/compat/config/resolve-config.test.ts
@@ -172,88 +172,87 @@ test('theme keys can reference other theme keys using the theme function regardl
   })
 })
 
-test('theme keys can read from the CSS theme', () => {
-  let originalWarn = console.warn
-  try {
-    console.warn = vi.fn()
-    let theme = new Theme()
-    theme.add('--color-green', 'green')
+test('theme keys can read from the CSS theme', ({ onTestFinished }) => {
+  let warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+  onTestFinished(() => {
+    warn.mockReset()
+  })
 
-    let design = buildDesignSystem(theme)
+  let theme = new Theme()
+  theme.add('--color-green', 'green')
 
-    let config = resolveConfig(design, [
-      {
-        config: {
-          theme: {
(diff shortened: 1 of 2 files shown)


packages/tailwindcss/src/compat/config/resolve-config.test.ts
```diff
@@ -173,78 +173,87 @@ test('theme keys can reference other theme keys using the theme function regardl
 })
 
 test('theme keys can read from the CSS theme', () => {
-  let theme = new Theme()
-  theme.add('--color-green', 'green')
+  let originalWarn = console.warn
```
thecrypticace, line 176: I'd probably just mock `console` for the entire test file — it'll help cleanup the test and makes adding / testing additional warnings in the future simpler:

```ts
  const warn = vi.spyOn(console, 'log').mockImplementation(() => undefined);

  afterAll(() => warn.mockReset());
```

```ts
    expect(warn).toHaveBeenCalledWith(
       'Using the plugin object parameter as the theme function is deprecated. Please use the `theme` property instead.',
     )
```

philipp-spiess: Not a fan of mocking apis like this too broadly since it can lead to us hiding warnings somewhere where we don't want it. Can put this test into a describe block to scope this and still use the same pattern though 👍 (or can use the `onTestFinished` hook to clean up and have it inside the single test only.. 🤔)

packages/tailwindcss/src/compat/config/resolve-config.test.ts
```diff
@@ -172,7 +172,12 @@ test('theme keys can reference other theme keys using the theme function regardl
   })
 })
 
-test('theme keys can read from the CSS theme', () => {
+test('theme keys can read from the CSS theme', ({ onTestFinished }) => {
+  let warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+  onTestFinished(() => {
+    warn.mockReset()
+  })
```
thecrypticace, line 179: ooh I forgot this existed 💯

thecrypticace (approved):

PR #14664, opened 2024-10-14
Migrate `theme(…)` calls to `var(…)` or the modern `theme(…)` syntax

RobinMalfait:
This PR adds a codemod to convert `theme(…)` calls to `var(…)` calls. If we can't safely do this, then we try to convert the `theme(…)` syntax (dot notation) to the modern `theme(…)` syntax (with CSS variable-like syntax).

### Let's look at some examples:

**Simple example:**

Input:
```html
<div class="bg-[theme(colors.red.500)]"></div>
```

Output:
```html
<div class="bg-[var(--color-red-500)]"></div>
```

---

**With fallback:**

Input:
```html
<div class="bg-[theme(colors.red.500,theme(colors.blue.500))]"></div>
```

Output:
```html
<div class="bg-[var(--color-red-500,var(--color-blue-500))]"></div>
```

---

**With modifiers:**

Input:
```html
<div class="bg-[theme(colors.red.500/75%)]"></div>
```

Output:
```html
<div class="bg-[var(--color-red-500)]/75"></div>
```

We can special case this, because if you are using that modifier syntax we _assume_ it's being used in a `theme(…)` call referencing a color. This means that we can also convert it to a modifier on the actual candidate.

---

**With modifier, if a modifier is already present:**

Input:
```html
<div class="bg-[theme(colors.red.500/75%)]/50"></div>
```

Output:
```html
<div class="bg-[theme(--color-red-500/75%)]/50"></div>
```

In this case we can't use the `var(…)` syntax because that requires us to move the opacity modifier to the candidate itself. In this case we could use math to figure out the expected modifier, but that might be too confusing. Instead, we convert to the modern `theme(…)` syntax.

---

**Multiple `theme(…)` calls with modifiers:**

Input:
```html
<div class="bg-[theme(colors.red.500/75%,theme(colors.blue.500/50%))]"></div>
```

Output:
```html
<div class="bg-[theme(--color-red-500/75%,theme(--color-blue-500/50%))]"></div>
```

In this case we can't convert to `var(…)` syntax because then we lose the opacity modifier. We also can't move the opacity modifier to the candidate itself e.g.: `/50` because we have 2 different variables to worry about.

In this situation we convert to the modern `theme(…)` syntax itself.

---

**Inside variants:**

Input:
```html
<div class="max-[theme(spacing.20)]:flex"></div>
```

Output:
```html
<div class="max-[theme(--spacing-20)]:flex"></div>
```

Unfortunately we can't convert to `var(…)` syntax reliably because in some cases (like the one above) the value will be used inside of an `@media (…)` query and CSS doesn't support that at the time of writing this PR.

So to be safe, we will not try to convert `theme(…)` to `var(…)` in variants, but we will only upgrade the `theme(…)` call itself to modern syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- _Upgrade (experimental)_: Migrate `theme(…)` calls in classes to `var(…)` or to the modern `theme(…)` syntax ([#14664](https://github.com/tailwindlabs/tailwindcss/pull/14664))
+
 ### Fixed
 
 - Ensure `theme` values defined outside of `extend` in JS configuration files overwrite all existing values for that namespace ([#14672](https://github.com/tailwindlabs/tailwindcss/pull/14672))
--- a/packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
+++ b/packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
@@ -3,8 +3,12 @@ import { expect, test } from 'vitest'
 import { themeToVar } from './theme-to-var'
 
 test.each([
+  // Keep candidates that don't contain `theme(…)` or `theme(…, …)`
+  ['[color:red]', '[color:red]'],
+
   // Convert to `var(…)` if we can resolve the path
   ['[color:theme(colors.red.500)]', '[color:var(--color-red-500)]'], // Arbitrary property
+  ['[color:theme(colors.red.500)]/50', '[color:var(--color-red-500)]/50'], // Arbitrary property + modifier
   ['bg-[theme(colors.red.500)]', 'bg-[var(--color-red-500)]'], // Arbitrary value
   ['bg-[size:theme(spacing.4)]', 'bg-[size:var(--spacing-4)]'], // Arbitrary value + data type hint
 
@@ -77,6 +81,20 @@ test.each([
 
   // `theme(…)` calls valid in v3, but not in v4 should still be converted.
   ['[--foo:theme(fontWeight.semibold)]', '[--foo:theme(fontWeight.semibold)]'],
+
+  // Invalid cases
+  ['[--foo:theme(colors.red.500/50/50)]', '[--foo:theme(colors.red.500/50/50)]'],
+  ['[--foo:theme(colors.red.500/50/50)]/50', '[--foo:theme(colors.red.500/50/50)]/50'],
+
+  // Partially invalid cases
+  [
+    '[--foo:theme(colors.red.500/50/50)_theme(colors.blue.200)]',
+    '[--foo:theme(colors.red.500/50/50)_var(--color-blue-200)]',
+  ],
+  [
+    '[--foo:theme(colors.red.500/50/50)_theme(colors.blue.200)]/50',
+    '[--foo:theme(colors.red.500/50/50)_var(--color-blue-200)]/50',
+  ],
 ])('%s => %s', async (candidate, result) => {
   let designSystem = await __unstable__loadDesignSystem('@import "tailwindcss";', {
     base: __dirname,
--- a/packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
+++ b/packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
@@ -1,0 +1,86 @@
+import { __unstable__loadDesignSystem } from '@tailwindcss/node'
+import { expect, test } from 'vitest'
+import { themeToVar } from './theme-to-var'
+
+test.each([
+  // Convert to `var(…)` if we can resolve the path
+  ['[color:theme(colors.red.500)]', '[color:var(--color-red-500)]'], // Arbitrary property
+  ['bg-[theme(colors.red.500)]', 'bg-[var(--color-red-500)]'], // Arbitrary value
+  ['bg-[size:theme(spacing.4)]', 'bg-[size:var(--spacing-4)]'], // Arbitrary value + data type hint
+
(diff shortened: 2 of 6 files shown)


packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.ts
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        // Update path to be the first part
+        path = pathPart
+      }
+
+      let variable = `--${keyPathToCssProperty(toKeyPath(path))}` as const
+      if (!designSystem.theme.get([variable])) return null
+
+      return fallback ? `var(${variable}, ${fallback})` : `var(${variable})`
```
thecrypticace, line 101: I feel like there may be some design where this shared bits between convert and convertBasic aren't duplicated but perhaps the duplication is fine.

RobinMalfait: Let me think about this one. It's definitely a little bit awkward because we are analyzing _while_ replacing. But depending on later information the final result might change.

At least all the necessary logic is there, but a clean 2 step pass might be enough.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Upgrade (experimental)_: Migrate `@media screen(…)` when running codemods ([#14603](https://github.com/tailwindlabs/tailwindcss/pull/14603))
 - _Upgrade (experimental)_: Inject `@config "…"` when a `tailwind.config.{js,ts,…}` is detected ([#14635](https://github.com/tailwindlabs/tailwindcss/pull/14635))
 - _Upgrade (experimental)_: Migrate `aria-*`, `data-*`, and `supports-*` variants from arbitrary values to bare values ([#14644](https://github.com/tailwindlabs/tailwindcss/pull/14644))
+- _Upgrade (experimental)_: Migrate `theme(…)` calls to `var(…)` or the modern `theme(…)` syntax ([#14664](https://github.com/tailwindlabs/tailwindcss/pull/14664))
```
philipp-spiess, line 23: The change log entry is a bit misleading since I would assume all `theme()` functions are migrated when it's only `theme()` function inside candidates.

Might want to clarify this or (even better) follow up with the CSS `theme()` function implementation right away? hopefully should be able to reuse most of the implementation.

packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  ['bg-[theme(colors.red.500/50%)]/50', 'bg-[theme(--color-red-500/50%)]/50'],
+
+  // Variants, we can't use `var(…)` especially inside of `@media(…)`. We can
+  // still upgrade the `theme(…)` to the modern syntax.
+  ['max-[theme(spacing.4)]:flex', 'max-[theme(--spacing-4)]:flex'],
+
+  // This test in itself doesn't make much sense. But we need to make sure
+  // that this doesn't end up as the modifier in the candidate itself.
+  ['max-[theme(spacing.4/50)]:flex', 'max-[theme(--spacing-4/50)]:flex'],
+])('%s => %s', async (candidate, result) => {
```
philipp-spiess, line 77: So there's a class of `theme()` keypaths that are valid in v3 (and are valid in our interop layer) but don't have a keypath equivalent in v4. Things like: `theme(fontWeight.semibold)` or `theme(cursor.progress)`. Let's add some tests here to see they are still migrated. 

Maybe what we can do is to use the dict in `default-theme.ts` and resolve these to the actual values?

CHANGELOG.md
RobinMalfait: > The change log entry is a bit misleading since I would assume all `theme()` functions are migrated when it's only `theme()` function inside candidates.

Do you mean that you expect all the `theme()` functions to be gone? Because in case you have `max-[theme(spacing.4)]:flex` then we can't replace the `theme()` with var because the browser doesn't support this. We _could_ sort of resolve the `theme(spacing.4)` value, but then we end up with `max-[1rem]:flex` which isn't "dynamic" anymore if you change your spacing scale.

So in this case, we migrate `max-[theme(spacing.4)]` to the modern theme function `max-[theme(--spacing.4)]`. 

Is this what you mean?

RobinMalfait: Got it!

It's only about the `theme(…)` in classes, _not_ about the `theme(…)` in CSS.

packages/@tailwindcss-upgrade/src/template/codemods/theme-to-var.test.ts
RobinMalfait: While on call, one thing we noticed is that `theme(fontWeight.semibold)` is hardcoded in the font-weight utility definition to be `600` and doesn't have a corresponding CSS variable. This means that we can't migrate `theme(fontWeight.semibold)` to a `var(…)` or even a `theme(--font-weight-semibold)` because there is no CSS variable. We could inline it to `600` directly since it's hardcoded.

However, there is a catch... _if_ you define `--font-weight-semibold: 100` as a CSS variable inside your `@theme { … }`, then the value is not hardcoded anymore, but the CSS variable takes precedence.

This means that we _will_ migrate `theme(fontWeight.semibold)` to `theme(--font-weight-semibold)` or even `var(--font-weight-semibold)` but _only_ if you have this CSS variable configured.

Long story short:
1. If you migrate to `var(…)` or `theme(--…)`, and remove the CSS variable from your `@theme { … }` later, then the default `fontWeight.semibold` won't apply.
2. If you _don't_ migrate, but inline the value `theme(fontWeight.semibold)` becomes `600`. Then adding the CSS variable to your `@theme { … }` later, then the variable won't take effect because we already hardcoded the value to `600`.

thecrypticace (approved):

PR #14667, opened 2024-10-14
Support complex compound variants in `group-*`, `peer-*`, `has-*`, and `not-*`

thecrypticace:
This PR adds support for more complex variant definitions when using the four compound variants: `group-*`, `peer-*`, `has-*`, and `not-*`. Before this PR these variants were limited in what compound variants they supported — often limited to single rules with single selectors.

Now, all of these variants support (nearly) the full range of all variant definitions.

Examples TBD — opening to get some eyes on stuff

TODOs:
- [ ] Bail when there's an at rule that is not invertible

### DIFF
--- a/packages/tailwindcss/src/selectors/negate.ts
+++ b/packages/tailwindcss/src/selectors/negate.ts
@@ -1,0 +1,87 @@
+import type { AstNode } from '../ast'
+import { rule, walkDepth } from '../ast'
+import { DefaultMap } from '../utils/default-map'
+import { segment } from '../utils/segment'
+import { and, lit, not, or, type Expr } from './expr'
+import { flattenNesting } from './nesting'
+import { toDNF } from './rewriting'
+
+export function negateRules(ast: AstNode[]): AstNode[] {
+  // 1. Flatten any nesting
+  // 2. Isolate rules such that no siblings are present except at the root
+  let flattened = flattenNesting(structuredClone(ast))
+
+  // 3. Convert the AST into an expression graph
+  // 4. Negate the graph
+  let graph = not(toGraph(flattened))
+
+  // 5. Convert the graph to Disjunctive Normal Form (DNF)
+  graph = toDNF(graph)
+
+  // 6. Convert the graph back into a structured AST
+  // The current graph looks like this
+  // (A && B && …) || (D && E && …) || …
+
+  // Each conjunction represents one set of nested rules
+  // Each disjunction represents a separate set of rules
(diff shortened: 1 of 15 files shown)


thecrypticace:
@RobinMalfait @philipp-spiess It's not like in a 100% reviewable state but wanted to get the general thoughts on things so far?

packages/tailwindcss/src/variants.ts
```diff
@@ -205,99 +207,138 @@ export function createVariants(theme: Theme): Variants {
...
+  // }
+  //
+  // &:not(:hover) {
+  //   /* declarations */
+  // }
+  // ```
+  //
+  // Which as you can see can introduce arbitrary levels of nesting and
+  // complexity. This is why we do not support inverting at-rules.
   variants.compound('not', (ruleNode, variant) => {
```
RobinMalfait, line 283: One thing to figure out with this `not` variant is that it _can_ compounds other variants. Right now all `compound()` variants will check if the variant we are nesting have the `compounds: true` property. 

This makes sense for `peer-*`, `group-*`, and `has-*`, since those operate on selectors and most variants with `compound: false` (like breakpoints) don't have selectors (so `group-md` doesn't make sense).

But in this case, for `not`, it does make sense. Because right now you cannot do `not-portrait` or `not-md`.

RobinMalfait: Same goes for custom `@variant` definitions in your CSS. They are marked as `compounds: false`.

packages/tailwindcss/src/selectors/negate.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  return flattenNesting(newAst)
+}
+
+function invertSelector(selector: string): string {
+  if (selector[0] === '@') {
+    return selector
+      .replace('@media', '@media not')
+      .replace('@supports', '@supports not')
+      .replace('@document', '@document not')
```
RobinMalfait, line 51: If a `@media` already has a `not`, then this could produce:

```css
@media not not screen {
  /* … */
}
```

Which is invalid CSS and Lightning CSS will throw it out. 

In this case, I think we have a few options:
1. Wrap it in `@media not (_original_here_)`.
2. Inspect the params, and if it already has a `not`, delete it

thecrypticace:
We've decide that we're not gonna pursue this due to added complexity in the codebase

I'll be landing a separate PR to cleanup some things around compound variants instead — as well as some of the bug fixes landed along the way in this PR.

PR #14668, opened 2024-10-15
Fix Plugin type issue

mrmckeb:
This is a quick fix to an issue where the types for PluginsConfig don't match those used in plugins like [`@tailwindcss/container-queries`](https://github.com/tailwindlabs/tailwindcss-container-queries).

This was caught by TypeScript with [`exactOptionalPropertyTypes`]( https://www.typescriptlang.org/tsconfig/exactOptionalPropertyTypes.html) enabled, where TypeScript checks if `undefined` can be supplied as a value for optional types.

I felt that it made more sense to fix this here, as it makes the core types more flexible, as opposed to each plugin needing to fix this when/if they hit it.


philipp-spiess:
@mrmckeb Thanks for the PR! This change looks reasonable but I’m not sure I understand what current issue this is fixing. I was looking at the `@tailwindcss/container-queries` plugin you referenced but I don't see it using a `config: undefined` somewhere 🤔

mrmckeb:
Hi @philipp-spiess,

No problem, I can add more detail :)

The types for the plugin, as per the DTS file, are:
```
declare const _default: {
    handler: import("tailwindcss/types/config").PluginCreator;
    config?: Partial<import("tailwindcss/types/config").Config> | undefined;
};
export = _default;
```

So in this stricter mode of TypeScript, this is flagged as an issue.

I'm not 100% sure where the `undefined` comes from, but it seems to be from an older version of Tailwind (`insiders`, from 10+ months ago), but I feel this is a probably still a good fix given that other plugins may have the same types.

mrmckeb:
Hi @philipp-spiess, any more feedback on this one?

philipp-spiess:
@mrmckeb Hey! sorry for not getting back earlier. Do you mind adding a changelog entry, something along those lines? 

```
- Ensure the TypeScript types for `PluginsConfig` allow `undefined` values ([#14668](https://github.com/tailwindlabs/tailwindcss/pull/14668))
```

I'm having some issues pushing to your branch directly 🙈

mrmckeb:
Sorry @philipp-spiess - I was a little sick last week and this dropped off of my plate.

I've updated the branch and added that entry in.

You do have permission to write to this branch, so I'm not sure why you couldn't sorry!
image

philipp-spiess:
@mrmckeb Thanks a ton! Will merge this in but given our current focus on v4, it might take a while before we do a new v3 release unless something urgent needs fixing.

philipp-spiess (approved):

philipp-spiess (approved):

PR #14670, opened 2024-10-15
Ensure `font-stretch` utilities only accepts positive integer bare values

RobinMalfait:
This PR fixes an issue where bare values in the `font-stretch` utility such as `font-stretch-50.5%` compiled. But for bare values, we want these values to be positive integers only so that we don't have too many special characters.

If you want to use `50.5%`, you can still use `font-stretch-[50.5%]` as a utility.


philipp-spiess (approved):
TY

PR #14672, opened 2024-10-15
Reset `default` @theme values for non `extend` JS theme config

philipp-spiess:
Imagine the following setup:

```css
/* src/input.css */
@import "tailwindcss";
@config "../tailwind.config.ts";
@theme {
  --color-red-500: #ef4444;
}
```

```ts
/* tailwind.config.ts */
export default {
  theme: {
    colors: {
      red: {
        600: '#dc2626'
      } 
    },
    extend: {
      colors: {
        400: '#f87171'
      }
    }
  }
}
```

Since the theme object in the JS config contains `colors` in the non-`extends` block, you would expect this to _not pull in all the default colors imported via `@import "tailwindcss";`_. This, however, wasn't the case right now since all theme options were purely _additive_ to the CSS.

This PR makes it so that non-`extend` theme keys _overwrite default CSS theme values_. The emphasis is on `default` here since you still want to be able to overwrite your options via `@theme {}` in user space.

This now generates the same CSS that our upgrade codemods would also generate as this would apply the new CSS right after the `@import "tailwindcss";` rule resulting in:

```css
@import "tailwindcss";
@theme {
  --color-*: initial;
  --color-red-400: #f87171;
  --color-red-600: #dc2626;
}
@theme {
  --color-red-500: #ef4444;
}
```

## Keyframes

This PR also adds a new core API to unset keyframes the same way. We previously had no option of doing that but while working on the above codemods we noticed that keyframes should behave the same way:

```css
@import "tailwindcss";
@theme {
  --keyframes-*: initial;
  @keyframes spin {
    to {
      transform: rotate(361deg);
    }
  }
}
```

To do this, the keyframes bookeeping was moved from the main Tailwind CSS v4 file into the `Theme` class. 

_I’m not sure super of the API yet but we would need a way for the codemods to behave the same as out interop layer here. Option B is that we don't reset keyframes the same way we reset other theme variables_.


packages/tailwindcss/src/theme.ts
```diff
@@ -89,9 +93,15 @@ export class Theme {
     return `--${this.prefix}-${key.slice(2)}`
   }
 
-  #clearNamespace(namespace: string) {
+  clearNamespace(namespace: string, clearOptions: ThemeOptions) {
```
philipp-spiess, line 96: Made this public so we can selectively only clean `@theme default` values... See the test case as to why I think this is necessary.

adamwathan (approved):

PR #14679, opened 2024-10-15
Speed up template migrations

thecrypticace:
This PR does two things:
- Computes UTF-16 string positions in Rust rather than in JS — eliminating a significant number of traversals of the input string
- Applies replacements to the content in ascending order so we only ever move forward through the source string — this lets v8 optimize string concatenation


philipp-spiess (approved):
[wow](https://app.graphite.dev//settings/meme-library?org=tailwindlabs)

PR #14687, opened 2024-10-16
`@apply` is not allowed inside `@keyframes`

RobinMalfait:
This PR makes sure that you cannot use `@apply` inside `@keyframes`. 

While some utilities can be used in `@keyframes`, the moment you introduce a variant, that's not going to work anymore because they need to operate on selectors which `@keyframes` don't have.

This PR now removes all usages of `@apply` in `@keyframes`.

### DIFF
--- a/packages/tailwindcss/src/apply.ts
+++ b/packages/tailwindcss/src/apply.ts
@@ -1,4 +1,4 @@
-import { walk, type AstNode } from './ast'
+import { walk, WalkAction, type AstNode } from './ast'
 import { compileCandidates } from './compile'
 import type { DesignSystem } from './design-system'
 import { escape } from './utils/escape'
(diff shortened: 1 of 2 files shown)


RobinMalfait:
* **#14687** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

thecrypticace (approved):

lorypelli:
why don't you just disallow modifiers, I used `backdrop-filter` in an animation and I'd prefer to use tailwind class (because I'm also setting the `-webkit` vendor prefix as tailwind does for that property, but I'd prefer to have it handled by tailwind, any ideas?

PR #14697, opened 2024-10-17
Bump dependencies to fix vulnerabilities

RobinMalfait:
This PR fixes some package vulnerabilities that you will see when running `npm install`. This PR fixes that by bumping dependencies to get rid of the vulnerabilities.


thecrypticace (changes requested):
Need to `npm install` in `standalone-cli` too to update its lockfile.

thecrypticace (approved):

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Bump dependencies to fix vulnerabilities ([#14697](https://github.com/tailwindlabs/tailwindcss/pull/14697))
```
philipp-spiess, line 12: Did we end up bundling vulnerable code here? "Fix vulnerabilities" makes it sound like we had a vulnerability in Tailwind CSS which might be a bit too alarming.

In the past we used the wording "Bump versions for security vulnerabilities" which gave it a more passive sound. WDYT?

PR #14700, opened 2024-10-17
Migrate static plugins with options to CSS

philipp-spiess:
This PR extends our JS configuration to CSS migration by also allowing `plugins` with options.  

An example of such config would be:

```js
import myPlugin from "./myPlugin";

export default {
  plugins: [
    myPlugin({
      class: "tw",
    }),
  ],
} satisfies Config;
```

If the option object contains only values allowed in our CSS API, we can convert this to CSS entirely:

```css
@plugin './myPlugin' {
  class: 'tw';
}
```

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/format-nodes.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/format-nodes.ts
@@ -18,7 +18,15 @@ export function formatNodes(): Plugin {
     // Format the nodes
     await Promise.all(
       nodesToFormat.map(async (node) => {
-        node.replaceWith(parse(await format(node.toString(), { parser: 'css', semi: true })))
+        node.replaceWith(
+          parse(
+            await format(node.toString(), {
+              parser: 'css',
+              semi: true,
+              singleQuote: true,
+            }),
+          ),
+        )
       }),
     )
   }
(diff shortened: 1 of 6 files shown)


packages/@tailwindcss-upgrade/src/codemods/format-nodes.ts
```diff
@@ -18,7 +18,15 @@ export function formatNodes(): Plugin {
     // Format the nodes
     await Promise.all(
       nodesToFormat.map(async (node) => {
-        node.replaceWith(parse(await format(node.toString(), { parser: 'css', semi: true })))
+        node.replaceWith(
+          parse(
+            await format(node.toString(), {
+              parser: 'css',
+              semi: true,
+              singleQuote: true,
```
philipp-spiess, line 26: I noticed that we inconsistently used single quotes for the `@plugin` and `@config` directives but double-quotes for CSS values so this is just a consistency change.

packages/@tailwindcss-upgrade/src/codemods/migrate-config.ts
```diff
@@ -63,7 +63,22 @@ export function migrateConfig(
...
+        } else {
+          css += `@plugin '${relative}' {\n`
+          for (let [property, value] of Object.entries(plugin.options)) {
+            css += `  ${property}: ${
+              Array.isArray(value)
+                ? value.map((v) => (typeof v === 'string' ? quoteString(v) : '' + v)).join(', ')
+                : typeof value === 'string'
+                  ? quoteString(value)
+                  : '' + value
+            };\n`
```
thecrypticace, line 78: I think I'd split this into multiple concatenations w/ ifs wrapping things rather than inlining all of them. It's feels a bit hard to follow like this.

RobinMalfait (approved):

philipp-spiess:
### Merge activity

* **Oct 18, 9:10 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14700).

PR #14703, opened 2024-10-17
Ensure existing spaces in attribute selectors are valid

RobinMalfait:
This PR fixes an issue where spaces in a selector generated invalid CSS.

Lightning CSS will throw those incorrect lines of CSS out, but if you are in an environment where Lightning CSS doesn't run then invalid CSS is generated.

Given this input:

```html
data-[foo_=_"true"]:flex
```

This will be generated:

```css
.data-\[foo_\=_\"true\"\]\:flex[data-foo=""true] {
  display: flex;
}
```

With this PR in place, the generated CSS will now be:

```css
.data-\[foo_\=_\"true\"\]\:flex[data-foo="true"] {
  display: flex;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Upgrade (experimental)_: Migrate `plugins` with options to CSS ([#14700](https://github.com/tailwindlabs/tailwindcss/pull/14700))
 
+### Fixed
+
+- Ensure existing spaces in attribute selectors are valid ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
+
 ### Changed
 
 - _Upgrade (experimental)_: Don't create `@source` rules for `content` paths that are already covered by automatic source detection ([#14714](https://github.com/tailwindlabs/tailwindcss/pull/14714))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Ensure existing spaces in attribute selectors are valid ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
+- Allow spaces spaces around operators in attribute selector variants ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
 
 ### Changed
 
--- a/packages/tailwindcss/src/variants.test.ts
+++ b/packages/tailwindcss/src/variants.test.ts
@@ -1985,6 +1985,7 @@ test('aria', async () => {
       'aria-checked:flex',
       'aria-[invalid=spelling]:flex',
       'aria-[valuenow=1]:flex',
+      'aria-[valuenow_=_"1"]:flex',
 
       'group-aria-[modal]:flex',
       'group-aria-checked:flex',
@@ -2059,6 +2060,10 @@ test('aria', async () => {
 
     .aria-\\[valuenow\\=1\\]\\:flex[aria-valuenow="1"] {
       display: flex;
+    }
+
+    .aria-\\[valuenow_\\=_\\"1\\"\\]\\:flex[aria-valuenow="1"] {
+      display: flex;
     }"
   `)
   expect(await run(['aria-checked/foo:flex', 'aria-[invalid=spelling]/foo:flex'])).toEqual('')
@@ -2069,6 +2074,8 @@ test('data', async () => {
     await run([
       'data-disabled:flex',
       'data-[potato=salad]:flex',
+      'data-[potato_=_"salad"]:flex',
+      'data-[potato_^=_"salad"]:flex',
       'data-[foo=1]:flex',
       'data-[foo=bar_baz]:flex',
       "data-[foo$='bar'_i]:flex",
@@ -2155,6 +2162,14 @@ test('data', async () => {
       display: flex;
     }
 
+    .data-\\[potato_\\=_\\"salad\\"\\]\\:flex[data-potato="salad"] {
+      display: flex;
+    }
+
+    .data-\\[potato_\\^\\=_\\"salad\\"\\]\\:flex[data-potato^="salad"] {
+      display: flex;
+    }
+
     .data-\\[foo\\=1\\]\\:flex[data-foo="1"] {
       display: flex;
     }
@@ -2171,7 +2186,13 @@ test('data', async () => {
       display: flex;
     }"
   `)
-  expect(await run(['data-disabled/foo:flex', 'data-[potato=salad]/foo:flex'])).toEqual('')
+  expect(
+    await run([
+      'data-[foo_^_=_"bar"]:flex', // Can't have spaces between `^` and `=`
+      'data-disabled/foo:flex',
+      'data-[potato=salad]/foo:flex',
+    ]),
+  ).toEqual('')
 })
 
 test('portrait', async () => {
--- a/packages/tailwindcss/src/variants.test.ts
+++ b/packages/tailwindcss/src/variants.test.ts
@@ -2076,6 +2076,7 @@ test('data', async () => {
       'data-[potato=salad]:flex',
       'data-[potato_=_"salad"]:flex',
       'data-[potato_^=_"salad"]:flex',
+      'data-[potato="^_="]:flex',
       'data-[foo=1]:flex',
       'data-[foo=bar_baz]:flex',
       "data-[foo$='bar'_i]:flex",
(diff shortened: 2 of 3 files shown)


RobinMalfait:
* **#14703** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/tailwindcss/src/variants.ts
```diff
@@ -901,8 +901,10 @@ export function createVariants(theme: Theme): Variants {
 function quoteAttributeValue(value: string) {
   if (value.includes('=')) {
     value = value.replace(/(=.*)/g, (_fullMatch, match) => {
+      let value = match.slice(1).trim()
+
       // If the value is already quoted, skip.
-      if (match[1] === "'" || match[1] === '"') {
+      if (value[0] === "'" || value[0] === '"') {
```
RobinMalfait, line 907: `_` values in the value are already converted to ` ` which means that we can just trim the valid (if we ignore the first `=` character)

adamwathan:
Do we actually want people to be able to write stuff like `data-[foo_=_"true"]:flex` in the first place? You can't put spaces there in CSS so I don't know that it's something we need to support in Tailwind either. I think I'd rather just toss the candidate personally but maybe that's more work?

RobinMalfait:
@adamwathan I noticed it while working on the codemods and created an internal issue so that I didn't forget about it. Fun thing is, that you can put spaces in that position in CSS (see https://jsfiddle.net/ezu8qjdk/) and since arbitrary values should be valid CSS (as an escape hatch if the value isn't used), I figured that it is best to at least generate the correct CSS if you put it in that way.

We can of course throw it out (or even fix it at a codemod level), but since it's valid CSS it might be good to keep it? I'm happy with either solution, just want to fix the scenario that exists today where the currently generated CSS is invalid.

RobinMalfait:
The part I'm missing is that you can't put spaces if you use `[data-foo ^ = "bar"]` (but `[data-foo ^= "bar"]` is valid). Will add an explicit test for this case.

thecrypticace (approved):

packages/tailwindcss/src/variants.test.ts
```diff
@@ -2155,6 +2162,14 @@ test('data', async () => {
       display: flex;
     }
 
+    .data-\\[potato_\\=_\\"salad\\"\\]\\:flex[data-potato="salad"] {
+      display: flex;
+    }
+
+    .data-\\[potato_\\^\\=_\\"salad\\"\\]\\:flex[data-potato^="salad"] {
```
philipp-spiess, line 2169: Is it lightningcss that did remove the spaces again here or did we? I feel like if the argument is "we want to support arbitrary CSS because CSS would also allow that", we should also make sure we output the spaces again in the resulting CSS. Conceptually we just pass-through even though we do auto-quote where we have to.

RobinMalfait: Yep Lightning CSS is cleaning that up for us: https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.data-%5C%5C%5Bpotato_%5C%5C%5E%5C%5C%3D_%5C%5C%5C%22salad%5C%5C%5C%22%5C%5C%5D%5C%5C%3Aflex%5Bdata-potato%20%5E%3D%20%5C%22salad%5C%22%5D%20%7B%5Cn%20%20color%3A%20red%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D


I don't think re-emitting the _exact_ space is necessary (even though we do). This is what we generate:
```css
.data-\\[potato_\\^\\=_\\"salad\\"\\]\\:flex {
  &[data-potato ^= "salad"] {
    display: flex;
  }
}
```

Only thing I care about is that if you pass in valid CSS that valid CSS comes out and that we didn't generate invalid CSS along the way.

adamwathan:
> Fun thing is, that you can put spaces in that position in CSS

Well then…

continue-tv-land-gif-by-impastor

packages/tailwindcss/src/variants.ts
```diff
@@ -900,9 +906,16 @@ export function createVariants(theme: Theme): Variants {
 
 function quoteAttributeValue(value: string) {
   if (value.includes('=')) {
+    // Do not allow invalid operators in attribute values
+    if (/[*$^|~]\s+=/.test(value)) {
```
adamwathan, line 910: Should we try to always precompile regular expressions?

RobinMalfait: That's something we could do, it's also a non-global one (`/g`) so it's stateless.

This only ever runs in `data-*` or `aria-*` variants when using arbitrary values that contain an `=` so I'm assuming this will be used very rarely. Especially because the `aria-foobar` already translates to `aria-[foobar="true"]`, but this doesn't go via this function since it's a bare value instead of an arbitrary value.

adamwathan: I think this condition would break this variant:

```
data-[potato="^_="]
```

Feeling like the better overall solution to this problem should rely on `segment` — do we just need to segment the value on `=` and quote the last chunk if it isn't already quoted? Maybe we don't need to protect people from writing `data-[potato_^_=_foo]` and just let that be user error that generates junk.

adamwathan, line 910: I think this condition would break this variant:

```
data-[potato="^_="]
```

Feeling like the better overall solution to this problem should rely on `segment` — do we just need to segment the value on `=` and quote the last chunk if it isn't already quoted? Maybe we don't need to protect people from writing `data-[potato_^_=_foo]` and just let that be user error that generates junk.

adamwathan (changes requested):

packages/tailwindcss/src/variants.ts
RobinMalfait: Reworked it by using `segment` instead of the regex.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Upgrade (experimental)_: Migrate `plugins` with options to CSS ([#14700](https://github.com/tailwindlabs/tailwindcss/pull/14700))
 
+### Fixed
+
+- Ensure existing spaces in attribute selectors are valid ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
```
philipp-spiess, line 16:
```suggestion
- Support spaces around attribute selectors in arbitrary variants ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
```

Maybe this is a bit more clear? WDYT?

philipp-spiess (approved):
Looks good now and feels like a simple fix. Just one idea regarding the change log entry.

packages/tailwindcss/src/variants.ts
```diff
@@ -898,32 +898,35 @@ export function createVariants(theme: Theme): Variants {
...
-        if (
-          match[match.length - 2] === ' ' &&
-          (trailingCharacter === 'i' ||
-            trailingCharacter === 'I' ||
-            trailingCharacter === 's' ||
-            trailingCharacter === 'S')
-        ) {
-          return `="${match.slice(1, -2)}" ${match[match.length - 1]}`
-        }
+    // Handle regex flags on unescaped values
```
adamwathan, line 912: I know this isn't new to this PR but this comment confused me:

```suggestion
    // Handle case sensitivity flags on unescaped values
```

CHANGELOG.md
adamwathan, line 16: Agree with @philipp-spiess, maybe something like this?

```suggestion
- Allow spaces spaces around operators in attribute selector variants ([#14703](https://github.com/tailwindlabs/tailwindcss/pull/14703))
```

"Arbitrary variants" is specifically stuff like `[&>.foo]` whereas `data-[dessert_=_potato]` is an arbitrary value for a functional variant.

packages/tailwindcss/src/variants.ts
```diff
@@ -898,32 +898,35 @@ export function createVariants(theme: Theme): Variants {
...
+    if (value.length > 1) {
```
adamwathan, line 913: Is it important that we are operating on the non-trimmed value here? It's not obvious to me why we need to have both.

RobinMalfait: I ported it as-is, but can indeed simplify it a bit more. Pushing an update 👍

adamwathan (approved):

PR #14704, opened 2024-10-17
Run tests without optimization

RobinMalfait:
This PR improves the tests by not running the optimization step (via Lightning CSS) in places where we know that nothing should be generated.

In fact, if _we_ generate invalid CSS where we expected that nothing was being generated, then Lightning CSS will throw the CSS out. This means that we have a bug somewhere in our code, and we should fix it but we don't know about it.

Bonus points: this improves the performance of the tests a tiny bit.

### DIFF
--- a/packages/tailwindcss/src/test-utils/run.ts
+++ b/packages/tailwindcss/src/test-utils/run.ts
@@ -6,14 +6,10 @@ export async function compileCss(css: string, candidates: string[] = [], options
   return optimizeCss(build(candidates)).trim()
 }
 
-export async function run(candidates: string[]) {
+export async function run(candidates: string[], { optimize = true } = {}) {
   let { build } = await compile('@tailwind utilities;')
-  return optimizeCss(build(candidates)).trim()
-}
-
-export async function build(candidates: string[]) {
-  let { build } = await compile('@tailwind utilities;')
-  return build(candidates).trim()
+  let css = build(candidates)
+  return optimize ? optimizeCss(css).trim() : css.trim()
 }
 
 export function optimizeCss(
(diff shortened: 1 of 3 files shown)


thecrypticace (approved):

packages/tailwindcss/src/test-utils/run.ts
```diff
@@ -6,9 +6,10 @@ export async function compileCss(css: string, candidates: string[] = [], options
   return optimizeCss(build(candidates)).trim()
 }
 
-export async function run(candidates: string[]) {
+export async function run(candidates: string[], { optimize = true } = {}) {
```
philipp-spiess, line 9: Not a big fan of having to add an object with `optimize` to every call site. How do you feel about this?

I’m wondering if we:
- Should make this the default (or is the number of `run()` calls that do need `optimize: true` even larger than this subset?
- Should instead add a second `runOptimized` export instead of adding a boolean variable?

thecrypticace: Also not a fan but not enough to really care. I think unoptimized is *not* the default. Pretty sure more tests require an optimized CSS result than not but I don't have a good idea of what the function should be called.

RobinMalfait: Funny thing is, I did use a new method initially (https://github.com/tailwindlabs/tailwindcss/pull/14704/commits/14817cba32d36097fff0bc886302a67e69d21624) but `run` vs `runOptimized` or vs `build` all felt wrong, so did a fallback here because typically you run it without.

You could argue that we just don't run the tests through Lightning CSS (which will make the tests bigger but faster).

RobinMalfait:
Closing this, if we want this we can re-enable it later.

PR #14714, opened 2024-10-18
`content` rules from the JS config that are also covered by the automatic source detection should not be migrated to CSS

philipp-spiess:
This PR changes the migration of `content` rules in the JS config to CSS codemods.

When a `content` rule is processed which matches files that are _also matched by the automatic content discovery in v4_, we do not need to emit CSS for that rule. 

Take, for example this v3 configuration file:

```ts

module.exports = {
  content: [
    './src/**/*.{html,js}', 
    './node_modules/my-external-lib/**/*.{html}'
  ],
} satisfies Config
```

Provided the base directories match up, the first rule will also be covered by the automatic content discovery in v4 and thus we only need to convert the second rule to CSS:

```css
@import "tailwindcss";
@source '../node_modules/my-external-lib/**/*.{html}';
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- _Upgrade (experimental)_: `content` rules from the JS configuration that are also covered by the automatic source detection are no longer migrated to CSS ([#14714](https://github.com/tailwindlabs/tailwindcss/pull/14714))
+- _Upgrade (experimental)_: Don't create `@source` rules for `content` paths that are already covered by automatic source detection ([#14714](https://github.com/tailwindlabs/tailwindcss/pull/14714))
 
 ## [4.0.0-alpha.28] - 2024-10-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Upgrade (experimental)_: Migrate `plugins` with options to CSS ([#14700](https://github.com/tailwindlabs/tailwindcss/pull/14700))
 
+### Changed
+
+- _Upgrade (experimental)_: `content` rules from the JS configuration that are also covered by the automatic source detection are no longer migrated to CSS ([#14714](https://github.com/tailwindlabs/tailwindcss/pull/14714))
+
 ## [4.0.0-alpha.28] - 2024-10-17
 
 ### Added
--- a/packages/@tailwindcss-upgrade/src/migrate-js-config.ts
+++ b/packages/@tailwindcss-upgrade/src/migrate-js-config.ts
@@ -1,6 +1,7 @@
+import { Scanner } from '@tailwindcss/oxide'
 import fs from 'node:fs/promises'
 import { dirname } from 'path'
-import type { Config } from 'tailwindcss'
+import { type Config } from 'tailwindcss'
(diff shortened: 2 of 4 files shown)


philipp-spiess:
* **#14714** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

philipp-spiess:
Yo what is this branch name, let me see if I can truncate this in Graphite 🙈

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Changed
+
+- _Upgrade (experimental)_: `content` rules from the JS configuration that are also covered by the automatic source detection are no longer migrated to CSS
```
```suggestion
- _Upgrade (experimental)_: Don't create `@source` rules for `content` paths that are already covered by automatic source detection
```

packages/@tailwindcss-upgrade/src/migrate-js-config.ts
```diff
@@ -158,16 +159,35 @@ function createSectionKey(key: string[]): string {
...
+      if (!autoContentFiles.includes(sourceFile)) {
+        autoContentContainsAllSourceFiles = false
+        break
+      }
+    }
+
+    if (autoContentContainsAllSourceFiles) {
+      warn(
+        'The `content` configuration `${content}` is already included in the automatic content file discovery and will not be migrated.',
+      )
```
adamwathan, line 187: Actually seeing this I think I'd rather just not log anything, it almost reads like the upgrade couldn't be completed and I'm not sure how to say it more clearly without it being too verbose.

packages/@tailwindcss-upgrade/src/migrate-js-config.ts
```diff
@@ -253,3 +273,15 @@ function keyframesToCss(keyframes: Record<string, unknown>): string {
...
+  let scanner = new Scanner({ detectSources: { base } })
+  scanner.scan()
+  return scanner.files
+}
+
+function listSourceContentFiles(source: { base: string; pattern: string }): string[] {
+  let scanner = new Scanner({ sources: [source] })
+  scanner.scan()
+  return scanner.files
+}
```
adamwathan, line 287: Should we avoid the word "content" for code that's just looking at the modern stuff?

Are these names clear enough?

```suggestion
function autodetectedSourceFiles(base: string) {
  let scanner = new Scanner({ detectSources: { base } })
  scanner.scan()
  return scanner.files
}

function patternSourceFiles(source: { base: string; pattern: string }): string[] {
  let scanner = new Scanner({ sources: [source] })
  scanner.scan()
  return scanner.files
}
```

packages/@tailwindcss-upgrade/src/migrate-js-config.ts
philipp-spiess: @adamwathan Yeah I think this makes sense.

philipp-spiess, line 187:
```suggestion
```

packages/@tailwindcss-upgrade/src/migrate-js-config.ts
philipp-spiess: Sounds good. I went with _content_ in the first iteration since we are looking at the `content` array and this is the `content` field migration, but yeah in terms of v4, the term _content_ does not really exists.

adamwathan (approved):

PR #14716, opened 2024-10-18
Vite: Fix issues when loading files via static asset queries

### LINKED ISSUE #14558: [v4] Something has broken since alpha.15, causing imports in CSS files and vite URL imports to not work

v4.0.0-alpha.25

vite 5.4.8

bun 1.1.27

Chrome

macOS

**Reproduction URL**

Unfortunately Stackblitz doesn't support creating a repro since it can't run the tailwind wasm module used in the @tailwindcss/vite plugin.

**Describe your issue**

With v4.0.0-alpha.15, I can have a css file that does:

```
@import "tailwindcss";
@import "common/base.css";
```

and use that stylesheet as a [vite URL import](https://vitejs.dev/guide/assets#explicit-url-imports):

```
import stylesUrl from "./root.css?url";
```

However, when I just upgrade to v4.0.0-alpha.25, I get

1. `Uncaught SyntaxError: The requested module '/app/root.css?url' does not provide an export named 'default'` because the vite URL imports have broken.

If I just import normally without the URL suffix (`import "./root.css";`),

2. Now not all the styles from common/base.css which is imported in ./root.css apply, and trying to do something basic like reassign a CSS variable in ./root.css no longer works:

```
@theme {
	--color-bg: orange;
}
```

That works back on v4.0.0-alpha.15.

Please let me know if I can provide more info, or if instead of Stackblitz, there is another provider that does support wasm.

philipp-spiess:
Fixes: #14558

This PR fixes an issue where our Vite plugin would crash when trying to load stylesheets via certain static asset query parameters:

```ts
import raw from './style.css?raw'
import url from './style.css?url'
```

The proper behavior for our extension is to _not touch these file at all_. The `?raw` identifier should never transform anything and the `?url` one will emit a module which points to the asset URL. However, if that URL is loaded as a stylesheet, another transform hook is called and the file is properly transformed. I verified this in the Vite setup and have added an integration test ensuring these two features work as expected.

I've also greatly reduced the complexity of the Vite playground to make it easier to set up examples like this in the future.


philipp-spiess:
* **#14716** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -261,9 +261,13 @@ function getExtension(id: string) {
...
   let isCssFile =
-    extension === 'css' ||
-    (extension === 'vue' && id.includes('&lang.css')) ||
-    (extension === 'astro' && id.includes('&lang.css'))
+    (extension === 'css' ||
+      (extension === 'vue' && id.includes('&lang.css')) ||
+      (extension === 'astro' && id.includes('&lang.css'))) &&
+    // Don't intercept ?url and ?raw static asset handling
+    !id.includes('?url') &&
+    !id.includes('?raw')
```
adamwathan, line 269: Is there ever a case where these are not at the end of the string? If not maybe `endsWith` is clearer?

bitofbreeze: or could combine into a single check with some regex like `.test(/\?(url|raw)$/)`

aaronadamsCA: This seems to be the regex Vite uses:

https://github.com/vitejs/vite/blob/257883fccc36f27cc879af249eeba353ebe94cdd/packages/vite/src/node/constants.ts#L86

(Sorry, stumbled across this while searching for something unrelated)

philipp-spiess: @adamwathan yeah I think it can be anywhere in the query string. Doesn't even have to be first place so we also need to look for `&url` or `&raw`. Going with a modified version of the regex instead ty @aaronadamsCA (we don't have to worry about workers/shared workers though)

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -261,9 +263,12 @@ function getExtension(id: string) {
...
   let extension = getExtension(id)
   let isCssFile =
-    extension === 'css' ||
-    (extension === 'vue' && id.includes('&lang.css')) ||
-    (extension === 'astro' && id.includes('&lang.css'))
+    (extension === 'css' ||
+      (extension === 'vue' && id.includes('&lang.css')) ||
+      (extension === 'astro' && id.includes('&lang.css'))) &&
+    // Don't intercept special static asset resources
+    !id.match(SPECIAL_QUERY_RE)
```
philipp-spiess, line 270:
```suggestion
    !SPECIAL_QUERY_RE.test(id)
```

thecrypticace (approved):

RobinMalfait (approved):

PR #14717, opened 2024-10-18
Make `decodeArbitraryValue` overridable

RobinMalfait:
This PR allows us to override the `decodeArbitraryValue` function used when parsing candidates and variants.

I'm not 100% happy with this (open for suggestions!) but it was either this, or introducing some class where we mock a method or as a last resort, copy the whole `parseCandidate` implementation and make the necessary changes there when running codemods.

While the last one seems to be the cleanest, I'm afraid that the implementation will diverge over time.

PR #14718, opened 2024-10-18
Override `designSystem` methods by introducing a `mockDesignSystem`

RobinMalfait:
We can definitely change the name, also open for suggestions here.

The idea is as follows:

1. We reuse the existing `designSystem` (and therefore keep all the existing functionality)
2. We override the `decodeArbitraryValues` function such that we don't add spaces around operators anymore. This is a divergence from the actual implementation. This allows us to keep the whitespace from the candidate without trying to be smart about it.
3. When parsing arbitrary variants, we wrap them in `&:is(…)`. We do get rid of this wrapper in this `mockDesignSystem` function as well. This has a downside that if people literally wrote `[&:is(p)]:flex` that it will be converted to `[p]:flex` but I think that's reasonable.

Another thing that this unlocks is that by overriding the `parseCandidate` and `parseVariant` functions, that we can remove the layer of caching. This means that we don't have to cache anymore, and even better, we don't have to use `structuredClone(candidate)` anymore. We used `structuredClone` before to make sure that we didn't mutate existing candidates in the cached design system. But now everything is fresh.

Some examples:

Input:
```html
<div class="[p]:flex"></div>
<div class="has-[p]:flex"></div>
<div class="px-[theme(spacing.4)-1px]"></div>
```

Output before:
```html
<div class="[&:is(p)]:flex"></div>
<div class="has-[&:is(p)]:flex"></div>
<div class="px-[var(--spacing-4)_-_1px]"></div>
```

Output after:
```html
<div class="[p]:flex"></div>
<div class="has-[p]:flex"></div>
<div class="px-[var(--spacing-4)-1px]"></div>
```

### DIFF
--- a/packages/@tailwindcss-upgrade/src/mock-design-system.ts
+++ b/packages/@tailwindcss-upgrade/src/mock-design-system.ts
@@ -1,0 +1,84 @@
+import {
+  parseCandidate,
+  parseVariant,
+  type Candidate,
+  type Variant,
+} from '../../tailwindcss/src/candidate'
+import type { DesignSystem } from '../../tailwindcss/src/design-system'
+import { convertUnderscoresToWhitespace } from '../../tailwindcss/src/utils/decode-arbitrary-value'
+import * as ValueParser from '../../tailwindcss/src/value-parser'
+
+export function mockDesignSystem(designSystem: DesignSystem): DesignSystem {
+  // Custom `parseCandidate` implementation that does two things:
+  // 1. Has custom `decodeArbitraryValue` that does not add whitespace around
+  //    operators.
+  // 2. Does not cache the parsing of candidates
+  designSystem.parseCandidate = (candidate) => {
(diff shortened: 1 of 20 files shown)


RobinMalfait:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14718?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#14718** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#14717** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `next`

<h2></h2>

adamwathan:
Going to close this in favor of #14720 — I think that's the cleaner way to approach this problem 👍

PR #14720, opened 2024-10-18
Minify arbitrary values when printing candidates

RobinMalfait:
This PR will optimize and simplify the candidates when printing the candidate again after running codemods.

When we parse a candidate, we will add spaces around operators, for example `p-[calc(1px+1px)]]` will internally be handled as `calc(1px + 1px)`. Before this change, we would re-print this as: `p-[calc(1px_+_1px)]`.

This PR changes that by simplifying the candidate again so that the output is `p-[calc(1px+1px)]`. In addition, if _you_ wrote `p-[calc(1px_+_1px)]` then we will also simplify it to the concise form `p-[calc(1px_+_1px)]`.

Some examples:

Input:
```html
<div class="[p]:flex"></div>
<div class="[&:is(p)]:flex"></div>
<div class="has-[p]:flex"></div>
<div class="px-[theme(spacing.4)-1px]"></div>
```

Output before:
```html
<div class="[&:is(p)]:flex"></div>
<div class="[&:is(p)]:flex"></div>
<div class="has-[&:is(p)]:flex"></div>
<div class="px-[var(--spacing-4)_-_1px]"></div>
```

Output after:
```html
<div class="[p]:flex"></div>
<div class="[p]:flex"></div>
<div class="has-[p]:flex"></div>
<div class="px-[var(--spacing-4)-1px]"></div>
```

---

This is alternative implementation to #14717 and #14718
Closes: #14717 
Closes: #14718

### DIFF
--- a/packages/@tailwindcss-upgrade/src/template/candidates.test.ts
+++ b/packages/@tailwindcss-upgrade/src/template/candidates.test.ts
@@ -108,40 +108,63 @@ const candidates = [
   ['bg-[#0088cc]/[0.5]', 'bg-[#0088cc]/[0.5]'],
   ['bg-[#0088cc]!', 'bg-[#0088cc]!'],
   ['!bg-[#0088cc]', 'bg-[#0088cc]!'],
+  ['bg-[var(--spacing)-1px]', 'bg-[var(--spacing)-1px]'],
+  ['bg-[var(--spacing)_-_1px]', 'bg-[var(--spacing)-1px]'],
+  ['bg-[-1px_-1px]', 'bg-[-1px_-1px]'],
   ['w-1/2', 'w-1/2'],
+
+  // Keep spaces in strings
+  ['content-["hello_world"]', 'content-["hello_world"]'],
+  ['content-[____"hello_world"___]', 'content-["hello_world"]'],
 ]
 
 const variants = [
-  '', // no variant
-  '*:',
-  'focus:',
-  'group-focus:',
+  ['', ''], // no variant
+  ['*:', '*:'],
+  ['focus:', 'focus:'],
+  ['group-focus:', 'group-focus:'],
 
-  'hover:focus:',
-  'hover:group-focus:',
-  'group-hover:focus:',
-  'group-hover:group-focus:',
+  ['hover:focus:', 'hover:focus:'],
+  ['hover:group-focus:', 'hover:group-focus:'],
+  ['group-hover:focus:', 'group-hover:focus:'],
+  ['group-hover:group-focus:', 'group-hover:group-focus:'],
 
-  'min-[10px]:',
-  // TODO: This currently expands `calc(1000px+12em)` to `calc(1000px_+_12em)`
-  'min-[calc(1000px_+_12em)]:',
+  ['min-[10px]:', 'min-[10px]:'],
 
-  'peer-[&_p]:',
-  'peer-[&_p]:hover:',
-  'hover:peer-[&_p]:',
-  'hover:peer-[&_p]:focus:',
-  'peer-[&:hover]:peer-[&_p]:',
+  // Normalize spaces
+  ['min-[calc(1000px_+_12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px_+12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px+_12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px___+___12em)]:', 'min-[calc(1000px+12em)]:'],
+
+  ['peer-[&_p]:', 'peer-[&_p]:'],
+  ['peer-[&_p]:hover:', 'peer-[&_p]:hover:'],
+  ['hover:peer-[&_p]:', 'hover:peer-[&_p]:'],
+  ['hover:peer-[&_p]:focus:', 'hover:peer-[&_p]:focus:'],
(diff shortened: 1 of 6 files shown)


RobinMalfait:
* **#14720** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@tailwindcss-upgrade/src/template/candidates.test.ts
```diff
@@ -108,40 +108,63 @@ const candidates = [
...
-  'min-[10px]:',
-  // TODO: This currently expands `calc(1000px+12em)` to `calc(1000px_+_12em)`
-  'min-[calc(1000px_+_12em)]:',
-
-  'peer-[&_p]:',
-  'peer-[&_p]:hover:',
-  'hover:peer-[&_p]:',
-  'hover:peer-[&_p]:focus:',
-  'peer-[&:hover]:peer-[&_p]:',
+  ['', ''], // no variant
```
RobinMalfait, line 122: Created a tuple with `in` and `out` (similar to the candidate above)

packages/@tailwindcss-upgrade/src/template/candidates.test.ts
```diff
@@ -108,40 +108,63 @@ const candidates = [
...
+  ['group-hover:focus:', 'group-hover:focus:'],
+  ['group-hover:group-focus:', 'group-hover:group-focus:'],
+
+  ['min-[10px]:', 'min-[10px]:'],
+
+  // Normalize spaces
+  ['min-[calc(1000px_+_12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px_+12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px+_12em)]:', 'min-[calc(1000px+12em)]:'],
+  ['min-[calc(1000px___+___12em)]:', 'min-[calc(1000px+12em)]:'],
```
adamwathan, line 138: It looks like we're missing tests for the cases where it is unsafe to remove the spaces (I don't know what these are off the top of my head but I know they exist).

Can we figure out what those cases are and make sure we handle them properly? Or do they actually not exist and CSS is just stupid for not letting us omit these stupid spaces? 😄

adamwathan: Tried to look into this and I can't actually find any situations where it's ambiguous in reality? There were some talks about eventually supporting stuff like `calc(min-content * 50%)` and then the `-` in `min-content` is ambiguous as to whether it's subtraction or part of a keyword, but that isn't actually supported in CSS and it looks like `calc-size(…)` is what they are going to use for that stuff instead.

So fuck it, probably fine 😁

RobinMalfait: I was thinking about this as well and very hard to come up with some tests.

I did see that we have `to-zero` as an edge case that we handle. Let me add a test for that.

packages/@tailwindcss-upgrade/src/template/candidates.ts
```diff
@@ -138,3 +139,73 @@ function escapeArbitrary(input: string) {
     .replaceAll('_', String.raw`\_`) // Escape underscores to keep them as-is
     .replaceAll(' ', '_') // Replace spaces with underscores
 }
+
+function trimWhitespace(input: string) {
```
adamwathan, line 143: Looks like this and `escapeArbitrary` are never used separately, maybe should just make this one function like `toCandidateArbitraryValue` or something? Not totally sure on the name but really we're taking a CSS value and converting it to the arbitrary value of an actual class string.

Either way I don't think `trimWhitespace` is the right name since to me at least "trim" implies whitespace at the start and end, not whitespace inside the string.

RobinMalfait: `minifyArbitraryValue` or `printArbitraryValue`. It's also used in variant and modifier positions so don't think the `candidate` in the name is a good idea.

Merging the functionality however is a good idea.

RobinMalfait: Or `optimizeArbitraryValue`

RobinMalfait: You know what, we already have `printVariant` and `printCandidate`, let's go with `printArbitraryValue`

packages/@tailwindcss-upgrade/src/template/candidates.ts
```diff
@@ -133,8 +134,75 @@ function printVariant(variant: Variant) {
...
+      // Operators
+      (node.value === '+' || node.value === '-' || node.value === '*' || node.value === '/')
+    ) {
+      let idx = parentArray.indexOf(node) ?? -1
+
+      // This should not be possible
+      if (idx === -1) return
+
+      let previous = parentArray[idx - 1]
+      if (previous?.kind !== 'separator' || previous.value !== ' ') return
```
adamwathan, line 157: What actually happens when we return `undefined` here? We're still interpolating the return value of this function into the class name at the call site.

RobinMalfait: The surrounding `walk()` will keep walking (equivalent to `WalkAction.Continue`) because it could be that you have multiple `calc(…)` calls for example and if we can't upgrade the first one, we might be able to upgrade the second one.

adamwathan (approved):

PR #14734, opened 2024-10-21
Update Important Selector Strategy to apply styles to wrapper component also

nickfujita:
Enables tailwind styles to also be applied to the wrapper component containing the important selector.

- Adds the wrapper element with the selector to also be included in the output styles.
- Adjusts removal of important selector before reapplication for @apply selectors

More info
https://github.com/tailwindlabs/tailwindcss/discussions/14735


nickfujita:
@RobinMalfait Thanks for self assigning this one. I was able to resolve the issue with the one test not passing that I originally had. All tests are now passing.

nickfujita:
Hey @RobinMalfait just wanted to see if I could follow up with you on this enhancement. We've been using our fork successfully with these changes, and would love to get back to main branch soon.

RobinMalfait:
Hey!

A small update: most projects don't require the `#app` to be on the element or on the parent, but typically rely on the `#app` being on the parent. This PR seems to be a good fix if you require that behavior, but for most projects this means that the CSS will be bigger than it needs to be.

I think because this problem solves a problem you are having, that a fork is not a bad idea in this case. Alternatively you could use `patch-package` in your repo such that you can keep using Tailwind CSS, but apply this one change.

Going to close this PR for now, and if more people are running into this, we can revisit this behavior.

Thanks again for the PR, really appreciate that!

PR #14738, opened 2024-10-21
Remove `layer(utilities)` if imports contain `@utility`

RobinMalfait:
We have a migration that adds the `layer(…)` next to the `@import` depending on the order of original values. For example:
```css
@import "tailwindcss/utilities":
@import "./foo.css":
@import "tailwindcss/components":
```

Will be turned into:
```css
@import "tailwindcss":
@import "./foo.css" layer(utilities):
```

Because it used to exist between `utilities` and `components`. Without this it would be _after_ `components`.

This results in an issue if an import has (deeply) nested `@utility` at-rules after migrations. This is because if this is generated:
```css
/* ./src/index.css */
@import "tailwindcss";
@import "./foo.css" layer(utilities);

/* ./src/foo.css */
@utility foo {
  color: red;
}
```

Once we interpret this (and thus flatten it), the final CSS would look like:
```css
@layer utilities {
  @utility foo {
    color: red;
  }
}
```

This means that `@utility` is not top-level and an error would occur.

This fixes that by removing the `layer(…)` from the import if the imported file (or any of its children) contains an `@utility`. This is to ensure that once everything is imported and flattened, that all `@utility` at-rules are top-level.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -151,6 +151,11 @@ async function run() {
 
     // Cleanup `@import "…" layer(utilities)`
     for (let sheet of stylesheets) {
+      // If the `@import` contains an injected `layer(…)` we need to remove it
+      if (!Array.from(sheet.importRules).some((node) => node.raws.tailwind_injected_layer)) {
+        continue
+      }
+
       let hasAtUtility = false
 
       // Only remove the `layer(…)` next to the import, if any of the children
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -149,6 +149,42 @@ async function run() {
       error(`${e}`)
     }
 
+    // Cleanup `@import "…" layer(utilities)`
+    for (let sheet of stylesheets) {
+      let hasAtUtility = false
+
+      // Only remove the `layer(…)` next to the import, if any of the children
+      // contains an `@utility`. Otherwise the `@utility` will not be top-level.
+      {
+        sheet.root.walkAtRules('utility', () => {
+          hasAtUtility = true
+          return false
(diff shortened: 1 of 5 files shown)


RobinMalfait:
* **#14738** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@tailwindcss-upgrade/src/stylesheet.ts
```diff
@@ -197,26 +197,26 @@ export class Stylesheet {
...
+      if (isConvertible) {
+        convertiblePaths.push(path)
       } else {
-        nonConvertablePaths.push(path)
+        nonConvertiblePaths.push(path)
       }
     }
 
-    return { convertablePaths, nonConvertablePaths }
+    return { convertiblePaths, nonConvertiblePaths }
```
thecrypticace, line 219: lol what a typo 🤣

RobinMalfait: To be fair, `convertable` makes more sense to my non-native-English brain lol

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -151,6 +151,11 @@ async function run() {
 
     // Cleanup `@import "…" layer(utilities)`
     for (let sheet of stylesheets) {
+      // If the `@import` contains an injected `layer(…)` we need to remove it
+      if (!Array.from(sheet.importRules).some((node) => node.raws.tailwind_injected_layer)) {
+        continue
+      }
+
```
thecrypticace, line 158: I think you can also skip stylesheets with parent stylesheets yeah?

thecrypticace (approved):

PR #14745, opened 2024-10-21
Prepare preflight for border compatibility

RobinMalfait:
This PR prepares the `preflight.css` so that we can introduce border style compatibility in a future PR.

PR #14746, opened 2024-10-21
Add codemod for border style compatibility

RobinMalfait:
This PR adds a codemod that ensures that the border styles from Tailwind CSS v3 work as expected once your project is migrated to Tailwind CSS v4.

In Tailwind CSS v3, the default border color is `colors.gray.200` and in Tailwind CSS v4 the default border color is `currentColor`.

Similarly in Tailwind CSS v3, DOM elements such as `input`, `select`, and `textarea` have a border width of `0px`, in Tailwind CSS v4, we don't change the border width of these elements and keep them as `1px`.

If your project happens to already use the same value for the default border color (`currentColor`) as we use in Tailwind CSS v4, then nothing happens. But this is very unlikely, so we will make sure that we honor your `borderColor.DEFAULT` value.

If you didn't change the default values in your `tailwind.config.js`, then we will inject compatibility CSS using the default Tailwind CSS v3 values to ensure the default color and width are applied correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Upgrade (experimental)_: Migrate `max-w-screen-*` utilities to `max-w-[var(…)]`([#14754](https://github.com/tailwindlabs/tailwindcss/pull/14754))
 - _Upgrade (experimental)_: Migrate `@variants` and `@responsive` directives ([#14748](https://github.com/tailwindlabs/tailwindcss/pull/14748))
 - _Upgrade (experimental)_: Migrate `@screen` directive ([#14749](https://github.com/tailwindlabs/tailwindcss/pull/14749))
+- _Upgrade (experimental)_: Generate compatibility styles for legacy default border color ([#14746](https://github.com/tailwindlabs/tailwindcss/pull/14746))
+- _Upgrade (experimental)_: Generate compatibility styles for legacy default border width on form elements ([#14746](https://github.com/tailwindlabs/tailwindcss/pull/14746))
 
 ### Fixed
 
--- a/integrations/upgrade/js-config.test.ts
+++ b/integrations/upgrade/js-config.test.ts
@@ -1,4 +1,4 @@
-import { expect } from 'vitest'
+import { describe, expect } from 'vitest'
 import { css, html, json, test, ts } from '../utils'
 
 test(
@@ -91,6 +91,40 @@ test(
       "
       --- src/input.css ---
       @import 'tailwindcss';
+
+      /*
+        The default border color has changed to \`currentColor\` in Tailwind CSS v4,
+        so we've added these compatibility styles to make sure everything still
+        looks the same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add an explicit border
+        color utility to any element that depends on these defaults.
+      */
+      @layer base {
+        *,
+        ::after,
+        ::before,
+        ::backdrop,
+        ::file-selector-button {
+          border-color: var(--color-gray-200, currentColor);
+        }
+      }
+
+      /*
+        Form elements have a 1px border by default in Tailwind CSS v4, so we've
+        added these compatibility styles to make sure everything still looks the
+        same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add \`border-0\` to
+        any form elements that shouldn't have a border.
+      */
+      @layer base {
+        input:where(:not([type='button'], [type='reset'], [type='submit'])),
+        select,
+        textarea {
+          border-width: 0;
+        }
+      }
 
       @source '../node_modules/my-external-lib/**/*.{html}';
 
@@ -212,6 +246,40 @@ test(
       --- src/input.css ---
       @import 'tailwindcss';
 
+      /*
+        The default border color has changed to \`currentColor\` in Tailwind CSS v4,
+        so we've added these compatibility styles to make sure everything still
+        looks the same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add an explicit border
+        color utility to any element that depends on these defaults.
+      */
+      @layer base {
+        *,
+        ::after,
+        ::before,
+        ::backdrop,
+        ::file-selector-button {
+          border-color: var(--color-gray-200, currentColor);
+        }
+      }
+
+      /*
+        Form elements have a 1px border by default in Tailwind CSS v4, so we've
+        added these compatibility styles to make sure everything still looks the
+        same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add \`border-0\` to
+        any form elements that shouldn't have a border.
+      */
+      @layer base {
+        input:where(:not([type='button'], [type='reset'], [type='submit'])),
+        select,
+        textarea {
+          border-width: 0;
+        }
+      }
+
       @plugin '@tailwindcss/typography';
       @plugin '../custom-plugin' {
         is-null: null;
@@ -279,6 +347,40 @@ test(
       --- src/input.css ---
       @import 'tailwindcss';
 
+      /*
+        The default border color has changed to \`currentColor\` in Tailwind CSS v4,
+        so we've added these compatibility styles to make sure everything still
+        looks the same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add an explicit border
+        color utility to any element that depends on these defaults.
+      */
+      @layer base {
+        *,
+        ::after,
+        ::before,
+        ::backdrop,
+        ::file-selector-button {
+          border-color: var(--color-gray-200, currentColor);
+        }
+      }
+
+      /*
+        Form elements have a 1px border by default in Tailwind CSS v4, so we've
+        added these compatibility styles to make sure everything still looks the
+        same as it did with Tailwind CSS v3.
+
+        If we ever want to remove these styles, we need to add \`border-0\` to
+        any form elements that shouldn't have a border.
+      */
+      @layer base {
+        input:where(:not([type='button'], [type='reset'], [type='submit'])),
+        select,
+        textarea {
+          border-width: 0;
+        }
+      }
+
       @theme {
         --color-gray-50: oklch(0.985 0 0);
         --color-gray-100: oklch(0.97 0 0);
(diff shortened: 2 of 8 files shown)


RobinMalfait:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14746?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#14746** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#14745** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14745?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `next`

<h2></h2>

packages/@tailwindcss-upgrade/src/codemods/migrate-border-compatibility.ts
```diff
@@ -0,0 +1,213 @@
...
+   * these compatibility styles to your project.
+   *
+   * It is recommended to migrate your project, and delete this layer once you
+   * are done.
+   *
+   * - All elements with a 'border' class will default to 'currentColor' in
+   *   Tailwind CSS v4, instead of 'border-gray-200' in Tailwind CSS v3.
+   * - All inputs, selects, and textareas will have a 'border-width: 1px' in
+   *   Tailwind CSS v4, instead of 'border-width: 0' in Tailwind CSS v3.
+   */
```
RobinMalfait, line 27: Can probably simplify this, but wanted to include _why_ we are doing this and how you could/should proceed.

packages/@tailwindcss-upgrade/src/migrate.ts
```diff
@@ -36,12 +37,13 @@ export async function migrateContents(
 
   return postcss()
     .use(migrateAtApply(options))
-    .use(migrateThemeToVar(options))
     .use(migrateMediaScreen(options))
     .use(migrateAtLayerUtilities(stylesheet))
     .use(migrateMissingLayers())
     .use(migrateTailwindDirectives(options))
     .use(migrateConfig(stylesheet, options))
+    .use(migrateBorderCompatibility(options))
```
RobinMalfait, line 45: Requires the tailwind directives to `@import` migrations

packages/@tailwindcss-upgrade/src/migrate.ts
```diff
@@ -36,12 +37,13 @@ export async function migrateContents(
...
+    .use(migrateThemeToVar(options))
```
RobinMalfait, line 46: Using `theme(…)` inside of the compatibility CSS, so needed to move this after the `migrateBorderCompatibility` migration.

packages/@tailwindcss-upgrade/src/codemods/migrate-border-compatibility.ts
```diff
@@ -0,0 +1,200 @@
...
+  @layer base {
```
RobinMalfait, line 27: In my testing the `@layer base` isn't strictly required (once the previous PR where we updated `preflight.css` is in place). But I think this is still nicer because:

1. It is about the reset that already lives in the `base` layer
2. It groups the compatibility CSS together, which is nice if you look at the CSS files we've updated

integrations/upgrade/js-config.test.ts
```diff
@@ -468,3 +642,218 @@ test(
...
+         * - All inputs, selects, and textareas will have a 'border-width: 1px' in
+         *   Tailwind CSS v4, instead of 'border-width: 0' in Tailwind CSS v3.
+         */
+        @layer base {
+          *,
+          ::after,
+          ::before,
+          ::backdrop,
+          ::file-selector-button {
+            border-color: var(--border-color, var(--color-gray-200, currentColor));
```
RobinMalfait, line 774: This now contains `var(--border-color)` because it existed in your own tailwind.config.js file. We could also always generate this instead of just `var(--color-gray-200, currentColor)`.

integrations/upgrade/js-config.test.ts
```diff
@@ -468,3 +642,218 @@ test(
...
+         * - All inputs, selects, and textareas will have a 'border-width: 1px' in
+         *   Tailwind CSS v4, instead of 'border-width: 0' in Tailwind CSS v3.
+         */
+        @layer base {
+          input:where(:not([type='button'], [type='reset'], [type='submit'])),
+          select,
+          textarea {
+            border-width: 0;
+          }
+        }
```
RobinMalfait, line 850: Notice that the default styles for the `border-color` are not here because they resolve to `currentColor` which is also the default in v4. Therefore if we would've added it, it would be additional CSS doing the same thing.

packages/@tailwindcss-upgrade/src/codemods/migrate-border-compatibility.test.ts
```diff
@@ -0,0 +1,235 @@
+import { __unstable__loadDesignSystem } from '@tailwindcss/node'
+import dedent from 'dedent'
+import postcss from 'postcss'
+import { expect, it } from 'vitest'
+import { formatNodes } from './format-nodes'
+import { migrateBorderCompatibility } from './migrate-border-compatibility'
+
+const css = dedent
+
+async function migrate(input: string) {
```
philipp-spiess, line 10: Let's add a test here as well with a `Config` that does define `borderColor.DEFAULT` and one where said value uses a `theme()` function. I saw code related to it in your implementation 🙂

philipp-spiess (commented):
LGTM, but we may need to tweak the messaging a bit. I provided a suggestion inline

packages/@tailwindcss-upgrade/src/codemods/migrate-border-compatibility.test.ts
RobinMalfait: In this file it's a bit tricky since we resolve user configs, so I added dedicated tests for all of these in the integration tests:
https://github.com/tailwindlabs/tailwindcss/blob/e6c13b968986f0469fe7d67aa78c4d2c213978be/integrations/upgrade/js-config.test.ts#L646

The backwards compatibility CSS uses the `theme(…)` function, which gets replaced which you can also see in that file above.

I can of course add some tests to this file as well where I mock the user config. But it was easier to write integration tests for.

philipp-spiess: oh I skipped over this file since there were so many diffs because of the new comments 🙈 Thanks!

integrations/upgrade/js-config.test.ts
philipp-spiess: @RobinMalfait IIRC in the call yesterday we discussed maybe inlining the value of `--border-color` and omitting it from the created CSS (since it doesn't do anything for the core utilities anyways). WDYT about that approach?

integrations/upgrade/js-config.test.ts
```diff
@@ -468,3 +642,218 @@ test(
...
+
+        @theme {
+          --border-color: currentColor;
+        }
```
philipp-spiess, line 854: might not need that in this case either

packages/@tailwindcss-upgrade/src/codemods/migrate-border-compatibility.test.ts
RobinMalfait: Thank you GitHub for hiding the diff 😂

integrations/upgrade/js-config.test.ts
RobinMalfait: This is a side effect of another codemod where we convert the js config to a CSS-based `@theme` config. 

We have another card in Basecamp where we should purge these kinds of unnecessary values so I feel like handling this where we remove it should not be part of this codemod. 🤔

integrations/upgrade/js-config.test.ts
RobinMalfait: Yeah not a bad idea. I started with resolving the value using `theme(…)` and since it generated the variable and used it looked nice because you can still update the default.

However, the moment you get rid of the backwards compatibility CSS, then the variable is useless and vice versa.

Let me inline it instead.

packages/@tailwindcss-upgrade/src/index.test.ts
```diff
@@ -149,6 +184,39 @@ it('should migrate a stylesheet (with imports)', async () => {
...
+    @layer base {
+      *,
+      ::after,
+      ::before,
+      ::backdrop,
+      ::file-selector-button {
+        border-color: var(--color-gray-200, currentColor);
+      }
+    }
+    /*
```
philipp-spiess, line 205:
```suggestion
    }

    /*
```

Seems like something screws up the empty line between the two blocks 😢

packages/@tailwindcss-upgrade/src/index.test.ts
```diff
@@ -175,6 +243,38 @@ it('should migrate a stylesheet (with preceding rules that should be wrapped in
     /**! My license comment */
     @import 'tailwindcss';
     @config './tailwind.config.js';
+    /*
```
philipp-spiess, line 246: Unsure why there is no empty line above this one here?

RobinMalfait: Fun part is that the space between the `@import` and `@config` is also gone while it should be there. 

So I feel like whitespace handling might need a dedicated PR to get it all right. But I'll take a look at the other cases to see if it's easily fixable or not.

Bigger problem is that in isolation it's correct, but the accumulation of changes is not.

packages/@tailwindcss-upgrade/src/index.test.ts
RobinMalfait: Figured it out. In this case the test is performing migrations and once done formats the nodes on a stringified version. The actual integration tests don't have this issue.

packages/@tailwindcss-upgrade/src/index.test.ts
RobinMalfait: Figured it out. In this case the test is performing migrations and once done formats the nodes on a stringified version. The actual integration tests don't have this issue.

packages/@tailwindcss-upgrade/src/index.test.ts
philipp-spiess: If it works in the integration tests this is fine!

philipp-spiess (approved):

philipp-spiess:
### Merge activity

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Upgrade (experimental)_: Migrate `plugins` with options to CSS ([#14700](https://github.com/tailwindlabs/tailwindcss/pull/14700))
 - _Upgrade (experimental)_: Allow JS configuration files with `corePlugins` options to be migrated to CSS ([#14742](https://github.com/tailwindlabs/tailwindcss/pull/14742))
+- _Upgrade (experimental)_: Add codemod for border style compatibility ([#14746](https://github.com/tailwindlabs/tailwindcss/pull/14746))
```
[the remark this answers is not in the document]
RobinMalfait: Ah yep, I condensed it into a single entry, I like this, thanks!

PR #14748, opened 2024-10-22
Add codemod for migrating `@variants` and `@responsive` directives

RobinMalfait:
This PR migrates the `@variants` and `@responsive` directives.

In Tailwind CSS v2, these were used to generate certain variants of responsive variants for the give classes. In Tailwind CSS v3, these still worked but were implemented as a no-op such that these directives don't end up in your final CSS.

In Tailwind CSS v4, these don't exist at all anymore, so we can safely get rid of them by replacing them with their contents.

Input:
```css
@variants hover, focus {
  .foo {
    color: red;
  }
}

@responsive {
  .bar {
    color: blue;
  }
}
```

Output:
```css
.foo {
  color: red;
}

.bar {
  color: blue;
}
```

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
@@ -400,3 +400,19 @@ it('should drop `@tailwind variants;`', async () => {
     `),
   ).toEqual('')
 })
+
+it('should replace `@responsive` with its children', async () => {
+  expect(
+    await migrate(css`
+      @responsive {
+        .foo {
+          color: red;
+        }
+      }
+    `),
+  ).toMatchInlineSnapshot(`
+    ".foo {
+      color: red;
+    }"
+  `)
+})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Upgrade (experimental)_: Migrate `plugins` with options to CSS ([#14700](https://github.com/tailwindlabs/tailwindcss/pull/14700))
 - _Upgrade (experimental)_: Allow JS configuration files with `corePlugins` options to be migrated to CSS ([#14742](https://github.com/tailwindlabs/tailwindcss/pull/14742))
+- _Upgrade (experimental)_: Migrate `@variants` and `@responsive` directives ([#14748](https://github.com/tailwindlabs/tailwindcss/pull/14748))
 
 ### Fixed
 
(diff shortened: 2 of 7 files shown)


RobinMalfait:
* **#14749** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14749?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14748** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14748?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

RobinMalfait:
The confusing part here is that if you use `@variants hover`
```css
@variants hover {
  .foo {
    color: red;
  }
}
```

If you then use it in your html as:
```html
<div class="hover:foo focus:foo foo">hello world</div>
```

Then the following CSS is being generated:
```css
.foo {
  color: red;
}
.hover\:foo:hover {
  color: red;
}
.focus\:foo:focus {
  color: red;
}
```

Which means that the `variants` listed in `@variants` don't matter. But `foo` is treated as a utility. So maybe the better solution is to wrap this in `@layer utilities` and migrate it to an `@utility` instead 🤔

philipp-spiess:
@RobinMalfait Ah, yeah, I think you're right we should migrate this to `@utility` to not break stuff. Fun that it doesn't matter which variants you used anymore but makes it easier for us at least 🙈

RobinMalfait:
Haha yeah that is a bit confusing. In v2 we generated everything ahead-of-time, and later introduced the just-in-time in v3. So in v2 we needed that information to generate all the CSS.

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
```diff
@@ -400,3 +400,37 @@ it('should drop `@tailwind variants;`', async () => {
...
+          color: red;
+        }
+      }
+    `),
+  ).toMatchInlineSnapshot(`
+    "@variants hover, focus {
+      .foo {
+        color: red;
+      }
+    }"
```
thecrypticace, line 418: Either the test name is wrong or the snapshot is wrong

philipp-spiess (changes requested):
Need to add the `migrateVariantsDirective` rule to the defaults now 🙈  Let's add a full integration test to make sure it indeed works

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
RobinMalfait: Deleted it now; This was moved to a separate file. Should be all good now.

RobinMalfait:
Should be good to go now

philipp-spiess (approved):

philipp-spiess:
### Merge activity

PR #14750, opened 2024-10-22
Add `postcss` as a dependency of `@tailwindcss/postcss`

RobinMalfait:
This PR adds `postcss` as a dependency of `@tailwindcss/postcss`. 

If you are in an environment with Next.js where you can use the `@tailwindcss/postcss` package, then `postcss` is required.

If you have `postcss` in your `package.json`, then everything is fine, however if you don't then you will get an error that `postcss` cannot be found.

Note: this only happens when you are using `npm`, if you are using `pnpm`, then the `postcss` package can be resolved correctly and you won't run into issues. This is also why the integration tests just worked (because we use `pnpm`).

To make sure that this package works for most people in most environments, we explicitly add `postcss` as a dependency of `@tailwindcss/postcss`.

---

I wanted to create an integration test for this to make sure this works, but we are currently using `pnpm` with some of `pnpm`'s features to make sure that we can override dependencies that point to `.tgz` files.


philipp-spiess (approved):

RobinMalfait:
* **#14750** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

PR #14752, opened 2024-10-22
Adjust CSS resolver heuristics

philipp-spiess:
When we implemented the CSS import resolution system, we found out a nasty detail about CSS imports in that files without a relative path prefix would still be relative to the source file. E.g.:

```css
@import 'foo.css';
```

Should first look for the file `foo.css` in the same directory. To make this cost as cheap as possible, we limited this by a heuristics to only apply the auto-relative imports for files with a file extension.

Naturally, while testing v4 on more templates, we found that it's common for people to omit the file extension when loading css file. The above could also be written as such:

```css
@import 'foo';
```

So we are lifting the heuristics for better interop with the ecosystem. Note that we still look for specific imports (`tailwindcss` and `tailwindcss/*`) where we do not do the additional lookup since this is a module level import we expect most users to be adding in their app.


philipp-spiess:
* **#14752** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14752?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

philipp-spiess:
Going to close this one in favor of a codemod that updates the CSS file to append `./` if the file is locally. This will make compilations much faster while avoiding unexpected behavior.

PR #14755, opened 2024-10-22
Upgrade: Rewrite imports of relative files to use relative file paths

philipp-spiess:
When we implemented the CSS import resolution system, we found out a detail about CSS imports in that files without a relative path prefix would still be relative to the source file. E.g.:

```css
@import 'foo.css';
```

Should first look for the file `foo.css` in the same directory. To make this cost as cheap as possible, we limited this by a heuristics to only apply the auto-relative imports for files with a file extension.

Naturally, while testing v4 on more templates, we found that it's common for people to omit the file extension when loading css file. The above could also be written as such:

```css
@import 'foo';
```

To improve this, we have two options:

- We either remove the heuristics, making every `@import` more expensive because we have to check for relative files.
- We upgrade our codemods to rewrite `@import` statements to be explicitly relative.

Because we really care about performance, we opted to go with the latter option. This PR adds the codemod and removes the heuristics so we resolve CSS files similar to how you would resolve JS files.


philipp-spiess:
* **#14755** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@tailwindcss-upgrade/src/codemods/migrate-import.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      }),
+    )
+    .process(input, { from: expect.getState().testPath })
+    .then((result) => result.css)
+}
+
+it('prints relative file imports as relative paths', async () => {
+  expect(
+    await migrate(css`
+      @import 'fixtures/test';
```
RobinMalfait, line 27:
```suggestion
      @import 'fixtures/test';
      @import './fixtures/test';
```

RobinMalfait (commented):
Looks good, we do bail when it starts with `.` which means that
```css
@import "./foo"
```

Stays as-is instead of resulting in:
```css
@import "./foo.css"
```

RobinMalfait (approved):

philipp-spiess:
### Merge activity

CHANGELOG.md
```diff
@@ -30,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- No longer resolve filenames without a prefix as relative files (e.g. `@import 'styles.css'` instead of `@import './styles.css'`) ([#14755](https://github.com/tailwindlabs/tailwindcss/pull/14755))
```
adamwathan, line 34:
```suggestion
- Require a relative path prefix for importing relative CSS files (e.g. `@import './styles.css'` instead of `@import 'styles.css'`) ([#14755](https://github.com/tailwindlabs/tailwindcss/pull/14755))
```

adamwathan:
Can't quite tell from the changes — do we support this now or no?

```css
@import "some-relative-file.css";
```

IMO we shouldn't if we do currently.

adamwathan:
> With this codemod, this will be converted to:
> 
> ```css
> @import "./some-relative-file.css";
> ```
> 
> If the `some-relative-file.css` can be found on disk.

Yeah but say you don't run the codemod, does our `@import` handling handle this?

```css
@import "some-relative-file.css";
```

IMO it should error.

adamwathan (approved):

philipp-spiess:
@adamwathan Yeah this change removed the logic where it automatically appends the `./` so we now error with this in the code.

lubomirblazekcz:
Both 

> > With this codemod, this will be converted to:
> > ```css
> > @import "./some-relative-file.css";
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If the `some-relative-file.css` can be found on disk.
> 
> Yeah but say you don't run the codemod, does our `@import` handling handle this?
> 
> ```css
> @import "some-relative-file.css";
> ```
> 
> IMO it should error.

Both is valid css import and should work though. Just found out that this broke bunch of imports in my css sice the update to latest beta. Should I create a new issue?

PR #14765, opened 2024-10-23
Refactor `@media` handling

RobinMalfait:
This PR is an internal refactor to improve the way we handle at-rules. Essentially we check for the `@` symbol and bail early if that's not the case.

We also improve the way `@media` is being handled by first checking for `@media`, and then handle all the params separately.

Last but not least, we also inverted the `@theme` check condition such that it is handled the same way we handle everything else.

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -258,54 +258,55 @@ async function parseCss(
       return WalkAction.Skip
     }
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
* **#14765** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14765?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

thecrypticace (approved):

PR #14766, opened 2024-10-23
Hide skipped tests

RobinMalfait:
Very small quality of life improvement, but let's hide skipped tests. The summary still shows that there are skipped tests, but this makes the UI much better. Especially when debugging a single test, then you don't need to scroll up all the time.

PR #14767, opened 2024-10-23
Handle new `source(…)`

RobinMalfait:
This PR introduces a new `source(…)` argument and improves on the existing `@source`. The goal of this PR is to make the automatic source detection configurable, let's dig in.

By default, we will perform automatic source detection starting at the current working directory. Auto source detection will find plain text files (no binaries, images, ...) and will ignore git-ignored files.

If you want to start from a different directory, you can use the new `source(…)` next to the `@import "tailwindcss/utilities" layer(utilities) source(…)`.

E.g.:

```css
/* ./src/styles/index.css */
@import 'tailwindcss/utilities' layer(utilities) source('../../');
```

Most people won't split their source files, and will just use the simple `@import "tailwindcss";`, because of this reason, you can use `source(…)` on the import as well:

E.g.:

```css
/* ./src/styles/index.css */
@import 'tailwindcss' source('../../');
```

Sometimes, you want to rely on auto source detection, but also want to look in another directory for source files. In this case, yuo can use the `@source` directive:

```css
/* ./src/index.css */
@import 'tailwindcss';

/* Look for `blade.php` files in `../resources/views` */
@source '../resources/views/**/*.blade.php';
```

However, you don't need to specify the extension, instead you can just point the directory and all the same automatic source detection rules will apply.

```css
/* ./src/index.css */
@import 'tailwindcss';

@source '../resources/views';
```

If, for whatever reason, you want to disable the default source detection feature entirely, and only want to rely on very specific glob patterns you define, then you can disable it via `source(none)`.

```css
/* Completely disable the default auto source detection */
@import 'tailwindcss' source(none);

/* Only look at .blade.php files, nothing else  */
@source "../resources/views/**/*.blade.php";
```

Note: even with `source(none)`, if your `@source` points to a directory, then auto source detection will still be performed in that directory. If you don't want that, then you can simply add explicit files in the globs as seen in the previous example.

```css
/* Completely disable the default auto source detection */
@import 'tailwindcss' source(none);

/* Run auto source detection in `../resources/views` */
@source "../resources/views";
```

### DIFF
--- a/crates/oxide/src/lib.rs
+++ b/crates/oxide/src/lib.rs
@@ -229,10 +229,10 @@ impl Scanner {
 
                 expression
                     .into_iter()
-                    .filter_map(|expanded_pattern| expanded_pattern.map(|x| x.into()).ok())
+                    .filter_map(Result::ok)
                     .map(move |pattern| GlobEntry {
                         base: source.base.clone(),
-                        pattern,
+                        pattern: pattern.into(),
                     })
                     .collect::<Vec<_>>()
             })
--- a/crates/oxide/src/lib.rs
+++ b/crates/oxide/src/lib.rs
@@ -62,9 +62,6 @@ pub struct GlobEntry {
 
 #[derive(Debug, Clone, Default)]
 pub struct Scanner {
-    /// Auto content configuration
-    detect_sources: Option<DetectSources>,
-
     /// Glob sources
     sources: Option<Vec<GlobEntry>>,
 
@@ -86,9 +83,8 @@ pub struct Scanner {
 }
 
 impl Scanner {
-    pub fn new(detect_sources: Option<DetectSources>, sources: Option<Vec<GlobEntry>>) -> Self {
+    pub fn new(sources: Option<Vec<GlobEntry>>) -> Self {
         Self {
-            detect_sources,
             sources,
             ..Default::default()
         }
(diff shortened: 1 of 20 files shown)


RobinMalfait:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14767?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#14767** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#14766** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14766?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14765** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14765?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `next`

<h2></h2>

crates/oxide/src/lib.rs
```diff
@@ -219,6 +219,36 @@ impl Scanner {
...
+
+        // Find all `@source` globs that point to a directory. If so, promote the source to auto
+        // source detection instead.
+        if let Some(sources) = &mut self.sources {
+            for source in sources {
+                // If a glob ends with `**/*`, then we just want to register the base path as a new
+                // base.
+                if source.pattern.ends_with("**/*") {
+                    source.pattern = source.pattern.trim_end_matches("**/*").to_owned();
+                }
```
thecrypticace, line 231: 1. `trim_end_matches` can be done unconditionally — no need to check `ends_with` first. I'd imagine the compiler is smart enough to omit the extra heap allocation in optimized builds.
2. `trim_end_matches` removes every occurrence which I think is fine. But the presence of the `ends_with` makes me think we wanted to do `strip_suffix` instead.

crates/oxide/src/lib.rs
```diff
@@ -219,6 +219,36 @@ impl Scanner {
...
+
+                let path = PathBuf::from(&source.base).join(&source.pattern);
+                let Some(folder_name) = path.file_name() else {
+                    continue;
+                };
+
+                // Contains a file extension, e.g.: `foo.html`, therefore we don't want to
+                // detect sources here.
+                if folder_name.to_str().unwrap().contains(".") {
```
thecrypticace, line 240: While this is likely to be correct _most_ of the time — folder names can definitely have dots in them. I think the correct check here requires using the filesystem.

crates/oxide/src/lib.rs
```diff
@@ -231,7 +218,58 @@ impl Scanner {
...
+                };
+
+                expression
+                    .into_iter()
+                    .filter_map(|expanded_pattern| expanded_pattern.map(|x| x.into()).ok())
+                    .map(move |pattern| GlobEntry {
+                        base: source.base.clone(),
+                        pattern,
+                    })
+                    .collect::<Vec<_>>()
```
thecrypticace, line 237:
```suggestion
                expression
                    .into_iter()
                    .filter_map(Result::ok)
                    .map(move |pattern| GlobEntry {
                        base: source.base.clone(),
                        pattern: pattern.into(),
                    })
                    .collect::<Vec<_>>()
```

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -379,9 +384,26 @@ class Root {
...
+        // Disable auto source detection
+        if (this.compiler.root === 'none') {
+          this.includeCandidatesFromModuleGraph = false
+          return []
+        }
+
+        // No root specified, use the module graph
+        if (this.compiler.root === null) {
+          return []
+        }
```
thecrypticace, line 397:
```suggestion
        if (this.compiler.root === null) {
          this.includeCandidatesFromModuleGraph = false

          return []
        }
```

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -379,9 +384,26 @@ class Root {
...
+
+        // TODO: In a follow up PR we want this filter this against the module graph.
+        this.includeCandidatesFromModuleGraph = false
```
thecrypticace, line 400:
```suggestion
        this.includeCandidatesFromModuleGraph = true
```

RobinMalfait:
closed in favor of https://github.com/tailwindlabs/tailwindcss/pull/14820

PR #14776, opened 2024-10-24
Don't escape underscores for the first parameter of `var()`

philipp-spiess:
This PR updates our arbitrary value decoder to:

- No longer require an escaping for underscores in the first parameter of `var()`. Example:

    ```
    ml-[var(--spacing-1_5,_1rem)]
    ```

- Ensures that properties before an eventual `url()` are properly unescaped. Example:
    ```
    bg-[no-repeat_url(./image.jpg)]
    ```

I will ensure that this properly works for the migrate use case in a follow-up PR in the stack.

## Test Plan

Added unit tests as well as tests for the variant decoder. Additionally this PR also adds a higher-level test using the public Tailwind APIs to ensure this is properly propagated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,20 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Changed
 
-- No longer require escaping underscores for CSS variables used in `var()` ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
+- Don't convert underscores in the first argument to `var()` to spaces ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
 
 ### Fixed
 
 - Ensure individual logical property utilities are sorted later than left/right pair utilities ([#14777](https://github.com/tailwindlabs/tailwindcss/pull/14777))
 - Don't migrate important modifiers inside conditional statements in Vue and Alpine (e.g. `<div v-if="!border" />`) ([#14774](https://github.com/tailwindlabs/tailwindcss/pull/14774))
-<<<<<<< HEAD
 - Ensure third-party plugins with `exports` in their `package.json` are resolved correctly ([#14775](https://github.com/tailwindlabs/tailwindcss/pull/14775))
-||||||| parent of 9bf8e915 (Don't escape underscores for the first parameter of `var()`)
-=======
 - Ensure underscores in the `url()` function are never unescaped ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
->>>>>>> 9bf8e915 (Don't escape underscores for the first parameter of `var()`)
 - _Upgrade (experimental)_: Ensure `@import` statements for relative CSS files are actually migrated to use relative path syntax ([#14769](https://github.com/tailwindlabs/tailwindcss/pull/14769))
 
 ## [4.0.0-alpha.29] - 2024-10-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- No longer require escaping underscores for CSS variables used in `var()` ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
+
 ### Fixed
 
 - Ensure individual logical property utilities are sorted later than left/right pair utilities ([#14777](https://github.com/tailwindlabs/tailwindcss/pull/14777))
 - Don't migrate important modifiers inside conditional statements in Vue and Alpine (e.g. `<div v-if="!border" />`) ([#14774](https://github.com/tailwindlabs/tailwindcss/pull/14774))
+<<<<<<< HEAD
 - Ensure third-party plugins with `exports` in their `package.json` are resolved correctly ([#14775](https://github.com/tailwindlabs/tailwindcss/pull/14775))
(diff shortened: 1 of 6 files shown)


philipp-spiess:
* **#14776** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/tailwindcss/src/utils/decode-arbitrary-value.test.ts
```diff
@@ -55,8 +66,8 @@ describe('adds spaces around math operators', () => {
     ['calc(24px+(-1rem))', 'calc(24px + (-1rem))'],
     ['calc(24px_+_(-1rem))', 'calc(24px + (-1rem))'],
     [
-      'calc(var(--10-10px,calc(-20px-(-30px--40px)-50px)',
-      'calc(var(--10-10px,calc(-20px - (-30px - -40px) - 50px)',
+      'calc(var(--10-10px,calc(-20px-(-30px--40px)-50px)))',
+      'calc(var(--10-10px,calc(-20px - (-30px - -40px) - 50px)))',
```
philipp-spiess, line 70: This test case didn't have the proper number of closing quotes. I remember that @RobinMalfait once mentioned that we do validate this already at the parser level.

packages/tailwindcss/src/utils/decode-arbitrary-value.ts
```diff
@@ -1,13 +1,16 @@
...
 
 export function decodeArbitraryValue(input: string): string {
-  // We do not want to normalize anything inside of a url() because if we
-  // replace `_` with ` `, then it will very likely break the url.
-  if (input.startsWith('url(')) {
-    return input
+  // There's definitely no functions in the input, so bail early
+  if (input.indexOf('(') === -1) {
+    return convertUnderscoresToWhitespace(input)
   }
```
philipp-spiess, line 8: This check was moved up from being inside the `addWhitespaceAroundMathOperators`, since neither the `url()` nor the `var()` handling need to run if the arbitrary value has no parenthesis.

CHANGELOG.md
```diff
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Added
+
+- No longer require escaping underscores for CSS variables used in `var()` ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
```
adamwathan, line 12: I think we might want to put this in CHANGED:

```suggestion
### Changed

- Don't convert underscores in the first argument to `var()` to spaces ([#14776](https://github.com/tailwindlabs/tailwindcss/pull/14776))
```

packages/tailwindcss/src/candidate.test.ts
```diff
@@ -1059,24 +1059,50 @@ it('should not replace `_` inside of `url()`', () => {
...
         },
-        "variants": [],
-      },
-    ]
-  `)
+      ]
+    `)
+})
+
+it('should not replace `_` for the first value of `var()`', () => {
```
adamwathan, line 1083:
```suggestion
it('should not replace `_` in the first argument to `var()`', () => {
```

packages/tailwindcss/src/index.test.ts
```diff
@@ -115,6 +115,34 @@ describe('compiling CSS', () => {
       ),
     ).toMatchSnapshot()
   })
+
+  test('unescapes underscores to spaces inside arbitrary values expect for url() and first parameter of var()', async () => {
```
adamwathan, line 119:
```suggestion
  test('unescapes underscores to spaces inside arbitrary values except for `url()` and first argument of `var()`', async () => {
```

packages/tailwindcss/src/utils/decode-arbitrary-value.test.ts
RobinMalfait: I think we only validate the parens in your CSS file (and throw), but we don't validate if it happens in a candidate.

### Input:

```html
<div class="[--foo:calc(var(--10-10px,calc(-20px-(-30px--40px)-50px)]"></div>
```

### Output:

```css
.\[--foo\:calc\(var\(--10-10px\,calc\(-20px-\(-30px--40px\)-50px\)\] {
  --foo: calc(var(--10-10px,calc(-20px - (-30px - -40px) - 50px);
}
```

packages/tailwindcss/src/utils/decode-arbitrary-value.ts
```diff
@@ -1,13 +1,16 @@
+import * as ValueParser from '../value-parser'
 import { addWhitespaceAroundMathOperators } from './math-operators'
 
 export function decodeArbitraryValue(input: string): string {
-  // We do not want to normalize anything inside of a url() because if we
-  // replace `_` with ` `, then it will very likely break the url.
-  if (input.startsWith('url(')) {
-    return input
+  // There's definitely no functions in the input, so bail early
```
adamwathan, line 5:
```suggestion
  // There are definitely no functions in the input, so bail early
```

packages/tailwindcss/src/utils/decode-arbitrary-value.test.ts
philipp-spiess: I see, lol!

With this change the closing parentheses would be automatically inserted actually (because we can't encode the absence of them in the `ValueParser`. But I think it would be better if we throw away candidates like this 🤔

adamwathan (approved):

PR #14788, opened 2024-10-24
Handle feedback from #14783

RobinMalfait:
This PR is a continuation of #14783 to handle the feedback on that PR.

1. Update the test to be more realistic
2. Updated the comment

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -1560,7 +1560,7 @@ test(
       `,
       'tailwind.config.js': js`module.exports = {}`,
       'src/index.css': css`
-        @import 'tailwindcss/utilities';
+        @import 'tailwindcss/components';
 
         /* No layer expected */
         @import './my-components.css';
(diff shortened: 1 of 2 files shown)


integrations/upgrade/index.test.ts
```diff
@@ -1600,16 +1600,16 @@ test(
...
       @import './my-components.css';
 
       /* No layer expected */
       @import './my-utilities.css';
 
       /* Expecting a layer */
-      @import './my-other.css' layer(utilities);
+      @import './my-other.css' layer(components);
+
+      @import 'tailwindcss/utilities' layer(utilities);
```
RobinMalfait, line 1612: It's moved to the bottom because we replace `@import 'tailwindcss/utilities';` with the new at-rule. Currently we just drop `@import 'tailwindcss/components';` but we could replace that rule as well instead such that this import lives at the top. 🤔

adamwathan (approved):

PR #14798, opened 2024-10-25
[Failing Test] - Fix multiple data attributes

luisrudge:

For example: v4.0.0-alpha.30

For example: "postcss": "^8.4.47",

**Describe your issue**

image

The upgrade was successful, however I can see in the diffs that only the first instance of the attribute was updated, not the other ones.


luisrudge:
Sorry, I wasn't able to run the tests locally, so I decided to open a PR with a failing test instead. I hope it's ok

thecrypticace:
Hey! So unfortunately these new tests are incorrect. The test you modified is intended to run on individual classes — not class lists. In my testing _known_ classes are indeed converted even when there's multiple classes that need to be changed in a class list.

What I suspect is happening is that we're not detecting `fade-in-0`, `fade-out-0`, `zoom-in-95`, and `zoom-out-95` as known utilities.

Would you mind opening an issue with a repo that reproduces the issue when upgraded? Or at the very least your config file?

Gonna close the PR but an issue that we can use to reproduce what you saw would be super helpful!

luisrudge:
thanks for the clarification! I tried to run the tests on my fork, but it didn't work as well 😓

PR #14808, opened 2024-10-28
Bump `prettier-plugin-tailwindcss` to latest version during upgrade

RobinMalfait:
This PR adds a migration to bump the `prettier-plugin-tailwindcss` version to the latest version when upgrading your project. This is to ensure that the plugin is compatible with the latest version of Tailwind CSS.

Note: we will only do this _if_ you already used the `prettier-plugin-tailwindcss` plugin in your project.

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -1629,3 +1629,33 @@ test(
     `)
   },
 )
+
+test(
+  'migrating the prettier-plugin-tailwindcss version',
+  {
+    fs: {
+      'package.json': json`
+        {
+          "dependencies": {
+            "tailwindcss": "workspace:^",
+            "@tailwindcss/upgrade": "workspace:^"
+          },
+          "devDependencies": {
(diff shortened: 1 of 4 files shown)


integrations/upgrade/index.test.ts
```diff
@@ -1629,3 +1629,33 @@ test(
...
+  },
+  async ({ fs, exec }) => {
+    await exec('npx @tailwindcss/upgrade --force')
+
+    let pkg = JSON.parse(await fs.read('package.json'))
+
+    expect(pkg.devDependencies).toMatchObject({
+      'prettier-plugin-tailwindcss': expect.any(String),
+    })
+    expect(pkg.devDependencies['prettier-plugin-tailwindcss']).not.toEqual('0.5.0')
```
RobinMalfait, line 1659: Was using a snapshot before, but that would break every time we bump the version. Alternatively we could parse the version and ensure it's _higher_ compared to what it is in this test.

thecrypticace (approved):

PR #14820, opened 2024-10-29
Auto source detection improvements

RobinMalfait:
This PR introduces a new `source(…)` argument and improves on the existing `@source`. The goal of this PR is to make the automatic source detection configurable, let's dig in.

By default, we will perform automatic source detection starting at the current working directory. Auto source detection will find plain text files (no binaries, images, ...) and will ignore git-ignored files.

If you want to start from a different directory, you can use the new `source(…)` next to the `@import "tailwindcss/utilities" layer(utilities) source(…)`.

E.g.:

```css
/* ./src/styles/index.css */
@import 'tailwindcss/utilities' layer(utilities) source('../../');
```

Most people won't split their source files, and will just use the simple `@import "tailwindcss";`, because of this reason, you can use `source(…)` on the import as well:

E.g.:

```css
/* ./src/styles/index.css */
@import 'tailwindcss' source('../../');
```

Sometimes, you want to rely on auto source detection, but also want to look in another directory for source files. In this case, yuo can use the `@source` directive:

```css
/* ./src/index.css */
@import 'tailwindcss';

/* Look for `blade.php` files in `../resources/views` */
@source '../resources/views/**/*.blade.php';
```

However, you don't need to specify the extension, instead you can just point the directory and all the same automatic source detection rules will apply.

```css
/* ./src/index.css */
@import 'tailwindcss';

@source '../resources/views';
```

If, for whatever reason, you want to disable the default source detection feature entirely, and only want to rely on very specific glob patterns you define, then you can disable it via `source(none)`.

```css
/* Completely disable the default auto source detection */
@import 'tailwindcss' source(none);

/* Only look at .blade.php files, nothing else  */
@source "../resources/views/**/*.blade.php";
```

Note: even with `source(none)`, if your `@source` points to a directory, then auto source detection will still be performed in that directory. If you don't want that, then you can simply add explicit files in the globs as seen in the previous example.

```css
/* Completely disable the default auto source detection */
@import 'tailwindcss' source(none);

/* Run auto source detection in `../resources/views` */
@source "../resources/views";
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support specifying the base path for automatic source detection using a `source(…)` function on `@tailwind utilities` or `@import "tailwindcss"` ([#14820](https://github.com/tailwindlabs/tailwindcss/pull/14820))
+- Support disabling automatic source detection with `source(none)` ([#14820](https://github.com/tailwindlabs/tailwindcss/pull/14820))
+- Support passing directories to `@source` without needing to pass a complete glob ([#14820](https://github.com/tailwindlabs/tailwindcss/pull/14820))
 - _Upgrade (experimental)_: Bump `prettier-plugin-tailwindcss` to latest version during upgrade ([#14808](https://github.com/tailwindlabs/tailwindcss/pull/14808))
 
 ### Fixed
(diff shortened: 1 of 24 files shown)


adamwathan (approved):

xcrap:
I have a Laravel project where I separated the frontend and backend.

I have the following paths for example.

Backend:
/resources/components/admin/*
/resources/views/admin/*
/resources/js/custom.js

Frontend:
/resources/components/website/*
/resources/view/website/*
/resources/js/admin.js

I was loading individual tailwind.config.js for each part, can we load multiple sources ? 
Does the @config "../../tailwind.admin.config.js"; still exists in v4 or it's gone.

Thanks.

philipp-spiess:
@xcrap `@config` still works in v4 but it's recommended you move the config to a CSS file instead. If you have any issues with it, feel free to create a new issue, it seems unrelated to this PR and since this PR is already closed, you might not get a response otherwise (I just happen to see it because I was going over closed PRs after my vacation)

xcrap:
> @xcrap `@config` still works in v4 but it's recommended you move the config to a CSS file instead. If you have any issues with it, feel free to create a new issue, it seems unrelated to this PR and since this PR is already closed, you might not get a response otherwise (I just happen to see it because I was going over closed PRs after my vacation)

Thank you, I'm using configs in the CSS, thanks for the reply :)

PR #14850, opened 2024-10-31
Generate CSS after Vite handles imports during prod builds

### LINKED ISSUE #14839: `@tailwindcss/vite@alpha` used in vite vue, `:deep()` selector can not  use in build mode, dev mode is ok

@tailwindcss/vite  `4.0.0-alpha.31`

"vite": "^5.4.10"    "vue": "^3.5.12"   "@vitejs/plugin-vue": "^5.1.4",

20.18.0

 Chrome

Windows

[**Reproduction URL**](https://github.com/huchaomin/tw4alpha-vite-deep-select-bug)
it's an empty Vite/vue project.

:::bash
pnpm i
pnpm build
:::

Snipaste_2024-10-31_10-44-16

when i dont use `@tailwindcss/vite`  in `vite.config.ts`,  build again

Snipaste_2024-10-31_10-46-07

Snipaste_2024-10-31_10-44-16

 `:deep()` selector works!!!!

need help, love Tailwind, thanks!

thecrypticace:
When we started handling `@import` ourselves in #14446 we inadvertently broke Vite's URL asset url rewriting inside of some imported CSS files. 

For example, given the following files:
```css
/* ./main.css */
@import "./dir/imported.css";

/* ./dir/imported.css */
.foo {
  background: url(../image.png);
}

/* ./image.png */
/* … */
```

We'd flatten the import CSS and you'd end up with the following:
```css
/* main.css */
.foo {
  background: url(../image.png);
}

/* ./image.png */
/* … */
```

And, as you can see, `../image.png` does not exist relative to the "final" CSS. It should've been rewritten to `./image.png` but was not. Letting Vite handle the imports will rewrite these urls properly — which is what this PR does. It achieves this by running our CSS generation when typtical Vite plugins do rather than before (via `enforce: 'pre'`).

Additionally, because we were running the generation step early, Vue had not handled `:deep(…)` selectors in `<style scoped>` blocks yet and we were caching the unprocessed CSS. This is also fixed by this PR.

Ideally both of these things would be fixable by changing how we handle imports but that will require additional APIs so this is a reasonable stopgap measure until we can do that.

Fixes #14839

### DIFF
--- a/integrations/utils.ts
+++ b/integrations/utils.ts
@@ -29,7 +29,7 @@ interface ExecOptions {
 
 interface TestConfig {
   fs: {
-    [filePath: string]: string
+    [filePath: string]: string | Uint8Array
   }
 }
 interface TestContext {
@@ -279,8 +279,14 @@ export function test(
           })
         },
         fs: {
-          async write(filename: string, content: string): Promise<void> {
+          async write(filename: string, content: string | Uint8Array): Promise<void> {
             let full = path.join(root, filename)
+            let dir = path.dirname(full)
+            await fs.mkdir(dir, { recursive: true })
+
+            if (typeof content !== 'string') {
+              return await fs.writeFile(full, content)
+            }
 
             if (filename.endsWith('package.json')) {
               content = await overwriteVersionsInPackageJson(content)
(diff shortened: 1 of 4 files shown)


philipp-spiess:
We decided to keep the pre processing phase for now and fixed the issues with the following PRs:

- #14877
- #14871

PR #14860, opened 2024-11-02
Refactor `Cursor::new` to initialize `curr` and `next` directly from `input`

BiswajitThakur:


philipp-spiess:
Hey @BiswajitThakur! Thanks for your contribution but we're not going to accept it at this time. The existing version is, IMO, a bit easier to understand and it's unclear if this change really makes something easier (e.g. what if we have to change the struct later?). An argument can be made that avoiding the function call is faster but without a benchmark this is also hard to know for sure.

PR #14866, opened 2024-11-03
Sort upgraded CSS

RobinMalfait:
During the migration process, a lot of changes to the CSS file happen. Some parts are converted, some parts are deleted and some new CSS is added.

To make sure we are generating a sensible and good looking CSS file, we will sort the final CSS and pretty print it.

The order we came up with looks like this:

```css
/* Imports */
@import "tailwindcss";
@import "../other.css";

/* Configuration */
@config "../path/to/tailwindcss.config.js";

@plugin "my-plugin-1";
@plugin "my-plugin-2";

@source "./foo/**/*.ts";
@source "./bar/**/*.ts";

@variant foo {}
@variant bar {}

@theme {}

/* Border compatibility CSS */
@layer base {}

/* Utilities */
@utility foo {}
@utility bar {}

/* Rest of your own CSS if any */
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure adjacent rules are merged together after handling nesting when generating optimized CSS ([#14873](https://github.com/tailwindlabs/tailwindcss/pull/14873))
 - _Upgrade (experimental)_: Install `@tailwindcss/postcss` next to `tailwindcss` ([#14830](https://github.com/tailwindlabs/tailwindcss/pull/14830))
 - _Upgrade (experimental)_: Remove whitespace around `,` separator when print arbitrary values ([#14838](https://github.com/tailwindlabs/tailwindcss/pull/14838))
-- _Upgrade (experimental)_: Sort upgraded CSS ([#14866](https://github.com/tailwindlabs/tailwindcss/pull/14866))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure adjacent rules are merged together after handling nesting when generating optimized CSS ([#14873](https://github.com/tailwindlabs/tailwindcss/pull/14873))
 - _Upgrade (experimental)_: Install `@tailwindcss/postcss` next to `tailwindcss` ([#14830](https://github.com/tailwindlabs/tailwindcss/pull/14830))
 - _Upgrade (experimental)_: Remove whitespace around `,` separator when print arbitrary values ([#14838](https://github.com/tailwindlabs/tailwindcss/pull/14838))
+- _Upgrade (experimental)_: Sort upgraded CSS ([#14866](https://github.com/tailwindlabs/tailwindcss/pull/14866))
 
 ### Changed
 
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -398,7 +398,6 @@ test(
         If we ever want to remove these styles, we need to add an explicit border
         color utility to any element that depends on these defaults.
       */
-
       @layer base {
         *,
         ::after,
@@ -1002,6 +1001,7 @@ test(
           border-color: var(--color-gray-200, currentColor);
         }
       }
+
       /*
         Form elements have a 1px border by default in Tailwind CSS v4, so we've
         added these compatibility styles to make sure everything still looks the
@@ -1193,6 +1193,7 @@ test(
 
       --- ./src/a.1.utilities.1.css ---
       @import './a.1.utilities.utilities.css';
+
       @utility foo-from-a {
         color: red;
       }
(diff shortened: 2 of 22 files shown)


packages/@tailwindcss-upgrade/src/codemods/sort-buckets.ts
```diff
@@ -0,0 +1,142 @@
...
+          injectInto('variant', node)
+        } else if (node.type === 'atrule' && node.name === 'source') {
+          injectInto('source', node)
+        } else if (node.type === 'atrule' && node.name === 'theme') {
+          injectInto('theme', node)
+        } else if (node.type === 'atrule' && node.name === 'config') {
+          injectInto('config', node)
+        } else if (node.type === 'atrule' && node.name === 'plugin') {
+          injectInto('plugin', node)
+        }
```
thecrypticace, line 88: Given that node.name always becomes the bucket name in these handful of cases I kinda feel like these could be simplified. I'm indifferent to actually doing this tho — just pointing it out.

packages/@tailwindcss-upgrade/src/codemods/migrate-tailwind-directives.test.ts
```diff
@@ -412,7 +414,7 @@ it('should replace `@responsive` with its children', async () => {
     `),
   ).toMatchInlineSnapshot(`
     ".foo {
-      color: red;
-    }"
+        color: red;
+      }"
```
thecrypticace, line 418: This is formatted kinda weird isn't it?

RobinMalfait: Ah yep, I'll see if I can clean this up. What's happening is that:

1. This is part of the "user" CSS bucket, which we don't format
2. The `@responsive` just replaces the outer layer with its children. This results in CSS like this:

```css
··.format·{
····color:·red;
··}
```
Added `·` to show the spaces, but that's why it's being formatted this way.

RobinMalfait: Solved it by marking each node inside `@responsive` with `tailwind_pretty`

thecrypticace (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure adjacent rules are merged together after handling nesting when generating optimized CSS ([#14873](https://github.com/tailwindlabs/tailwindcss/pull/14873))
 - _Upgrade (experimental)_: Install `@tailwindcss/postcss` next to `tailwindcss` ([#14830](https://github.com/tailwindlabs/tailwindcss/pull/14830))
 - _Upgrade (experimental)_: Remove whitespace around `,` separator when print arbitrary values ([#14838](https://github.com/tailwindlabs/tailwindcss/pull/14838))
+- _Upgrade (experimental)_: Sort upgraded CSS ([#14866](https://github.com/tailwindlabs/tailwindcss/pull/14866))
```
philipp-spiess, line 32: The current message makes it sound like the order was wrong or something but the change is only really for esthetic reasons, what do you think about this? I tried to not use the word order since it has an inherent meaning in CSS and am trying to lean on the "restructure" terminology

```suggestion
- _Upgrade (experimental)_: Restructure the generated CSS to make it easier to read ([#14866](https://github.com/tailwindlabs/tailwindcss/pull/14866))
```

Also this doesn't really fix anything so maybe it shouldn't be in the fixed category either I think 🤔 

I am still not convinced this is a very good message, maybe maybe this just doesn't need a changelog entry at all haha

integrations/upgrade/index.test.ts
philipp-spiess, line 1: Reading this diff is bliss 😄

packages/@tailwindcss-upgrade/src/codemods/format-nodes.ts
```diff
@@ -1,33 +1,73 @@
...
+      // Do not format the user bucket to ensure we keep the user's formatting
+      // intact.
+      if (node.type === 'atrule' && node.name === 'tw-bucket' && node.params === 'user') {
+        output.push(contents)
+        continue
+      }
+
+      // Format buckets
+      if (node.type === 'atrule' && node.name === 'tw-bucket') {
+        output.push(await format(contents, { parser: 'css', semi: true, singleQuote: true }))
```
philipp-spiess, line 51: Maybe we can move out this part into a const object so we ensure we don't diverge the prettier config between the two branches (and avoid creating the same object over and over):

```suggestion
        output.push(await format(contents, FORMATTING_OPTIONS))
```

packages/@tailwindcss-upgrade/src/codemods/migrate-at-layer-utilities.ts
```diff
@@ -192,7 +192,6 @@ export function migrateAtLayerUtilities(stylesheet: Stylesheet): Plugin {
       clone.params = cls
 
       // Mark the node as pretty so that it gets formatted by Prettier later.
```
philipp-spiess, line 194:
```suggestion
```

philipp-spiess (commented):
Let's gooo it's so good to read the diff of the integration test snapshots haha. Some comments inline, mostly not sure about the change log thing 🤔

CHANGELOG.md
RobinMalfait: Yeah hmm, maybe this is just an internal refactor. Especially because the codemods are not idempotent right now meaning you only will be able to run them once anyway. So it's still a single 'upgrade' and not a fix to previous upgrades from the user's perspective.

PR #14871, opened 2024-11-05
Apply non-Tailwind CSS transforms in Vite plugin

### LINKED ISSUE #14796: @tailwindcss/vite 4.0.0-alpha.25 and above breaks Scoped CSS styles in Vue production builds (whereas in dev-builds they work)
Currently using `4.0.0-alpha.30` but the issue occurs since `4.0.0-alpha.25`.
The issue is **not** present in `4.0.0-alpha.24`

`vite 5.4.10`

`v20.18.0`

Chrome

Windows

**Reproduction URL**

https://github.com/sceee/tailwind-vue-scoped-build-issue

Reproduction steps:
1. Clone repo
2. `npm i`
3. `npm run dev`
4. Check browser page, focus on the links beneath "You did it" - it displays the "Vite + Vue 3" links green and the "Some fancy link" red
image
This is expected as the "Some fancy link" receives a scoped style that can also be seen in the devtools
image
which is defined as scoped here:
https://github.com/sceee/tailwind-vue-scoped-build-issue/blob/31c440f57f16fef1671da722f2c6921f08f7ac2d/src/App.vue#L21-L24

5. Stop HMR webserver
6. Run `npm run build`
7. Run `npm run preview` to serve the production build
8. Check browser page of the preview - it now displays the "Vite + Vue3" links also (incorrectly) red
image
See that the (initially) style is now no longer scoped in the browser devtools:
image
Instead it is now a global style.

As written above, this was introduced with `4.0.0-alpha.25`.
If you downgrade to `4.0.0-alpha.24`, this issue no longer happens and the scoped style remains during production builds.

_I could imagine it has something to do with the dependency changes for @tailwindcss/vite from .24 to .25 (regarding the `postcss` removal) but this is just a random guess._

**Describe your issue**

The issue is that css defined in a scoped `<style scoped> ... </style>` tag results only in scoped css when being served as dev builds.
In production builds, these styles are no longer scoped, instead they are just "global" css and therefore unintentionally affect other elements.

See reproduction steps abov - both (dev and production builds) should emit the same scoped css like
image
instead of being different between dev and production builds.

### LINKED ISSUE #14839: `@tailwindcss/vite@alpha` used in vite vue, `:deep()` selector can not  use in build mode, dev mode is ok

philipp-spiess:
Fixes: #14839
Fixes: #14796

This PR fixes an issue in the Vite extension where we previously only ran a small list of allow-listed plugins for the second stage transform in the build step. This caused some CSS features to unexpectedly not work in production builds (one such example is Vue's `:deep(...)` selector).

To fix this, I changed the allow listed plugins that we do want to run to a block list to filter out some plugins we know we don't want to run (e.g. the Tailwind Vite plugin for example or some built-in Vite plugins that are not necessary).

## Test plan

This PR adds a new integration test suite to test interop with a custom Vite transformer that looks like this:

```js
{
  name: 'recolor',
  transform(code, id) {
    if (id.includes('.css')) {
      return code.replace(/red/g, 'blue')
    }
  },
}
```

I also validated that this does indeed fix the Vue `:deep(...)` selector related issue that we were seeing by copying the repro of #14839 into our playground:

You can see in the screenshot above that the `:deep()` selector overwrites the scoped styles as expected in both the dev mode and the prod build (screenshotted).

Furthermore I reproduced the issue reported in https://github.com/tailwindlabs/tailwindcss/issues/14796 and was able to confirm that in a production build, the styling works as expected:

Screenshot 2024-11-06 at 14 26 50

Lastly, I created a repository out of the biggest known-to-me Vite projects: [Astro, Nuxt, Remix, SolidStart, and SvelteKit](https://github.com/philipp-spiess/tailwind-playgrounds) and verified that both dev and prod builds show no issue and the candidate list is properly appended in each case.


philipp-spiess:
* **#14871** <a href="https://app.graphite.dev/github/pr/tailwindlabs/tailwindcss/14871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

philipp-spiess:
Need to test the following setups before we land this:

- [ ] Svelte
- [ ] Astro
- [ ] Remix
- [ ] Nuxt

philipp-spiess:
Can confirm that #14796 also works with this change:

Screenshot 2024-11-06 at 14 26 50

integrations/vite/other-transforms.test.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        expect(styles).toContain(css`
+          .foo {
+            background-color: blue;
+          }
+        `)
+      })
+    })
+  })
+}
```
thecrypticace, line 153: Have we decided on whether or not these transforms are expected to run on generated utilities? Can we add a test for that explicitly?

thecrypticace:
This is looking good. I really like that the change is fairly simple. I left some comments about stuff.

Aside: I **really** think we should investigate the two-pass approach we talked about in a future PR because I still feel real weird about running other plugins transforms _after_ the transform phase. However, in the short term, I think this is probably (hopefully) okay.

integrations/vite/vue.test.ts
```diff
@@ -65,5 +71,7 @@ test(
     expect(files).toHaveLength(1)
 
     await fs.expectFileToContain(files[0][0], [candidate`underline`, candidate`foo`])
+
+    await fs.expectFileNotToContain(files[0][0], [':deep(.bar)'])
```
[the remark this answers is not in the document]
philipp-spiess: @thecrypticace I cherry picked this commit from you haha but yeah agree, will look into it.

integrations/vite/other-transforms.test.ts
philipp-spiess: It currently will, I don't know if we want to make this explicit by adding a test because this feels like something we might want to change later (with the two-pass approach).

I can add a test to document it tho and add a comment.

thecrypticace: yeah I'd add a test and a comment noting that we're okay with changing the behavior.

adamwathan (approved):

integrations/vite/other-transforms.test.ts
```diff
@@ -0,0 +1,172 @@
...
+          color: red;
+        }
+      `,
+    },
+  }
+}
+
+for (let transformer of ['postcss', 'lightningcss'] as const) {
+  describe(transformer, () => {
+    test.only(`production build`, createSetup(transformer), async ({ fs, exec }) => {
```
adamwathan, line 63:
```suggestion
    test(`production build`, createSetup(transformer), async ({ fs, exec }) => {
```

PR #14877, opened 2024-11-05
Rewrite urls in CSS files when using Vite

### LINKED ISSUE #14784: [4 alpha] breaks the default Vite `url(...)` references in imported CSS files.

For example: alpha.29 (latest)

For example: Vite 5.4.10 (latest)

For example: v20.9.0

Mainly Safari, however it's not related to the issue 🙃

MacOS

**Reproduction URL**

https://github.com/slavarazum/tailwind4-alpha-vite/

**Describe your issue**

Vite provides [@import inline and rebasing](https://vite.dev/guide/features.html#import-inlining-and-rebasing) for CSS.
I'm trying to import font from the Fontsource library: https://fontsource.org/fonts/nunito which references font files using `url(...)`.

As a result we have such rebased urls:
CleanShot 2024-10-24 at 17 32 22@2x

After installing Tailwind, importing it, and enabling the Vite plugin, the URLs remain with the default related paths:
CleanShot 2024-10-24 at 17 35 46@2x

thecrypticace:
Fixes #14784

This is an alternative to #14850 in which we actually perform url rewriting / rebasing ourselves. We ported a large portion of the URL-rewriting code from Vite (with attribution) to use here with some minor modifications. We've added test cases for the url rewriting so verifying individual cases is easy. We also wrote integration tests for Vite that use PostCSS and Lightning CSS that verify that files are found and inlined or relocated/renamed as necessary.

We also did some manual testing in the Playground to verify that this works as expected across several CSS files and directories which you can see a screenshot from here:

Screenshot 2024-11-05 at 10 25 16


packages/tailwindcss/src/ast.ts
```diff
@@ -202,7 +202,7 @@ export function walkDepth(
   }
 }
 
-export function toCss(ast: AstNode[]) {
+export function toCss(ast: AstNode[], { printUtilitiesNode = false } = {}) {
```
philipp-spiess, line 205: Still feel conflicted about this. Like we talked, this should ultimately not be a concern of the printer (the printer should be agnostic to Tailwind, really). 

Maybe one other idea is that we allow the call sites where we want the `@tailwind utilities` directives to be removed to _mutate_ the `at-rule` with something like `params = ''` or so and then use that as hint to not print it? 

So basically right before [this line](https://github.com/tailwindlabs/tailwindcss/blob/next/packages/tailwindcss/src/index.ts#L509) we do something like this:

```ts
if (utilitiesNode) { 
  // Avoid printing the utility node
  utilitiesNode.name = ''
}
```

Then in the `toCSS` we skip at-rules with `name === ''`. Since the `at-rule` always contains the prefixed `@` this is really not valid otherwise anyways. WDYT?

philipp-spiess (commented):
Left some small notes but LGTM. I did co-author it with you tho so someone else should look over it 😄

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix crash when using `@source` containing `..` ([#14831](https://github.com/tailwindlabs/tailwindcss/pull/14831))
 - Ensure instances of the same variant with different values are always sorted deterministically (e.g. `data-focus:flex` and `data-active:flex`) ([#14835](https://github.com/tailwindlabs/tailwindcss/pull/14835))
 - Ensure `--inset-ring=*` and `--inset-shadow-*` variables are ignored by `inset-*` utilities ([#14855](https://github.com/tailwindlabs/tailwindcss/pull/14855))
+- Rewrite urls in CSS files when using Vite ([#14877](https://github.com/tailwindlabs/tailwindcss/pull/14877))
```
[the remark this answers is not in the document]
thecrypticace: It actually uses _both_ rebase and rewrite terminology in the same file but neither of these terms appear in the Vite documentation. Not sure if there's something even more clear tho.

packages/tailwindcss/src/ast.ts
thecrypticace: Yeah I still don't like this either. I kinda still want a "contents" node of sorts that only prints its children and we'd "switch" the utility node to it in `compile()`. I wonder if we could use the existing context node for that because it's basically that. Thoughts?

RobinMalfait: There are other nodes being printed here that are Tailwind CSS specific. E.g.: declarations like `--tw-sort`. We could add another pass that gets rid of them.

I think since we only care about the utilitiesNode reference that we fill up with the actual utilities, I think changing it to a context node instead might be a nicer way of handling as @thecrypticace mentioned.

I also don't like the mutation of `utilitiesNode.name` or using the `printUtilitiesNode` flag. Both feel like a bandaid solution.

CHANGELOG.md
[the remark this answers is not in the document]
philipp-spiess: https://v2.vitejs.dev/guide/features.html#css This calls it "URL rebasing". That's where I got it from.

packages/tailwindcss/src/index.ts
```diff
@@ -452,6 +453,23 @@ async function parseCss(
     firstThemeRule.nodes = nodes
   }
 
+  // Replace the `@tailwind utilities` node with a conext since it prints
+  // children directly.
```
philipp-spiess, line 457:
```suggestion
  // Replace the `@tailwind utilities` node with a context since it prints
  // children directly.
```

packages/tailwindcss/src/index.ts
```diff
@@ -452,6 +453,23 @@ async function parseCss(
...
+    node.context = {}
+
+    // Remove additional `@tailwind utilities` nodes
+    walk(ast, (node, { replaceWith }) => {
+      if (node.kind !== 'at-rule') return
+      if (node.name !== '@tailwind') return
+      if (!node.params.startsWith('utilities')) return
+
+      replaceWith([])
+    })
```
philipp-spiess, line 470: Can you think of any case where there are two `@tailwind utilities` nodes in the CSS file that is not "you screwed up your imports"? Wondering if we should just error on this case instead (since we already have a walk that will find `@tailwind utilities`.)

thecrypticace: We could. There's an explicit test for this though which is the only reason I even added this bit.

CHANGELOG.md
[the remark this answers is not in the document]
thecrypticace: Ah that's the old docs. In the code "rebasing" terminology appears to only apply to Sass/Less but it's whatever imo.

packages/tailwindcss/src/index.ts
philipp-spiess: Ah, makes sense. Yeah then maybe let's keep this behavior but I think we could still remove this walk and remove the node as part of the walk where we find the `utilitiesNode`:

```diff
     // Find `@tailwind utilities` so that we can later replace it with the
     // actual generated utility class CSS.
     if (
-      utilitiesNode === null &&
       node.name === '@tailwind' &&
       (node.params === 'utilities' || node.params.startsWith('utilities'))
     ) {
+      if (utilitiesNode !== null) {
+        replaceWith([])
+        return
       }
       let params = segment(node.params, ' ')
       for (let param of params) {
         if (param.startsWith('source(')) {
           let path = param.slice(7, -1)
           // Keyword: `source(none)`
           if (path === 'none') {
             root = path
             continue
           }
           // Explicit path: `source('…')`
           if (
             (path[0] === '"' && path[path.length - 1] !== '"') ||
             (path[0] === "'" && path[path.length - 1] !== "'") ||
             (path[0] !== "'" && path[0] !== '"')
           ) {
             throw new Error('`source(…)` paths must be quoted.')
           }
           root = {
             base: context.sourceBase ?? context.base,
             pattern: path.slice(1, -1),
           }
         }
       }
       utilitiesNode = node
     }
```

thecrypticace: ah ya good idea

thecrypticace: Updated

adamwathan (approved):

PR #14894, opened 2024-11-06
Remove `theme(inline)` option from CSS API

RobinMalfait:
This PR removes the `theme(inline)` option for now. We might re-introduce this in the future but for now we get rid of it to ensure there are fewer knobs to turn.

We still have the `inline` option internally because it is used in the v3 compat layer.


thecrypticace (approved):

RobinMalfait:
We were a bit too fast on this one. As the original PR points out: https://github.com/tailwindlabs/tailwindcss/pull/14095

This is a needed feature in some cases.

PR #14896, opened 2024-11-06
Fix crash during upgrade when content globs escape root of project

RobinMalfait:
This PR fixes an issue where globs in you `content` configuration escape the current "root" of the project.

This can happen if you have a folder, and you need to look up in the tree (e.g.: when looking at another package in a monorepo, or in case of a Laravel project where you want to look at mail templates).

This applies a similar strategy we already implement on the Rust side.

1. Expand braces in the globs
2. Move static parts of the `pattern` to the `base` of the glob entry object

---

Given a project setup like this:
```
.
├── admin
│   ├── my-tailwind.config.ts
│   └── src
│       ├── abc.jpg
│       ├── index.html
│       ├── index.js
│       └── styles
│           └── input.css
├── dashboard
│   ├── src
│   │   ├── index.html
│   │   ├── index.js
│   │   ├── input.css
│   │   └── pickaday.css
│   └── tailwind.config.ts
├── package-lock.json
├── package.json
├── postcss.config.js
└── unrelated
    └── index.html

7 directories, 14 files
```

If you then have this config:
```ts
// admin/my-tailwind.config.ts
export default {
  content: {
    relative: true,
    files: ['./src/**/*.html', '../dashboard/src/**/*.html'],
                            //  ^^  this is the important part, which escapes
                            //      the current root of the project.
  },
  theme: {
    extend: {
      colors: {
        primary: 'red',
      },
    },
  },
}
```

Then before this change, running the command looks like this:
image

After this change, running the command looks like this:
image

### DIFF
--- a/packages/@tailwindcss-upgrade/src/utils/hoist-static-glob-parts.ts
+++ b/packages/@tailwindcss-upgrade/src/utils/hoist-static-glob-parts.ts
@@ -55,7 +55,8 @@ function splitPattern(pattern: string): [staticPart: string | null, dynamicPart:
 
   let lastSlashPosition: number | null = null
 
-  for (let [i, c] of pattern.split('').entries()) {
+  for (let i = 0; i < pattern.length; i++) {
+    let c = pattern[i];
     if (c === '/') {
       lastSlashPosition = i
     }
--- a/packages/@tailwindcss-upgrade/src/utils/hoist-static-glob-parts.ts
+++ b/packages/@tailwindcss-upgrade/src/utils/hoist-static-glob-parts.ts
@@ -1,0 +1,78 @@
+import braces from 'braces'
+import path from 'node:path'
+
+interface GlobEntry {
+  base: string
+  pattern: string
+}
+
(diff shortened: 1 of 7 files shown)


packages/@tailwindcss-upgrade/src/utils/hoist-static-glob-parts.ts
```diff
@@ -0,0 +1,78 @@
...
+//  Split on second input: ("../project-b", "**/*.js")
+function splitPattern(pattern: string): [staticPart: string | null, dynamicPart: string | null] {
+  // No dynamic parts, so we can just return the input as-is.
+  if (!pattern.includes('*')) {
+    return [pattern, null]
+  }
+
+  let lastSlashPosition: number | null = null
+
+  for (let [i, c] of pattern.split('').entries()) {
```
philipp-spiess, line 58: Wonder if this is a bit easier to read if it's equivalent?

```suggestion
  for (let i = 0; i < pattern.length; i++) {
    let c = pattern[i];
```

philipp-spiess (approved):

PR #14903, opened 2024-11-07
Add new `**` variant

RobinMalfait:
This PR adds a new `**` variant to target any level of children.

This is very similar to the `*` variant, the big difference is that:

- `*` applies to direct children only
- `**` applies to any level of children

Thought of this because of all the recent work we did around globs. So a good analogy for this is glob syntax where you have the exact same difference. `*.html` vs `**/*.html`.


thecrypticace (commented):
I'm a fan. Can we verify that Oxide picks this up?

RobinMalfait:
> ensure `*:flex` and `**:flex` are picked up

@thecrypticace sure thing! Added an integration test: https://github.com/tailwindlabs/tailwindcss/pull/14903/commits/b1ab2b828b09829850709f4022ec12b35f181206

thecrypticace (approved):

PR #14915, opened 2024-11-08
Support marking utilities and theme values as deprecated

thecrypticace:
wip

IntelliSense PR: https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1084


thecrypticace:
Thinking out loud here. It's possible we'll want the deprecation information to be more than just a boolean. Maybe a list of suggested classes? I'm hesitant to put "message" strings in here though. I'd prefer those kinds of things were generated by the IntelliSense extension.

thecrypticace:
So right now this is implemented in a very simple manner — affecting the classList metadata that intellisense returns. However, given that we have bare value support there are technically an infinite number of potential deprecated classes now or in the future. I could also see a case in which a class becomes deprecated base of a specific modifier or something.

I think this means we may need something like a `designSystem.classMetadata(string[])` API (an array because it would be better for IntelliSense to submit "Work" in bulk).

philipp-spiess:
We'll need to rebase this once we decide to make progress, going to close this for now but we'll come back to that later

PR #14919, opened 2024-11-08
Revert lightningcss change

philipp-spiess:
In https://github.com/tailwindlabs/tailwindcss/commit/c5b6df2a273b7debaf28f4d56b23f6f2bdfe4b0d we upgraded `lightningcss` which caused `pnpm install` to fail for me because the patch no longer worked. I re-applied the patch, doesn't seem like anything has changed in the diff (just some formatting, probably because I opened it in a new VS Code window...)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   "pnpm": {
     "patchedDependencies": {
       "@parcel/watcher@2.4.1": "patches/@parcel__watcher@2.4.1.patch",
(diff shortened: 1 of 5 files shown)


philipp-spiess:
Looks like there are some actual issues when we upgrade `lightningcss` so for now going to revert it back to `1.26.0`.

RobinMalfait (approved):

PR #14935, opened 2024-11-09
Forward modifier from `has-*` and `not-*` variants to compounded variant

RobinMalfait:
This PR fixes an issue where using a modifier in combination with `has-*` or `not-*` doesn't work as expected.

If you have the following HTML:
```html
<div class="group/parent" data-visible="">
  <div class="group-data-visible/parent:text-red-500"></div>
</div>
```

Then the following CSS is generated:
```css
.group-data-visible\/parent\:text-red-500:is(:where(.group\/parent)[data-visible] *) {
  color: var(--color-red-500);
}
```

But if you want to use it with `not-*`, then nothing is being generated:
```html
<div class="group/parent" data-visible="">
  <div class="not-group-data-visible/parent:text-red-500"></div>
</div>
```

```css
/* <EMPTY> */
```

But with this PR, we will forward the modifier `parent` from the `not` variant, to the `group` variant, so the following CSS is generated:
```css
.not-group-data-visible\\/parent\\:text-red-500:not(:where(.group\\/parent)[data-visible] *) {
  color: var(--color-red-500);
}
```

This is also implemented for `has-*`


packages/tailwindcss/src/variants.ts
```diff
@@ -443,7 +443,14 @@ export function createVariants(theme: Theme): Variants {
...
 
-    if (variant.modifier) return null
+    if (variant.modifier) {
+      if (variant.variant.kind === 'compound' || variant.variant.kind === 'functional') {
+        variant.variant.modifier = variant.modifier
+        variant.modifier = null
+      } else {
+        return null
+      }
+    }
```
thecrypticace, line 453: It feels super weird to me to mutate the variant we're getting. Could this be handled in parseCandidate or something?

thecrypticace: Though I guess we'd need some additional metadata for that to work right. I guess as long as this isn't guaranteed to "break" non-compounded versions of the same variant (do we have a test for that?) then it's probably fine.

thecrypticace: Yeah the mutation breaks stuff:

Screenshot 2024-11-15 at 14 08 38

thecrypticace (changes requested):
hopefully can fix this w/o too much hassle

packages/tailwindcss/src/variants.ts
RobinMalfait: Handling it during parsing might break things as well, especially when we re-print a candidate.

packages/tailwindcss/src/candidate.ts
```diff
@@ -557,6 +557,13 @@ export function parseVariant(variant: string, designSystem: DesignSystem): Varia
         case 'compound': {
           if (value === null) return null
 
+          // Forward modifier from `not-*` and `has-*` variants to the
+          // compound variant.
+          if (modifier !== null && (root === 'not' || root === 'has')) {
+            value = `${value}/${modifier}`
+            modifier = null
+          }
+
```
RobinMalfait, line 566: This is more correct, but feels so wrong in this spot

thecrypticace: yeah agree it does feel wrong 🤔

Could in theory add a `forwardsModifier()` method to Variants and control it from there? Gonna think on this some

RobinMalfait:
Closing because this would only work 1 level deep. We can re-visit this (maybe with a different syntax) if we want to apply modifiers to compounds variants.

PR #14875, opened 2024-11-05
Migrate default utilities to have a value suffix

RobinMalfait:
This PR adds a migration for migrating the changes we implemented in https://github.com/tailwindlabs/tailwindcss/pull/14849

This is the migration we perform:

| Old               | New                |
| ----------------- | ------------------ |
| `shadow`          | `shadow-sm`        |
| `shadow-sm`       | `shadow-xs`        |
| `shadow-xs`       | `shadow-2xs`       |
| `inset-shadow`    | `inset-shadow-sm`  |
| `inset-shadow-sm` | `inset-shadow-xs`  |
| `inset-shadow-xs` | `inset-shadow-2xs` |
| `drop-shadow`     | `drop-shadow-sm`   |
| `drop-shadow-sm`  | `drop-shadow-xs`   |
| `rounded`         | `rounded-sm`       |
| `rounded-sm`      | `rounded-xs`       |
| `blur`            | `blur-sm`          |
| `blur-sm`         | `blur-xs`          |

Also added an integration test to ensure that `shadow` is properly migrated to `shadow-sm`, and doesn't get migrated to `shadow-xs` (because `shadow-sm` is migrated to `shadow-xs`).

PR #14976, opened 2024-11-12
Do not migrate legacy classes with custom values

RobinMalfait:
This PR fixes an issue where we migrated classes such as `rounded` to `rounded-sm` (see: https://github.com/tailwindlabs/tailwindcss/pull/14875)

However, if you override the values in your `tailwind.config.js` file, then the migration might not be correct.

This PR makes sure to only migrate the classes if you haven't overridden the values in your `tailwind.config.js` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure that CSS inside Svelte `<style>` blocks always run the expected Svelte processors when using the Vite extension ([#14981](https://github.com/tailwindlabs/tailwindcss/pull/14981))
-- _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
 - _Upgrade (experimental)_: Ensure it's safe to migrate `blur`, `rounded`, or `shadow` ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
+- _Upgrade (experimental)_: Do not rename classes using custom defined theme values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
 
 ## [4.0.0-alpha.33] - 2024-11-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure that CSS inside Svelte `<style>` blocks always run the expected Svelte processors when using the Vite extension ([#14981](https://github.com/tailwindlabs/tailwindcss/pull/14981))
+- _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
 
 ## [4.0.0-alpha.33] - 2024-11-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure that CSS inside Svelte `<style>` blocks always run the expected Svelte processors when using the Vite extension ([#14981](https://github.com/tailwindlabs/tailwindcss/pull/14981))
 - _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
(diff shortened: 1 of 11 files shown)


RobinMalfait:
Initial implementation: just check whether you have `theme[themekey]` or `theme.extend[themekey]` defined. An improvement to this is actually comparing the value from the "before" state to the value of the "after" state. if those are the same, then we can safely migrate the value.

integrations/utils.ts
```diff
@@ -421,6 +421,20 @@ export function test(
...
+          await context.exec('git add --all', { cwd: root })
+          await context.exec('git commit -m "before migration"', { cwd: root })
+        } catch (error: any) {
+          console.error(error)
+          console.error(error.stdout?.toString())
+          console.error(error.stderr?.toString())
+          throw error
+        }
+      }
+
```
RobinMalfait, line 437: Did this to improve the logs in `.only` mode, and an easier before/after state in `.debug` mode.

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
```
RobinMalfait, line 12: Maybe something like:
```suggestion
- _Upgrade (experimental)_: Do not migrate legacy classes with custom defined theme values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
```
🤔

packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!SEEDED.has(designSystem)) {
+    for (let old in LEGACY_CLASS_MAP) {
+      designSystem.utilities.static(old, () => [])
+    }
+    SEEDED.add(designSystem)
+  }
+
+  let defaultDesignSystem = await __unstable__loadDesignSystem('@import "tailwindcss";', {
+    base: __dirname,
+  })
```
philipp-spiess, line 66: Should move this outside of the function scope, it's doing this for every candidate otherwise

packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.ts
```diff
@@ -0,0 +1,110 @@
...
+  _userConfig: Config,
+  rawCandidate: string,
+): Promise<string> {
+  // Prepare design system with the unknown legacy classes
+  if (!SEEDED.has(designSystem)) {
+    for (let old in LEGACY_CLASS_MAP) {
+      designSystem.utilities.static(old, () => [])
+    }
+    SEEDED.add(designSystem)
+  }
```
philipp-spiess, line 62: Not needed anymore

RobinMalfait: Updated the comment, the reason we do need this is to make the migration easier by registering the `shadow-sm` (or all old classes) as a static utility so that the `candidate.root` is the actual class.

If we don't do this, then we have to deal with all kinds of candidate types (which might be fine). But that also means looking at the value and migrating those.

At the end of the day it's easier to migrate:
```json
[
  {
    "kind": "static",
    "root": "shadow-sm",
    "variants": [],
    "important": false,
    "raw": "shadow-sm"
  },
]
```

Compared to this:
```json
[
  {
    "kind": "functional",
    "root": "shadow",
    "modifier": null,
    "value": {
      "kind": "named",
      "value": "sm",
      "fraction": null
    },
    "variants": [],
    "important": false,
    "raw": "shadow-sm"
  }
]
```

CHANGELOG.md
philipp-spiess: What about
 
```suggestion
- _Upgrade (experimental)_: Do not rename classes using custom defined theme values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
```

packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.ts
philipp-spiess: I wonder if this might cause issues downstream though since the `designSystem` is passed through the whole migration pipeline and now it has `shadow-sm` added everywhere wrongly, just for this migration.

RobinMalfait: Had an initial fix for this: https://github.com/tailwindlabs/tailwindcss/pull/14976/commits/d644a3f06be300f014fca49cadafe033526a6f35

Then improved it further: https://github.com/tailwindlabs/tailwindcss/pull/14976/commits/4bf1b7a508921e937adb0c7f9de6a40cea672700

philipp-spiess (approved):

PR #14979, opened 2024-11-12
Ensure it's safe to perform suffix-less migrations

RobinMalfait:
This PR makes sure that migrations from suffix-less candidates (e.g.: `rounded`, `blur`, `shadow`) are safe to be migrated. 

In some code snippets that's not always the case.

Given the following code snippet:
```tsx
type Star = [
  x: number,
  y: number,
  dim?: boolean,
  blur?: boolean,
  rounded?: boolean,
  shadow?: boolean,
]

function Star({ point: [cx, cy, dim, blur, rounded, shadow] }: { point: Star }) {
  return <svg class="rounded shadow blur" filter={blur ? 'url(…)' : undefined} />
}
```

Without this change, it would result in:
```tsx
type Star = [
  x: number,
  y: number,
  dim?: boolean,
  blur-sm?: boolean,
  rounded-sm?: boolean,
  shadow-sm?: boolean,
]

function Star({ point: [cx, cy, dim, blur-sm, rounded-sm, shadow-sm] }: { point: Star }) {
  return <svg class="rounded-sm shadow-sm blur-sm" filter={blur-sm ? 'url(…)' : undefined} />
}
```

But with this change, it results in:
```tsx
type Star = [
  x: number,
  y: number,
  dim?: boolean,
  blur?: boolean,
  rounded?: boolean,
  shadow?: boolean,
]

function Star({ point: [cx, cy, dim, blur, rounded, shadow] }: { point: Star }) {
  return <svg class="rounded-sm shadow-sm blur-sm" filter={blur ? 'url(…)' : undefined} />
}
```

Notice how the classes inside the `class` attribute _are_ converted, but the ones in the types or as part of the JavaScript code (e.g.: `filter={blur ? 'url(…)' : undefined}`) are not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
+- _Upgrade (experimental)_: Ensure it's safe to perform suffix-less candidate migrations ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
 
 ## [4.0.0-alpha.33] - 2024-11-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
-- _Upgrade (experimental)_: Ensure it's safe to perform suffix-less candidate migrations ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
+- _Upgrade (experimental)_: Ensure it's safe to migrate `blur`, `rounded`, or `shadow` ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
 
 ## [4.0.0-alpha.33] - 2024-11-11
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - _Upgrade (experimental)_: Do not migrate legacy classes with custom values ([#14976](https://github.com/tailwindlabs/tailwindcss/pull/14976))
+- _Upgrade (experimental)_: Ensure it's safe to perform suffix-less candidate migrations ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
```
philipp-spiess, line 17:
```suggestion
- _Upgrade (experimental)_: Ensure it's safe to migrate `blur`, `rounded`, or `shadow` ([#14979](https://github.com/tailwindlabs/tailwindcss/pull/14979))
```

"suffix-less candidate" is maybe a bit confusing. It's only the three for now, right?

packages/@tailwindcss-upgrade/src/template/codemods/important.ts
```diff
@@ -46,63 +34,15 @@ export function important(
...
-      // The printCandidate function will already put the exclamation mark in
-      // the right place, so we just need to mark this candidate as requiring a
-      // migration.
-      return printCandidate(designSystem, candidate)
     }
+
+    // The printCandidate function will already put the exclamation mark in the
+    // right place, so we just need to mark this candidate as requiring a
+    // migration.
+    return printCandidate(designSystem, candidate)
```
philipp-spiess, line 45: Let's move this back inside the `if` condition, otherwise all candidates that are not ending with an exclamation mark are force-reprinted which does seem unexpected (e.g. if we have a serialize/deserialize issue it now affects all candidates)

philipp-spiess (changes requested):
LGTM! One comment about the moved `printCandidate` which I think should be moved back into the condition (otherwise we force every candidate through the candidate deseiralization/serialization logic.

packages/@tailwindcss-upgrade/src/template/codemods/important.ts
RobinMalfait: Aha, I messed it up due to the refactor and deleted the wrong `}`. Fixed now!

philipp-spiess (approved):

PR #14981, opened 2024-11-12
Vite: Don't track candidate changes for Svelte `<style>` tags

### LINKED ISSUE #14965: [v4] Vite plugin breaks svelte global animation keyframes handling

v4.0.0-alpha.32

Vite with SvelteKit 2.8.0 (latest)

For example: v23.1.0

Chrome

MacOS

**Reproduction URL**

https://github.com/emmbm/tailwind-svelte-keyframes

**Describe your issue**

Updating from `@tailwindcss/vite@4.0.0-alpha.31` to `@tailwindcss/vite@4.0.0-alpha.32` breaks svelte's expected handling (read here _renaming_) of global keyframes defined in style blocks.

In svelte components, styles are scoped by default. This scoping also applies to keyframes, [unless they are defined with a `-global-` prefix which svelte then removes during build](https://svelte.dev/docs/svelte/global-styles#:global()). This behavior is broken in the current `next` version of tailwind.

philipp-spiess:
Closes #14965

This PR changes the way we register Tailwind CSS as a Svelte preprocessor when using the Vite plugin. The idea is to reduce the bookkeeping for interacting with CSS inside `<style>` tags so that we have a more consistent behavior and make sure the Svelte-specific post-processing (e.g. local class mangling) works as expected.

Prior to this change, we were running Tailwind CSS as a Svelte preprocessor and then we would transform the file again when necessary inside the Vite `transform` hook. This is necessary to have the right list of candidates when we build the final CSS, but it did cause some situation to not apply the Svelte post-processors anymore. The repro for this seemed to indicate a timing specific issue and I did notice that specifically the code where we invalidate modules in Vite would cause unexpected processing orders.

We do, however, not officially support rendering utilities (`@tailwind utilities;`) inside `<style>` tag. This is because the `<style>` block is scoped by default and emitting utilities will always include utilities for all classes in your whole project. For this case, we highly recommend creating as separate `.css` file and importing it explicitly.

With this limitation in place, the additional bookkeeping where we need to invalidate modules because the candidate list has changed is no longer necessary and removing it allows us to reduce the complexity of the Svelte integration.

## Test Plan

https://github.com/user-attachments/assets/32c8e91f-ab21-48c6-aeaf-2582273b9bac

Not seen in the test plan above I also tested the `pnpm build --watch` step of the Vite project. This does require the `pnpm preview` server to restart but the build artifact are updated as expected.

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -308,6 +308,8 @@ function isPotentialCssRootFile(id: string) {
     (extension === 'css' ||
       (extension === 'vue' && id.includes('&lang.css')) ||
       (extension === 'astro' && id.includes('&lang.css')) ||
+      // We want to process Svelte `<style>` tags to properly add dependency
+      // tracking for imported files.
       isSvelteStyle(id)) &&
     // Don't intercept special static asset resources
     !SPECIAL_QUERY_RE.test(id)
@@ -412,6 +414,8 @@ class Root {
   // The resolved path given to `source(…)`. When not given this is `null`.
   private basePath: string | null = null
 
+  public overwriteCandidates: string[] | null = null
+
   constructor(
     private id: string,
     private getSharedCandidates: () => Map<string, Set<string>>,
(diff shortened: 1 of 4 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -304,13 +308,18 @@ function isPotentialCssRootFile(id: string) {
     (extension === 'css' ||
       (extension === 'vue' && id.includes('&lang.css')) ||
       (extension === 'astro' && id.includes('&lang.css')) ||
-      (extension === 'svelte' && id.includes('&lang.css'))) &&
+      isSvelteStyle(id)) &&
```
philipp-spiess, line 311: Add a comment why we don't want to skip it here

philipp-spiess:
Want to make it so that the root created for svelte files only scans the template file

thecrypticace (approved):

PR #14982, opened 2024-11-12
Add missing `layer(…)` to imports above Tailwind directives

RobinMalfait:
This PR fixes an issue where imports above Tailwind directives didn't get a `layer(…)` argument.

Given this CSS:
```css
@import "./typography.css";
@tailwind base;
@tailwind components;
@tailwind utilities;
```

It was migrated to:
```css
@import "./typography.css";
@import "tailwindcss";
```

But to ensure that the typography styles end up in the correct location, it requires the `layer(…)` argument.

This PR now migrates the input to:
```css
@import "./typography.css" layer(base);
@import "tailwindcss";
```

Test plan:
---

Added an integration test where an import receives the `layer(…)`, but an import that eventually contains `@utility` does not receive the `layer(…)` argument. This is necessary otherwise the `@utility` will be nested when we are processing the inlined CSS.

Running this on the Commit template, we do have a proper `layer(…)`
image

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -1667,7 +1667,6 @@ test(
 
       --- ./src/index.css ---
       @import './tailwind-setup.css';
-      @import './tailwind-setup.utilities.css';
 
       --- ./src/base.css ---
       @import 'tailwindcss/theme' layer(theme);
@@ -1708,13 +1707,11 @@ test(
       --- ./src/tailwind-setup.css ---
       @import './base.css';
       @import './components.css';
+      @import './components.utilities.css';
       @import './utilities.css';
+      @import './utilities.utilities.css';
 
       @config '../tailwind.config.ts';
-
-      --- ./src/tailwind-setup.utilities.css ---
-      @import './components.utilities.css';
-      @import './utilities.utilities.css'
 
       --- ./src/typography.css ---
       .typography {
@@ -1809,7 +1806,6 @@ test(
 
       --- ./src/index.css ---
       @import './tailwind-setup.css';
-      @import './tailwind-setup.utilities.css';
 
       --- ./src/base.css ---
       @import 'tailwindcss/theme' layer(theme);
@@ -1850,15 +1846,13 @@ test(
       --- ./src/tailwind-setup.css ---
       @import './base.css';
       @import './components.css';
+      @import './components.utilities.css';
       @import './utilities.css';
+      @import './utilities.utilities.css';
 
       @theme {
         --color-my-red: red;
       }
-
-      --- ./src/tailwind-setup.utilities.css ---
-      @import './components.utilities.css';
-      @import './utilities.utilities.css'
 
       --- ./src/typography.css ---
       .typography {
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -454,6 +454,132 @@ test(
 )
 
 test(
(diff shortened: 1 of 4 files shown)


integrations/upgrade/index.test.ts
```diff
@@ -453,6 +453,130 @@ test(
...
+        color: red;
+      }
+
+      @utility bar-mix {
+        color: red;
+      }
+
+      html {
+        color: blue;
+      }
```
RobinMalfait, line 569: Something to think about: should this file be split into an import _with_ the `layer(…)` and one with the utilities to ensure that this `html {}` base CSS is in the correct spot. 🤔

RobinMalfait: I think this is right, otherwise `html {}` would be in the correct spot. I implemented it in the followup commits.

integrations/upgrade/index.test.ts
```diff
@@ -1571,8 +1698,9 @@ test(
       }
 
       --- ./src/components.css ---
-      @import './typography.css';
+      @import './typography.css' layer(components);
```
RobinMalfait, line 1701: This has `layer(components)` because this is coming from the components file where nothing sits above it (in that file), but there is a `@layer components` and `@tailwind components` below it.

_if_ we want to look at as-if it was flattened, then it exists between `base` and `components`. This means that if this code existed in the same file we would've generated `layer(base)` instead. 🤔

RobinMalfait: This is also what we get on the Commit template with this PR:
image

Without knowing the rest of the files, it looks correct. But I wonder if `layer(base)` is more correct.

RobinMalfait: Talked with @philipp-spiess and @thecrypticace and we came to the conclusion that this is the least surprising outcome where you look at CSS files in isolation.

philipp-spiess (commented):
Maybe we can fix the case where the `tailwind-setup.css` file is split, otherwise this looks solid

integrations/upgrade/index.test.ts
```diff
@@ -1706,16 +1836,19 @@ test(
...
 
+      --- ./src/components.utilities.css ---
       @utility foo {
         color: red;
       }
```
RobinMalfait, line 1844: I think that technically since the only non `@utility` is an `@import` _with_ a `layer(…)` that this split shouldn't be required. But that feels like an optimization for a separate PR (if we even need that.).

integrations/upgrade/index.test.ts
```diff
@@ -1584,6 +1712,10 @@ test(
 
       @config '../tailwind.config.ts';
 
+      --- ./src/tailwind-setup.utilities.css ---
+      @import './components.utilities.css';
+      @import './utilities.utilities.css'
```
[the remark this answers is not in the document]
RobinMalfait: Good catch, _a_ split made sense, but the split should've stopped at the Tailwind root file indeed. Handled here: https://github.com/tailwindlabs/tailwindcss/pull/14982/commits/b62c49990dfdc3aa37fb0b6b2afaf8b3e6d6dc41

You can see the difference in the integration tests: https://github.com/tailwindlabs/tailwindcss/pull/14982/commits/87c0bb29e55a939cad795ad318c6300d65c4452e

Re-ran it on the Commit template and the output is the same (because no splits happened there, but it also didn't break anything)

philipp-spiess (approved):

PR #14989, opened 2024-11-13
Ensure `@config` is injected in common ancestor sheet

RobinMalfait:
This PR fixes an issue where an `@config` was injected in a strange location if you have multiple CSS files with Tailwind directives. 

Let's say you have this setup:
```css
/* ./src/index.css */
@import "./tailwind-setup.css";

/* ./src/tailwind-setup.css */
@import "./base.css";
@import "./components.css";
@import "./utilities.css";

/* ./src/base.css */
@tailwind base;

/* ./src/components.css */
@tailwind components;

/* ./src/utilities.css */
@tailwind utilities;
```

In this case, `base.css`, `components.css`, and `utilities.css` are all considered Tailwind roots because they contain Tailwind directives or imports.

Since there are multiple roots, the nearest common ancestor should become the tailwind root (where `@config` is injected). In this case, the nearest common ancestor is `tailwind-setup.css` (not `index.css` because that's further away).

Before this change, we find the common ancestor between `base.css` and `components.css` which would be `index.css` instead of `tailwind-setup.css`.

In a next iteration, we compare `index.css` with `utilities.css` and find that there is no common ancestor (because the `index.css` file has no parents). This resulted in the `@config` being injected in `index.css` and in `utilities.css`.

Continuing with the rest of the migrations, we migrate the `index.css`'s `@config` away, but we didn't migrate the `@config` from `utilities.css`.

With this PR, we don't even have the `@config` in the `utilities.css` file anymore.

Test plan
---

1. Added an integration test with a non-migrateable config file to ensure that the `@config` is injected in the correct file.
2. Added an integration test with a migrateable config file to ensure that the CSS config is injected in the correct file. h/t @philipp-spiess 
3. Ran the upgrade on the https://commit.tailwindui.com project and ensured that 
  1. The `@config` does not exist in the `utilities.css` file (this was the first bug we solved)
  2. The `@config` is replaced in the `tailwind.css` file correctly.

image

image

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -1468,6 +1468,139 @@ test(
 )
 
 test(
+  'injecting `@config` in the shared root, when a tailwind.config.{js,ts,…} is detected',
+  {
+    fs: {
+      'package.json': json`
+        {
+          "dependencies": {
+            "@tailwindcss/upgrade": "workspace:^"
+          }
+        }
+      `,
+      'tailwind.config.ts': js`
+        export default {
+          content: ['./src/**/*.{html,js}'],
+          plugins: [
+            () => {
+              // custom stuff which is too complicated to migrate to CSS
+            },
+          ],
+        }
+      `,
+      'src/index.html': html`
+        <div
+          class="!flex sm:!block bg-gradient-to-t bg-[--my-red]"
+        ></div>
+      `,
+      'src/index.css': css`@import './tailwind-setup.css';`,
+      'src/tailwind-setup.css': css`
+        @import './base.css';
+        @import './components.css';
+        @import './utilities.css';
+      `,
+      'src/base.css': css`
+        html {
+          color: red;
+        }
+        @tailwind base;
+      `,
+      'src/components.css': css`
+        @import './typography.css';
+        @layer components {
+          .foo {
+            color: red;
+          }
+        }
+        @tailwind components;
+      `,
+      'src/typography.css': css`
+        .typography {
+          color: red;
+        }
+      `,
+      'src/utilities.css': css`
+        @layer utilities {
+          .bar {
+            color: red;
+          }
+        }
+        @tailwind utilities;
+      `,
+    },
+  },
+  async ({ exec, fs }) => {
+    await exec('npx @tailwindcss/upgrade --force')
+
+    expect(await fs.dumpFiles('./src/**/*.{html,css}')).toMatchInlineSnapshot(`
+      "
+      --- ./src/index.html ---
+      <div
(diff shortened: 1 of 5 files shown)


integrations/upgrade/index.test.ts
```diff
@@ -1467,6 +1467,139 @@ test(
...
+        }
+      `,
+      'tailwind.config.ts': js`
+        export default {
+          content: ['./src/**/*.{html,js}'],
+          plugins: [
+            () => {
+              // custom stuff which is too complicated to migrate to CSS
+            },
+          ],
```
philipp-spiess, line 1488:
```suggestion
          theme: {
            extend: {
              colors: {
                'my-red': 'red',
              },
            },
          },
```

When we update the config to something that we _can_ convert to, we noticed that the `@theme` now ends up in `src/index.css` and that the `@config` is still inside `src/tailwind-setup.css`.

RobinMalfait: Added a failing test for this: https://github.com/tailwindlabs/tailwindcss/pull/14989/commits/a8a66a4aa62ad7d79bcc9a23ef9e104b47498f0e

Then fixed it: https://github.com/tailwindlabs/tailwindcss/pull/14989/commits/dc5cd8db12545c89852a2dae853d9193f537210f

Bonus points: basically deleted most of the code in the `migrate-config` file

thecrypticace (approved):

packages/tailwindcss/src/compat/plugin-api.test.ts
```diff
@@ -2973,8 +2973,7 @@ describe('matchUtilities()', () => {
       return compiled.build(candidates)
     }
 
-    expect(optimizeCss(await run(['@w-1','hover:@w-1'])).trim())
```
philipp-spiess, line 2976: Oh how did I smuggle that in huh

integrations/upgrade/index.test.ts
```diff
@@ -1467,6 +1467,276 @@ test(
...
+      }
+
+      @layer base {
+        html {
+          color: red;
+        }
+      }
+
+      --- ./src/components.css ---
+      @import './typography.css';
```
philipp-spiess, line 1709: Just for completeness: This is still wrong, we expect this to be moved into a layer. This is planned for a follow-up PR.

packages/@tailwindcss-upgrade/src/codemods/migrate-config.ts
```diff
@@ -82,41 +84,6 @@ export function migrateConfig(
     root.append(cssConfig.nodes)
   }
 
-  function migrate(root: Root) {
```
philipp-spiess, line 85: very nice cleanup

philipp-spiess (approved):

PR #14999, opened 2024-11-14
Add codemod and interop for legacy `container` component configu

philipp-spiess:
This PR adds support for handling v3 [`container` customizations ](https://tailwindcss.com/docs/container#customizing). This is done by adding a custom utility to extend the core `container` utility. A concrete example can be taken from the added integration test.

### Input

```ts
export default {
  content: ['./src/**/*.html'],
  theme: {
    container: {
      center: true,
      padding: {
        DEFAULT: '2rem',
        '2xl': '4rem',
      },
      screens: {
        md: '48rem', // Matches a default --breakpoint
        xl: '1280px',
        '2xl': '1536px',
      },
    },
  },
}
```

### Output

```css
@import "tailwindcss";

@utility container {
  margin-inline: auto;
  padding-inline: 2rem;

  @media (width >= theme(--breakpoint-sm)) {
    max-width: none;
  }

  @media (width >= 48rem) {
    max-width: 48rem;
  }

  @media (width >= 1280px) {
    max-width: 1280px;
  }

  @media (width >= 1536px) {
    max-width: 1536px;
    padding-inline: 4rem;
  }
}
````

## Test Plan

This PR adds extensive tests to the compat layer as part of unit tests. Additionally it does at a test to the codemod setup that shows that the right `@utility` code is generated. Furthermore I compared the implementation against v3 on both the compat layer and the custom `@utility`:

https://github.com/user-attachments/assets/44d6cbfb-4861-4225-9593-602b719f628f

### DIFF
--- a/packages/tailwindcss/src/compat/container.ts
+++ b/packages/tailwindcss/src/compat/container.ts
@@ -58,7 +58,11 @@ export function buildCustomContainerUtilityRules(
     }
 
     for (let [key, value] of Object.entries(screens)) {
-      breakpointOverwrites.set(key, [decl('max-width', value)])
+      let coreBreakpoint = breakpoints.find(([k]) => k === key)
+      if (coreBreakpoint && value === coreBreakpoint[1]) {
+        continue
+      }
+      breakpointOverwrites.set(key, [decl('max-width', `var(--breakpoint-${key})`)])
     }
   }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/tailwindcss/src/compat/container.ts [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    for (let [key, value] of Object.entries(screens)) {
+      // We're inlining the breakpoint values because the screens configured in
+      // the `container` option do not have to match the ones defined in the
+      // root `screen` setting.
+      breakpointOverwrites.set(key, {
+        rule: `(width >= ${value})`,
+        nodes: [decl('max-width', value)],
+      })
```
thecrypticace, line 74: Can we just use `atRule` here rather than an object that's turned into one further down?

adamwathan (approved):

PR #15001, opened 2024-11-14
Improve multi-root `@config` linking

RobinMalfait:
This PR improves the discoverability of Tailwind config files when we are trying to link them to your CSS files.

When you have multiple "root" CSS files in your project, and if they don't include an `@config` directive, then we tried to find the Tailwind config file in your current working directory.

This means that if you run the upgrade command from the root of your project, and you have a nested folder with a separate Tailwind setup, then the nested CSS file would link to the root Tailwind config file.

Visually, you can think of it like this:

```
.
├── admin
│   ├── src
│   │   └── styles
│   │       └── index.css       <-- This will be linked to (1)
│   └── tailwind.config.js      (2)
├── src
│   └── styles
│       └── index.css           <-- This will be linked to (1)
└── tailwind.config.js          (1)
```

If you run the upgrade command from the root of your project, then the `/src/styles/index.css` will be linked to `/tailwind.config.js` which is what we expect.

But `/admin/src/styles/index.css` will _also_ be linked to `/tailwind.config.js`

With this PR we improve this behavior by looking at the CSS file, and crawling up the parent tree. This mens that the new behavior looks like this:

```
.
├── admin
│   ├── src
│   │   └── styles
│   │       └── index.css       <-- This will be linked to (2)
│   └── tailwind.config.js      (2)
├── src
│   └── styles
│       └── index.css           <-- This will be linked to (1)
└── tailwind.config.js          (1)
```

Now `/src/styles/index.css` will be linked to `/tailwind.config.js`, and `/admin/src/styles/index.css` will be linked to `/admin/tailwind.config.js`.

When we discover the Tailwind config file, we will also print a message to the user to let them know which CSS file is linked to which Tailwind config file.

This should be a safe improvement because if your Tailwind config file had a different name, or if it lived in a sibling folder then Tailwind wouldn't find it either and you already required a `@config "…";` directive in your CSS file to point to the correct file.

In the unlikely event that it turns out that 2 (or more) CSS files resolve to the same to the same Tailwind config file, then an upgrade might not be safe and some manual intervention might be needed. In this case, we will show a warning about this.

image

Test plan:
---

- Added an integration test that properly links the nearest Tailwind config file by looking up the tree
- Added an integration test that resolves 2 or more CSS files to the same config file, resulting in an error where manual intervention is needed
- Ran it on the Tailwind UI codebase

Running this on Tailwind UI's codebase it looks like this:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure `flex` is suggested ([#15014](https://github.com/tailwindlabs/tailwindcss/pull/15014))
 - _Upgrade (experimental)_: Resolve imports when specifying a CSS entry point on the command-line ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
-- _Upgrade (experimental)_: Resolve nearest Tailwind config file, when CSS file does not contain `@config` ([#15001](https://github.com/tailwindlabs/tailwindcss/pull/15001))
+- _Upgrade (experimental)_: Resolve nearest Tailwind config file when CSS file does not contain `@config` ([#15001](https://github.com/tailwindlabs/tailwindcss/pull/15001))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure `flex` is suggested ([#15014](https://github.com/tailwindlabs/tailwindcss/pull/15014))
 - _Upgrade (experimental)_: Resolve imports when specifying a CSS entry point on the command-line ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
+- _Upgrade (experimental)_: Resolve nearest Tailwind config file, when CSS file does not contain `@config` ([#15001](https://github.com/tailwindlabs/tailwindcss/pull/15001))
 
 ### Changed
 
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -1,3 +1,4 @@
+import { stripVTControlCharacters } from 'node:util'
 import { expect } from 'vitest'
 import { candidate, css, html, js, json, test, ts } from '../utils'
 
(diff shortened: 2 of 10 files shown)


RobinMalfait:
This involves printing relative paths, so to make sure they work as expected I temporarily enabled the Windows tests to ensure everything works.

You can see it passing here: https://github.com/tailwindlabs/tailwindcss/actions/runs/11854912553
As seen in this PR: https://github.com/tailwindlabs/tailwindcss/pull/15001/commits/c005e243e113ac2fc2eae0c1e9aa45aa3b826d4d

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -124,6 +123,9 @@ async function run() {
       let config = await prepareConfig(sheet.linkedConfigPath, { base })
       configBySheet.set(sheet, config)
 
+      info(
+        `Migrating JavaScript configuration file: ${highlight(relative(config.configFilePath, base))}`,
+      )
```
RobinMalfait, line 128: We could be dealing with multiple config files, therefore this log has been moved inside of the loop.

Slightly updated the message and included the linked config file.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -140,9 +142,10 @@ async function run() {
     // Migrate source files, linked to config files
     {
       // Template migrations
-
-      info('Migrating templates using the provided configuration file.')
       for (let config of configBySheet.values()) {
+        info(
+          `Migrating templates using the provided configuration file: ${highlight(relative(config.configFilePath, base))}.`,
+        )
```
RobinMalfait, line 148: We could be dealing with template files originating from different Tailwind config files, therefore this log has been moved inside of the loop.

Slightly updated the message and included the linked config file.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -163,12 +166,13 @@ async function run() {
         await Promise.allSettled(
           files.map((file) => migrateTemplate(config.designSystem, config.userConfig, file)),
         )
-      }
 
-      success('Template migration complete.')
+        success('↳ Template migration complete.')
```
RobinMalfait, line 170: This is also moved into the loop so you can see each migration complete. I also included `↳` which looked a bit nicer, but this can be changed of course.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -163,12 +166,13 @@ async function run() {
...
+      }
     }
 
     // Migrate each CSS file
+    info('Migrating stylesheets')
```
RobinMalfait, line 175: We didn't have a start message for the stylesheet migrations, but we do for others, so I made it more complete by adding this.

This is just because I was looking at the output and noticed it.

packages/@tailwindcss-upgrade/src/migrate-postcss.ts
```diff
@@ -291,7 +284,7 @@ async function detectJSConfigPath(base: string): Promise<null | string> {
     let fullPath = path.resolve(base, file)
     try {
       await fs.access(fullPath)
-      return file
+      return fullPath
```
RobinMalfait, line 287: We already know the absolute path, so returning this means that we don't have to re-compute the absolute path in other places where it is more error prone because we could resolve it against the wrong `base`.

integrations/upgrade/index.test.ts
```diff
@@ -1530,19 +1555,15 @@ test(
...
-      @config '../tailwind.config.ts';
+      @import './base.css';
+      @import './utilities.css';
 
-      --- ./src/root.5.css ---
-      @import './root.5/tailwind.css';
+      @config './tailwind.config.ts';
 
       --- ./src/root.4/base.css ---
-      @import 'tailwindcss/theme' layer(theme);
```
philipp-spiess, line 1544: Where did the `tailwindcss/theme` import go? 🤔

packages/@tailwindcss-upgrade/src/index.ts
philipp-spiess: @RobinMalfait I like the `↳`

packages/@tailwindcss-upgrade/src/template/prepare-config.ts
```diff
@@ -94,15 +94,36 @@ const DEFAULT_CONFIG_FILES = [
...
+
   throw new Error(
-    'No configuration file found. Please provide a path to the Tailwind CSS v3 config file via the `--config` option.',
+    `No configuration file found for ${highlight(relative(start))}. Please provide a path to the Tailwind CSS v3 config file via the ${highlight('--config')} option.`,
   )
 }
+
+// Yields all parent paths from `from` to `to` (inclusive on both ends)
+function* parentPaths(from: string, to: string = from) {
+  let fromAbsolute = path.resolve(from)
```
philipp-spiess, line 115: Should this have a check to ensure that `from.startsWith(to)`? otherwise this might never terminate

packages/@tailwindcss-upgrade/src/migrate.ts
```diff
@@ -337,16 +337,54 @@ export async function linkConfigs(
...
+      for (let sheet of sheets) {
+        problematicStylesheets.add(sheet)
+      }
+    }
+  }
+
+  // There are multiple "root" files without `@config` directives. Manual
+  // intervention is needed to link to the correct Tailwind config files.
+  if (problematicStylesheets.size > 1) {
+    let msg = `You have multiple stylesheets that do not have an \`@config\`.\n`
```
philipp-spiess, line 372: Should this include something like "you have multiple stylesheets that we resolved to the same JavaScript config file"? multiple stylesheets alone is not an issue I think, especially when we can find multiple config files 🤔

integrations/upgrade/index.test.ts
```diff
@@ -1593,6 +1618,110 @@ test(
...
+  async ({ exec }) => {
+    let output = await exec('npx @tailwindcss/upgrade --force', {}, { ignoreStdErr: true }).catch(
+      (e) => e.toString(),
+    )
+
+    output = stripVTControlCharacters(output)
+      .replace(/tailwindcss v(.*)/g, 'tailwindcss') // Remove the version number from the error message
+      .replace(/\\/g, '/') // Make Windows paths look like Unix paths
+
+    expect(output).toMatchInlineSnapshot(`
```
philipp-spiess, line 1699: should this be a `toIncludes`?
Getting some node version specific warnings in the CLI when i run it locally lol, prob i need to downgrade my node version to fix this:


```
+ (node:10064) ExperimentalWarning: CommonJS module /Users/philipp/.local/share/mise/installs/node/23.1.0/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /Users/philipp/.local/share/mise/installs/node/23.1.0/lib/node_modules/npm/node_modules/supports-color/index.js using require().
+ Support for loading ES Module in require() is an experimental feature and might change at any time
+ (Use `node --trace-warnings ...` to show where the warning was created)
```

integrations/upgrade/index.test.ts
RobinMalfait: Oh good catch!

It's because I was running into issues with the imports because `tailwindcss/base` couldn't be found. Then I realised that it should be `tailwindcss/preflight` but that's only in v4, not in v3.

I will fix the import on line 1455. Because one thing we do is rely on `tailwindcss: workspace` but it should actually be tailwindcss v3. 

I'll fix this, thanks!

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure `flex` is suggested ([#15014](https://github.com/tailwindlabs/tailwindcss/pull/15014))
+- _Upgrade (experimental)_: Resolve imports from passed CSS file(s) ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
 - _Upgrade (experimental)_: Resolve imports when specifying a CSS entry point on the command-line ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
```
RobinMalfait, line 14:
```suggestion
- _Upgrade (experimental)_: Resolve imports when specifying a CSS entry point on the command-line ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure `flex` is suggested ([#15014](https://github.com/tailwindlabs/tailwindcss/pull/15014))
 - _Upgrade (experimental)_: Resolve imports when specifying a CSS entry point on the command-line ([#15010](https://github.com/tailwindlabs/tailwindcss/pull/15010))
+- _Upgrade (experimental)_: Resolve nearest Tailwind config file, when CSS file does not contain `@config` ([#15001](https://github.com/tailwindlabs/tailwindcss/pull/15001))
```
thecrypticace, line 18:
```suggestion
- _Upgrade (experimental)_: Resolve nearest Tailwind config file when CSS file does not contain `@config` ([#15001](https://github.com/tailwindlabs/tailwindcss/pull/15001))
```

adamwathan (approved):

PR #15010, opened 2024-11-15
Resolve imports from CSS file

RobinMalfait:
This PR adds an improvement to the upgrade tool to make sure that if you pass a single CSS file, that the upgrade tool resolves all the imports in that file and processes them as well.

Test plan:
---

Created a project where `index.css` imports `other.css`. Another `leave-me-alone.css` is created to proof that this file is _not_ changed. Running the upgrade guide using `index.css` also migrates `other.css` but not `leave-me-alone.css`.

Here is a video so you don't have to manually create it:

https://github.com/user-attachments/assets/20decf77-77d2-4a7c-8ff1-accb1c77f8c1

### DIFF
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -1,10 +1,9 @@
 #!/usr/bin/env node
 
-import { globby, isGitIgnored } from 'globby'
+import { globby } from 'globby'
 import fs from 'node:fs/promises'
 import path from 'node:path'
 import postcss from 'postcss'
-import atImport from 'postcss-import'
 import { formatNodes } from './codemods/format-nodes'
 import { sortBuckets } from './codemods/sort-buckets'
 import { help } from './commands/help'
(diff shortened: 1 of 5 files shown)


packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -80,19 +79,44 @@ async function run() {
...
 
-    // Load and parse all stylesheets
-    for (let result of loadResults) {
-      if (result.status === 'rejected') {
-        error(`${result.reason}`)
+      // Already handled
+      if (sheetsByFile.has(file)) continue
+
+      // We don't want to process ignored files (like node_modules)
+      if (isIgnored(file)) continue
```
RobinMalfait, line 93: When you don't pass any files, then we used globby with the `gitignore: true` option, so I wanted to add that here as well.

I added an explicitly ignored CSS file to the integration tests to make sure we don't touch it.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -80,19 +79,44 @@ async function run() {
...
+      // Track the sheet by its file
+      sheetsByFile.set(file, sheet)
+
+      // We process the stylesheet which will also process its imports and
+      // inline everything. We still want to handle the imports separately, so
+      // we just use the postcss-import messages to find the imported files.
+      //
+      // We can't use the `sheet.root` directly because this will mutate the
+      // `sheet.root`
+      let processed = await postcss().use(atImport()).process(sheet.root.toString(), { from: file })
```
thecrypticace, line 110: Is there a chance this would fail and we'd want to just skip it? Or should it hard crash or w/e?

thecrypticace: Also shouldn't we be able to do some of this stuff in `analyze()` instead of using postcss-import or nah?

RobinMalfait: Oh, hmm, let me try to move it. Might work indeed 🤔

thecrypticace (approved):

packages/@tailwindcss-upgrade/src/migrate.ts
```diff
@@ -65,13 +66,28 @@ export async function migrate(stylesheet: Stylesheet, options: MigrateOptions) {
...
+    // We don't want to process ignored files (like node_modules)
+    if (isIgnored(file)) {
+      return null
+    }
 
-  for (let sheet of stylesheets) {
-    if (sheet.file) {
-      stylesheetsByFile.set(sheet.file, sheet)
+    try {
+      let sheet = Stylesheet.loadSync(file)
```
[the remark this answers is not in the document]
RobinMalfait: Yep exactly.

PR #15025, opened 2024-11-17
Add new `in-*` variant

RobinMalfait:
This PR adds a new `in-*` variant that allows you to apply utilities when you are in a certain selector.

While doing research for codemods, we notice that some people use `group-[]:flex` (yep, the arbitrary value is empty…). The idea behind is that people want to know if you are in a `.group` or not.

Similarly, some people use `group-[]/name:flex` to know when you are in a `.group/name` class or not.

This new `in-*` variant allows you to do that without any hacks.

If you want to check whether you are inside of a `p` tag, then you can write `in-[p]:flex`. If you want to check that you are inside of a `.group`, you can write `in-[.group]`.

This variant is also a compound variant, which means that you can write `in-data-visible:flex` which generates the following CSS:
```css
:where([data-visible]) .in-data-visible\:flex {
  display: flex;
}
```

This variant also compounds with `not-*`, for example: `not-in-[.group]:flex`.

Additionally, this PR also includes a codemod to convert `group-[]:flex` to `in-[.group]:flex`.

---

This was proposed before for v3 in #13912


packages/tailwindcss/src/variants.ts
```diff
@@ -711,6 +711,18 @@ export function createVariants(theme: Theme): Variants {
 
   staticVariant('inert', ['&:is([inert], [inert] *)'])
 
+  variants.functional('in', (ruleNode, variant) => {
+    if (!variant.value || variant.modifier) return null
+
+    // Named values should be alphanumeric. This prevents `in-foo-bar` from
+    // being used as a variant.
+    if (variant.value.kind === 'named' && !/^[a-zA-Z]+$/i.test(variant.value.value)) {
```
thecrypticace, line 719: We should lift the regex out of the function to ensure its only compiled once

RobinMalfait: Did some benchmarking and a manual loop is faster. So used that instead (also pushed the benchmark file).

RobinMalfait: If we make it a compound variant and only allow `in-<other-variant>` like `in-data-visible` or `in-[p]` then we don't need the `isAlpha` check because `in-p` requires `p` to be a variant.

packages/tailwindcss/src/variants.ts
```diff
@@ -711,6 +711,35 @@ export function createVariants(theme: Theme): Variants {
...
+      for (let parent of path.slice(0, -1)) {
+        if (parent.kind !== 'rule') continue
+
+        didApply = false
+        return WalkAction.Stop
+      }
+
+      // Replace `&` in target variant with `*`, so variants like `&:hover`
+      // become `:where(*:hover) &`. The `*` will often be optimized away.
+      node.selector = `:where(${node.selector.replaceAll('&', '*')}) &`
```
RobinMalfait, line 732: This is basically the same implementation as the `has-*` variant, but the selector is a bit different.

thecrypticace (changes requested):
Need to make sure suggestions work for `in-*`

packages/tailwindcss/src/variants.ts
```diff
@@ -711,6 +711,35 @@ export function createVariants(theme: Theme): Variants {
...
+
+      // Track that the variant was actually applied
+      didApply = true
+    })
+
+    // If the node wasn't modified, this variant is not compatible with
+    // `in-*` so discard the candidate.
+    if (!didApply) return null
+  })
+
```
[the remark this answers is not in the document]
RobinMalfait: Smart!

thecrypticace (dismissed):

philipp-spiess (approved):

PR #15033, opened 2024-11-18
Implement compat for `<alpha-value>` from v3

thecrypticace:
This implements backwards compatibility for colors that use the old `<alpha-value>` feature from v3. We can do this by replacing `<alpha-value>` with `1` because we use `color-mix` to actually apply opacity modifiers in v4.


philipp-spiess (approved):
Great stuff

PR #15036, opened 2024-11-19
Add consistent base styles for buttons and form controls

philipp-spiess:
This PR introduces consistent base styles for buttons and form controls in Tailwind CSS v4.

## Motivation

In v3, form elements lack default styles, which can be confusing—especially when certain elements, like a text input without a placeholder or value, are rendered completely invisible on the page.

The goal of this change is to provide reasonable default styles for buttons, inputs, selects, and textareas that are (mostly) consistent across all browsers while remaining easy to customize with your own styles.

This improvement should make Tailwind more accessible for developers new to the framework and more convenient in scenarios where you need to quickly create demos (e.g., using Tailwind Play).

## Light and dark mode support

These styles support both light and dark mode, achieved using the `light-dark()` CSS function. While browser support for this function is still somewhat limited, Lightning CSS transpiles it to a CSS variable-based approach that works in older browsers.

For this approach to function correctly, a default `color-scheme` must be set in your CSS (as explained in [the Lightning CSS documentation](https://lightningcss.dev/transpilation.html#light-dark()-color-function)). This PR addresses this requirement by setting the `color-scheme` to `light` on the `html` element in Preflight.

image

image

## Breaking changes

While we don’t expect these changes to significantly impact v3 users upgrading to v4, there may be minor differences for those relying on the simpler v3 styles.

For example, Preflight now applies a `border-radius` to buttons and form controls. If you weren’t explicitly setting the border radius to `0` in your project, you’ll need to do so to restore the previous look.

Thankfully, reverting to the v3 styles is straightforward—just add the following reset to your CSS:

```css
@layer base {
  input,
  textarea,
  select,
  button {
    border: 0px solid;
    border-radius: 0;
    padding: 0;
    color: inherit;
  }
}
```

It’s worth noting that this reset doesn't touch the `::file-selector-button` styles that were added in this PR. This is because it's not possible to reliably "undo" these styles and restore the original user-agent styles (which is what was used in v3), as these are different in each browser. However, these new styles actually match the defaults in most browsers pretty closely, so hopefully this just won't be an issue.

## Codemod

This PR includes a codemod that automatically inserts the above mentioned v3 reset to help avoid breaking changes during the upgrade. The codemod will insert the following CSS:

```css
/*
  In Tailwind CSS v4, basic styles are applied to form elements by default. To
  maintain compatibility with v3, the following resets have been added:
*/
@layer base {
  input,
  textarea,
  select,
  button {
    border: 0px solid;
    border-radius: 0;
    padding: 0;
    color: inherit;
  }
}
```

## Testing

These changes have been tested across a wide range of browsers, including Chrome, Safari, Firefox, Edge, and Opera on macOS and Windows, as well as Safari, Chrome, Firefox, and several lesser-known browsers on iOS and Android.

However, some quirks still exist in certain mobile browsers, such as iOS Safari, which adds too much bottom padding below date and time inputs:

Screenshot 2024-11-20 at 3 57 20 PM

The only reliable way to address these issues is by applying `appearance: none` to these form controls. However, this felt too opinionated for Preflight, so we’ve opted to leave such adjustments to user-land implementations.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/migrate-forms-reset.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/migrate-forms-reset.ts
@@ -1,0 +1,49 @@
+import dedent from 'dedent'
+import postcss, { type Plugin, type Root } from 'postcss'
+
+const css = dedent
+
+const FORMS_RESET_CSS = css`
+  /*
+    In Tailwind CSS v4, form elements have basic styling applied to them. For
+    compatibility with v3, we've applied the following resets:
+  */
+  @layer base {
+    input,
+    textarea,
+    select,
+    button {
+      border: 0px solid;
+      border-radius: 0;
+      padding: 0;
+      background-color: transparent;
+    }
+  }
+`
+
+export function migrateFormsReset(): Plugin {
+  function migrate(root: Root) {
+    let isTailwindRoot = false
+    root.walkAtRules('import', (node) => {
+      if (
+        /['"]tailwindcss['"]/.test(node.params) ||
+        /['"]tailwindcss\/preflight['"]/.test(node.params)
+      ) {
+        isTailwindRoot = true
+        return false
+      }
+    })
+
+    if (!isTailwindRoot) return
+
+    let compatibilityCssString = `\n@tw-bucket compatibility {\n${FORMS_RESET_CSS}\n}\n`
+    let compatibilityCss = postcss.parse(compatibilityCssString)
+
+    root.append(compatibilityCss)
+  }
+
+  return {
+    postcssPlugin: '@tailwindcss/upgrade/migrate-forms-reset',
+    OnceExit: migrate,
+  }
+}
--- a/packages/@tailwindcss-upgrade/src/codemods/migrate-forms-reset.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/migrate-forms-reset.ts
@@ -5,8 +5,8 @@ const css = dedent
 
 const FORMS_RESET_CSS = css`
   /*
-    In Tailwind CSS v4, form elements have basic styling applied to them. For
-    compatibility with v3, we've applied the following resets:
+    In Tailwind CSS v4, basic styles are applied to form elements by default. To
+    maintain compatibility with v3, the following resets have been added:
   */
   @layer base {
     input,
--- a/packages/tailwindcss/preflight.css
+++ b/packages/tailwindcss/preflight.css
@@ -23,6 +23,7 @@
   5. Use the user's configured `sans` font-feature-settings by default.
   6. Use the user's configured `sans` font-variation-settings by default.
   7. Disable tap highlights on iOS.
+  8. Set a default `color-scheme`.
 */
 
 html,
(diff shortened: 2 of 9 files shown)


packages/@tailwindcss-upgrade/src/codemods/migrate-forms-reset.ts
```diff
@@ -0,0 +1,49 @@
+import dedent from 'dedent'
+import postcss, { type Plugin, type Root } from 'postcss'
+
+const css = dedent
+
+const FORMS_RESET_CSS = css`
+  /*
+    In Tailwind CSS v4, form elements have basic styling applied to them. For
+    compatibility with v3, we've applied the following resets:
```
philipp-spiess, line 9: This could use some tweaks 🙂

reinink, line 9: Not sure this is better but maybe this?

```suggestion
    In Tailwind CSS v4, basic styles are applied to form elements by default. To
    maintain compatibility with v3, the following resets have been added:
```

packages/tailwindcss/preflight.css
```diff
@@ -283,67 +320,137 @@ progress {
...
+
+:where(select:not([multiple], [size])) option,
+:where(select:not([multiple], [size])) optgroup {
+  color: FieldText;
+  background-color: Field;
+}
+
+:where(select[multiple], select[size]) optgroup {
+  font-weight: bold;
+}
```
thecrypticace, line 385:
```suggestion
:where(select:is([multiple], [size])) optgroup {
  font-weight: bold;
}
```

packages/tailwindcss/preflight.css
```diff
@@ -283,67 +320,137 @@ progress {
...
+
+:where(select[multiple], select[size]) optgroup option {
+  padding-inline-start: 20px;
+}
```
thecrypticace, line 389:
```suggestion
:where(select:is([multiple], [size])) optgroup option {
  padding-inline-start: 20px;
}
```

thecrypticace (approved):

RobinMalfait (commented):
Still testing, but let's add a changelog entry 👍

RobinMalfait (approved):

philipp-spiess:
@madmoizo Hey! Curious what the difference were for you/how did this affect your site? Do you use the `<meta>` tag to set the default `color-scheme`?

philipp-spiess:
Regarding the opt out: this is just a default now, to opt out you'd just overwrite this in your css:

```
html {
  color-scheme: light dark;
}
```

However, still curious how you noticed the issue.

philipp-spiess:
@madmoizo Completely understand your feedback here. The `color-scheme` change is definitely something that we're still talking about internally. The change here is also limited because of the limited browser support for the `light-dark()` function and available polyfills. I'll update the thread here when we know more!

> For example, in the reset code you suggested, you forgot <option>:

`option` wasn't something we reset in v3 either it seems like 😬 
 https://github.com/tailwindlabs/tailwindcss/blob/main/src/css/preflight.css#L167-L183

philipp-spiess:
@madmoizo Hey there! Just wanted to update our discussion that we've decided to revert this PR in #15100. Thanks a ton for your feedback on this.

PR #15069, opened 2024-11-21
Bring back some deprecated utilities

thecrypticace:
These utilities are deprecated (and were removed) but we're brining them back so they keep working:

- `flex-grow-*`
- `flex-shrink-*`
- `decoration-slice`
- `decoration-clone`
- `overflow-ellipsis`


packages/tailwindcss/src/utilities.ts
```diff
@@ -1929,6 +1965,10 @@ export function createUtilities(theme: Theme) {
   ])
 
   staticUtility('text-ellipsis', [['text-overflow', 'ellipsis']])
+
+  /** @deprecated */
+  staticUtility('overflow-ellipsis', [['text-overflow', 'ellipsis']])
```
philipp-spiess, line 1970: We should add this here: https://github.com/tailwindlabs/tailwindcss/blob/next/packages/tailwindcss/src/compat/legacy-utilities.ts

thecrypticace: ohhh I forgot we'd split these. Thanks for the reminder!

packages/tailwindcss/src/compat/legacy-utilities.test.ts
```diff
@@ -131,3 +131,83 @@ test('max-w-screen', async () => {
...
+      -webkit-box-decoration-break: slice;
+      box-decoration-break: slice;
+    }"
+  `)
+})
+
+test('overflow-ellipsis', async () => {
+  expect(await run(['overflow-ellipsis'])).toMatchInlineSnapshot(`
+    ".overflow-ellipsis {
+      text-overflow: linear-gradient(var(--tw-gradient-stops));
```
philipp-spiess, line 152: This looks wrong? 🤔

thecrypticace: lol um

thecrypticace: fixed

philipp-spiess (approved):

PR #15084, opened 2024-11-22
Ensure `@import layer(…) source(…)` work together

RobinMalfait:
This PR fixes an issue where the combining `source(…)` and `layer(…)` didn't work well together.

```css
@import "tailwindcss/utilities" layer(utilities);
```

and

```css
@import "tailwindcss/utilities" source(none);
```

Worked well individually, but together they did not:

```
@import "tailwindcss/utilities" layer(utilities) source(none);
```

This is because when we handled the `@import`, if you see a `layer(…)`, the import handling already converts it to:

```css
@layer utilities {
  /* Contents of the imported file */
} 
```

However, if you have multiple params, it would be converted to:

```css
@media layer(utilities) source(none) {
  /* Contents of the imported file */
} 
```

In this case, we never handled the `@media layer(…)` case correctly and we have to make sure to convert it to an `@layer utilities {…}` ourselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Ensure `@import layer(…) source(…)` work together ([#15084](https://github.com/tailwindlabs/tailwindcss/pull/15084))
 
 ## [4.0.0-beta.1] - 2024-11-21
 
--- a/integrations/cli/index.test.ts
+++ b/integrations/cli/index.test.ts
@@ -433,7 +433,7 @@ test(
         /* - Only './src' should be auto-scanned, not the current working directory */
         /* - .gitignore'd paths should be ignored (node_modules) */
         /* - Binary extensions should be ignored (jpg, zip) */
-        @import 'tailwindcss/utilities' source('./src');
+        @import 'tailwindcss/utilities' source('./src') layer(utilities);
 
         /* (2) */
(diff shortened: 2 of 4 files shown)


philipp-spiess:
> However, if you have multiple params, it would be converted to:
>
> @media layer(utilities) source(none) {
>   /* Contents of the imported file */
> } 

This seems like a bug in the at-import code then, no? 

Layer's are supposed to be extracted into this way and it should instead create:

```css
@layer (utilities) {
  @media source(none) {
    /* Contents of the imported file */
  }
}
```

https://developer.mozilla.org/en-US/docs/Web/CSS/@import#syntax

philipp-spiess (changes requested):
This PR only fixes

```css
@import 'tailwindcss/utilities' source('./paths') layer(utilities);
```

and not

```css
@import 'tailwindcss/utilities' layer(utilities) source('./paths');
```

I think 🙈  I think it's fine to require layer to be the first part as per the at-import spec though, but there is still a bug in the at-import resolver 😬

RobinMalfait:
Ah you are right, I was thinking about that people might do this themselves or if we got this if other tools handled imports. But you are right let me fix it there instead!

thecrypticace (changes requested):
This should be handled in the `substituteAtImports` step

philipp-spiess:
Oh ignore my comment `@import 'tailwindcss/utilities' layer(utilities) source('./paths');` already works, but putting source first doesn't since that's not valid css and we rely on media query propagation right now. I'll clean this up a bit so we can allow these CSS spec functions (`layer` and `supports`) at any place in the param argument list.

PR #15100, opened 2024-11-22
Revert new default form styles

reinink:
Closes #15071

This PR reverts the changes in #15036 which add consistent base styles for buttons and form controls to Preflight.

While this felt like a good idea (for the reasons explained in that PR), practically this is just too disruptive of a change for people upgrading from v3 to v4.

While updating some of our projects to v4 we found ourselves adding classes to undo styles more often than we expected, and it also felt inconsistent to have to use a different set of classes to style a link or a button when we wanted them to look the same.

We also decided it feels a little strange that you could change the border color of an element without ever specifying that it should have a border, for example this just feels a little wrong:

```html
<button class="border-blue-500">
```

We also needed to set a default `color-scheme` value for any of this stuff to work which breaks the ability to use the `color-scheme` meta tag.

Since this change was a fairly major breaking change and we aren't feeling much benefit from it, it doesn't feel worth making this change for v4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Changed
+
+- Revert the new base styles for buttons and form controls ([#15100](https://github.com/tailwindlabs/tailwindcss/pull/15100))
 
 ## [4.0.0-beta.1] - 2024-11-21
 
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -121,23 +121,6 @@ test(
       @import 'tailwindcss';
 
       /*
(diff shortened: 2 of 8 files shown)


philipp-spiess (commented):
LGTM, just missing a change log entry

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Changed
+
+- Revert the new base styles for buttons and form controls ([#15100](https://github.com/tailwindlabs/tailwindcss/pull/15100))
```
philipp-spiess, line 12: Here's my quick attempt. If you're happy, we can merge this.

adamwathan (approved):

CHANGELOG.md
reinink: Yup that works! Thanks dude.

PR #15125, opened 2024-11-22
Improve integration tests (stability + performance)

RobinMalfait:
This PR improves the integration tests in two ways:
1. Make the integration tests more reliable and thus less flakey
2. Make the integration tests faster (by introducing concurrency)

Tried a lot of different things to make sure that these tests are fast and stable.

---

The biggest issue we noticed is that some tests are flakey, these are tests with long running dev-mode processes where watchers are being used and/or dev servers are created.
To solve this, all the tests that spawn a process look at stdout/stderr and wait for a message from the process to know whether we can start making changes. 

For example, in case of an Astro project, you get a `watching for file changes` message. In case of Nuxt project you can wait for an `server warmed up in` and in case of Next.js there is a `Ready in` message.

These depend on the tools being used, so this is hardcoded per test instead of a magically automatic solution.

These messages allow us to wait until all the initial necessary work, internal watchers and/or dev servers are setup before we start making changes to the files and/or request CSS stylesheets before the server(s) are ready.

---

Another improvement is how we setup the dev servers. Before, we used to try and get a free port on the system and use a `--port` flag or a `PORT` environment variable. Instead of doing this (which is slow), we rely on the process itself to show a URL with a port. Basically all tools will try to find a free port if the default port is in use. We can then use the stdout/stderr messages to get the URL and the port to use.

To reduce the amount of potential conflicts in ports, we used to run every test and every file sequentially to basically guarantee that ports are free. With this new approach where we rely on the process, I noticed that we don't really run into this issue again (I reran the tests multiple times and they were always stable)

image
Note: these tests run Linux, Windows and macOS in this branch just for testing purposes. Once this is done, we will only run Linux tests on PRs and run all 3 of them on the `next` branch.

We do make the tests concurrent by default now, which in theory means that there could be conflicts (which in practice means that the process has to do a few more tries to find a free port). To reduce these conflicts, we split up the integration tests such that Vite, PostCSS, CLI, … tests all run in a separate job in the GitHub actions workflow. 

image

Comparing this branch against the `next` branch, this is what CI looks like right now:

| `next` | `feat/improve-integration-tests` |
| --- | --- |
| image | image |

There also was a point in time where I introduced sequential tests such that all spawned processes still run after each other, but so far I didn't run into issues if we keep them concurrent so I dropped that code.

Some small changes I made to make things more reliable:
1. When relying on stdout/stderr messages, we split lines on `\n` and we strip all the ANSI escapes which allows us to not worry about special ANSI characters when finding the URL or a specific message to wait for.
2. Once a test is done, we `child.kill()` the spawned process. If that doesn't work, for whatever reason, we run a `child.kill('SIGKILL')` to force kill the process. This could technically lead to some memory or files not being cleaned up properly, but once CI is done, everything is thrown away anyway.
3. As you can see in the screenshots, I used some nicer names for the workflows.

| `next` | `feat/improve-integration-tests` |
| --- | --- |
| image | image |

They also look a bit nicer in the PR overview as well:
image

The very last commit just filters out Windows and macOS tests again for PRs (but they are executed on the `next` branch.

---

### Nest steps

I think for now we are in a pretty good state, but there are some things we can do to further improve everything (mainly make things faster) but aren't necessary. I also ran into issue while trying it so there is more work to do.

1. More splits — instead of having a Vite folder and PostCSS folder, we can go a step further and have folders for Next.js, Astro, Nuxt, Remix, …
2. Caching — right now we have to run the build step for every OS on every "job". We can re-use the work here by introducing a setup job that the other jobs rely on. @thecrypticace and I tried it already, but were running into some Bun specific Standalone CLI issues when doing that.
3. Remote caching — we could re-enable remote caching such that the `build` step can be full turbo (e.g.: after a PR is merged in `next` and we run everything again)


integrations/vitest.global-setup.ts
```diff
@@ -0,0 +1,63 @@
+import http from 'http'
+import net from 'node:net'
+import type { GlobalSetupContext } from 'vitest/node'
+
+// Random port to start from
+let CURRENT_PORT = 3000
+
+export default async function setup({ provide }: GlobalSetupContext) {
+  // Start mini server to provide free-ports-as-a-service
```
thecrypticace, line 9: This is a WILD solution to this problem 😱

integrations/cli/config.test.ts
```diff
@@ -161,7 +161,10 @@ test(
     },
   },
   async ({ fs, spawn }) => {
-    await spawn('pnpm tailwindcss --input src/index.css --output dist/out.css --watch')
+    let process = await spawn(
+      'pnpm tailwindcss --input src/index.css --output dist/out.css --watch',
+    )
+    await process.onStderr((m) => m.includes('Done in'))
```
RobinMalfait, line 167: All `spawn(…)` tests will wait for a specific message to ensure we wait for everything to be ready before making changes. This improves the reliability of the tests.

integrations/cli/index.test.ts
```diff
@@ -491,7 +492,7 @@ test(
       'pages/nested/foo.jsx': 'content-["pages/nested/foo.jsx"] content-["BAD"]',
     },
   },
-  async ({ fs, exec }) => {
+  async ({ fs, exec, expect }) => {
```
RobinMalfait, line 495: Due to the concurrency, we have to use the `expect` provided by the test itself instead of the globally imported `expect`.

integrations/package.json
```diff
@@ -4,7 +4,6 @@
   "private": true,
   "devDependencies": {
     "dedent": "1.5.3",
-    "fast-glob": "^3.3.2",
-    "kill-port": "^2.0.1"
```
RobinMalfait, line 8: We removed the `getFreePort()` which relied on the `kill-port` package.

integrations/postcss/next.test.ts
```diff
@@ -126,26 +127,35 @@ test(
...
-      await spawn(`pnpm next dev ${bundler === 'turbo' ? '--turbo' : ''} --port ${port}`)
+    async ({ fs, spawn, expect }) => {
+      let process = await spawn(`pnpm next dev ${bundler === 'turbo' ? '--turbo' : ''}`)
+
+      let url = ''
+      await process.onStdout((m) => {
+        let match = /Local:\s*(http.*)/.exec(m)
+        if (match) url = match[1]
+        return Boolean(url)
+      })
```
RobinMalfait, line 138: We can potentially abstract this in a function but the regex is different based on the project. It also depends on the project if stdout or stderr is used. Because this is only a handful of lines, I just duplicated them.

integrations/utils.ts
```diff
@@ -448,6 +425,9 @@ export function test(
 test.only = (name: string, config: TestConfig, testCallback: TestCallback) => {
   return test(name, config, testCallback, { only: true })
 }
+test.skip = (name: string, config: TestConfig, testCallback: TestCallback) => {
+  return test(name, config, testCallback, { skip: true })
+}
```
RobinMalfait, line 430: Introduced this during debugging and while this is not used right now, I think this is still useful (similar to how `.only` and `.debug` is not used right now).

integrations/vite/index.test.ts
```diff
@@ -14,680 +14,686 @@ import {
...
-          let files = await fs.glob('project-a/dist/**/*.css')
-          expect(files).toHaveLength(1)
-          filename = files[0][0]
-        })
+    },
+    async ({ root, spawn, fs, expect }) => {
+      let process = await spawn('pnpm vite build --watch', {
+        cwd: path.join(root, 'project-a'),
+      })
+      await process.onStdout((m) => m.includes('built in'))
```
RobinMalfait, line 330: This is an example where we weren't properly waiting for the initial build to be ready.

integrations/vite/nuxt.test.ts
```diff
@@ -37,42 +36,62 @@ const SETUP = {
   },
 }
 
-test('dev mode', SETUP, async ({ fs, spawn, getFreePort }) => {
-  let port = await getFreePort()
-  await spawn(`pnpm nuxt dev --port ${port}`)
+test('dev mode', SETUP, async ({ fs, spawn, expect }) => {
+  let process = await spawn('pnpm nuxt dev', {
+    env: {
+      TEST: 'false', // VERY IMPORTANT OTHERWISE YOU WON'T GET OUTPUT
```
RobinMalfait, line 42: Vitest (I think) sets a `process.env.TEST` environment variable. If this is set to `true`, then `nuxt dev` doesn't show any output. Why? I'm not sure, but it was confusing because it worked while running manually. It also worked with copying the `spawn()` setup in a custom `whatisgoingon.js` file...

integrations/postcss/next.test.ts
thecrypticace: Seems fine to me.

integrations/vite/resolvers.test.ts
```diff
@@ -1,143 +1,148 @@
 import { describe, expect } from 'vitest'
```
thecrypticace, line 1: This test still uses a global expect — this is probably fine but maybe we should use the one from the test for consistency.

thecrypticace (approved):

PR #15212, opened 2024-11-27
Remove redundant `line-height` from `body`

MartijnCuppens:
It was added in #2729 to override line heights set on the body by modern-normalize. However,  it appears that modern-normalize never included any line-height definitions—only a font-family rule was present.

Ref: https://github.com/sindresorhus/modern-normalize/blob/v1.1.0/modern-normalize.css


philipp-spiess:
> However, it appears that modern-normalize never included any line-height definitions—only a font-family rule was present.

Maybe I misunderstand something but there's a line-height rule here: https://github.com/sindresorhus/modern-normalize/blob/v1.1.0/modern-normalize.css#L33

MartijnCuppens:
Yup, and the `<body>` inherits that line height by default, there's no need to explicitly set this.

philipp-spiess (approved):
Thanks! This looks good to me. Did some more testing across Safari/Chrome/Firefox just in case.

PR #15273, opened 2024-12-02
Postcss: Run plugin in `Once` hook

### LINKED ISSUE #15138: [v4] Plugins registered after `@tailwindcss/postcss` do not work

`@tailwindcss/postcss@4.0.0-beta.2`

`postcss@8.4.49`

Node 18, 20, 22

N/A

Windows 11 23H2 (Node 20) 

Ubuntu 22.04.5 LTS (Node 18, 20, 22)

**Reproduction URL**

https://github.com/maizzle/tw4-with-plugins

**Describe your issue**

Running into an issue where PostCSS plugins used after Tailwind CSS v4.0-beta.2 do not work.

For example, this should add `!important` to properties and resolve CSS variables:

```js
import postcss from "postcss";
import tailwindcss from "@tailwindcss/postcss";
import cssVariables from "postcss-css-variables";

postcss([
    tailwindcss, 
    cssVariables,
    addImportantPlugin, // see `index.js`
    ]).process(
        `
            @import "tailwindcss/theme";
        `, 
        {from: undefined}
    )
    .then(result => result.css);
```

However, the test fails:

```diff
.text-red-200 {
+   color: oklch(0.885 0.062 18.334) !important; // expected
-   color: var(--color-red-200);
}
```

Initially I thought this might be an async issue because of the `posthtml-postcss` plugin which uses Promises, but the issue persists even when using `postcss` directly.

philipp-spiess:
Closes #15138

This PR changes the postcss client to run in the `Once` hook instead of `OnceExit`. This makes sure the postcss order in v4 matches that of v3.

Conceptually this also makes more sense, since we expect tailwindcss to be run as one of the first plugins in the pipeline (where `OnceExit` would run it almost at the end).

To make sure it's still possible to use `postcss-import` before and have it resolve to the right paths, we also needed to change the `postcss-fix-relative-paths` plugin to run in the `Once` order (`postcss-import` also uses `Once` order so the order).

## Test Plan

This issue had many ways in which it can manifest. I added a unit test to ensure the plugin order works but here's a concrete example when using the postcss plugin in Vite.

### Before

Image `url()`s were not properly handled since the postcss plugin to transform these was run before Tailwind CSS could generate the class for it:

Screenshot 2024-12-02 at 14 55 42

### After

Screenshot 2024-12-02 at 14 53 52


CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Ensure other plugins can run after `@tailwindcss/postcss` ([#15273](https://github.com/tailwindlabs/tailwindcss/pull/15273))
+- Rebase `url()` inside imported CSS files when using Vite with the `@tailwindcss/postcss` extension ([#15273](https://github.com/tailwindlabs/tailwindcss/pull/15273))
```
thecrypticace, line 13: Do we have a test for this specifically?

thecrypticace (approved):

PR #15275, opened 2024-12-02
Vite: Don't rebase absolute `url()`s

philipp-spiess:
Closes #15269

This PR fixes an issue where our Vite extension was rebasing absolute urls inside `@import`-ed files. We forgot to cover this when we implemented the URL rebasing.

## Test Plan

We validated that this fixes the repro in #15269:

Screenshot 2024-12-02 at 18 07 35

Also added a unit test for this.


thecrypticace (approved):

hugentobler:
Thank you for fixing

PR #15279, opened 2024-12-02
Fix resolution of imported CSS files in Vite SSR builds

AlemTuzlak:
It seems this change makes the below mentioned issue go away and seems to generate the correct tailwindcss output file durring the ssr build. Big shoutout to @rossipedia for figuring this out.

I think you guys might want to take this over and own it as I'm not sure about other implications and if it breaks any other framework.

Fixes #15237

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,9 @@ jobs:
         on-next-branch:
           - ${{ github.ref == 'refs/heads/next' }}
         exclude:
-          - on-next-branch: false
+          - on-next-branch: true
             runner: windows-latest
-          - on-next-branch: false
+          - on-next-branch: true
             runner: macos-14
 
     runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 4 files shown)


thecrypticace:
Ah this is interesting. I was just heads down looking into this and found it was related to SSR when the user config explicitly overrides `ssr.resolve.conditions` but came up with a different fix:
```
ssrConfig: {
  resolve: {
    conditions: ['style', 'development|production'],
  },
},
```

This fixes it in Vite 5 but isn't needed at all Vite 6 because it just works as expected. Gonna look into the `aliasOnly` thing a bit because that would be a cleaner fix assuming it's the right one.

thecrypticace:
I think this fix is the correct one. Looking into it (and the original problem) the `aliasOnly` flag causes Vite to consider aliases but not SSR stuff, externals, fs paths, etc… which shouldn't be a problem here.

Just needs tests (don't worry I'm gonna add these)!

thecrypticace (approved):

PR #14029, opened 2024-07-20
Upgrade lilconfig version to enable ESM support

thibautsabot:
This PR https://github.com/tailwindlabs/tailwindcss/pull/12455 was merged last year by a [lilconfig](https://github.com/antonk52/lilconfig) maintainer.

However, looking at the merge commit [here](https://github.com/tailwindlabs/tailwindcss/commit/f75c3988b3ea862c2388e83f9372386e68913e06), only the `package-lock.json` file was changed, the `package.json` didn't make it through.

Upgrading to lilconfig v3 allows user to use ESM for importing postcss configs.
(For example, when using `"type": "module"` in `package.json`)

On this branch, I used the latest version (which includes security patches), alongside a small bump to `postcss-load-config`.

PR #15289, opened 2024-12-03
fix: update lilconfig for ESM and Windows support

shellscape:
This PR updates `lilconfig` from v2.1.0 to v3.1.3. Significant improvements to this package include Windows absolute path support as well as ESM config files support. This supersedes https://github.com/tailwindlabs/tailwindcss/pull/14029 which has fallen behind and has conflicts with the upstream branch. This is a critical update for Next.js apps running in development which have dependencies on packages that use an updated version of `lilconfig`. I understand that v4 will not be using `lilconfig` but it's an important update for users on v3.x in the meantime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure the TypeScript types for `PluginsConfig` allow `undefined` values ([#14668](https://github.com/tailwindlabs/tailwindcss/pull/14668))
 
+# Changed
+
+- Bumped lilconfig to v3.x ([#15289](https://github.com/tailwindlabs/tailwindcss/pull/15289))
+
 ## [3.4.15] - 2024-11-14
 
 - Bump versions for security vulnerabilities ([#14697](https://github.com/tailwindlabs/tailwindcss/pull/14697))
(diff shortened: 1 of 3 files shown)


thecrypticace:
We only use lilconfig in the CLI. How would this affect Next.js apps using PostCSS directly?

thecrypticace:
(gonna get this updated regardless — just wondering)

thecrypticace (approved):

shellscape:
Triage for this issue https://github.com/shellscape/jsx-email/issues/257 concluded that next.js/webpack was eagerly bundling 2.1.0 as the predominant lilconfig version and was not bundling both 2.1.0 and 3.1.2 (from jsx-email) during development. That resulted in errors when users attempted to use a jsx-email config file in the same app that included tailwind. I was able to reproduce that in fresh next.js apps consistently, and was able to workaround by forcing pnpm/npm to install 3.1.2 only via overrides. Super weird case where webpack is misbehaving in this context. I didn't get a chance to triage what in next.js is causing the bundling issue in dev, but that's next on my list.

Thank you very much for tackling this so quickly.

thecrypticace:
Got it. Seems like a Next.js bug there but yeah we should update. Sorry for letting this sit out so long.

thecrypticace:
Will get a new v3 release out today 👍

Issue #15045, opened 2024-11-19
[v4] `addComponents` is adding styles to `@layer utilities` instead of `@layer components`


saadeghi:
**What version of Tailwind CSS are you using?**

v4.0.0-alpha.34

**What build tool (or framework if it abstracts the build tool) are you using?**

`@tailwindcss/cli`

**Reproduction URL**

https://github.com/saadeghi/tw4-component-layer-issue

**Describe your issue**

These are the layers in output CSS file:
```
@layer theme, base, components, utilities;
```
**Expectation**  
It's expected for `addComponents` to add styles to `@layer components`  

**Current behavior**  
Currently `addComponents` adds styles to `@layer utilities`, similar to `addUtilities`

Plugin example:  
https://github.com/saadeghi/tw4-component-layer-issue/blob/master/myplugin.js

Generated style:  
https://github.com/saadeghi/tw4-component-layer-issue/blob/9b7a944690a35d55c7406756e30cc98c7a239623/output.css#L516

RobinMalfait:
Hey!

In v4 this is done on purpose. The biggest benefit of the split in v3 is that you can always override a component with more specific utilities. However, in v4 we improved sorting of all the utilities, which means that this rule should still apply but it reduces the amount of concepts in the system.

Can you share some of the issues you are seeing because of this?

saadeghi:
Thanks for the quick answer.

Sure.  
The issue is order of styles which is important because order is important in CSS.  
I updated the example: https://github.com/saadeghi/tw4-component-layer-issue

For example here I have 2 plugins:  
https://github.com/saadeghi/tw4-component-layer-issue/blob/master/input.css

`addComponents` adds both of them to `@layer utilities`:  
https://github.com/saadeghi/tw4-component-layer-issue/blob/064182b7d683bde9c9bfb9c618a334ed33136202/output.css#L512

But I have no control over the order:
```
@layer utilities {
  .bigcard {
    background: green;
    padding: 2rem;
  }
  .p-10 {
    padding: calc(var(--spacing) * 10);
  }
  .card {
    background: yellow;
  }
}
```
For example why one plugin came before `p-10` and the other came after `p-10`?  
How are they being sorted  
Changing the order in [`input.css`](https://github.com/saadeghi/tw4-component-layer-issue/blob/master/input.css) doesn't change anything.

saadeghi:
**Scenario**  
[...]

PR #15293, opened 2024-12-04
feat: [v3] add support for `oklch oklab lch lab` color spaces

RyanYANG52:
Hi, currently in v3 doesn't support `oklab(var(--foo) / var(--alpha))`, I create this simple PR to add support for it.

Because v4 is still in beta, it would be very nice to try out some of the new color space in our current projects, before migrating to v4

### DIFF
--- a/src/util/color.js
+++ b/src/util/color.js
@@ -13,6 +13,9 @@ let RGB = new RegExp(
 let HSL = new RegExp(
   `^(hsla?)\\(\\s*((?:${VALUE.source})(?:deg|rad|grad|turn)?|${CUSTOM_PROPERTY.source})(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
 )
+let OTHER_MODES = new RegExp(
+  `^(lch|lab|oklab|oklch)\\(\\s*(${VALUE.source}|${CUSTOM_PROPERTY.source})(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?(?:${ALPHA_SEP.source}(${VALUE.source}|${CUSTOM_PROPERTY.source}))?\\s*\\)$`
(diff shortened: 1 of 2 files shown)


philipp-spiess:
Hey! Thanks for the PR. I'll bring it up with the team but just to set expectations I don't think it's very likely we're going to add new features to v3 at this point and rather focus our effort on getting v4 ready, I hope you understand!

Since you mentioned the v4 beta: Did you have a chance yet to try it out in your project? As you said, v4 has much better support for these wide gamut color spaces out of the box including using the `color-mix()` function for opacity etc. 

If you're running into any issues that prevent you from upgrading, please let us know!

RyanYANG52:
I understand, thanks for reply.
I tried v4, it's very nice, great work.

our project is very large svelte monorepo, and we currently has lots of `@apply` in svelte `<style>`, and a custom plugin for some components and theme we use, I try to migrate once, too many file to change and test because of this change https://github.com/tailwindlabs/tailwindcss/discussions/15205  and I am not sure about the our plugin's custom theme and component will be able to import by https://github.com/tailwindlabs/tailwindcss/pull/14981
```css
@import "tailwindcss/theme" theme(reference); /* custom theme define by js plugin */
@tailwind utilities;  /* custom componets define by js plugin */
```


https://github.com/tailwindlabs/tailwindcss/issues/15045 this one may have some change behavior as well

so we decided to wait for v4 rc release, and then to find a way to upgrade

philipp-spiess:
@RyanYANG52 Gotcha! Yeah the best practice for Svelte etc. is to not have `@tailwind utilities` inside the `<style>` block ever since it would generate a list of _all utilities used by your project_ and that's not something you want to scope to an individual component. 

To load your existing theme and plugins etc. for use with `@apply` we're adding a new API in the next beta release though that you might find helpful: https://github.com/tailwindlabs/tailwindcss/pull/15228

RyanYANG52:
nice, I will try it out once it's released.

RobinMalfait:
Hey!

Now that Tailwind CSS v4 is out, it's unlikely we will be adding new features to v3. I would recommend to try and upgrade to v4 where this already works. We wrote an upgrade guide and included some tooling to upgrade your projects to minimize the amount of work you have to do yourself. More info: https://tailwindcss.com/docs/upgrade-guide

Since you are working with Svelte, this section of the upgrade guide will be relevant to you as well: https://tailwindcss.com/docs/upgrade-guide#using-apply-with-vue-svelte-or-css-modules

Going to close this for now, but I do appreciate the PR!

PR #15294, opened 2024-12-04
Add `getThemeEntries` API

thecrypticace:
This adds an API that lets us provide `@theme` variable completions from within core. The old list was hardcoded in IntelliSense but since the names for some of these changed recently we need a way to control this from within Tailwind CSS itself.

IntelliSense PR: https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1104

Do we think this is worth a changelog entry? Kinda thinking not but idk

### DIFF
--- a/packages/tailwindcss/src/design-system.ts
+++ b/packages/tailwindcss/src/design-system.ts
@@ -1,7 +1,14 @@
 import { toCss } from './ast'
 import { parseCandidate, parseVariant, type Candidate, type Variant } from './candidate'
 import { compileAstNodes, compileCandidates } from './compile'
-import { getClassList, getVariants, type ClassEntry, type VariantEntry } from './intellisense'
+import {
+  getClassList,
+  getThemeEntries,
+  getVariants,
+  type ClassEntry,
+  type ThemeEntry,
+  type VariantEntry,
+} from './intellisense'
 import { getClassOrder } from './sort'
 import type { Theme, ThemeKey } from './theme'
 import { Utilities, createUtilities, withAlpha } from './utilities'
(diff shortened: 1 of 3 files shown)


philipp-spiess:
@thecrypticace The way this is implement now, were just moving a hardcoded list from the intellisense repo to core but the list is still hardcoded and will likely go out of sync easily, or am I missing something here? 🤔 

> Do we think this is worth a changelog entry? Kinda thinking not but idk

I'd say if it fixes an issue that users are seeing then yeah but to my understanding this won't do anything unless you also update intellisense?

thecrypticace:
Yes we're moving the list into core even if it's hardcoded right now. I did update some of the entries (we should double check to see which ones we want to include / change / whatever).

> I'd say if it fixes an issue that users are seeing then yeah but to my understanding this won't do anything unless you also update intellisense?

That is correct.

philipp-spiess:
@thecrypticace Forgive me stupidity but it's unclear what problem this is really solving? I don't think we're currently planning on making any more changes to the variables and adding an allow list in core just because _it might_ seems a bit much especially since it's still fixable by updating the intellisense extension anyways? Since this is referenced in the `DesignSystem`, all those strings are now always inlined in every build making potential future CDN builds bigger as well, for example. Not to mention the higher API surface.

thecrypticace:
I can just update the hardcoded list in IntelliSense instead if we want to. But the goal was to allow control from core. Whether we're going to update the list or not (I could see us _adding_ entries to the list because what we have right now was cherry picked based on the default theme).

The API would in theory allow us to tailor the suggestions based on the current state of the theme. For example, if we wanted to suggest `--accent-color-` because the theme is already using that variable we could.

PR #15297, opened 2024-12-04
Use AST transformations in `@tailwindcss/postcss`

RobinMalfait:
This PR improves the `@tailwindcss/postcss` integration by using direct AST transformations between our own AST and PostCSS's AST. This allows us to skip a step where we convert our AST into a string, then parse it back into a PostCSS AST.

The only downside is that we still have to print the AST into a string if we want to optimize the CSS using Lightning CSS. Luckily this only happens in production (`NODE_ENV=production`).

This also introduces a new private `compileAst` API, that allows us to accept an AST as the input. This allows us to skip the PostCSS AST -> string -> parse into our own AST step.

To summarize:

Instead of:
- Input: `PostCSS AST` -> `.toString()` -> `CSS.parse(…)` -> `Tailwind CSS AST`
- Output: `Tailwind CSS AST` -> `toCSS(ast)` -> `postcss.parse(…)` -> `PostCSS AST`

We will now do this instead:
- Input: `PostCSS AST` -> `transform(…)` -> `Tailwind CSS AST`
- Output: `Tailwind CSS AST` -> `transform(…)` -> `PostCSS AST`

---

Running this on Catalyst, the time spent in the `@tailwindcss/postcss` looks like this:
- Before: median time per run: 19.407687 ms
- After: median time per run: 11.8796455 ms

This is tested on Catalyst which roughly generates ~208kb worth of CSS in dev mode.

While it's not a lot, skipping the stringification and parsing seems to improve this step by ~40%.

Note: these times exclude scanning the actual candidates and only time the work needed for parsing/stringifying the CSS from and into ASTs. The actual numbers are a bit higher because of the Oxide scanner reading files from disk. But since that part is going to be there no matter what, it's not fair to include it in this benchmark.

### DIFF
--- a/packages/@tailwindcss-node/src/compile.ts
+++ b/packages/@tailwindcss-node/src/compile.ts
@@ -7,14 +7,79 @@ import { pathToFileURL } from 'node:url'
 import {
   __unstable__loadDesignSystem as ___unstable__loadDesignSystem,
   compile as _compile,
+  compileAst as _compileAst,
   Features,
 } from 'tailwindcss'
+import type { AstNode } from '../../tailwindcss/src/ast'
 import { getModuleDependencies } from './get-module-dependencies'
 import { rewriteUrls } from './urls'
 
 export { Features }
 
 export type Resolver = (id: string, base: string) => Promise<string | false | undefined>
+
+export async function compileAst(
+  ast: AstNode[],
+  {
+    base,
+    onDependency,
+    shouldRewriteUrls,
+
+    customCssResolver,
+    customJsResolver,
+  }: {
+    base: string
+    onDependency: (path: string) => void
+    shouldRewriteUrls?: boolean
+
+    customCssResolver?: Resolver
+    customJsResolver?: Resolver
+  },
+) {
+  let compiler = await _compileAst(ast, {
+    base,
+    async loadModule(id, base) {
+      return loadModule(id, base, onDependency, customJsResolver)
+    },
+    async loadStylesheet(id, base) {
+      let sheet = await loadStylesheet(id, base, onDependency, customCssResolver)
+
+      if (shouldRewriteUrls) {
+        sheet.content = await rewriteUrls({
+          css: sheet.content,
+          root: base,
+          base: sheet.base,
+        })
+      }
+
+      return sheet
+    },
+  })
+
+  // Verify if the `source(…)` path exists (until the glob pattern starts)
+  if (compiler.root && compiler.root !== 'none') {
+    let globSymbols = /[*{]/
+    let basePath = []
+    for (let segment of compiler.root.pattern.split('/')) {
+      if (globSymbols.test(segment)) {
+        break
+      }
+
+      basePath.push(segment)
+    }
+
+    let exists = await fsPromises
+      .stat(path.resolve(base, basePath.join('/')))
+      .then((stat) => stat.isDirectory())
+      .catch(() => false)
+
+    if (!exists) {
+      throw new Error(`The \`source(${compiler.root.pattern})\` does not exist`)
+    }
+  }
+
+  return compiler
+}
 
 export async function compile(
   css: string,
--- a/packages/@tailwindcss-node/src/compile.ts
+++ b/packages/@tailwindcss-node/src/compile.ts
@@ -18,30 +18,29 @@ export { Features }
 
 export type Resolver = (id: string, base: string) => Promise<string | false | undefined>
 
-export async function compileAst(
-  ast: AstNode[],
-  {
-    base,
-    onDependency,
-    shouldRewriteUrls,
+export interface CompileOptions {
+  base: string
+  onDependency: (path: string) => void
+  shouldRewriteUrls?: boolean
 
-    customCssResolver,
-    customJsResolver,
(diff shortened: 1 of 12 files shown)


packages/@tailwindcss-node/src/compile.ts
```diff
@@ -7,15 +7,80 @@ import { pathToFileURL } from 'node:url'
...
 } from 'tailwindcss'
+import type { AstNode } from '../../tailwindcss/src/ast'
 import { getModuleDependencies } from './get-module-dependencies'
 import { rewriteUrls } from './urls'
 
 export { Features }
 
 export type Resolver = (id: string, base: string) => Promise<string | false | undefined>
 
+export async function compileAst(
```
philipp-spiess, line 21: is it possible we share some code here between the two implementations?

packages/tailwindcss/src/css-parser.ts
```diff
@@ -434,15 +435,7 @@ export function parse(input: string) {
...
-            let importantIdx = buffer.indexOf('!important', colonIdx + 1)
-            parent.nodes.push({
-              kind: 'declaration',
-              property: buffer.slice(0, colonIdx).trim(),
-              value: buffer
-                .slice(colonIdx + 1, importantIdx === -1 ? buffer.length : importantIdx)
-                .trim(),
-              important: importantIdx !== -1,
-            } satisfies Declaration)
+            parent.nodes.push(parseDeclaration(buffer, colonIdx))
```
philipp-spiess, line 438: nice cleanup

packages/tailwindcss/src/index.ts
```diff
@@ -580,7 +581,7 @@ export async function compile(
   // resulted in a generated AST Node. All the other `rawCandidates` are invalid
   // and should be ignored.
   let allValidCandidates = new Set<string>()
-  let compiledCss = features !== Features.None ? toCss(ast) : css
+  let compiled = features !== Features.None ? optimizeAst(ast) : input
```
philipp-spiess, line 584: I think we can actually get rid of this `optimizeAst()` call here. There are two scenarios to consider:

1. A subsequent `build()` is called with no changes: This is the case where we currently safe one computation. In that case we can also move the `optimizeAst()` inot the `build()` function though and add something like (`if (!compiled) features !== Features.None ? optimizeAst(ast) : input`) there inline
2. A subsequent `build()` is called **with changes**: I'd say this happens very often since the first build probably more often than not does require candidates which will require a rebuild. In this case we're currently computing the initial `optimizeAst` pass for no reason since it's going to be replaced soon anyway.

This is possible since the `compileAst()` function does never return the compiled CSS, you can only do so by calling `build()`.

packages/tailwindcss/src/ast.ts
```diff
@@ -208,18 +208,151 @@ export function walkDepth(
   }
 }
 
-export function toCss(ast: AstNode[]) {
-  let atRoots: string = ''
+// Optimize the AST for printing where all the special nodes that require custom
+// handling are handled such that the printing is a 1-to-1 transformation.
+export function optimizeAst(ast: AstNode[]) {
```
philipp-spiess, line 213: Hm now that I see this in the whole PR I'm not so super convinced about this name anymore. The reason being that we refer to `optimize` as the step in post-compilation where we run `lightningcss`. Now we're overloading the term with something used during the compilation. Maybe `applyFallbacksToAst` or something like this? 🤔 I also dig your initial name of `prepareAstForPrinting` honestly lol. Find it less confusing then overloading the term.

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -205,23 +211,29 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
...
+              context.optimizedAst = postcss.parse(
+                optimizeCss(toCss(ast), {
+                  minify: typeof optimize === 'object' ? optimize.minify : true,
+                }),
+                result.opts,
+              )
+              env.DEBUG && console.timeEnd('[@tailwindcss/postcss] Optimize CSS')
+            } else {
+              // Convert our AST to a PostCSS AST
+              context.cachedAst = cssAstToPostCssAst(ast, root.source)
```
philipp-spiess, line 230: should this be wrapped in a timing call as well for debug cases?

packages/tailwindcss/src/ast.ts
RobinMalfait: Hmm I see. The problem is that this is doing multiple things. While `applyFallbacksToAst` is correct, we are also in addition to tracking the fallbacks making sure that `@property` is handled once, that `Context` nodes become transparent and that `AtRoot` nodes are hoisted.

`prepareAstForPrinting` is a bit more vague which is good to describe that we are doing multiple things. But the `forPrinting` part is off because in `@tailwindcss/postcss` that would be `prepareAstForTransformingIntoOtherAST` 😅

packages/@tailwindcss-postcss/src/index.ts
RobinMalfait: Yeah good catch!

philipp-spiess (approved):
Superb stuff

PR #15303, opened 2024-12-04
Improve debug logs

RobinMalfait:
This PR improves the debug logs for the `@tailwindcss/postcss` integration. It uses custom instrumentation to provide a nested but detailed overview of where time is spent during the build process.

The updated logs look like this:
```
[0.15ms] [@tailwindcss/postcss] src/app/geistsans_9fc57718.module.css
[0.14ms]   ↳ Quick bail check
[0.02ms] [@tailwindcss/postcss] src/app/geistsans_9fc57718.module.css
[0.01ms]   ↳ Quick bail check

[0.03ms] [@tailwindcss/postcss] src/app/geistmono_b9f59162.module.css
[0.02ms]   ↳ Quick bail check
[0.12ms] [@tailwindcss/postcss] src/app/geistmono_b9f59162.module.css
[0.11ms]   ↳ Quick bail check

[42.09ms] [@tailwindcss/postcss] src/app/globals.css
[ 0.01ms]   ↳ Quick bail check
[12.12ms]   ↳ Setup compiler
[ 0.11ms]     ↳ PostCSS AST -> Tailwind CSS AST
[11.99ms]     ↳ Create compiler
[ 0.07ms]   ↳ Register full rebuild paths
[ 0.06ms]   ↳ Setup scanner
[ 7.51ms]   ↳ Scan for candidates
[ 5.86ms]   ↳ Register dependency messages
[ 5.88ms]   ↳ Build utilities
[ 8.34ms]   ↳ Optimization
[ 0.23ms]     ↳ AST -> CSS
[ 4.20ms]     ↳ Lightning CSS
[ 3.89ms]     ↳ CSS -> PostCSS AST
[ 1.97ms]   ↳ Update PostCSS AST
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix dependency related warnings when using `@tailwindcss/postcss` on Windows ([#15321](https://github.com/tailwindlabs/tailwindcss/pull/15321))
 - Skip creating a compiler for CSS files that should not be processed ([#15340](https://github.com/tailwindlabs/tailwindcss/pull/15340))
+- Improve debug logs to get better insights ([#15303](https://github.com/tailwindlabs/tailwindcss/pull/15303))
 
 ## [4.0.0-beta.6] - 2024-12-06
 
--- a/crates/oxide/src/lib.rs
+++ b/crates/oxide/src/lib.rs
@@ -24,7 +24,7 @@ pub mod paths;
 pub mod scanner;
 
 static SHOULD_TRACE: sync::LazyLock<bool> = sync::LazyLock::new(
-    || matches!(std::env::var("DEBUG"), Ok(value) if value.eq("*") || value.eq("1") || value.eq("true") || value.contains("tailwind")),
+    || matches!(std::env::var("DEBUG"), Ok(value) if value.eq("*") || (value.contains("tailwindcss:oxide") && !value.contains("-tailwindcss:oxide"))),
 );
 
 fn init_tracing() {
--- a/packages/@tailwindcss-cli/src/commands/build/index.ts
+++ b/packages/@tailwindcss-cli/src/commands/build/index.ts
@@ -1,12 +1,11 @@
 import watcher from '@parcel/watcher'
-import { compile, env } from '@tailwindcss/node'
+import { compile, env, Instrumentation } from '@tailwindcss/node'
 import { clearRequireCache } from '@tailwindcss/node/require-cache'
 import { Scanner, type ChangedContent } from '@tailwindcss/oxide'
 import { Features, transform } from 'lightningcss'
 import { existsSync, type Stats } from 'node:fs'
 import fs from 'node:fs/promises'
 import path from 'node:path'
-import { Instrumentation } from '../../../../@tailwindcss-node/src/instrumentation'
 import type { Arg, Result } from '../../utils/args'
 import { Disposables } from '../../utils/disposables'
 import {
--- a/packages/@tailwindcss-cli/src/commands/build/index.ts
+++ b/packages/@tailwindcss-cli/src/commands/build/index.ts
@@ -1,11 +1,12 @@
 import watcher from '@parcel/watcher'
-import { compile, env, instrumentation as I } from '@tailwindcss/node'
+import { compile, env } from '@tailwindcss/node'
 import { clearRequireCache } from '@tailwindcss/node/require-cache'
 import { Scanner, type ChangedContent } from '@tailwindcss/oxide'
 import { Features, transform } from 'lightningcss'
 import { existsSync, type Stats } from 'node:fs'
 import fs from 'node:fs/promises'
 import path from 'node:path'
+import { Instrumentation } from '../../../../@tailwindcss-node/src/instrumentation'
 import type { Arg, Result } from '../../utils/args'
 import { Disposables } from '../../utils/disposables'
 import {
@@ -66,8 +67,8 @@ async function handleError<T>(fn: () => T): Promise<T> {
 }
 
 export async function handle(args: Result<ReturnType<typeof options>>) {
-  env.DEBUG && I.reset()
-  env.DEBUG && I.start('[@tailwindcss/cli]')
+  using I = new Instrumentation()
+  env.DEBUG && I.start('[@tailwindcss/cli] (initial build)')
 
   let base = path.resolve(args['--cwd'])
 
@@ -106,7 +107,7 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
     optimizedCss: '',
   }
 
-  async function write(css: string, args: Result<ReturnType<typeof options>>) {
+  async function write(css: string, args: Result<ReturnType<typeof options>>, I: Instrumentation) {
     let output = css
 
     // Optimize the output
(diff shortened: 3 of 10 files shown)


packages/@tailwindcss-node/src/instrumentation.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  expect.assertions(1)
+
+  I.report((output) => {
+    expect(stripVTControlCharacters(output).replace(/\[.*\]/g, '[0.xxms]')).toMatchInlineSnapshot(`
+      "Hits:
+        Potato × 4
+
+      Timers:
+      [0.xxms] Foo
+      [0.xxms]   ↳ Bar × 100"
```
philipp-spiess, line 31: This is neat!

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -161,14 +194,16 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
 
             // Look for candidates used to generate the CSS
             context.scanner = new Scanner({ sources })
+            env.DEBUG && I.end('Setup scanner')
```
philipp-spiess, line 197: wonder if we should reassign `env.DEBUG` to a local variable so we save on the repeated property access

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -66,6 +67,9 @@ async function handleError<T>(fn: () => T): Promise<T> {
 }
 
 export async function handle(args: Result<ReturnType<typeof options>>) {
+  using I = new Instrumentation()
```
RobinMalfait, line 70: Ideally this is only created when `DEBUG` is true but I don't think that's possible

crates/oxide/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub mod paths;
 pub mod scanner;
 
 static SHOULD_TRACE: sync::LazyLock<bool> = sync::LazyLock::new(
-    || matches!(std::env::var("DEBUG"), Ok(value) if value.eq("*") || value.eq("1") || value.eq("true") || value.contains("tailwind")),
+    || matches!(std::env::var("DEBUG"), Ok(value) if value.eq("*") || (value.contains("tailwindcss:oxide") && !value.contains("-tailwindcss:oxide"))),
```
RobinMalfait, line 27: When using `DEBUG=1` it also shows a lot of oxide related tracing logs. This change only shows them if you are using `DEBUG=*` or `DEBUG=tailwindcss:oxide` by following the `DEBUG` conventions.

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -226,32 +233,36 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
             fullRebuildPaths = inputFilePath ? [inputFilePath] : []
 
             // Create a new compiler, given the new `input`
-            ;[compiler, scanner] = await createCompiler(input)
+            ;[compiler, scanner] = await createCompiler(input, I)
 
             // Scan the directory for candidates
-            env.DEBUG && console.time('[@tailwindcss/cli] Scan for candidates')
+            DEBUG && I.start('Scan for candidates')
```
RobinMalfait, line 239: These are called _after_ the initial `I.start('[@tailwincss/cli]')`, this allows us to remove the prefix because these logs will be indented.

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -62,27 +64,54 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
...
+              node.name === 'theme' ||
+              node.name === 'config' ||
+              node.name === 'plugin' ||
+              node.name === 'apply'
+            ) {
+              canBail = false
+              return false
+            }
+          })
+          if (canBail) return
```
RobinMalfait, line 86: This is the quick bail where we are sure that Tailwind CSS should not be used. This prevents creating the compiler entirely.

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -207,34 +241,43 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
...
+              let ast = optimizeCss(css, {
+                minify: typeof optimize === 'object' ? optimize.minify : true,
+              })
+              DEBUG && I.end('Lightning CSS')
+
+              DEBUG && I.start('CSS -> PostCSS AST')
+              context.optimizedPostCssAst = postcss.parse(ast, result.opts)
+              DEBUG && I.end('CSS -> PostCSS AST')
+
+              DEBUG && I.end('Optimization')
```
RobinMalfait, line 269: Same as before, but split up explicitly

packages/tsconfig.base.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
-    "lib": ["ESNext", "DOM"],
-    "target": "ESNext",
+    "target": "es2022",
+    "lib": ["es2022", "esnext.disposable", "dom"],
```
RobinMalfait, line 4: These are needed to compile the new `using` keyword

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Skip creating a compiler for CSS files that should not be processed ([#15303](https://github.com/tailwindlabs/tailwindcss/pull/15303))
 - Ensure `@import "…" reference` never generates utilities ([#15307](https://github.com/tailwindlabs/tailwindcss/pull/15307))
+- Improve debug logs to get better insights ([#15303](https://github.com/tailwindlabs/tailwindcss/pull/15303))
```
RobinMalfait, line 14:
```suggestion
- Ensure `@import "…" reference` never generates utilities ([#15307](https://github.com/tailwindlabs/tailwindcss/pull/15307))
- Skip creating a compiler for CSS files that should not be processed ([#15303](https://github.com/tailwindlabs/tailwindcss/pull/15303))
- Improve debug logs to get better insights ([#15303](https://github.com/tailwindlabs/tailwindcss/pull/15303))
```

packages/@tailwindcss-postcss/src/index.ts
philipp-spiess: Should this be in a separate commit?

packages/@tailwindcss-cli/src/commands/build/index.ts
RobinMalfait: If you look at the compiled output (because we can't use the native version just yet), what it essentially does is it wraps the code in a try/catch/finally and calls the dispose function in the finally block.

E.g.: https://www.typescriptlang.org/play/?#code/DYUwLgBAZg9jEF4IG8BQEJhAZzACgEoUIBfAGnQgG0BlATwFsAjGYAOgBMBLbABxmwgAuoRSUMAYxgA7bKxBtgMAOZ4A5Bxkg1BSiVT7UUAK7SJYLjIgALEMCWi0GY9i7TlEAB6JocSp7YsXEIDIA

The cool part is that you don't have to worry about all the spots where you used an early return to flush the changes to the terminal because the moment you return, the `finally` block is executed as well.

But yeah, I don't know/think you can create one conditionally because of this try/catch setup 😬

packages/@tailwindcss-postcss/src/index.ts
RobinMalfait: I think you are right, it started this way and then diverged into 2 PRs. Created a separate one here: https://github.com/tailwindlabs/tailwindcss/pull/15340

packages/@tailwindcss-cli/src/commands/build/index.ts
thecrypticace: I think it's fine. It's like a few objects and closures being created. The overhead for this is massively dwarfed by everything else. And reporting doesn't happen unless in debug mode either so that code does nothing.

thecrypticace (approved):
yay

PR #15349, opened 2024-12-09
Include `:open` in the open variant

lukewarlow:
Update the `open` variant to addtionally match the new `:open` pseudo class.

See spec at https://drafts.csswg.org/selectors-4/#open-state

See Chromium intent to ship: https://groups.google.com/a/chromium.org/g/blink-dev/c/Pdo4oOXYKcA

`:open` matches open `<details>` and `<dialog>` elements (already covered by [open]), but also matches select elements with their picker open and inputs with their picker open.

Thanks to `:is()` taking a forgiving selector list, this is safe to add even with limited browser support.


lukewarlow:
@RobinMalfait now that v4 is out, is this addition something you'd be open to?

RobinMalfait:
Yep, I think this is a good idea, especially because `:is(…)` is forgiving. Thanks!

RobinMalfait (approved):
I like it, thanks!

PR #15380, opened 2024-12-12
Reset `<th>` alignment & font weight

MartijnCuppens:
Set the `<th>` alignment to inherit rather than centering it. Also reset the font weight like we do with `h1`,`h2`,...

Demo: https://codepen.io/MartijnCuppens/pen/PoqjNxx

Bootstrap reference: https://github.com/twbs/bootstrap/pull/30323
Issue with Safari: https://github.com/whatwg/html/issues/7978
Resolved issue: https://bugs.webkit.org/show_bug.cgi?id=138577 (you can check it with the codepen demo)

### DIFF
--- a/packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
+++ b/packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
@@ -460,6 +460,11 @@ exports[`\`@import 'tailwindcss'\` is replaced with the generated CSS 1`] = `
     border-collapse: collapse;
   }
 
+  th {
+    text-align: inherit;
+    font-weight: inherit;
+  }
+
   :-moz-focusring {
     outline: auto;
   }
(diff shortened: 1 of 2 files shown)


BraveNima:
how so

philipp-spiess:
@MartijnCuppens Hey! It seems like this was fixed in Safari in 2022 or am I missing something?

Screenshot 2025-02-26 at 11 22 25

Since Tailwind CSS v4 [requires Safari 16.4+ (from March 2023)](https://tailwindcss.com/docs/compatibility), it seems like this change is no longer necessary so I'm going to close this PR but feel free to show me otherwise. Thank you!

MartijnCuppens:
Yup, the issue with Safari is changed, that's why https://github.com/twbs/bootstrap/pull/30323 wasn't needed anymore. 

This PR was about resetting the alignment and font weight of table headings, so that they are not centered and bold anymore. Definitely a breaking change btw, but It felt more consistent with `h1,h2,...` which have their font weight reset too.

philipp-spiess:
Ahh gotcha I misunderstood. Yeah I think the ship for shipping that has kinda sailed for now, that's definitely a bigger breaking change and would require a v5 release :/

PR #15421, opened 2024-12-17
Work around issue with Node 22 and Jiti

### LINKED ISSUE #15374: Plugins no longer load when using Node 22.12.0 and 23.*

v3.4.16

tailwind cli

v22.12.0
v23.*

N/A

Windows

**Reproduction URL**

N/A

**Describe your issue**

It seems module resolution changed in Node [22.12.0](https://github.com/nodejs/node/releases/tag/v22.12.0) (as well as v23) and now tailwind no longer loads plugins. The only one I'm using is `@tailwindcss/forms` and with Node v22.11.0 it's included in the output, but with v22.12.0 or v23.* it's not.

My `package.json` is set to `type": "module"` and my config is named `tailwind.config.js`.

I can resolve this issue two different ways:

1. Rename my config to use `.mjs`

2. Pass the `--no-experimental-require-module` flag when calling tailwind like so

    ```console
    NODE_OPTIONS=\"--no-experimental-require-module\" tailwindcss -i ./Styles/base.css --postcss -o ./wwwroot/css/site.css
    ```

ref: https://github.com/nodejs/node/pull/55085
ref: https://github.com/nodejs/node/issues/56155

thecrypticace:
Fixes #15374

If we always use Jiti the problem should, in theory, go away (I hope). It does mean that loading configs is slower than it would be if they're written in CJS but 🤷‍♂️ 

Wanna get this running with the integration tests to see if anything breaks.

### DIFF
--- a/src/lib/load-config.ts
+++ b/src/lib/load-config.ts
@@ -33,6 +33,10 @@ function lazyJiti() {
 
 export function loadConfig(path: string): Config {
   let config = (function () {
+    // Always use jiti for now. There is a a bug that occurs in Node v22.12+
+    // where imported files return invalid results
+    return lazyJiti()(path)
+
     // Always use jiti for ESM or TS files
     if (
       path &&


RobinMalfait (approved):
I think this is a good quickfix solution

src/lib/load-config.ts
```diff
@@ -33,6 +33,10 @@ function lazyJiti() {
 
 export function loadConfig(path: string): Config {
   let config = (function () {
+    // Always use jiti for now. There is a a bug that occurs in Node v22.12+
+    // where imported files return invalid results
+    return lazyJiti()(path)
+
     // Always use jiti for ESM or TS files
     if (
```
yoyo837, line 41: Dead code after the new line.

vhscom: That's one way to create a code smell. Approved. ;D

PR #15455, opened 2024-12-19
Add functional utility syntax

RobinMalfait:
This PR adds support for functional utilities constructed via CSS.

# Registering functional utilities in CSS

To register a functional utility in CSS, use the `@utility potato-*` syntax, where the `-*` signals that this is a functional utility:

```css
@utility tab-* {
  tab-size: --value(--tab-size-*);
}
```

## Resolving values

The special `--value(…)` function is used to resolve the utility value.

### Resolving against `@theme` values

To resolve the value against a set of theme keys, use `--value(--theme-key-*)`:

```css
@theme {
  --tab-size-1: 1;
  --tab-size-2: 2;
  --tab-size-4: 4;
  --tab-size-github: 8;
}

@utility tab-* {
  /* tab-1, tab-2, tab-4, tab-github */
  tab-size: --value(--tab-size-*);
}
```

### Bare values

To resolve the value as a bare value, use `--value({type})`, where `{type}` is the data type you want to validate the bare value as:

```css
@utility tab-* {
  /* tab-1, tab-76, tab-971 */
  tab-size: --value(integer);
}
```

### Arbitrary values

To support arbitrary values, use `--value([{type}])` (notice the square brackets) to tell Tailwind which types are supported as an arbitrary value:

```css
@utility tab-* {
  /* tab-[1], tab-[76], tab-[971] */
  tab-size: --value([integer]);
}
```

### Supporting theme values, bare values, and arbitrary values together

All three forms of the `--value(…)` function can be used within a rule as multiple declarations, and any declarations that fail to resolve will be omitted in the output:

```css
@theme {
  --tab-size-github: 8;
}

@utility tab-* {
  tab-size: --value([integer]);
  tab-size: --value(integer);
  tab-size: --value(--tab-size-*);
}
```

This makes it possible to treat the value differently in each case if necessary, for example translating a bare integer to a percentage:

```css
@utility opacity-* {
  opacity: --value([percentage]);
  opacity: calc(--value(integer) * 1%);
  opacity: --value(--opacity-*);
}
```

The `--value(…)` function can also take multiple arguments and resolve them left to right if you don't need to treat the return value differently in different cases:

```css
@theme {
  --tab-size-github: 8;
}

@utility tab-* {
  tab-size: --value(--tab-size-*, integer, [integer]);
}

@utility opacity-* {
  opacity: calc(--value(integer) * 1%);
  opacity: --value(--opacity-*, [percentage]);
}
```

### Negative values

To support negative values, register separate positive and negative utilities into separate declarations:

```css
@utility inset-* {
  inset: calc(--var(--spacing) * --value([percentage], [length]));
}

@utility -inset-* {
  inset: calc(--var(--spacing) * --value([percentage], [length]) * -1);
}
```

## Modifiers

Modifiers are handled using the `--modifier(…)` function which works exactly like the `--value(…)` function but operates on a modifier if present:

```css
@utility text-* {
  font-size: --value(--font-size-*, [length]);
  line-height: --modifier(--line-height-*, [length], [*]);
}
```

If a modifier isn't present, any declaration depending on a modifier is just not included in the output.

## Fractions

To handle fractions, we rely on the CSS `ratio` data type. If this is used with `--value(…)`, it's a signal to Tailwind to treat the value + modifier as a single value:

```css
/* The CSS `ratio` type is our signal to treat the value + modifier as a fraction */
@utility aspect-* {
  /* aspect-square, aspect-3/4, aspect-[7/9] */
  aspect-ratio: --value(--aspect-ratio-*, ratio, [ratio]);
}
```


imjeehoo:
In Resolving against @theme values, is tab-3 possible by inferring from tab-1 or is it typo of tab-4?

thecrypticace:
it was a typo — fixed.

packages/tailwindcss/src/index.ts
```diff
@@ -188,9 +191,307 @@ async function parseCss(
         )
       }
 
-      customUtilities.push((designSystem) => {
-        designSystem.utilities.static(name, () => structuredClone(node.nodes))
-      })
+      // Functional utilities. E.g.: `tab-size-*`
```
thecrypticace, line 194: I wonder if it's a good time to pull this stuff out into a separate file? This is a fairly large chunk of code.

thecrypticace (changes requested):
This example, from the tests:
```css
@utility example-* {
  --value-as-number: value(number);
  --value-as-percentage: value(percentage);
  --value-as-ratio: value(ratio);
}
```

produces this for `example-2/3`:
```css
.example-2\\/3 {
  --value-as-number: 2;
  --value-as-ratio: 2 / 3;
}
```

which feels incorrect. I'm pretty sure it should be:
```css
.example-2\\/3 {
  --value-as-ratio: 2 / 3;
}
```

Especially since this:
```css
@utility example-* {
  --value-as-number: value(number);
}
```

prints no CSS at all for `example-2/3`.

thecrypticace:
Also, we're going to want to change `value(…)` -> `--value(…)` and `modifier(…)` -> `--modifier(…)`

packages/tailwindcss/src/index.ts
```diff
@@ -188,9 +191,307 @@ async function parseCss(
...
+                  if (valueNode.kind !== 'function') return
+
+                  // Value function, e.g.: `value(integer)`
+                  if (valueNode.value === 'value') {
+                    usedValueFn = true
+
+                    for (let arg of valueNode.nodes) {
+                      // Resolving theme value, e.g.: `value(--color)`
+                      if (
+                        candidate.value?.kind === 'named' &&
```
philipp-spiess, line 245: We know that `candidate.value` is not `null` because of the check on top. Maybe we can store `value` in a separate value at that point so we can avoid the repeated property access and nullability checks in here?

packages/tailwindcss/src/index.ts
```diff
@@ -188,9 +191,307 @@ async function parseCss(
...
+                    }
+
+                    // Drop the declaration in case we couldn't resolve the value
+                    usedValueFn ||= false
+                    replaceDeclarationWith([])
+                    return ValueParser.ValueWalkAction.Stop
+                  }
+
+                  // Modifier function, e.g.: `modifier(integer)`
+                  else if (valueNode.value === 'modifier') {
```
philipp-spiess, line 361: There's quite some repetition between the modifier and the value function, maybe we can avoid some code duplication here by adding a more generic resolver function?

philipp-spiess (commented):
Awesome stuff! @thecrypticace brought up some good points

RobinMalfait:
> This example, from the tests:
> 
> ```css
> @utility example-* {
>   --value-as-number: value(number);
>   --value-as-percentage: value(percentage);
>   --value-as-ratio: value(ratio);
> }
> ```
> 
> produces this for `example-2/3`:
> 
> ```css
> .example-2\\/3 {
>   --value-as-number: 2;
>   --value-as-ratio: 2 / 3;
> }
> ```
> 
> which feels incorrect. I'm pretty sure it should be:
> 
> ```css
> .example-2\\/3 {
>   --value-as-ratio: 2 / 3;
> }
> ```
> 
> Especially since this:
> 
> ```css
> @utility example-* {
>   --value-as-number: value(number);
> }
> ```
> 
> prints no CSS at all for `example-2/3`.

Yeah that's true and confusing. The idea is that if you have `example-2/3` that you either have to use `--value(number)` _and_ `--modifier(number)` to get the `2` and `3` out. OR you have to use `--value(ratio)` and then the others shouldn't resolve because that's a conflict. I think we can refine that.

packages/tailwindcss/src/index.ts
RobinMalfait: Merged the codepaths. There was some handling that isn't exactly the same (around `ratio` and `fraction`, but the differences are not too bad).

packages/tailwindcss/src/index.ts
```diff
@@ -26,12 +26,15 @@ import { substituteFunctions } from './css-functions'
...
 import { Theme, ThemeOptions } from './theme'
+import { inferDataType, isPositiveInteger, isValidSpacingMultiplier } from './utils/infer-data-type'
 import { segment } from './utils/segment'
+import * as ValueParser from './value-parser'
 import { compoundsForSelectors } from './variants'
 export type Config = UserConfig
 
 const IS_VALID_PREFIX = /^[a-z]+$/
 const IS_VALID_UTILITY_NAME = /^[a-z][a-zA-Z0-9/%._-]*$/
+const IS_VALID_FUNCTIONAL_UTILITY_NAME = /^-?[a-z][a-zA-Z0-9/%._-]*-\*$/
```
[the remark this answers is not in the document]
RobinMalfait: Yeah, what do you think, just drop the `A-Z` part?

Complex part is if you have `tab-*` but you use `tab-myTabSize`, should we validate that?

thecrypticace (approved):

packages/tailwindcss/src/index.ts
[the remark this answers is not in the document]
RobinMalfait: So after looking into this, what we disallow right now (in Oxide) is camelCasing _if_ it's the first segment of the candidate. So `fooBar` would not work. `foo-Bar` would work, `hover:fooBar` would also work.

This is a bit confusing in general, but feels like outside of the scope of this PR to fix. However, @thecrypticace and I talked about this, and we believe the proper way to solve this, is to provide the `root`s of the `static` and `functional` utilities and the variants to Oxide such that we have a much smaller and more correct list to work with. There could of course still be false positives (e.g.: `text-md` that looks okay, but might not exist). In that case we could provide the full theme as well, but that seems overkill.

Will explore this after the v4 launch, probably.

packages/tailwindcss/src/utilities.ts
```diff
@@ -4569,3 +4584,360 @@ export function createUtilities(theme: Theme) {
...
+    //                                arbitrary value of type number in order.
+    //
+    // Rules:
+    //
+    // - If `--value(…)` does not resolve to a valid value, the declaration is
+    //   removed.
+    // - If a `--value(ratio)` resolves, the `--modifier(…)` cannot be used.
+    // - If a candidate looks like `foo-2/3`, then the `--value(ratio)` should
+    //   be used OR the `--value(…)` and `--modifier(…)` must be used. But not
+    //   both.
```
philipp-spiess, line 4608: is there any valid use case for resolving a ratio like `foo-2/3` as `--value(…)` and `--modifier(…)`? If not I wonder if this might simplify things a bit since we can just determine in advance wether the input looks like a `ratio` and never resolve the `--value(number)` in this case.

packages/tailwindcss/src/utilities.ts
```diff
@@ -4569,3 +4584,360 @@ export function createUtilities(theme: Theme) {
...
+            modifiers: designSystem.theme
+              .keysInNamespaces(modifierThemeKeys)
+              .map((x) => x.replaceAll('_', '.')),
+          },
+        ] satisfies SuggestionGroup[]
+      })
+    }
+  }
+
+  if (IS_VALID_UTILITY_NAME.test(name)) {
```
philipp-spiess, line 4832:
```suggestion
  if (IS_VALID_STATIC_UTILITY_NAME.test(name)) {
```

should we rename this too now for consistency with the previous `IS_VALID_FUNCTIONAL_UTILITY_NAME` branch?

philipp-spiess (commented):
This is looking really good already! Left a question inline about the `ratio` branches, cause I wonder if we can simplify this a bit if we can avoid resolving `--value(integer)` at all when a ratio is used. Not sure if we can, though

packages/tailwindcss/src/utilities.ts
RobinMalfait: Sadly there is, if you define your font-sizes as numbers instead of tshirt sizes. E.g.: `text-4/7` could be `text-4/7` (font-size of `4/7` whatever that means) or `text-4` with line-height of `7`.

philipp-spiess (approved):

stafyniaksacha:
Could we pipe this to @apply directive ? 

Using new variables:
```css
@theme {
  --tab-size-1: 1;
  --tab-size-2: 2;
  --tab-size-4: 4;
  --tab-size-github: 8;
}

@utility tab-* {
  /* tab-1, tab-2, tab-4, tab-github */
  tab-size: --value(--tab-size-*);
}

@utility foo {
  @apply tab-1
}
```


Using existing variables:
```css
@theme {
  // ...
  --color-amber-500: #fff;
  // ...
}

@utility card-* {
  /* card-sky-100 */
  background-color: --value(--color-*);
}

@utility bar {
  @apply card-sky-100
}
```

mscofield0:
> Could we pipe this to @apply directive ?
> 
> Using new variables:
> 
> ```css
> @theme {
>   --tab-size-1: 1;
>   --tab-size-2: 2;
>   --tab-size-4: 4;
>   --tab-size-github: 8;
> }
> 
> @utility tab-* {
>   /* tab-1, tab-2, tab-4, tab-github */
>   tab-size: --value(--tab-size-*);
> }
> 
> @utility foo {
>   @apply tab-1
> }
> ```
> 
> Using existing variables:
> 
> ```css
> @theme {
>   // ...
>   --color-amber-500: #fff;
>   // ...
> }
> 
> @utility card-* {
>   /* card-sky-100 */
>   background-color: --value(--color-*);
> }
> 
> @utility bar {
>   @apply card-sky-100
> }
> ```

I would also like an answer to this, I would assume any tailwind utility is able to be used in `@apply`.

PR #15456, opened 2024-12-19
Don’t detect arbitrary properties when preceded by an escape

thecrypticace:
This is a targeted bug fix uncovered by the v4 docs.

Given this code:
```html
<!-- [!code word:group-has-\\[a\\]\\:block] -->
```

We'd pick up `[a\\]\\:block]` as a candidate which would then make it far enough to get output to CSS and throw an error. This makes sure we don't try to start an arbitrary property if the preceding character is a `\`

cc @RobinMalfait this look okay?

### DIFF
--- a/crates/oxide/src/parser.rs
+++ b/crates/oxide/src/parser.rs
@@ -1636,7 +1636,7 @@ mod test {
     }
 
     #[test]
-    fn wip2() {
+    fn arbitrary_properties_are_not_picked_up_after_an_escape() {
         _please_trace();
         let candidates = run(
             r#"
--- a/crates/oxide/src/parser.rs
+++ b/crates/oxide/src/parser.rs
@@ -595,7 +595,7 @@ impl<'a> Extractor<'a> {
     fn parse_start(&mut self) -> ParseAction<'a> {
         match self.cursor.curr {
             // Enter arbitrary property mode
-            b'[' => {
+            b'[' if self.cursor.prev != b'\\' => {
                 trace!("Arbitrary::Start\t");
                 self.arbitrary = Arbitrary::Brackets {
                     start_idx: self.cursor.pos,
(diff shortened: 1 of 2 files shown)


crates/oxide/src/parser.rs
```diff
@@ -1634,4 +1634,15 @@ mod test {
             ]
         );
     }
+
+    #[test]
+    fn wip2() {
```
philipp-spiess, line 1639: Title needs updating 😅

philipp-spiess (commented):
Looks good but the test needs a proper title 👍

crates/oxide/src/parser.rs
thecrypticace: oh yeah lol good call

crates/oxide/src/parser.rs
```diff
@@ -755,7 +755,7 @@ impl<'a> Extractor<'a> {
 
     #[inline(always)]
     fn parse_char(&mut self) -> ParseAction<'a> {
-        if !matches!(self.arbitrary, Arbitrary::None) {
+      if !matches!(self.arbitrary, Arbitrary::None) {
```
RobinMalfait, line 758: Not sure what happened here? 🤔

crates/oxide/src/parser.rs
```diff
@@ -595,7 +595,7 @@ impl<'a> Extractor<'a> {
     fn parse_start(&mut self) -> ParseAction<'a> {
         match self.cursor.curr {
             // Enter arbitrary property mode
-            b'[' => {
+            b'[' if self.cursor.prev != b'\\' => {
```
RobinMalfait, line 598: Easy 😍

RobinMalfait (approved):

philipp-spiess (approved):

PR #15542, opened 2025-01-05
Ensure `@utility` is processed before using them

### LINKED ISSUE #15501: [v4] background utility class is not generated when also applied to body

v4.0.0-beta.8

Vite, but it is also reproducible in playground

22.12.0 (not sure about playground)

Chrome

Windows 11

**Reproduction URL**

https://play.tailwindcss.com/uDAJ3VP1OF

**Describe your issue**

- in the v4 css config, create a `@utility` that generate some background-related classname
  - in the playground link above: `@utility bg-th { @apply bg-red-500; }`
- in the v4 css config, apply this `@utility` to the `body`
  - in the playground link above: `@layer base { body { @apply bg-th; } }`
- in your html, use that same `@utility` class
  - in the playground link above: `<div class="bg-th">` 

- the **red-500** color is correctly applied to the `body`
- but the same **red-500**  is not applied to the div with `bg-th` class (in the playground link above it inherits its background color from its parent instead)

Without having a look at the codebase, I suspect this happens because when the `bg-th` utility is "applied" to the body it is also removed from the list of available classes, regardless of the fact that it is used at other places.
This is based on the fact that commenting the `@apply bg-th` line in the playground link above results in the `bg-th` class being correctly generated.

RobinMalfait:
This PR fixes an issue where using an `@utility` before it is defined, and _if_ that `@utility` contains `@apply`, that it won't result in the expected output. But results in an `@apply` rule that is not substituted. Additionally, if you have multiple levels of `@apply`, we have to make sure that everything is applied (no pun intended) in the right order.

Right now, the following steps are taken:

1. Collect all the `@utility` at-rules (and register them in the system as utilities).
2. Substitute `@apply` on the AST (including `@utility`'s ASTs) with the content of the utility.
3. Delete the `@utility` at-rules such that they are removed from the CSS output itself.

The reason we do it in this order is because handling `@apply` during `@utility`
handling means that we could rely on another `@utility` that is defined later
and therefore the order of the utilities starts to matter. This is not a bad
thing, but the moment you import multiple CSS files or plugins, this could
become hard to manage.

Another important step is that when using `@utility foo`, the implementation creates a `structuredClone` from its AST when first using the utility. The reason we do that is because `foo` and `foo!` generate different output and we don't want to accidentally mutate the same AST. This structured clone is the start of the problem in the linked issue (#15501).

If we don't do the structured clone, then substituting the `@apply` rules would work, but then `foo` and `foo!` will generate the same output, which is bad.

The linked issue has this structure:
```css
.foo {
  @apply bar;
}

@utility bar {
  @apply flex;
}
```

If we follow the steps above, this would substitute `@apply bar` first, which
results in:

```css
.foo {
  @apply flex;
}
```

But the `bar` utility, was already cloned (and cached) so now we end up with an `@apply` rule that is not substituted.

To properly solve this problem, we have to make sure that we collect all the `@apply` at-rules, and apply them in the correct order. To do this, we run a topological sort on them which ensures that all the dependencies are applied before substituting the current `@apply`.

This means that in the above example, in order to process `@apply bar`, we have to process the `bar` utility first.

If we run into a circular dependency, then we will throw an error like before. You'll notice that the error message in this PR is updated to a different spot. This one is a bit easier to grasp because it shows the error where the circular dependency _starts_ not where it _ends_ (and completes the circle). The previous message was not wrong (since it's a circle), but now it's a bit easier to reason about.

Fixes: #15501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `inset-shadow-*` suggestions in IntelliSense ([#15471](https://github.com/tailwindlabs/tailwindcss/pull/15471))
 - Only compile arbitrary values ending in `]` ([#15503](https://github.com/tailwindlabs/tailwindcss/pull/15503))
 - Improve performance and memory usage ([#15529](https://github.com/tailwindlabs/tailwindcss/pull/15529))
-- Ensure `@utility` is fully processed before it's used ([#15542](https://github.com/tailwindlabs/tailwindcss/pull/15542))
+- Ensure `@apply` rules are processed in the correct order ([#15542](https://github.com/tailwindlabs/tailwindcss/pull/15542))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `inset-shadow-*` suggestions in IntelliSense ([#15471](https://github.com/tailwindlabs/tailwindcss/pull/15471))
 - Only compile arbitrary values ending in `]` ([#15503](https://github.com/tailwindlabs/tailwindcss/pull/15503))
 - Improve performance and memory usage ([#15529](https://github.com/tailwindlabs/tailwindcss/pull/15529))
+- Ensure `@utility` is fully processed before it's used ([#15542](https://github.com/tailwindlabs/tailwindcss/pull/15542))
 
 ### Changed
 
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -538,24 +538,15 @@ async function parseCss(
     node.context = {}
   }
 
-  // Post-process `@utility` nodes.
-  //
-  // We have to do this as a separate step because if we replace the `@utility`
-  // nodes when we collect it in the system, then we can't substitute the
-  // `@apply` at-rules anymore.
-  //
-  // We also can't substitute the `@apply` at-rules while collecting the
-  // `@utility` rules, because if it relies on a utility that is defined later,
-  // then we wouldn't be able to resolve the applied utility.
-  //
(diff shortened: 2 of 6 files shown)


thecrypticace (changes requested):
This breaks when generating `b` and `c`:
```css
@utility a {
  @apply b;
}

@utility b {
  @apply c;
}

@utility c {
  @apply flex;
}
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `inset-shadow-*` suggestions in IntelliSense ([#15471](https://github.com/tailwindlabs/tailwindcss/pull/15471))
 - Only compile arbitrary values ending in `]` ([#15503](https://github.com/tailwindlabs/tailwindcss/pull/15503))
 - Improve performance and memory usage ([#15529](https://github.com/tailwindlabs/tailwindcss/pull/15529))
+- Ensure `@utility` is fully processed before it's used ([#15542](https://github.com/tailwindlabs/tailwindcss/pull/15542))
```
RobinMalfait, line 18:
```suggestion
- Ensure `@apply` rules are processed in the correct order ([#15542](https://github.com/tailwindlabs/tailwindcss/pull/15542))
```

packages/tailwindcss/src/index.ts
```diff
@@ -538,10 +538,8 @@ async function parseCss(
     node.context = {}
   }
 
-  // Replace `@apply` rules with the actual utility classes.
-  features |= substituteAtApply(ast, designSystem)
-
   features |= substituteFunctions(ast, designSystem.resolveThemeValue)
+  features |= substituteAtApply(ast, designSystem)
```
RobinMalfait, line 542: Swapped these two around, because `substituteAtApply` will result in a larger tree, which means that the `substituteFunctions` can operate on a smaller tree.

thecrypticace: yay

thecrypticace (approved):

PR #15553, opened 2025-01-06
Fix type errors and improve type safety in variant processing logic.

stormdotcom:
Expanding the Node interface to include broader types like declaration and other possible node kinds.
Adding null and undefined checks for node.selector, node.name, and node.params to prevent runtime errors.
Explicitly casting types for path and node.nodes in the walkDepth function to ensure compatibility.
Refining type handling for design.variants.entries() and design.variants.getCompletions() to align with expected structures

### DIFF
--- a/packages/tailwindcss/src/intellisense.ts
+++ b/packages/tailwindcss/src/intellisense.ts
@@ -1,160 +1,112 @@
-import { styleRule, walkDepth } from './ast'
-import { applyVariant } from './compile'
-import type { DesignSystem } from './design-system'
-
-interface ClassMetadata {
-  modifiers: string[]
-}
-
-export type ClassEntry = [string, ClassMetadata]
-
-export function getClassList(design: DesignSystem): ClassEntry[] {
-  let list: [string, ClassMetadata][] = []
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey!

Thanks for the PR, I'm curious to know what issues you were running into that prompted you to open this PR.

For what it's worth, this PR has multiple issues and _introduces_ type errors instead of fixing them. Additionally, these changes also introduce a different code style (make sure to run prettier for example).

Some issues I see:
1. Introduces custom types/interfaces — we rely on the types in the system already. This introduces new types that don't match the expected types from the system.
   1. For example this introduces a single `Node` interface, but that means that it is possible to create a `Declaration` that contains `nodes`, which doesn't really make sense from a CSS perspective. 
1. I think you might have run into these issues, which is probably why you added `as ABC` in some spots. Casting types like this often hide problem (these now cause type issues) 
image

While I really appreciate you opening a PR, I think we have to take a step back and start with the issue(s) you are running into and start from there. 

Can you open a new issue with the issues you ran into?

Thanks!

PR #15558, opened 2025-01-06
Add `@tailwindcss/browser` package

thecrypticace:
We're going to publish the CDN / browser build of v4 as an npm package so it can be loaded from things like unpkg. This sets up the package for that. The CDN is still intended only for development use. **Please do not use it in production.**

Note: like the v3 CDN, since it  reacts to changes in the DOM, some things like transitions can break if the transition relies on the class being defined before being added to the DOM for correct behavior.


thecrypticace:
@RobinMalfait Is there anything special I need to do to make sure publishing will work normally?

RobinMalfait:
> @RobinMalfait Is there anything special I need to do to make sure publishing will work?

Nope should be good! We publish via `pnpm` which uses the workspaces, so as long as they are setup it should be good. The only other thing is if we publish additional packages that need to be in sync with the `@tailwindcss/browser` version, but that's not the case so should also be fine. Otherwise we would've to edit the `version-packages` script.

You can run `pnpm --recursive publish --tag next --no-git-checks --dry-run` to test it (note: this only results in output for packages that need to be published. Already published packages will not show up and will be a no-op)

packages/@tailwindcss-browser/src/index.ts
```diff
@@ -0,0 +1,295 @@
...
+  // The input CSS did not change so the compiler does not need to be recreated
+  if (lastCss === css) return
+
+  lastCss = css
+
+  I.start('Compile CSS')
+  try {
+    compiler = await tailwindcss.compile(css, {
+      base: '/',
+      async loadStylesheet(id, base) {
```
philipp-spiess, line 104: This function doesn't use anything from outside its scope so we might want to move it into the module scope instead.

philipp-spiess (approved):
Pretty impressive how simple the code for this is. 👏

packages/@tailwindcss-browser/src/index.ts
```diff
@@ -0,0 +1,295 @@
...
+  I.end(`Build utilities`)
+}
+
+function rebuild(kind: 'full' | 'incremental') {
+  async function run() {
+    if (!compiler && kind !== 'full') {
+      return
+    }
+
+    let buildId = nextBuildId++
```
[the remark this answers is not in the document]
thecrypticace: At 1,000 builds per millisecond it would take 285.62 *years* to overflow past the max safe integer in JS so I think we're good. 😅

packages/@tailwindcss-browser/src/index.ts
thecrypticace: Yeah can do that 👍

package.json
```diff
@@ -39,7 +39,7 @@
     "dev": "turbo dev --filter=!./playgrounds/*",
     "test": "cargo test && vitest run --hideSkippedTests",
     "test:integrations": "vitest --root=./integrations",
-    "test:ui": "pnpm run --filter=tailwindcss test:ui",
+    "test:ui": "pnpm run --filter=tailwindcss test:ui && pnpm run --filter=@tailwindcss/browser test:ui",
```
RobinMalfait, line 42: Can/should we merge these? I think we can use `--filter` twice 🤔

RobinMalfait (approved):
Looks nice, I like the performance marks as well 
image

package.json
thecrypticace: If you do that it does some super super weird stuff to the console output

RobinMalfait: Just tried it, that is … interesting 😂

PR #15566, opened 2025-01-07
Upgrade: Do not extract class names from functions 

philipp-spiess:
This PR prevents the migration of utilities detected in function names, e.g.: the use of `shadow` inside `filter: 'drop-shadow(…)'`.

## Test plan

Have content like this in a project you're migrating using the upgrade tool:

```js
{
  filter: 'drop-shadow(0 0 0.5rem #000)'
}
```

This was verified by adding unit tests to the specific codemods and adding an integration test.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.test.ts
+++ b/packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.test.ts
@@ -39,3 +39,33 @@ test.each([
 
   expect(await legacyClasses(designSystem, {}, candidate)).toEqual(result)
 })
+
+test('does not replace classes in invalid positions', async () => {
+  let designSystem = await __unstable__loadDesignSystem('@import "tailwindcss";', {
+    base: __dirname,
+  })
+
+  async function shouldNotReplace(example: string, candidate = 'shadow') {
+    expect(
+      await legacyClasses(designSystem, {}, candidate, {
+        contents: example,
+        start: example.indexOf(candidate),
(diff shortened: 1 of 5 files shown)


packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.test.ts
```diff
@@ -39,3 +39,33 @@ test.each([
...
+  })
+
+  async function shouldNotDetect(example: string, candidate = 'shadow') {
+    expect(
+      await legacyClasses(designSystem, {}, candidate, {
+        contents: example,
+        start: example.indexOf(candidate),
+        end: example.indexOf(candidate) + candidate.length,
+      }),
+    ).toEqual('shadow')
```
thecrypticace, line 55: Should this be this or nah?

```suggestion
    ).toEqual(candidate)
```

packages/@tailwindcss-upgrade/src/template/codemods/legacy-classes.test.ts
```diff
@@ -39,3 +39,33 @@ test.each([
 
   expect(await legacyClasses(designSystem, {}, candidate)).toEqual(result)
 })
+
+test('does not match false positives', async () => {
```
thecrypticace, line 43: This test name feels a bit weird. If it doesn't match them they're no longer false positives. I can't immediately think of a better name tho

thecrypticace (approved):

PR #15571, opened 2025-01-08
Remove extra commas in `ring-inset`

zyyv:
Chore: Remove extra commas in `ring-inset`. 
This comma doesn't seem to have any purpose.


thecrypticace:
Hi! This is intentional — the extra comma serves as an empty fallback value.

PR #15579, opened 2025-01-09
Migrate `theme(…)` to `--theme(…)`, migrate `calc(var(--spacing)*x)` to `--spacing(x)`

RobinMalfait:
This PR improves the codemod tool to simplify 2 things:

1. Whenever you have a `theme(…)` call, we try to change it to a `var(…)`, but if that doesn't work for some reason, we will make sure to at least convert it to the more modern `--theme(…)`.
2. When converting `theme(spacing.2)`, we used to convert it to `calc(var(--spacing)*2)`, but now we will convert it to `--spacing(2)` instead.


RobinMalfait:
TODO: 
- [ ] Make sure that we _only_ use `--theme()` _if_ it's using the modern syntax, not the old dot-notation based syntax.

philipp-spiess (approved):

PR #15591, opened 2025-01-10
Include outline-color in default color properties to transition (v4 edition)

benface:
This was done for v3 (#10385), but then reverted (#10604) because it was a breaking change. This is the same change but for v4.


RobinMalfait:
Hey!

Thanks for the PR! Closed this PR in favor of #15690, but I realised too late (@thecrypticace noticed it actually) that this PR was open (so referenced it last minute), so sorry about not using your PR here. Still really appreciate this PR from you!

PR #15622, opened 2025-01-14
Remove `@property` fallbacks for Firefox

philipp-spiess:
This PR removes the `@property` fallbacks added in #13655. This is possible because we're targeting a minimum Firefox version of 128 which [includes support for \`@property\` rules](https://developer.mozilla.org/de/docs/Web/CSS/@property).

Screenshot 2025-01-14 at 11 36 44


RobinMalfait (approved):
😍

image

PR #15663, opened 2025-01-17
Implement `@variant`

RobinMalfait:
This PR replaces `@variant` with `@custom-variant` for registering custom variants via your CSS.

In addition, this PR introduces `@variant` that can be used in your CSS to use a variant while writing custom CSS.

E.g.:

```css
.btn {
  background: white;

  @variant dark {
    background: black;
  }
}
```

Compiles to:

```css
.btn {
  background: white;
}

  .btn {
    background: black;
  }
}
```

For backwards compatibility, the `@variant` rules that don't have a body and are
defined inline:

```css
@variant hocus (&:hover, &:focus);
```

And `@variant` rules that are defined with a body and a `@slot`:

```css
@variant hocus {
  &:hover, &:focus {
    @slot;
  }
}
```

Will automatically be upgraded to `@custom-variant` internally, so no breaking changes are introduced with this PR.

---

TODO:
- [x] ~~Decide whether we want to allow multiple variants and if so, what syntax should be used. If not, nesting `@variant <variant> {}` will be the way to go.~~ Only a single `@variant <variant>` can be used, if you want to use multiple, nesting should be used:

```css
.foo {
  @variant hover {
    @variant focus {
      color: red;
    }
  }
}
```

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -256,11 +256,23 @@ export function optimizeAst(ast: AstNode[]) {
 
     // Rule
     else if (node.kind === 'rule') {
-      let copy = { ...node, nodes: [] }
-      for (let child of node.nodes) {
-        transform(child, copy.nodes, depth + 1)
+      // Rules with `&` as the selector should be flattened
+      if (node.selector === '&') {
+        for (let child of node.nodes) {
+          let nodes: AstNode[] = []
+          transform(child, nodes, depth + 1)
+          parent.push(...nodes)
+        }
       }
-      parent.push(copy)
+
+      //
+      else {
+        let copy = { ...node, nodes: [] }
+        for (let child of node.nodes) {
+          transform(child, copy.nodes, depth + 1)
+        }
+        parent.push(copy)
+      }
     }
 
     // AtRule `@property`
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -547,16 +547,16 @@ async function parseCss(
       // Starting with the `&` rule node
       let node = styleRule('&', variantNode.nodes)
 
-      for (let variant of segment(variantNode.params, ':').reverse()) {
-        let variantAst = designSystem.parseVariant(variant)
-        if (variantAst === null) {
-          throw new Error(`Cannot use \`@variant\` with unknown variant: ${variant}`)
-        }
+      let variant = variantNode.params
 
-        let result = applyVariant(node, variantAst, designSystem.variants)
-        if (result === null) {
-          throw new Error(`Cannot use \`@variant\` with variant: ${variant}`)
-        }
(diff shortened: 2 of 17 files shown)


packages/tailwindcss/src/ast.ts
```diff
@@ -256,11 +256,23 @@ export function optimizeAst(ast: AstNode[]) {
...
-      for (let child of node.nodes) {
-        transform(child, copy.nodes, depth + 1)
+      // Rules with `&` as the selector should be flattened
+      if (node.selector === '&') {
+        for (let child of node.nodes) {
+          let nodes: AstNode[] = []
+          transform(child, nodes, depth + 1)
+          parent.push(...nodes)
+        }
+      }
```
RobinMalfait, line 266: Just an optimization to flatten:
```css
.foo {
  & {
    color: red;
  }
}
```

Into:
```css
.foo {
  color: red;
}
```

Which is more relevant when introducing this new `@variant` you can use.

packages/tailwindcss/src/index.ts
```diff
@@ -501,6 +541,30 @@ async function parseCss(
     node.context = {}
   }
 
+  // Replace the `@variant` at-rules with the actual variant rules.
+  if (variantNodes.length > 0) {
+    for (let variantNode of variantNodes) {
+      // Starting with the `&` rule node
+      let node = styleRule('&', variantNode.nodes)
+
+      for (let variant of segment(variantNode.params, ':').reverse()) {
```
RobinMalfait, line 550: To decide:

Right now this code means that we can use `@variant md:dark:hover {}`. If we _only_ want to allow a single `@variant`, then this could be simplified to just:
```suggestion
      for (let variant of [variantNode.params]) {
```

And get rid of the loop.

packages/tailwindcss/src/index.ts
```diff
@@ -501,6 +541,30 @@ async function parseCss(
...
+    for (let variantNode of variantNodes) {
+      // Starting with the `&` rule node
+      let node = styleRule('&', variantNode.nodes)
+
+      let variant = variantNode.params
+
+      let variantAst = designSystem.parseVariant(variant)
+      if (variantAst === null) {
+        throw new Error(`Cannot use \`@variant\` with unknown variant: ${variant}`)
+      }
```
thecrypticace, line 555: Could someone theoretically register a `hover:focus` variant through configs and then try using that with `@variant`?

We might want explicitly to error on uses of `:` if that's possible

thecrypticace (approved):

philipp-spiess (approved):
Awesome!

packages/tailwindcss/src/index.ts
RobinMalfait: So you can technically add a custom variant with a `:`, but you will never be able to use it because we split it on `:` anyway.

Added some validation (relatively strict) which we can loosen up down the line if needed.

thecrypticace (approved):

PR #15693, opened 2025-01-21
Prepare v4.0.0 release

thecrypticace:


RobinMalfait (approved):

vladshcherbin:
Was it by mistake?

Docs don't have v4 in version select, there's no tag or blog post about release and 4.0.0 doesn't have latest tag on npm 🤔

effulgentsia:
The title of this MR is "prepare", so assuming it wasn't a mistake, I think it means a 4.0.0 release is coming soon (hours? days?), not that it's already here.

florianbouvot:
It’s comming later today, for more info follow @adamwathan on X.

But you can still install and use 4.0.0

PR #15698, opened 2025-01-22
Merge changelogs

RobinMalfait:
This PR makes sure all v3 and earlier versions are visible in the changelog as
well as the v4 changelog entries.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Nothing yet
 
 ## [4.0.0] - 2025-01-21
+
```
philipp-spiess, line 13: Feels weird that this is empty, hmm. Should we include something here?

thecrypticace: Should we maybe repeat (some of?) the info from the GitHub release?

RobinMalfait: Added some of the contents of the GitHub release.

I also added compare links for the individual alpha/beta releases (but before v4 alpha 21, we didn't publish GitHub releases so those aren't included)

philipp-spiess (approved):

PR #15738, opened 2025-01-23
Ensure CSS variable shorthand uses valid CSS variables

RobinMalfait:
This PR fixes an issue where we didn't always correctly validated invalid CSS variables when using the CSS variable shorthand syntax.

This PR fixes that by ensuring we start with `--`. We can't validate that the variable exists at runtime, but we can validate if it looks like a valid CSS variable.

We can go a step further by validating if the CSS variable is valid syntax (e.g.: all characters are valid), but we can introduce this if this is causing actual issues in projects.

```
p-(a-b)
```

Used to compile to:
```css
.p-\(a-b\) {
  padding: var(a-b);
}
```

But not anymore.


packages/tailwindcss/src/candidate.test.ts
```diff
@@ -616,6 +623,13 @@ it('should parse a utility with an arbitrary value with parens including a typeh
   `)
 })
 
+it('should not parse a utility with an arbitrary value with parens including a typehint that does not start with --', () => {
+  let utilities = new Utilities()
+  utilities.functional('bg', () => [])
+
+  expect(run('bg-(color:my-color)', { utilities })).toMatchInlineSnapshot(`[]`)
+})
+
```
thecrypticace, line 632: We should add tests for stuff like bg-foo/(bar) to make sure it's thrown out too

RobinMalfait: Ah good call!

thecrypticace (approved):

Issue #15844, opened 2025-01-25
Cannot build plugin: types in `tailwindcss/plugin` are broken


cyberalien:
**What version of Tailwind CSS are you using?**

4.0.0

**What build tool (or framework if it abstracts the build tool) are you using?**

typescript: 5.7.2

**What version of Node.js are you using?**

20.14.0

**What operating system are you using?**

macOS 15.2

**Reproduction URL**

https://github.com/iconify/iconify-tailwind/blob/main/packages/plugin/src/plugin.ts

**Describe your issue**

Reproduction includes a very basic plugin:
```ts
import plugin from 'tailwindcss/plugin';

const exportedPlugin = plugin(() => {
	return ({ matchComponents }) => {
		matchComponents({
			test: (content: string) => {
				return {
					color: content,
				};
			},
		});
	};
});

export default exportedPlugin;
```
On line 3 TypeScript throws the following error:
```
src/plugin.ts:3:7 - error TS2742: The inferred type of 'exportedPlugin' cannot be named without a reference to '../node_modules/tailwindcss/dist/types-BTRmm49E.mjs'. This is likely not portable. A type annotation is necessary.

3 const exportedPlugin = plugin(() => {
        ~~~~~~~~~~~~~~
```
Building it using `tsc -b`.

Removed all unnecessary dependencies and packages to show smallest possible reproduction.

cyberalien:
Changed reproduction URL to a smaller reproduction without monorepo setup.

rijenkii:
Think this is a TS problem rather than tailwind, see possible workarounds: https://github.com/microsoft/TypeScript/issues/47663#issuecomment-1519138189

Specifically, this seems to work:
```diff
 import plugin from 'tailwindcss/plugin';

-const exportedPlugin = plugin(() => {
+const exportedPlugin: ReturnType<typeof plugin> = plugin(() => {
 	return ({ matchComponents }) => {
 		matchComponents({
 			test: (content: string) => {
 				return {
 					color: content,
 				};
 			},
 		});
 	};
 });

export default exportedPlugin;
```

Resulting `lib/plugin.d.ts`:
```typescript
import plugin from 'tailwindcss/plugin';
declare const exportedPlugin: ReturnType<typeof plugin>;
export default exportedPlugin;
```

cyberalien:
Fixed in https://github.com/tailwindlabs/tailwindcss/pull/15869

silverwind:
[...]

PR #15869, opened 2025-01-25
fix: default package exports

userquin:
This PR includes the following changes to the `packages/tailwindcss` package (check **Context**):
- fixes the default packages exports for node10 and node16 cjs types
- export types in `plugin.ts` module, cannot be used with Typescript 5.7
- add `"type": "commonjs"` to package.json
- add `./lib/util/flattenColorPalette.js` variant to package.json exports (`publishConfig`)

**Context**
The current version at npm registry has all types broken for all packages exports. The problem seems to be with tsup since it is not applying the `.cts` hack, all `d.ts` files using `export { X as default };`, check https://arethetypeswrong.github.io/?p=tailwindcss%404.0.0 .

With this PR (for context check this PR https://github.com/unjs/unbuild/pull/475 and this md file https://github.com/unjs/unbuild/blob/a011a7831e2004688bf44716f5e9c9377d08ebec/src/builders/rollup/plugins/cjs.md):

image

This PR includes a new script module that runs after tsup finish building the dist, extracting the default export and applying some replacements in the files including the default `export` correclty: `export = X;` and remaining named exports in the default export statement.

About exporting the types in the `plugin.ts` module, when using TypeScript 5.7 there is an error. Using the following code with any module resolution

```ts
import plugin from 'tailwindcss/plugin'

const exportedPlugin = plugin.withOptions(() => {
	return ({ matchComponents }) => {
		matchComponents({
			test: (content: string) => {
				return {
					color: content,
				};
			},
		});
	};
});

export default exportedPlugin;
```

and building it via `tsc -b`, there is an error about required annotation:

image

Using a local tgz from this PR, TypeScript is fine when building the module:

image

You can generate the local tgz using this PR in the following repository (use packages/plugin): https://github.com/iconify/iconify-tailwind

### DIFF
--- a/packages/tailwindcss/fix-default-cjs-exports.mjs
+++ b/packages/tailwindcss/fix-default-cjs-exports.mjs
@@ -16,36 +16,34 @@ const __dirname = dirname(fileURLToPath(import.meta.url))
  * export { type X, Y }; // <== this line will be only added if there are additional named exports
  * ```
  * @param dtsPath {string}
+ * @returns {Promise<void>}
  */
-function fixDefaultCjsExport(dtsPath) {
-  return readFile(dtsPath, { encoding: 'utf-8' })
-    .then((content) => {
-      let match = content.match(regexp)
-      if (!match) return undefined
-      const fullExport = match[0]
-      /**@type {string | undefined}*/
-      let defaultExport = undefined
-      /**@type {string[]} */
-      const namedExports = []
-      for (const exp of match[1].split(',').map((e) => e.trim())) {
-        match = exp.match(defaultExportRegexp)
-        if (match) {
-          defaultExport = exp.replace(match[0], '').trim()
-        } else {
-          namedExports.push(exp)
-        }
-      }
-      if (!defaultExport) return undefined
-      return namedExports.length === 0
-        ? content.replace(fullExport, `export = ${defaultExport}`)
-        : content.replace(
-            fullExport,
-            `export = ${defaultExport};\nexport { ${namedExports.join(', ')} }`,
-          )
-    })
-    .then((content) => {
-      return content ? writeFile(dtsPath, content, 'utf-8') : Promise.resolve()
-    })
+async function fixDefaultCjsExport(dtsPath) {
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/fix-default-cjs-exports.mjs
```diff
@@ -0,0 +1,65 @@
...
+// regex to find the default export in previous export {} expression
+const defaultExportRegexp = /\s*as\s+default\s*/
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+
+/**
+ * This function will replace `export { type X, Y, Z as default }` with:
+ * ```js
+ * export = X;
+ * export { Y, Z }; // <== this line will be only added if there are additional named exports
```
userquin, line 16: update this

userquin:
We should add `"type": "commonjs"`  to the package.json.

packages/tailwindcss/package.json
```diff
@@ -66,38 +66,85 @@
...
+        "import": {
+          "types": "./dist/colors.d.mts",
+          "default": "./dist/colors.mjs"
+        },
+        "require": {
+          "types": "./dist/colors.d.ts",
+          "default": "./dist/colors.js"
+        }
       },
       "./lib/util/flattenColorPalette": {
```
userquin, line 139: looks like we shoudl also add the `.js` variant

userquin:
We should add `"type": "commonjs"`  to the package.json.

userquin:
Looks like `@tailwindcss/node` package also broken, this is the `index.d.ts`  content (`index.d.mts` with same imports):

image

philipp-spiess:
@userquin Hey! I wonder if you're still working at this since the PR is open.

@thecrypticace and me have been thinking about this problem as well and our most recent thinking is that we may need to hand-write all exported types. :/

philipp-spiess:
Thanks @userquin! This is good insight. If you're not actively working on this I suggest we close this PR for now but we still have the bug report open (#15844) and plan to get back to that very soon. Really appreciate your hard work here, this is going to be super helpful!

PR #15921, opened 2025-01-27
Ensure font-size utilities with `none` modifier works e.g.: `text-sm/none`

### LINKED ISSUE #15911: Font size / line height shorthand doesn't apply `line-height` when leading value is `none`

v4.0.0

@tailwindcss/vite 4.0.0, Vite 6.0.11

Node.js 20.x

Appears in all browsers tested.

Windows 11

**Reproduction URL**

https://play.tailwindcss.com/ZmoWtl9NX5

**Describe your issue**

The `font-size` / `line-height` shorthand syntax (e.g., `text-3xl/none`) doesn't apply the `line-height` for `none` leading. The generated ruleset for, e.g. `text-3xl/none` doesn't contain the expected `line-height: 1`.

The example shows pairs of identical text at different sizes. Each pair shows:
- Left (red border): Using shorthand syntax (e.g., `text-3xl/none`)
- Right (blue border): Using separate classes (e.g., `text-3xl leading-none`)

Expected behavior:
- Both syntaxes should produce identical line heights
- The shorthand `text-3xl/none` should be equivalent to `text-3xl leading-none`
- The columns should look the same

Actual behavior:
- Text in the red column has its default line height.

Toggling the example to Tailwind v3 shows that this behavior is new with v4.

RobinMalfait:
This PR fixes an issue where classes such as `text-sm/none` don't work as expected. The reason for this is that `leading-none` is the only hardcoded leading utility and is not coming from the `@theme`. This means that `text-sm/none` tries to do a lookup for `none` but it won't resolve. 

This PR fixes that by allowing `none` as a modifier.

While working on this, I noticed that `text-sm/none` _did_ generate CSS:

```css
.text-sm\/none {
  font-size: var(--text-sm);
}
```

Notice that the `line-height` is missing. This means that any modifier that can't be resolved doesn't get the `line-height` set, but it _will_ generate CSS. In this case, no CSS should have been generated.

Otherwise, all of these generate CSS which will only bloat your CSS and won't
work as expected. E.g.: `text-sm/foo`, `text-sm/bar`, and `text-sm/baz`:

```css
.text-sm\/bar {
  font-size: var(--text-sm);
}
.text-sm\/baz {
  font-size: var(--text-sm);
}
.text-sm\/foo {
  font-size: var(--text-sm);
}
```

Fixes: #15911

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -332,7 +332,7 @@ For a deep-dive into everything that's new, [check out the announcement post](ht
 - Rename `drop-shadow` to `drop-shadow-sm` and `drop-shadow-sm` to `drop-shadow-xs` ([#14849](https://github.com/tailwindlabs/tailwindcss/pull/14849))
 - Rename `rounded` to `rounded-sm` and `rounded-sm` to `rounded-xs` ([#14849](https://github.com/tailwindlabs/tailwindcss/pull/14849))
 - Rename `blur` to `blur-sm` and `blur-sm` to `blur-xs` ([#14849](https://github.com/tailwindlabs/tailwindcss/pull/14849))
-- Remove fixed line-height theme values and derive `leading-*` utilites from `--spacing-*` scale ([#14857](https://github.com/tailwindlabs/tailwindcss/pull/14857))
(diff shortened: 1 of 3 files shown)


philipp-spiess (approved):
Nice!

PR #15941, opened 2025-01-27
Don’t look at ignore files outside initialized repos

### LINKED ISSUE #15876: [v4] Sources are not being taken into consideration with Next.js 15
Tailwind: `4.0.0`
Next.js: `15.1.6`
MacOS: `Sequoia 15.2`
Node: `v22.13.1`
pnpm: `9.15.3`
Arc Browser: `1.79.0`

**Reproduction URL**

https://github.com/AssisrMatheus/tailwindv4-next15-no-output-repro

Setup process:
1. `pnpx create-next-app@latest`
1. `pnpx @tailwindcss/upgrade@next`
1. `pnpm run dev`

**Describe your issue**

In a brand new next.js 15 project. Tailwind v4 is only building base css and theme variables, all classes used within source files are ignored.

I have tried to turn off turbopack with no success. Rolling back node to v20 also doesn't help. Tried using `postcss.config.json` file format instead of current one also with no success.

Current behaviour:  
Image

Expected behaviour:  
Image

thecrypticace:
Right now, when Oxide is scanning for files, it considers ignore files in the "root" directory it is scanning as well as all parent directories.

We honor .gitignore files even when not in a git repo as an optimization in case a project has been created, contains a .gitignore, but no repo has actually been initialized. However, this has an unintended side effect of including ignore files _ouside of a repo_ when there is one. This means that if you have a .gitignore file in your home folder it'll get applied even when you're inside a git repo which is not what you'd expect.

This PR addresses this by checking to see the folder being scanned is inside a repo and turns on a flag that ensures .gitignore files from the repo are the only ones used (global ignore files configured in git still work tho).

This still needs lots of tests to make sure things work as expected.

Fixes #15876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Suggest container query variants ([#15857](https://github.com/tailwindlabs/tailwindcss/pull/15857))
 - Disable bare value suggestions when not using the `--spacing` variable ([#15857](https://github.com/tailwindlabs/tailwindcss/pull/15857))
 - Ensure suggested classes are properly sorted ([#15857](https://github.com/tailwindlabs/tailwindcss/pull/15857))
+- Don’t look at ignore files outside initialized repos ([#15941](https://github.com/tailwindlabs/tailwindcss/pull/15941))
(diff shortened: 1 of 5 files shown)


RobinMalfait (commented):
Very nice! I wonder if we can add an integration test that mimics the behavior people were running into and that should now be fixed.

We should also run `cargo fmt` to format everything nicely.

RobinMalfait (approved):
Looks good, just need a changelog entry

RobinMalfait (approved):

PR #15751, opened 2025-01-23
chore: move tailwindcss from dependencies to peer dependencies

rtritto:

PR #15973, opened 2025-01-28
chore: move tailwindcss from dependencies to peer dependencies

rtritto:
Follow up of #15751

### DIFF
--- a/packages/@tailwindcss-cli/package.json
+++ b/packages/@tailwindcss-cli/package.json
@@ -35,7 +35,12 @@
     "enhanced-resolve": "^5.18.0",
     "lightningcss": "catalog:",
     "mri": "^1.2.0",
-    "picocolors": "^1.1.1",
+    "picocolors": "^1.1.1"
+  },
+  "devDependencies": {
+    "tailwindcss": "workspace:*"
+  },
+  "peerDependencies": {
     "tailwindcss": "workspace:*"
   }
 }
--- a/packages/@tailwindcss-node/package.json
+++ b/packages/@tailwindcss-node/package.json
@@ -38,7 +38,12 @@
   },
   "dependencies": {
     "enhanced-resolve": "^5.18.0",
-    "jiti": "^2.4.2",
+    "jiti": "^2.4.2"
+  },
+  "devDependencies": {
+    "tailwindcss": "workspace:*"
+  },
+  "peerDependencies": {
     "tailwindcss": "workspace:*"
   }
 }
--- a/packages/@tailwindcss-postcss/package.json
+++ b/packages/@tailwindcss-postcss/package.json
@@ -34,14 +34,17 @@
     "@tailwindcss/node": "workspace:^",
     "@tailwindcss/oxide": "workspace:^",
     "lightningcss": "catalog:",
-    "postcss": "^8.4.41",
-    "tailwindcss": "workspace:*"
+    "postcss": "^8.4.41"
   },
   "devDependencies": {
     "@types/node": "catalog:",
     "@types/postcss-import": "14.0.3",
     "dedent": "1.5.3",
     "internal-example-plugin": "workspace:*",
-    "postcss-import": "^16.1.0"
+    "postcss-import": "^16.1.0",
+    "tailwindcss": "workspace:*"
+  },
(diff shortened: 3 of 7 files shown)


rtritto:
FYI @philipp-spiess @RobinMalfait

RobinMalfait:
Hey!

The problem with `peerDependencies` is that you still need `@import "tailwindcss";` which means that you are now relying on a dependency that is not listed in your `package.json` (where you rely on transitive dependencies).

Apart from the "you have to keep dependencies in sync", are there any actual issues you are running into right now that this would solve?

Going to close this for now because I don't think we should change things unless there is an actual issue. I do appreciate your PR though!

rtritto:
@RobinMalfait Using `tailwindcss` as _dependency_ , for any reason (wanted or lack of attention) and with _dedupe-dependencies-on-install_ option disabled in a pacakge package manager, having the `tailwindcss@4.0.0` and `@tailwindcss/vite@4.0.0` already installed, you can update `tailwindcss` from `v4.0.0` to `v4.0.1`.
You will have 2 installed `tailwindcss` versions:
- `tailwindcss@4.0.1`
- `tailwindcss@4.0.0` used by `@tailwindcss/vite@4.0.0`

Using `tailwindcss` as _dependency_ instead of _peer dependency_ is an unwanted and dangerous behavior because (duped installed dependency) and you will not have any alert/warning.

With the _peer dependencies_ you can control the *preferred* version and install it only *once*.

Another example is with `@tailwindcss/standalone` where actually, with `tailwindcss` as _dependency_, you can't control the version of `tailwindcss` dependency that has the _caret_ (`^`) because it's always installed the latest `tailwindcss` version.

Note: I updated changes in the [commit](https://github.com/rtritto/tailwindcss/commit/f01b0724042a16b2d6c1ff8086d66eff10117604) (same branch [rtritto/tailwindcss#tailwindcss-peer-dep](https://github.com/rtritto/tailwindcss/tree/tailwindcss-peer-dep)).

rtritto:
@philipp-spiess @RobinMalfait any update to fix this issue?

philipp-spiess:
@rtritto Hey! I think we mentioned this in the first PR before but _we don't ever want to be in a position where `@tailwincss/vite` uses a different minor version of `taiwlindcss` than its own version_. We know this can lead to inlining the same dependency with two different versions but this is preferable to us than having the extensions crash with no proper error code or anything. 

We do recommend that you lock your dependencies and check in the lock file though. Then, if you upgrade one dependency, you'd see that it will add a new dependency to your project and can act accordingly.

Again, I don't think anything is broken by using a hard dependency other than yes you can end up with two different versions of the same package but that is still a better trade-off fur us than having to manage internal API changes. 

We might reconsider this in the future but for now we feel pretty good about the hard dependency so not going to make a change. Thanks!

rtritto:
> this is preferable to us than having the extensions crash with no proper error code or anything.

@philipp-spiess The package manager auto install the peer dependencies or display a warning. It's a controlled and safe standard de facto to handle this specific use case.
Some advantages of using peer dependencies:
- avoid duplicate of installed dependencies (my previous example of `tailwindcss@4.0.1` and `@tailwindcss/vite@4.0.0`) that
  - can lead the singleton issues (multiple instances that can break the runtime execution)
  - reduces the size of the installed dependencies
- control and flexibility using a specific version

> We do recommend that you lock your dependencies and check in the lock file though. Then, if you upgrade one dependency, you'd see that it will add a new dependency to your project and can act accordingly.

It's not just my problem. Using `tailwindcss` as dependency, that version used by the tool/plugin is hidden.
There are a lot of use case where using it as dependency is silent (no warning), unsafe (multiple instances that breaks) and with no control (can't select the version).
Another example is with `@tailwindcss/standalone` that has the dependency `tailwind@^<VERSION>` (note the caret) where the version range of the dependency `tailwindcss` was choosed. Imply that, changing `tailwindcss` from dependency, there is no version logic that changes, it's a simple change with no drawbacks.

Please take care about this change because it impacts the core of Tailwind ecosystem. Maybe we can hear the community's opinions in a new discussion.

FYI @RobinMalfait

rtritto:
I created #16135

PR #15974, opened 2025-01-28
Fix Oxide scanner bugs

### LINKED ISSUE #15632: [v4] Class Detection Bug

v4.0.0-beta.9

Vite v6.0.7

v23.2.0

Arc (Chromium)

macOS

**Reproduction URL**

- Clone this repo: https://github.com/okkdev/tailwindcss4-class-detection-bug-repro/
- run `npm dev`
- notice that `bg-red-200` is missing

**Describe your issue**

Tailwind doesn't detect the class inside this list `[class("bg-red-200")]` at https://github.com/okkdev/tailwindcss4-class-detection-bug-repro/blob/main/src/main.gleam#L33

It only works if there is a whitespace somewhere inside of the square braces. 
As shown here: https://github.com/okkdev/tailwindcss4-class-detection-bug-repro/blob/main/src/main.gleam#L35
and here: https://github.com/okkdev/tailwindcss4-class-detection-bug-repro/blob/main/src/main.gleam#L38

It would be awesome if there was a way to customize the detection to support more languages like Gleam, instead of having to rely on the automatic detection.

### LINKED ISSUE #15740: [v4] Angular conditional class not recognized by tailwind

v4.0.0

Angular v19.1

v23.5.0

Chrome

Windows

**Reproduction URL**

I made a small repository showcasing the bug: https://github.com/Eirmas/tailwind-4-bug

**Describe your issue**

Angular allows for adding conditional classes with the following syntax:
```html
<div [class.pt-4]="bool"></div>
```

It seems that Tailwind doesn't see the `pt-4` class being used here, and therefore doesn't add it to the stylesheet.

philipp-spiess:
Fixes #15632
Fixes #15740

This PR fixes a number of Oxide scanner bugs reported over various channels, specifically:

- When using the Svelte class shorthand split over various lines, we weren't extracting class names properly: 
   
   ```svelte
   <div
     class:underline={isUnderline}>
   </div>
   ```
- We now extract classes when using the class shortcut in Angular:
   
   ```html
   <div [class.underline]=\"bool\"></div>
   ```
- We now validate parentheses within arbitrary candidates so that we don't consume invalid arbitrary candidates anymore which allows us to parse the following case properly:
   
   ```js
   const classes = [wrapper("bg-red-500")]
   ```

## Test plan

Added unit tests


crates/oxide/src/lib.rs
```diff
@@ -447,7 +447,9 @@ fn read_changed_content(c: ChangedContent) -> Option<Vec<u8>> {
     };
 
     match extension {
-        Some("svelte") => Some(content.replace(" class:", " ")),
+        // Angular class shorthand
+        Some("html") => Some(content.replace("[class.", "[")),
+        Some("svelte") => Some(content.replace(" class:", " ").replace("\tclass:", " ")),
```
thecrypticace, line 452:
```suggestion
        Some("svelte") => Some(content.replace(" class:", " ").replace("\tclass:", " ").replace("\nclass:", " ")),
```

RobinMalfait (approved):

paolominel:
in version 4.0.1 the bug is still present (almost for angular shorthand syntax)

philipp-spiess:
@paolominel What's the exact file you've used that doesn't work?

paolominel:
I use tailwind classes in angular host component.
i tried something like:

`...,
host: {
"[class.underline]": "active()"
}`

active variable was a signal inside the relative angular component.

this class was not fired and not compiled during tailwind class scanning.
if i use not optional syntax ("class": "underline"), everything works well.

Issue #15005, opened 2024-11-14
@property isn't supported in shadow roots


blittle:
**What version of Tailwind CSS are you using?**

For example: `4.0.0-alpha.34`

**What build tool (or framework if it abstracts the build tool) are you using?**

Web components with shadow roots.

**What version of Node.js are you using?**

For example: v12.0.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

https://github.com/blittle/tw-shadow

**Describe your issue**

Tailwind v4 uses `@property` to define defaults for custom properties. At the moment, shadow roots do _not support_ `@property`_. It used to be explicitly denied in the spec, but it looks like there's talk on adding it: https://github.com/w3c/css-houdini-drafts/pull/1085

I don't know if this is something tailwind should fix, but it took me a while to find the issue, so it's probably worth keeping this issue to document the limitation.

Here is a work-around, [just attaching the `@property` definitions to the base document](https://benfrain.com/using-css-property-inside-shadowroot-web-components-workaround/). It would be nice if tailwind provided an easy way to import just that content.

ErwinAI:
Not 100% sure if this is related, but I noticed that in a shadow root, the defaults under `@theme` aren't available. I.e. I can't see `--spacing` in devtools, and it doesn't have any value.

Essentially rendering Tailwind unusable in shadow roots for most use cases.

rose21wiley:
Tailwind CSS v4’s @property rules aren't supported in shadow roots. A temporary fix involves defining these properties in a separate CSS file and using Vite to adjust them. Your Vite configuration snippet modifies inheritance settings and ensures proper implementation.

Great solution for handling this limitation!

philipp-spiess:
@blittle Thanks for the detailed writeup and sorry for the late answer here. 

From what I understand after doing some testing, it is indeed still required that the `@property` definitions are hoisted to the base document. 

The snippet you have for that also changes the `inherits` property. Do you remember why this was necessary? I was playing around it in a simple playground and found that this wasn't necessary for my tests. I didn't transform the sources via Vite though but added this snippet directly within the runtime of a custom component's `connectedCallback` hook:

```ts
let atProperties = styles.slice(styles.indexOf('@property'))
let style = document.createElement('style')
style.innerText = atProperties
document.head.appendChild(style)
```

kilobyte2007:
[...]

PR #15975, opened 2025-01-28
Add `:host` rule to `@theme` layer

### LINKED ISSUE #14478: [v4] Expose CSS properties also to :host

For example: v4.0.0-alpha.21

**Describe your issue**

Currently the `@theme` CSS properties are exposed to `:root`, which is great but it would be also good to expose it to `:host`, like this `:root, :host {}`. This way the CSS properties are also available in web components. 

Lets say you have a widget as web component `<custom-widget></custom-widget>` and inside there is CSS and tailwindcss classes in shadowroot, `:root` will not work - in this scenario the `:host` is the `:root` for the web component.

Eg. how it's done in Winduum here - https://github.com/winduum/winduum/blob/main/src/theme/default.css

### LINKED ISSUE #15799: tailwindcss and shadow dom not working well (--spacing)

hugo-vrijswijk:
Resolves #15799
Resolves #14478
Part-of #15005

Adds a `:host` selector for the `@theme` layer. This is necessary for the `@theme` layer to work correctly in shadow DOM.

Also updates the snapshots for the tests that were affected by this change (in a separate commit).

## Test plan

Tested via the Vite playground:

Screenshot 2025-01-29 at 15 06 49

Additionally made sure that `@property` defaults also work across Firefox, Chrome, and Safari (the `@property` definition from the root is pulled in) and added a UI spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Find utilities when using the Angular class shorthand syntax ([#15974](https://github.com/tailwindlabs/tailwindcss/pull/15974))
 - Find utilities when using functions inside arrays ([#15974](https://github.com/tailwindlabs/tailwindcss/pull/15974))
 - Ensure that `@tailwindcss/browser` does not pollute the global namespace ([#15978](https://github.com/tailwindlabs/tailwindcss/pull/15978))
+- Ensure CSS variables are registered on `:host` to be available within shadow roots ([#15975](https://github.com/tailwindlabs/tailwindcss/pull/15975))
 - _Upgrade_: Ensure JavaScript config files on different drives are correctly migrated ([#15927](https://github.com/tailwindlabs/tailwindcss/pull/15927))
 
 ## [4.0.0] - 2025-01-21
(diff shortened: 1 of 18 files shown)


lubomirblazekcz:
This should also fix https://github.com/tailwindlabs/tailwindcss/issues/14478

philipp-spiess:
@hugo-vrijswijk Thanks a ton for this PR! I went ahead and did a round of testing on the known repro cases and added a test plan to your description. I also added a UI spec to ensure we have some validation. Note that because of `@property` not being able to be scoped to a shadow root, you still have to have the stylesheet added to the browser root too.

thecrypticace (approved):

khopha:
> @hugo-vrijswijk Thanks a ton for this PR! I went ahead and did a round of testing on the known repro cases and added a test plan to your description. I also added a UI spec to ensure we have some validation. Note that because of `@property` not being able to be scoped to a shadow root, you still have to have the stylesheet added to the browser root too.

@philipp-spiess Thank you for the note. Does this mean that Tailwind CSS v4 is not fully compatible with Shadow DOM? (Because we still need to add it to the browser root too.) Is there any workaround that doesn't require adding it to the browser root, or is the only workaround to downgrade to v3?

khopha:
@hugo-vrijswijk 
Thank you for your workaround. I followed your approach, and it’s working fine now. Yes, it would be great if Tailwind could work properly or have an option to make it work in Shadow DOM.

lubomirblazekcz:
Sadly it's more of a spec problem, because @property at-rule doesn't work in shadow dom.

https://stackoverflow.com/questions/78717295/property-not-recognized-in-shadow-dom

khopha:
@lubomirblazekcz 
Have you faced the same issue? How do you deal with it?

lubomirblazekcz:
@khopha for now I copied the values manually to parent document as a temporary solution.

PR #15988, opened 2025-01-29
Fix `resolve_globs` crash when at root directory

Xetera:
Fixes a crash found in Dockerfiles when a build takes place at the root directory. It's not a good practice to keep your application logic in `/`, but it probably shouldn't cause a crash either.

I found this particularly difficult to write tests for because it would involve either running a glob on my real filesystem starting from `/` or mocking the calls to `fs` which as far as I can tell isn't supported in the codebase and would be out of scope to try to do here.

Fixes #15987


RobinMalfait (approved):
Good catch! Thank you very much

PR #16057, opened 2025-01-30
Ensure we process Tailwind CSS features when using `@reference`

### LINKED ISSUE #16056: `@tailwindcss/postcss` does not run on some files that use Tailwind features

v4.0.0, v4.0.1

webpack@5.97.1, postcss-cli@11.0.0

v18.17.1, v18.18.2

N/A

Linux in WSL

**Reproduction URL**

https://github.com/wongjn/tailwind-postcss

**Describe your issue**

An input CSS file does not get processed by Tailwind when it only has the Tailwind-specific at-rules of `@reference`, `@variant`, `@custom-variant`, `@utility` directives within it. I believe this is due to the `if` condition within the PostCSS plugin:

https://github.com/tailwindlabs/tailwindcss/blob/ea249957497c095ff8d228a3c913f608afbb9b28/packages/%40tailwindcss-postcss/src/index.ts#L74-L91

Use case is I have some separate files the don't need to compile any Tailwind utilities but should have reference to Tailwind theme tokens used elsewhere in the project.

Edit: Though `@keyframes` would get injected as per https://github.com/tailwindlabs/tailwindcss/issues/15956#issuecomment-2618659665 so perhaps this type of usage is unintended?

RobinMalfait:
This PR fixes an issue where if you only used `@reference` that we didn't process Tailwind CSS features.

We have a 'quick bail check', in the PostCSS plugin to quickly bail if we _konw_ that we don't need to handle any Tailwind CSS features. This is useful in Next.js applications where every single CSS file will be passed to the PostCSS plugin.

If you use custom font ins Next.js, each of those fonts will have a CSS file as well.

Before we introduced `@reference`, we used `@import "tailwindcss" reference`, which passed the bail check because `@import` was being used. Now we have `@reference` which wasn't included in the list.

This is now solved.

Fixes: #16056

### Test plan

Added a failing test that is now failing after the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Only generate positive `grid-cols-*` and `grid-rows-*` utilities ([#16020](https://github.com/tailwindlabs/tailwindcss/pull/16020))
-- Ensure we process Tailwind CSS features when only using `@reference` ([#16057](https://github.com/tailwindlabs/tailwindcss/pull/16057))
+- Ensure we process Tailwind CSS features when only using `@reference` or `@variant` ([#16057](https://github.com/tailwindlabs/tailwindcss/pull/16057))
 
 ## [4.0.1] - 2025-01-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Only generate positive `grid-cols-*` and `grid-rows-*` utilities ([#16020](https://github.com/tailwindlabs/tailwindcss/pull/16020))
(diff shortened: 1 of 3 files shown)


philipp-spiess (changes requested):
Needs to also handle `@variant` I think? You can use it with no config for the static variants.

RobinMalfait:
> Needs to also handle `@variant` I think? You can use it with no config for the static variants.

Ah yep good catch. Added a test and a fix.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Only generate positive `grid-cols-*` and `grid-rows-*` utilities ([#16020](https://github.com/tailwindlabs/tailwindcss/pull/16020))
+- Ensure we process Tailwind CSS features when only using `@reference` ([#16057](https://github.com/tailwindlabs/tailwindcss/pull/16057))
```
philipp-spiess, line 13:
```suggestion
- Ensure we process Tailwind CSS features when only using `@reference` or `@variant` ([#16057](https://github.com/tailwindlabs/tailwindcss/pull/16057))
```

philipp-spiess (approved):

PR #16069, opened 2025-01-30
Vite: Transform `<style>` blocks in html files

### LINKED ISSUE #16036: Cannot use HTML style tag with Vite to apply a theme

v4.0.1

Latest Vite with autoprefixer

NodeJS v22.10.0

Brave

macOS

**Reproduction URL**

```html
<!DOCTYPE html>
<html>
	<head>
		<style>
			@import "tailwindcss";
			@layer components {
				h1{
					@apply text-2xl;
				}
			}
		</style>
	</head>
	<body>
    <h1>Test</h1>
  </body>
</html>
```

Vite config
```js
module.exports = {
	root: "./example",
	base: "./",
	appType: "custom",
	build: {
		rollupOptions: {
			input: "./example/test.html",
		},
		outDir: "./example-min/",
		emptyOutDir: true,
		copyPublicDir: false,
	},
	plugins: [
		tailwindcss(),
	],
	css: {
		postcss: {
			plugins: [autoprefixer()],
		},
	},
}
```

**Describe your issue**

We have single HTML files which are report templates in a facility for analysis reportings that are styled with tailwindcss. These are later processed by Vite to reduce their size and to process tailwindcss classes. We aren't able to support external CSS or JS here since these templates are generated by another tool which has its own syntax and templates. When processing these HTML with Vite tailwindcss seems to ignore inline CSS even with `text/tailwindcss`. Using the CDN works fine, but our reports need to work offline (companies that size do not allow these machines to have internet access on pruporse ofc). The reason we use the style tag is because that way we can provided a versioned theme to each report template and do not need to touch our template files which consist of many smaller modules that we do not want to touch that often and because each customer can have its own styled theme or custom styles for individuallity.

philipp-spiess:
Fixes #16036

This adds a new rule to treat `<style>` blocks found within `.html` file as Tailwind CSS targets.

## Test plan

- Tested using the Vite extension (dev) and a new integration test (prod)

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -8,6 +8,7 @@ import type { Plugin, ResolvedConfig, Rollup, Update, ViteDevServer } from 'vite
 
 const DEBUG = env.DEBUG
 const SPECIAL_QUERY_RE = /[?&](raw|url)\b/
+const INLINE_STYLE_ID_RE = /[?&]index\=\d+\.css$/
 
 const IGNORED_DEPENDENCIES = ['tailwind-merge']
 
@@ -320,7 +321,7 @@ function isPotentialCssRootFile(id: string) {
   if (id.includes('/.vite/')) return
   let extension = getExtension(id)
   let isCssFile =
-    (extension === 'css' || id.includes('&lang.css')) &&
+    (extension === 'css' || id.includes('&lang.css') || id.match(INLINE_STYLE_ID_RE)) &&
     // Don't intercept special static asset resources
     !SPECIAL_QUERY_RE.test(id)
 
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -8,6 +8,7 @@ import type { Plugin, ResolvedConfig, Rollup, Update, ViteDevServer } from 'vite
 
 const DEBUG = env.DEBUG
 const SPECIAL_QUERY_RE = /[?&](raw|url)\b/
+const INLINE_STYLE_ID_RE = /[?&]index\=\d+\.css$/
```
RobinMalfait, line 11: Interesting, is this always the case? I'm assuming we see the link as `/index.html?index=0.css` where `0` is the index of the `<style>` block on the page?

RobinMalfait (approved):
Love it, nice!

packages/@tailwindcss-vite/src/index.ts
philipp-spiess: Yep, that's what I saw when i tested it 👍

PR #16078, opened 2025-01-30
Vite: Don't rebase urls that appear to be aliases

### LINKED ISSUE #16039: `@tailwindcss/vite` breaks `url()` using `resolve.alias`

4.0.1

Vite 6.0.11 with @tailwindcss/vite 4.0.1

For example: postcss-cli 8.3.1, Next.js 10.0.9, webpack 5.28.0

22.13.0

Chrome

Windows

**Reproduction URL**

https://stackblitz.com/edit/vitejs-vite-twabw18q?file=vite.config.ts,src%2Fstyles%2Fstyle.css&terminal=dev

1. Download the project from this button
    Image 
2. Run `npm i`
3. Run `npm run dev`
4. Open `http://localhost:5173`
5. The background image is not shown
    Image

**Describe your issue**

In Vite, I can use alias set with `resolve.alias` in `url()`. For example:
```css
.logo.vanilla {
  background-image: url('~/images/javascript.svg');
}
```
This does not work when `@tailwindcss/vite` is used. (It works when `@tailwind/postcss` is used with Vite)

If you comment out `plugins: [tailwind()],` from `vite.config.ts` in the reproduction, you can see the background image shown, which is the expected behavior.
Image

philipp-spiess:
Closes #16039

This PR changes our URL rebasing logic used with Vite so that it does not rebase URLs that look like common alias paths (e.g. urls starting in `~`, `@` or `#`, etc.). Unfortunately this is only an approximation and you can configure an alias for a path that starts with a regular alphabetical character (e.g. `foo` => `./my/foo`) so this isn't a perfect fix, however in practice most aliases will be prefixed with a symbol to make it clear that it's an alias anyways.

One alternative we have considered is to only rebase URLs that we know are relative (so they need to start with a `.`). This, however, will break common CSS use cases where urls are loaded like this:

```css
background: image-set(
  url('image1.jpg') 1x,
  url('image2.jpg') 2x
);
``` 

So making this change felt like we only trade one GitHub issue for another one. 

In a more ideal scenario we try to resolve the URL with the Vite resolver (we have to run the resolver and can't rely on the `resolve` setting alone due to packages like [`vite-tsconfig-paths`](https://www.npmjs.com/package/vite-tsconfig-paths)), however even then we can have relative paths being resolvable to different files based on wether they were rebased or not (e.g. when an image with the same filename exists in two different paths).

So ultimately we settled on extending the already existing blocklist (which we have taken from the Vite implementation) for now.

## Test plan

- Added unit test and it was tested with the Vite playground.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Only generate positive `grid-cols-*` and `grid-rows-*` utilities ([#16020](https://github.com/tailwindlabs/tailwindcss/pull/16020))
 - Ensure we process Tailwind CSS features when only using `@reference` or `@variant` ([#16057](https://github.com/tailwindlabs/tailwindcss/pull/16057))
+- Vite: Don't rebase urls that appear to be aliases ([#16078](https://github.com/tailwindlabs/tailwindcss/pull/16078))
 
 ## [4.0.1] - 2025-01-29
 
--- a/packages/@tailwindcss-node/src/urls.test.ts
+++ b/packages/@tailwindcss-node/src/urls.test.ts
@@ -24,6 +24,20 @@ test('URLs can be rewritten', async () => {
         background: url('/image.jpg');
         background: url("/image.jpg");
 
+        /* Potentially Vite-aliased URLs: ignored */
+        background: url(~/image.jpg);
+        background: url(~/foo/image.jpg);
+        background: url('~/image.jpg');
+        background: url("~/image.jpg");
+        background: url(#/image.jpg);
+        background: url(#/foo/image.jpg);
+        background: url('#/image.jpg');
+        background: url("#/image.jpg");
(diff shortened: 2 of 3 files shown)


ArnaudBarre:
Could the tailwind plugin skip the rebasing all together? It seems this is something that should be done by Vite core and not Tailwind. What's the need that made you add this? (I think we already discuss this but I don't remember the reason)

philipp-spiess:
@ArnaudBarre 

> Could the tailwind plugin skip the rebasing all together? It seems this is something that should be done by Vite core and not Tailwind. What's the need that made you add this? (I think we already discuss this but I don't remember the reason)

The reason we need to rebase is that we do `@import` resolving in pre-phase so the information about where the CSS file was located is gone by the time Vite gets it :/ And since the transformers are `string => string`, I don't think there's any type of meta information we could pass with the transformation step so that Vite knows that some chunk of the content comes from different files.

RobinMalfait (approved):

PR #16093, opened 2025-01-31
Discard invalid declarations when parsing CSS

thecrypticace:
I discovered this when triaging an error someone had on Tailwind Play.

1. When we see a `;` we often assume a valid declaration precedes it but that may not be the case
2. When we see the name of a custom property we assume everything that follows will be a valid declaration but that is not necessarily the case
3. A bare identifier inside of a rule is treated as a declaration which is not the case

This PR fixes all three of these by ignoring these invalid cases. Though some should probably be turned into errors.

### DIFF
--- a/packages/tailwindcss/src/css-parser.test.ts
+++ b/packages/tailwindcss/src/css-parser.test.ts
@@ -1041,6 +1041,34 @@ describe.each(['Unix', 'Windows'])('Line endings: %s', (lineEndings) => {
         },
       ])
     })
+
+    // TODO: These should probably be errors in some cases?
+    it('discards invalid declarations', () => {
+      // This shouldn't parse as a custom property declaration
+      expect(parse(`--foo`)).toEqual([])
+
+      // This shouldn't parse as a rule followed by a declaration
+      expect(parse(`@plugin "foo" {};`)).toEqual([
+        {
+          kind: 'at-rule',
+          name: '@plugin',
+          params: '"foo"',
+          nodes: [],
+        },
+      ])
+
+      // This shouldn't parse as consecutive declarations
+      expect(parse(`;;;`)).toEqual([])
+
+      // This shouldn't parse as a rule with a declaration inside of it
+      expect(parse(`.foo { bar }`)).toEqual([
+        {
+          kind: 'rule',
+          selector: '.foo',
+          nodes: [],
+        },
+      ])
+    })
   })
 
   describe('errors', () => {
--- a/packages/tailwindcss/src/css-parser.test.ts
+++ b/packages/tailwindcss/src/css-parser.test.ts
@@ -329,6 +329,28 @@ describe.each(['Unix', 'Windows'])('Line endings: %s', (lineEndings) => {
         ])
       })
 
+      it('should parse a custom property with an empty value', () => {
+        expect(parse('--foo:;')).toEqual([
+          {
+            kind: 'declaration',
+            property: '--foo',
+            value: '',
+            important: false,
+          },
+        ])
+      })
+
+      it('should parse a custom property with a space value', () => {
+        expect(parse('--foo: ;')).toEqual([
+          {
+            kind: 'declaration',
+            property: '--foo',
+            value: '',
+            important: false,
+          },
+        ])
+      })
+
       it('should parse a custom property with a block including nested "css"', () => {
         expect(
           parse(css`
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/css-parser.test.ts
```diff
@@ -329,6 +329,28 @@ describe.each(['Unix', 'Windows'])('Line endings: %s', (lineEndings) => {
...
+        expect(parse('--foo: ;')).toEqual([
+          {
+            kind: 'declaration',
+            property: '--foo',
+            value: '',
+            important: false,
+          },
+        ])
+      })
+
```
RobinMalfait, line 353: Added these just in case we broke something with unexpected semicolon detection

packages/tailwindcss/src/css-parser.test.ts
```diff
@@ -1097,5 +1119,39 @@ describe.each(['Unix', 'Windows'])('Line endings: %s', (lineEndings) => {
...
+    })
+
+    it('should error when a semicolon exists after an at-rule with a body', () => {
+      expect(() => parse('@plugin "foo" {} ;')).toThrowErrorMatchingInlineSnapshot(
+        `[Error: Unexpected semicolon]`,
+      )
+    })
+
+    it('should error when consecutive semicolons exist', () => {
+      expect(() => parse(';;;')).toThrowErrorMatchingInlineSnapshot(`[Error: Unexpected: \`;;;\`]`)
```
RobinMalfait, line 1148: We could also just throw the same `Unexpected semicolon` error, I'm indifferent on this one.

I think once we have proper span/location, we could highlight the full `;;;` section.

thecrypticace: I was thinking about this but I think the error is fine as is

packages/tailwindcss/src/css-parser.ts
```diff
@@ -286,10 +286,14 @@ export function parse(input: string) {
...
-        parent.nodes.push(declaration)
+      if (declaration) {
+        if (parent) {
+          parent.nodes.push(declaration)
+        } else {
+          ast.push(declaration)
+        }
       } else {
-        ast.push(declaration)
+        throw new Error(`Invalid custom property, expected a value`)
```
thecrypticace, line 296: Now that we're throwing errors maybe we can eliminate some of the nesting here?

```css
let declaration = parseDeclaration(buffer, colonIdx)
if (!declaration) {
  throw new Error(`Invalid custom property, expected a value`)
}

if (parent) {
  parent.nodes.push(declaration)
} else {
  ast.push(declaration)
}
```

packages/tailwindcss/src/css-parser.ts
```diff
@@ -435,7 +455,12 @@ export function parse(input: string) {
 
           // Attach the declaration to the parent.
           if (parent) {
-            parent.nodes.push(parseDeclaration(buffer, colonIdx))
+            let node = parseDeclaration(buffer, colonIdx)
+            if (node) {
+              parent.nodes.push(node)
+            } else {
+              throw new Error(`Invalid declaration: \`${buffer.trim()}\``)
+            }
```
thecrypticace, line 463: ditto about nesting here too

packages/tailwindcss/src/css-parser.ts
philipp-spiess: Yeah I think that would be a good change, the indentation is already quite rough here 👍

packages/tailwindcss/src/css-parser.ts
```diff
@@ -337,10 +341,26 @@ export function parse(input: string) {
...
+      } else if (input.charCodeAt(i - 1) === SEMICOLON || input.charCodeAt(i + 1) === SEMICOLON) {
+        let startSemiColonIdx = i
+        while (input.charCodeAt(startSemiColonIdx - 1) === SEMICOLON) {
+          startSemiColonIdx--
+        }
+        let endSemiColonIdx = i
+        while (input.charCodeAt(endSemiColonIdx) === SEMICOLON) {
+          endSemiColonIdx++
+        }
+        throw new Error(`Unexpected: \`${input.slice(startSemiColonIdx, endSemiColonIdx)}\``)
```
philipp-spiess, line 359: I'm not sure if this really brings more value then just saying `unexpected semicolon` haha

packages/tailwindcss/src/css-parser.ts
RobinMalfait: Nice, ok will update 👍

thecrypticace:
Can't approve the PR since I opened it originally lol but LGTM 👍

RobinMalfait (approved):

PR #16113, opened 2025-01-31
Feat: Adding a warning in the upgrade CLI when a config file contains non-migratable values.

baptisthecht:
I encountered an issue in several of my projects during migration. I was unable to migrate correctly because my config file contained the now-deprecated "safelist" key. I had to reverse-engineer the problem to identify what was blocking me. To help others facing the same issue, I added this warning to inform them that they need to edit their config file for a smooth automatic migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactor gradient implementation to work around [prettier/prettier#17058](https://github.com/prettier/prettier/issues/17058) ([#16072](https://github.com/tailwindlabs/tailwindcss/pull/16072))
 - Vite: Ensure hot-reloading works with SolidStart setups ([#16052](https://github.com/tailwindlabs/tailwindcss/pull/16052))
 - Vite: Fix a crash when starting the development server in SolidStart setups ([#16052](https://github.com/tailwindlabs/tailwindcss/pull/16052))
-- Adding a warning when a config file contains non-migratable values ([#16073](https://github.com/tailwindlabs/tailwindcss/pull/16073))
+- Adding a warning when a config file contains non-migratable values in upgrade cli ([#16073](https://github.com/tailwindlabs/tailwindcss/pull/16073))
 
 ## [4.0.1] - 2025-01-29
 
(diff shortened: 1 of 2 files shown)


philipp-spiess:
@baptisthecht Thanks! It would be good if we can add some warnings to the other cases where we opt-out of migrating to a CSS file as well and add a quick test to ensure this works. Would you be interested to take a stab at this?

baptisthecht:
> @baptisthecht Thanks! It would be good if we can add some warnings to the other cases where we opt-out of migrating to a CSS file as well and add a quick test to ensure this works. Would you be interested to take a stab at this?

Sure !

baptisthecht:
Should we keep the non-migratable Tailwind properties by merging the old tailwind.config.xxs file and warning that they are deprecated, or should we ignore the deprecated properties and proceed with the migration anyway?

philipp-spiess:
I thin it's fine to skip `corePlugins` like you said but I'm unsure of skipping _all unknown properties_ without an allow list. Could lead to some confusion since plugins in v3 you can read the whole config object so there might be some plugin specific settings: https://play.tailwindcss.com/1ZICA06UpK?file=config

baptisthecht:
I see.

So, should I just add a warning for unsupported properties and other excludable cases?
That could already be quite useful.

By the way, corePlugins is already skipped.

philipp-spiess:
@baptisthecht Yeah I think that makes sense, thanks!

Bhavdeep-Sai:
**Issue resolved by using the latest version and adjusting build configuration**

I encountered the same issue and was able to resolve it by updating Tailwind to the latest version. Additionally, I had to adjust my build configuration to ensure all classes are properly included in the production build.

Here’s what I did:

1. Updated Tailwind to the latest version:
   ```bash
   npm install tailwindcss@latest
   ```

2. Ensured the correct `content` paths in `tailwind.config.js`:
   ```js
   module.exports = {
     content: [
       "./index.html",
       "./src/**/*.{js,ts,jsx,tsx}", // adjust to your project structure
     ],
     theme: {
       extend: {},
     },
     plugins: [],
   }
   ```

3. Rebuilt the project using:
   ```bash
   npm run build
   ```

After these steps, everything worked fine in production, and the issue was resolved.

Hope this helps!

philipp-spiess:
Closing this one as the issue is stale but still happy to review a contribution here but ideally it should address all the code paths.

PR #16121, opened 2025-01-31
Do not emit empty rules/at-rules

RobinMalfait:
This PR is an optimization where it will not emit empty rules and at-rules. I noticed this while working on https://github.com/tailwindlabs/tailwindcss/pull/16120 where we emitted:
```css
:root, :host {
}
```

There are some exceptions for "empty" at-rules, such as:

```css
@charset "UTF-8";
@layer foo, bar, baz;
@custom-media --modern (color), (hover);
@namespace "http://www.w3.org/1999/xhtml";
```

These don't have a body, but they still have a purpose and therefore they will be emitted.

However, if you look at this:

```css
/* Empty rule */
.foo {
}

/* Empty rule, with nesting */
.foo {
  .bar {
  }
  .baz {
  }
}

/* Empty rule, with special case '&' rules */
.foo {
  & {
    &:hover {
    }
    &:focus {
    }
  }
}

/* Empty at-rule */
@media (min-width: 768px) {
}

/* Empty at-rule with nesting*/
@media (min-width: 768px) {
  .foo {
  }

  @media (min-width: 1024px) {
    .bar {
    }
  }
}
```

None of these will be emitted.

### DIFF
--- a/packages/tailwindcss/src/ast.test.ts
+++ b/packages/tailwindcss/src/ast.test.ts
@@ -138,7 +138,7 @@ it('should not emit empty rules once optimized', () => {
     }
 
     /* Exceptions: */
-    @charset 'UTF-8';
+    @charset "UTF-8";
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
     @namespace 'http://www.w3.org/1999/xhtml';
@@ -170,7 +170,7 @@ it('should not emit empty rules once optimized', () => {
         }
       }
     }
-    @charset 'UTF-8';
+    @charset "UTF-8";
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
     @namespace 'http://www.w3.org/1999/xhtml';
@@ -178,7 +178,7 @@ it('should not emit empty rules once optimized', () => {
   `)
 
   expect(toCss(optimizeAst(ast))).toMatchInlineSnapshot(`
-    "@charset 'UTF-8';
+    "@charset "UTF-8";
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
     @namespace 'http://www.w3.org/1999/xhtml';
--- a/packages/tailwindcss/src/ast.test.ts
+++ b/packages/tailwindcss/src/ast.test.ts
@@ -141,6 +141,7 @@ it('should not emit empty rules once optimized', () => {
     @charset 'UTF-8';
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
+    @namespace 'http://www.w3.org/1999/xhtml';
   `)
 
   expect(toCss(ast)).toMatchInlineSnapshot(`
@@ -172,6 +173,7 @@ it('should not emit empty rules once optimized', () => {
     @charset 'UTF-8';
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
+    @namespace 'http://www.w3.org/1999/xhtml';
     "
   `)
 
@@ -179,6 +181,7 @@ it('should not emit empty rules once optimized', () => {
     "@charset 'UTF-8';
     @layer foo, bar, baz;
     @custom-media --modern (color), (hover);
+    @namespace 'http://www.w3.org/1999/xhtml';
     "
   `)
 })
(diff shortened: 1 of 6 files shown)


packages/tailwindcss/src/ast.ts
```diff
@@ -297,7 +301,14 @@ export function optimizeAst(ast: AstNode[]) {
       for (let child of node.nodes) {
         transform(child, copy.nodes, depth + 1)
       }
-      parent.push(copy)
+      if (
+        copy.nodes.length > 0 ||
+        copy.name === '@layer' ||
+        copy.name === '@charset' ||
+        copy.name === '@custom-media'
```
thecrypticace, line 308:
```suggestion
        copy.name === '@custom-media' ||
        copy.name === '@namespace'
```

https://developer.mozilla.org/en-US/docs/Web/CSS/@namespace

RobinMalfait: Good catch 👍

packages/tailwindcss/src/ast.test.ts
```diff
@@ -95,3 +97,88 @@ it('should stop walking when returning `WalkAction.Stop`', () => {
...
+    @charset 'UTF-8';
+    @layer foo, bar, baz;
+    @custom-media --modern (color), (hover);
+    "
+  `)
+
+  expect(toCss(optimizeAst(ast))).toMatchInlineSnapshot(`
+    "@charset 'UTF-8';
+    @layer foo, bar, baz;
+    @custom-media --modern (color), (hover);
```
thecrypticace, line 181:
```suggestion
    @custom-media --modern (color), (hover);
    @namespace "http://www.w3.org/1999/xhtml";
```

packages/tailwindcss/src/ast.test.ts
```diff
@@ -95,3 +97,88 @@ it('should stop walking when returning `WalkAction.Stop`', () => {
...
+      .foo {
+      }
+      @media (min-width: 1024px) {
+        .bar {
+        }
+      }
+    }
+    @charset 'UTF-8';
+    @layer foo, bar, baz;
+    @custom-media --modern (color), (hover);
```
thecrypticace, line 174:
```suggestion
    @custom-media --modern (color), (hover);
    @namespace "http://www.w3.org/1999/xhtml";
```

packages/tailwindcss/src/ast.test.ts
thecrypticace: (I think I got the syntax right)

RobinMalfait: (you did not, but it's fine 😂)

RobinMalfait: Or you know what, it looks like `"` is required for `@charset` and can't have `'`. MDN doesn't say anything about `"` vs `'` for `@namespace`

RobinMalfait: Note: we just parse the value as-is, so if you use `@charset 'foo'` we keep it as-is.

thecrypticace (approved):

packages/tailwindcss/src/ast.ts
```diff
@@ -297,7 +301,15 @@ export function optimizeAst(ast: AstNode[]) {
       for (let child of node.nodes) {
         transform(child, copy.nodes, depth + 1)
       }
-      parent.push(copy)
+      if (
+        copy.nodes.length > 0 ||
+        copy.name === '@layer' ||
+        copy.name === '@charset' ||
+        copy.name === '@custom-media' ||
+        copy.name === '@namespace'
```
philipp-spiess, line 309: Should this just be all at-rules? 🤔

RobinMalfait:
```cs
@media (foo) {}
```
Should be removed imo

philipp-spiess (approved):

packages/tailwindcss/src/ast.ts
philipp-spiess: yeah i noticed this too, you just responded before i could update my comment heh

RobinMalfait: Hehe, I think in theory we could invert the condition and don't touch everything but the `@media` (but then we need to check for `@suspports` as well. 

In a perfect world `@layer foo {}` is also removed, (because you should use `@layer foo;` but we don't make that distinction.

PR #16144, opened 2025-02-01
Fix `@import url()` being stripped

### LINKED ISSUE #16136: @import statement is not working in v4.0.2

v4.0.2

N/A

N/A

N/A

N/A

**Reproduction URL**

https://play.tailwindcss.com/ykSBTsCL3K

**Describe your issue**

When we use @import url('https://...') for font files, it will be removed in the final CSS (compiled via tailwindcss)

wongjn:
Regression from #16121, where empty at-rules are stripped from output. There are some exceptions made (like `@layer`) but it seems `@import` was missed. This PR assumes that the non-`url()` `@import`s have already been dealt with up to this point.

Fixes #16136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
Shoot, good catch thanks! Will drop a release with this fix, thanks!

PR #16169, opened 2025-02-02
Update upgrade CLI to handle `containers` config correctly

joulev:
Right now, if you have

```js
{
  containers: {
    foo: "1rem",
  },
}
```

the generated CSS will be

```css
@theme {
  --containers-*: initial;
  --containers-foo: 1rem;
}
```

which is incorrect – it should be `container` not `containers`. Yes I spent a while thinking the Tailwind IntelliSense extension is not working, but in fact the extension is working fine and it's my CSS file that is invalid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Nothing yet!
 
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):

philipp-spiess:
Thanks for spotting this!

PR #16206, opened 2025-02-03
Ensure first argument to `var(…)` still unescapes `\_`

### LINKED ISSUE #16170: [v4] Inconsistent underscore

v4.0.3

Vite 6.0.5

v18.18.2

Chrome

WSL

**Reproduction URL**

https://play.tailwindcss.com/hmN4T4wuta

**Describe your issue**

Hi! I'm writing a React component library that styles Joy UI components the Tailwind CSS way.

Since the existing library was written to integrate with Tailwind CSS v3, I felt the need to guide how to integrate with v4, but I found some inconsistencies in the CSS build results, so I'm reporting the issue.

Please check the generated utility class in the attached playground link.

The first two classes (`h-[var(--\_root-size)]` and `text-[calc(0.2*var(--\_root-size))]`) both use `var(--\_root-size)` as an arbitrary value, but in the build output one refers to `--\_root-size`, while the other refers to `--_root-size`.

When I first noticed this difference, I suspected it had to do with the presence of `calc`, but in the case of the third class, the backslash remains even though `calc` and `var` are nested, unlike in the second class.

The last class is a bit more unusual, as even the two `--\_LinearProgress-padding`s written within one class differ in the presence or absence of backslashes in the build output.

It would be nice if the presence or absence of backslashes were consistent in build results.

philipp-spiess:
Resolves #16170

This PR fixes an issue where the previously opted-out escaping of the first argument for the `var(…)` function was not unescaped at all. This was introduced in https://github.com/tailwindlabs/tailwindcss/pull/14776 where the intention was to not require escaping of underscores in the var function (e.g. `ml-[var(--spacing-1_5)]`). However, I do think it still makes sense to unescape an eventually escaped underline for consistency.

## Test plan

The example from #1670 now parses as expected:

Screenshot 2025-02-03 at 13 51 35


packages/tailwindcss/src/candidate.test.ts
```diff
@@ -935,6 +935,154 @@ it('should parse a utility with an implicit variable as the modifier', () => {
   `)
 })
 
+it('should properly decode escaped underlines but not underlines to spaces for CSS variables in arbitrary positions', () => {
```
RobinMalfait, line 938:
```suggestion
it('should properly decode escaped underscores but not convert underscores to spaces for CSS variables in arbitrary positions', () => {
```

Not 100% sure what you wanted to say exactly, but replaced `underlines` with `underscores` and added missing word (or at least it felt like something was missing?)

philipp-spiess: hahaha yeah i think I wanted to say what you're saying 😆

RobinMalfait (approved):

PR #16210, opened 2025-02-03
Don't crash when setting JS theme value to `null`

### LINKED ISSUE #16035: [v4] "Cannot read properties of null (reading '50')" when using `theme.extend.colors.zinc: null`

 `"tailwindcss": "^4.0.1","@tailwindcss/vite": "^4.0.1",`

Astro

 `v22.13.1 (npm v10.9.2)`

Chrome

 macOS

**Describe your issue**

Tailwind 4 chokes when colors are set to `null` in the JS config.

I used to have code like this in my config to force only one color of gray and disable the autocomplete and class generation for all other colors:

```
  theme: {
    extend: {
      colors: {
        gray: colors.stone,
        slate: null,
        zinc: null,
        neutral: null,
        stone: null,
      },
    },
  },
```

With those settings I am getting errors like 
```
07:13:33 [ERROR] Cannot read properties of null (reading '50')
  Stack trace:
    at /Users/…/Development/…/src/layouts/base.css
    [...] See full stack trace in the browser, or rerun with --verbose.
```

This error shows when using the `npx @tailwindcss/upgrade@next` as well as when I update manually (referencing the js config from the css file) and use the vite Plugin in Astro.

---

I see the new way to disable default colors is https://tailwindcss.com/docs/colors#disabling-default-colors which does work once I remove the `null` lines from my js config.

philipp-spiess:
Closes #16035

In v3 it was possible to unset a specific color namespace by setting doing something like this:

```js
export default {
  theme: {
    extend: {
      colors: {
        red: null,
      },
    },
  },
}
```

This pattern would crash in v4 right now due to the theme access function not being able to work on the red property being a `null`. This PR fixes this crash.

However it leaves the behavior as-is for now so that the red namespace _defined via CSS will still be accessible_. This is technically different from v3 but fixing this would be more work as we only allow unsetting top-level namespaces in the interop layer (via the non-`extend`-theme-object). I would recommend migrating to the v4 API for doing these partial namespace resets if you want to get rid of the defaults in v4:

```css
@theme {
  --color-red-*: initial;
}
```

## Test plan

The crash was mainly captured via the test in `compat/config.test.ts` but I've added two more tests across the different levels of abstractions so that it's clear what `null` should be doing.


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Ensure that the `containers` JS theme key is added to the `--container-*` namespace. ([#16169](https://github.com/tailwindlabs/tailwindcss/pull/16169))
+- Fix a crash when setting JS theme values to `null` ([#16210](https://github.com/tailwindlabs/tailwindcss/pull/16210))
```
RobinMalfait, line 13:
```suggestion
- Fix a crash when setting JS theme values to `null` ([#16210](https://github.com/tailwindlabs/tailwindcss/pull/16210))
```

This was already present in the line below, but without a `.` at the end.

RobinMalfait (approved):
Small suggested change, otherwise looks good!

PR #16237, opened 2025-02-04
Ensure we always emit `@keyframes` correctly

### LINKED ISSUE #16227: [v4] Custom keyframes are not generated if the value of an animation variable contains a line break between the keyframe name and duration

v4.0.3

Vite 6.0.5

v18.18.2

Chrome

WSL

**Reproduction URL**

https://play.tailwindcss.com/PyXgGbc7uU?file=css

**Describe your issue**

If the keyframe name and duration are on different lines, the custom keyframe is not generated.

```
--animate-foo: <keyframe-name>
  <duration> <easing-function> <delay> <iteration-count> <direction> <fill-mode> <play-state>
```

If the keyframe name and duration are on the same line, the custom keyframe is generated.

```
--animate-foo: <keyframe-name> <duration>
  <easing-function> <delay> <iteration-count> <direction> <fill-mode> <play-state>
```

RobinMalfait:
This PR fixes an issue where we didn't always generate the `@keyframes`. Right now we only generate `@keyframes` _if_ they are being used as one of the `--animate-*` utilities.

However, if your `--animate-*` definition is pretty long such that it is defined across multiple lines, then we didn't always generate the `@keyframes` for it.

This is because the animation name would look like `'my-animation-name\n'` instead of `'my-animation-name'`.

Fixes: #16227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
-
 ### Fixed
 
 - Ensure that the `containers` JS theme key is added to the `--container-*` namespace. ([#16169](https://github.com/tailwindlabs/tailwindcss/pull/16169))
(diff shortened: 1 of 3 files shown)


philipp-spiess (approved):
🙈

PR #16211, opened 2025-02-03
Only expose used CSS variables

RobinMalfait:
This PR only exposes used CSS variables. 

My initial approach was to track the used variables, this was a bit messy because it meant that we had to walk part of the AST(s) in multiple places. We also had to be careful because sometimes if a variable exists in an AST, that doesn't mean that it's actually used. E.g.:

```css
h1 {
  color: var(--color-red-500); /* Definitely used, so let's keep it */
}

@utility foo {
  color: var(--color-blue-500); /* Hmm, used? */
}
```
In this last case, the `--color-blue-500` is part of the CSS AST, but as long as `foo` the utility is not used, it won't end up in your actual CSS file, therefore the variable is **not** used.

Alternatively, if the `foo` utility is used with an invalid variant (e.g.: `group-[>.foo]:foo`, then the `@utility foo` code will still run internally because variants are applied on top of the utility. This means that it looks like `var(--color-blue-500)` is being used.

Another annoying side effect was that because variables are conditionally generated, that the `@theme` -> `:root, :host` conversion had to happen for every build, instead of once in the `compile(…)` step.

---

To prevent all the messy rules and additional booking while walking of ASTs I thought about a different approach. We are only interested in variables that are actually used. The only way we know for sure, is right before the `toCss(…)` step. Any step before that could still throw away AST nodes.

However, we do have an `optimizeAst` step right before printing to simplify and optimize the AST. So the idea was to keep all the CSS variables in the AST, and only in the `optimizeAst` step we perform a kind of mark-and-sweep algorithm where we can first check which variables are _actually_ used (these are the ones that are left in the AST), and later we removed the ones that weren't part of known used list.

Moving the logic to this step feels a natural spot for this to happen, because we are in fact optimizing the AST. We were already walking the AST, so we can just handle these cases while we are walking without additional walks. Last but not least, this also means that there is only a single spot where need to track and remove variables.

Now, there is a different part to this story. If you use a variable in JS land for example, we also want to make sure that we keep the CSS variable in the CSS. To do this, we can mark variables as being used in the internal `Theme`.

The Oxide scanner will also emit used variables that it can find such as `var(--color-red-500)` and will emit `--color-red-500` as a "candidate". We can then proactively mark this one as used even though it may not be used anyway in the actual AST.

---

### Always including all variables

[...]

PR #16283, opened 2025-02-05
Improve performance of rebuilds

RobinMalfait:
This PR introduces a performance improvement we noticed while working on on: https://github.com/tailwindlabs/tailwindcss/pull/16211

We noticed that `substituteFunctions` was being called on every `node` after the `compileAstNodes` was done. However, the `compileAstNodes` is heavily cached.

By moving the `substituteFunctions` we into the cached `compileAstNodes` we sped up performance for Catalyst rebuilds from ~15ms to ~10ms.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -9,7 +9,6 @@ import {
   type StyleRule,
 } from './ast'
 import { type Candidate, type Variant } from './candidate'
-import { substituteFunctions } from './css-functions'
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
Oh, found a whoopsie there huh 😅

RobinMalfait:
@philipp-spiess haha just a nice feeling when you move some code around and improve performance 😅

Issue #15716, opened 2025-01-23
Migration tool doesn't recognize .scss files


ajuvonen:
**What version of Tailwind CSS are you using?**

3.4.17

**What build tool (or framework if it abstracts the build tool) are you using?**

Vite 6, Vue 3, postCSS

**What version of Node.js are you using?**

22

**What browser are you using?**

N/A

**What operating system are you using?**

Windows 11

**Reproduction URL**

No repro

**Describe your issue**

My project has a .scss file with Tailwindcss declaration, which is in turn imported in the main .ts file. Trying to run migration with `npx @tailwindcss/upgrade@next` returns the following error:

│ Searching for CSS files in the current directory and its subdirectories…

│ ↳ Cannot find any CSS files that reference Tailwind CSS.
│   Before your project can be upgraded you need to create a CSS file that imports Tailwind CSS or uses `@tailwind`.

The error disappears if I rename my .scss stylesheet to .css.

ghidoz:
Same here, in my Angular project, with scss.

MarcelKlammer:
https://tailwindcss.com/docs/compatibility

Don't use scss anymore, use plain css files.

e-oz:
It’s easier to say for a new project, but new projects don’t need this tool :)

ghidoz:
Yeah, what if I'm using a custom scss theme in Angular Material?

sheikalthaf:
Hey Angulars,

If your scss file contains **// comments** then it will break the tailwind v4 instead use **/\* comment \*/**.

@adamwathan FYI.

Kinselan:
I'm having the same issue -- .scss file is blocking the upgrade.

AntonioVentilii:
Any news here? Same problem

Hidduh:
Same issue here, I can temporarily change it to a regular .css file but this would require quite a few adjustments in these files.

e-oz:
Every Angular project with Angular Material uses mixins, so it means such projects simply can't upgrade to v4.

bugproof:
> Every Angular project with Angular Material uses mixins, so it means such projects simply can't upgrade to v4.

I think you can still use it. Just use regular .css for the rest of your project and other overrides. The main issue right now is with `@reference` I think as it requires massive rewrites at least in angular where you don't have direct access to vite config (don't ask me why - angular sucks).

jits:
[...]

PR #16285, opened 2025-02-05
Add .scss files to upgrade script.

GameLord2011:
I use an index.scss file in my project, and the upgrade script did not detect it.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -80,7 +80,7 @@ async function run() {
     if (files.length === 0) {
       info('Searching for CSS files in the current directory and its subdirectories…')
 
-      files = await globby(['**/*.css'], {
+      files = await globby(['**/*.css', '**/*.scss'], {
         absolute: true,
         gitignore: true,
       })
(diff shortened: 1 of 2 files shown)


GameLord2011:
WAIT DO NOT COMMIT THIS YET, I NEED TO ADD MORE CODE!

GameLord2011:
Can anyone help with adding the scss file format (switch `@import` for `@use` if scss and whatnot)?

wongjn:
See https://github.com/tailwindlabs/tailwindcss/issues/15716#issuecomment-2619593290

philipp-spiess:
Hey @GameLord2011. Thanks for the contribution but the [comment that @wongjn linked to](https://github.com/tailwindlabs/tailwindcss/issues/15716#issuecomment-2619593290) from my colleague @RobinMalfait is unfortunately still our stance on this.

Even if we fix the upgrade tool to work around some of the issues you already found (e.g. the `@use` stuff), SCSS is not something we support in Tailwind CSS v4: https://tailwindcss.com/docs/compatibility#sass-less-and-stylus. So the only way to make the upgrade tool work with SCSS files would be be to also convert all `.scss` files to `.css` which isn't even possible since not all SCSS features exist in Tailwind CSS.

Instead, our recommended migration approach is that you convert your SCSS project to CSS _before_ you migrate to Tailwind CSS v4.

GameLord2011:
I found a workaround where if you rename your scss stylesheets to css then
it will upgrade, except that that causes a broken build

PR #16295, opened 2025-02-06
Bun plugin

zackradisic:
> (Sorry for creating this PR without opening an issue first)

This PR implements a Bun plugin that makes TailwindCSS fast in Bun.

c612d2faa37f2cff03c8620a08bb2def7268851bb183c2fa8e2763246b9243e5

(bundling 2048 html + react + tailwind files)

# Overview

This plugins is comprised of two parts:

- A regular Bun bundler plugin which is the main driver of everything and invokes the native bundler plugin
- A [native bundler plugin](https://bun.sh/docs/bundler/plugins#native-plugins) which parallelizes the module graph scanning of candidates. Tailwind was part of the motivation for implementing this API.
 
## Native bundler plugin

The native bundler plugin is used to scan the module graph in parallel with the `Scanner` struct from `@tailwindcss/oxide`.

The main logic for this code is in the `tw_on_before_parse` function.

Native bundler plugins run in parallel on Bun's bundler threads and do not need to do UTF-16 <-> UTF-8 string conversions. This speeds up the plugin a lot.

Native bundler plugins are NAPI modules which export additional symbols (since NAPI modules themselves are dynamically loaded libraries which can be `dlopen()`'d). The [`bun-native-plugin`](https://crates.io/crates/bun-native-plugin) crate handles the boilerplate for creating one.

I placed the Bun plugin inside the existing `crates/node/lib.rs` (the `@tailwindcss/oxide` package). This reduces the need to create more compiled artifacts at the cost of a relatively small binary size change:

```sh
# original size
❯ ls -lhS dist/tailwindcss-oxide-darwin-arm64.tgz
-rw-r--r--@ 1 zackradisic  staff   2.1M Feb  4 20:42 dist/tailwindcss-oxide-darwin-arm64.tgz

# new size
❯ ls -lhS dist/tailwindcss-oxide-darwin-arm64.tgz
-rw-r--r--@ 1 zackradisic  staff   2.2M Feb  4 18:42 dist/tailwindcss-oxide-darwin-arm64.tgz
```

Please let me know if you would like me to split it out into its own separate package if you don't like the binary size change.

### Sharing state between the native plugin and JS

The scanned candidates and other state are held inside a NAPI External. The struct in the code that does
this is called `TailwindContextExternal`.

A NAPI External is a NAPI value which can be given to JS and which holds a `void*` data pointer. This data is inaccessible to JS, but a NAPI module can dereference the data and convert it to NAPI values.

This looks a bit like this on the Rust side:

```rust
/// Create the TailwindContextExternal and return it to JS wrapped in a Napi External.
///
/// Napi has an `External<T>` type which allows us to wrap it in an
/// external easily.
#[no_mangle]
#[napi]
pub fn twctx_create() -> External<TailwindContextExternal> {
  let external = External::new(TailwindContextExternal {
    module_graph_candidates: Default::default(),
    dirty: AtomicBool::new(false),
  });

  external
}
```

And the JS side:

```ts
// import napi functions which let us manipulate the external
import { twctxCreate, twctxIsDirty, twctxToJs } from '@tailwindcss/oxide'

// create the state, the returned value
// is a Napi External
const external = twctxCreate()

/* ... other code ... */

let moduleGraphCandidates = new Map<string, Set<string>>()
function getSharedCandidates() {
  // check if there are changes
  if (twctxIsDirty(external)) {
    // convert the state into js values
    let rawCandidates: Array<{ id: string; candidates: string[] }> = twctxToJs(external)
    for (let { id, candidates } of rawCandidates) {
      moduleGraphCandidates.set(id, new Set(candidates))
    }
  }
  return moduleGraphCandidates
}
```

### `napi-rs` version bump

The `napi-rs` crate was updated to version `2.16.15` so we can use the [`External::inner_from_raw()`](https://docs.rs/napi/latest/napi/bindgen_prelude/struct.External.html#method.inner_from_raw) function to turn an `External`'s `*mut c_void` pointer back into `TailwindContextExternal`.

## JS plugin

The JS plugin `@tailwindcss-bun/src/index.ts` uses logic copied over from the vite plugin implementation but modified to work with Bun's plugin API.

It invokes the native bundler plugin using the `.onBeforeParse` plugin API function:

```ts
// Called on every file which matches the filter
// before it is parsed by Bun
build.onBeforeParse(
  // filter which files the native plugin apply to
  { filter: NON_CSS_ROOT_FILE_RE },

  // pass the napi module, the symbol which points to the plugin main function,
  // and the external which holds the tailwind state
  { napiModule: addon, symbol: 'tw_on_before_parse', external },
)
```

One thing to note is that Bun's bundler currently does not have an API that is analogous to `.addWatchedFile()`, so there is currently no way to add additional files to the module graph.

## Testing

I added a `integrations/bun/index.test.ts` file, please let me know if you would like more tests


philipp-spiess:
Hey! Really excited for an official Bun plugin! We’re happy to collaborate and work on this together but to set the expectations right it will realistically take a few weeks before someone from our team can focus on this.

One reason for that is that we’re currently working on some changes that will break the existing Oxide <> Node APIs that are necessary fix some critical bugs. These changes will then impact the API design for the updated Oxide bindings necessary for the Bun plugin as well. 

I’m sorry that we can’t hop onto this right away but we have to prioritize some other important fixes for the v4 release right now, I hope you understand. We’re pretty stoked about this feature though and hope to get to it soon!

Nanome203:
Any update on this?

marcospgp:
would love to see this make it through! using tailwind with bun is a much better experience than vite, yet the former is not even listed in the installation docs, and the latter is the default suggestion 😭

ahmafi:
Where is this plugin coming from? I can't find the source of it. https://www.npmjs.com/package/bun-plugin-tailwind

Nanome203:
bump

choucavalier:
bump again! first class support of bun would be great :heart: thx @zackradisic for working on this!

neves:
Any follow up on this?

qodesmith:
Had to abandon `bun-plugin-tailwind` and go with a custom plugin that uses the Tailwind-maintained packages `@tailwindcss/node` and `@tailwindcss/oxide`. Hope this helps someone else out there:

```typescript
import type {BunPlugin} from 'bun'

import path from 'node:path'
import process from 'node:process'

import {compile, optimize} from '@tailwindcss/node'
import {Scanner} from '@tailwindcss/oxide'

/**
 * Fast skip for CSS files that don't use Tailwind. Anything containing one of
 * these directives goes through the Tailwind compiler; everything else is
 * returned untouched so Bun's native CSS loader handles it.
 */
const TAILWIND_DIRECTIVE =
  /@import\s+["']tailwindcss["']|@theme|@apply|@tailwind\b/

/**
 * Bun plugin that replaces `bun-plugin-tailwind` with a direct binding to the
 * official Tailwind v4 node packages. Because this plugin uses the
 * `tailwindcss` package from the project's own dependencies (not a bundled
 * copy), new utilities ship the same day Tailwind releases them.
 */
export function bunTailwindPlugin(): BunPlugin {
  const minify = process.env.NODE_ENV === 'production'

  return {
    name: 'bun-tailwind-plugin',
    setup(build) {
      build.onLoad({filter: /\.css$/}, async ({path: filePath}) => {
        const source = await Bun.file(filePath).text()
        if (!TAILWIND_DIRECTIVE.test(source)) return

        const base = path.dirname(filePath)
        const compiler = await compile(source, {
          from: filePath,
          base,
          shouldRewriteUrls: true,
          onDependency: () => {},
        })

        /**
         * Mirrors @tailwindcss/vite: the scanner needs both the compiler's
         * auto-detection root and any explicit `@source` directives. Passing
         * only `compiler.sources` yields an empty set, so generated CSS would
         * drop every utility that isn't in a @source file.
         */
        const rootSources = (() => {
          if (compiler.root === 'none') return []
          if (compiler.root === null) {
            return [{base: process.cwd(), pattern: '**/*', negated: false}]
          }
          return [{...compiler.root, negated: false}]
        })()
        const sources = rootSources.concat(compiler.sources)

        const candidates = new Scanner({sources}).scan()
        let css = compiler.build(candidates)
        if (minify) css = optimize(css, {minify: true}).code

        return {contents: css, loader: 'css'}
      })
    },
  }
}

// biome-ignore lint/style/noDefaultExport: Bun expects a default export
export default bunTailwindPlugin()
```

HotariTobu:
@zackradisic — do you intend to rebase this onto current `main`?
@philipp-spiess — if rebased, would your team review this, or should it be closed?

- [`bun-plugin-tailwind` 0.1.2 (2025-10-09)](https://www.npmjs.com/package/bun-plugin-tailwind/v/0.1.2) bundles its own copy of the Tailwind engine, so utilities added in v4.2.0 (`inset-s-*`, `inset-e-*`, `inset-bs-*`, `inset-be-*`, …) silently emit no CSS. Reported in #19940, closed as not planned because the plugin is not part of this repository.
- The package's `repository` / `bugs` metadata point to `packages/@tailwindcss-bun` — the directory this PR adds, which doesn't exist on `main`. Bug reports for the plugin have nowhere to go.

PR #16300, opened 2025-02-06
Properly skip over nodes when using `replaceNode`

### LINKED ISSUE #16298: [v4] "Maximum call stack size exceeded" when using @resource with @slot

4.0.3

angular: 19.1.4, tailwindcss: 4.0.3, tailwindcss/postcss: 4.0.3

23.7.0

Chrome

macOS

**Reproduction URL**

https://github.com/Joxeur2/tailwind-test.git

**Describe your issue**

I have an Angular application that I am migrating to Tailwind 4. I'm also using a purchased theme that is based on Tailwind and has also been migrated to Tailwind 4.
However, when I use the theme, I get the error "Maximum call stack size exceeded."

After some investigation, I managed to reproduce the issue and created a separate repo with minimal setup here: https://github.com/Joxeur2/tailwind-test.git

My problem occurs when these two things are present:

- I use a @resource pointing to my main.css to be able to use @apply
- I use a @slot in main.css

I imagine I could move my @variant along with my @slot into another CSS file and only reference the rest with @resource, but as I mentioned, I am using a purchased theme and I would really hate to have to mess with the theme’s code...

philipp-spiess:
Fixes #16298 

This PR fixes an issue where using an AST walk in combination with `replaceNode` and various `SkipAction` would either cause children to be visited multiple times or not visited at all even though it should. This PR fixes the issue which also means we can get rid of a custom walk for `@variant` inside the `@media` that was used to apply `@variant` because we never recursively visited children inside the `@media` rule.

Because we now can use the regular walk for `@variant`, we now properly convert `@variant` to `@custom-variant` inside `@reference`-ed stylesheet which also fixes #16298

## Test plan

Lots of tests added to ensure the combinations of `WalkAction` and `replaceWith()` works as expected.

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -137,31 +137,46 @@ export function walk(
     }
 
     path.push(node)
+    let replacedNode = false
+    let replacedNodeOffset = 0
     let status =
       visit(node, {
         parent,
         context,
         path,
         replaceWith(newNode) {
+          replacedNode = true
+
           if (Array.isArray(newNode)) {
             if (newNode.length === 0) {
               ast.splice(i, 1)
+              replacedNodeOffset = 0
             } else if (newNode.length === 1) {
               ast[i] = newNode[0]
+              replacedNodeOffset = 1
             } else {
               ast.splice(i, 1, ...newNode)
+              replacedNodeOffset = newNode.length
             }
           } else {
             ast[i] = newNode
+            replacedNodeOffset = 1
           }
-
-          // We want to visit the newly replaced node(s), which start at the
-          // current index (i). By decrementing the index here, the next loop
-          // will process this position (containing the replaced node) again.
-          i--
         },
       }) ?? WalkAction.Continue
     path.pop()
+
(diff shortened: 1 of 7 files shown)


packages/tailwindcss/src/ast.ts
```diff
@@ -137,39 +137,54 @@ export function walk(
     }
 
     path.push(node)
+    let replacedNode = false
```
RobinMalfait, line 140: We probably want to port this behavior to other `walk` implementations in the codebase:

https://github.com/tailwindlabs/tailwindcss/blob/d566dbdef658b8c98a8c596466d74f11c72295e2/packages/tailwindcss/src/compat/selector-parser.ts#L82-L130

and 

https://github.com/tailwindlabs/tailwindcss/blob/d566dbdef658b8c98a8c596466d74f11c72295e2/packages/tailwindcss/src/value-parser.ts#L53-L101

philipp-spiess: ay

RobinMalfait (approved):

PR #16306, opened 2025-02-06
Improve Oxide candidate extractor [0]

### LINKED ISSUE #15616: Valid Angular class binding not recognized

v4.0.0-beta.9

Angular v19, using the `@angular-devkit/build-angular:application` builder, which uses PostCSS under the hood

v22.12.0

Brave v1.73.105

macOS 15.2 (24C101)

**Reproduction URL**

https://github.com/JoepKockelkorn/tailwindcss-v4-beta-angular-class-bug

**Describe your issue**

If I use valid Angular class binding syntax to dynamically add a class in the following way, it is not recognized by tailwind:

- `npm i`
- `npm start`
- Go to http://localhost:4200/

```
<div [class.bg-red-500]="showRedBackground">This should have a red background</div>
```

There are workarounds, like using `[class]="{ 'bg-red-500': showRedBackground }"`, but ideally all valid Angular syntax which uses valid static tailwind syntax should work.

### LINKED ISSUE #16750: Some classes not being detected in source files with square brackets
Versions:
```
@tailwindcss/cli: 4.0.8
tailwindcss: 4.0.8
```

**Build tool**: None
**Node Version**: v23.7.0
**Browser**: N/A
**OS**: macOS (15.3.1)
**Reproduction URL**: https://github.com/caleb/tailwindcss4bug

**Issue**
Tailwind is not dectecting a class when it's in a source file with a specific layout.

Let's say we have a component that has this code:

```ruby
["h-[calc(100vh-(var(--spacing)*8)-(var(--spacing)*14))]",
 true],
```
Tailwind 4 will not detect the height class (version 3 did), and it won't include it in the output css file.

If we have either of these other layouts, then it does detect the class and generates the appropriate CSS:

```ruby
["h-[calc(100vh-(var(--spacing)*8)-(var(--spacing)*14))]", true],
```

```ruby
[ "h-[calc(100vh-(var(--spacing)*8)-(var(--spacing)*14))]",
 true],
```

### LINKED ISSUE #16790: [v4] Slim template regression - 2xl variant classes not extracted
There is a regression in v4 in Slim templates - the 2xl variant classes are not extracted from source files.

4.0.8

tailwindcss CLI

n/a

n/a

Linux

**Reproduction URL**

https://github.com/borama/tailwind4-slim-issue

**Describe your issue**

There is a regression similar to #14005 but present in Tailwind v4 (4.0.8). The `2xl:` variant classes are not extracted from sources by the Tailwind CLI.

Sample project:

```
$ tree
.
├── sample.slim
├── tw3.config.js
├── tw3.css
└── tw4.css

$ head *
==> sample.slim <==
.bg-blue-100.2xl:bg-red-100
 ' Hi!

==> tw3.config.js <==
module.exports = {
  content: ["sample.slim"]
}

==> tw3.css <==
@import "tailwindcss/base";
@import "tailwindcss/components";

==> tw4.css <==
@source "sample.slim";
```

#### 1. This works OK in v3 (tailwindcss v3.4.17):

```
$ tailwindcss -i tw3.css -c tw3.config.js

...
.bg-blue-100 {
  --tw-bg-opacity: 1;
  background-color: rgb(219 234 254 / var(--tw-bg-opacity, 1));
}

@media (min-width: 1536px) {
  .\32xl\:bg-red-100 {
    --tw-bg-opacity: 1;
    background-color: rgb(254 226 226 / var(--tw-bg-opacity, 1));
  }
}
```

#### 2. The same Slim template does NOT work in v4:

```
$ tailwindcss -i tw4.css

...
@layer utilities;

≈ tailwindcss v4.0.8

Done in 20ms
```

I.e. not only it does not extract the `2xl:bg-red-100` class, it ignores all classes in the Slim template on the same line.

#### 3. When the `2xl:` class is the first on the line, it works again

Interestingly, when I swap the order of the classes, making the `2xl:` class be the first on the line, it starts working:

```
$ cat sample.slim
.2xl:bg-red-100.bg-blue-100
 ' Hi!

$ tailwindcss -i tw4.css

...
@layer utilities {
  .bg-blue-100 {
    background-color: var(--color-blue-100);
  }
  .\32 xl\:bg-red-100 {
    @media (width >= 96rem) {
      background-color: var(--color-red-100);
    }
  }
}

≈ tailwindcss v4.0.8

Done in 20ms
```

### LINKED ISSUE #16801: Tailwindcss classes not generated when used with angular class binding.
v4.0.8

angular 19.1.7, postcss 8.5.3

node 23.8.0

Microsoft Edge

Linux

**Reproduction URL**
N/A

**Describe your issue**
```typescript
import { NgClass } from '@angular/common';
import { ChangeDetectionStrategy, Component, input } from '@angular/core';

@Component({
  selector: 'app-button, [app-button]',
  imports: [NgClass],
  templateUrl: './button.component.html',
  styleUrl: './button.component.scss',
  host: {
    '[attr.role]': "'button'",
    '[attr.variant]': 'variant()',
    '[ngClass]': `{"variant": variant(), "no-variant": !variant() }`,
    '[class]': '"bg-gradient-to-b px-6 py-3 rounded-3xl from-5%"',
    '[class.from-secondary-light]': `variant() === 'secondary'`,  // tailwindcss class not generated
    '[class.to-secondary]': `variant() === 'secondary'`,   // tailwind class not generated
    '[class.from-5%]': `variant() === 'secondary'`,   // tailiwind class not generated
    '[class.from-1%]': `variant() === 'primary'`,    // tailwind class not generated
    '[class.from-light-blue]': `variant() === 'primary'`,  // tailiwind class not generated
    '[class.to-primary]': `variant() === 'primary'`,  // tailwind class not generate
  },
  changeDetection: ChangeDetectionStrategy.OnPush,
})
export class ButtonComponent {
  variant = input<'primary' | 'secondary'>('secondary');

  /** The tailwindcss color to use as button shadow */
  shadow = input<string>('blue-500');

  /** The tailwindcss color to use as button background */
  background = input<string>('blue-600');
}

```

As shown on the piece of code above, any tailwindcss classes used with angular class binding ```[class.xxx-xxx]``` are not generated.

### LINKED ISSUE #16880: `.eps` files are included in source file detection

v4.0.6

@tailwindcss/cli v4.0.9

v23.7.0

Arc (chromium)

macOS

**Reproduction URL**

https://codesandbox.io/p/devbox/lffyvk

**Describe your issue**

When generating output CSS using Tailwind CLI (seems to also happen with the vite plugin), Tailwind includes .eps files and thus adds invalid syntax to the output CSS:
```css
.\[h3_ZF_m9\:O\] {
    h3_ZF_m9: O;
  }
  .\[ro0OH\:\>GA9\*\&Z\;oheh-\!a_V\)gL\?\*\,oC_FdhIkP\=\=SBF\,B\\\`ff\$P2amRVs\] {
    ro0OH: >GA9*&Z;oheh-!a V)gL?*,oC FdhIkP==SBF,B\`ff$P2amRVs;
  }
  .\[t0\:IuYnhV6c\] {
    t0: IuYnhV6c;
  }
```

I couldn't find a way to manually exclude the file from source file detection apart from adding it to the .gitignore -- maybe it would be worth adding a way to ignore individual misbehaving files?

RobinMalfait:
This PR adds a new candidate[^candidate] extractor with 2 major goals in mind:

1. It must be way easier to reason about and maintain.
2. It must have on-par performance or better than the current candidate extractor.

### Problem

Candidate extraction is a bit of a wild west in Tailwind CSS and it's a very critical step to make sure that all your classes are picked up correctly to ensure that your website/app looks good.

One issue we run into is that Tailwind CSS is used in many different "host" languages and frameworks with their own syntax. It's not only used in HTML but also in JSX/TSX, Vue, Svelte, Angular, Pug, Rust, PHP, Rails, Clojure, .NET, … the list goes on and all of these have different syntaxes. Introducing dedicated parsers for each of these languages would be a huge maintenance burden because there will be new languages and frameworks coming up all the time. The best thing we can do is make assumptions and so far we've done a pretty good job at that.

The only certainty we have is that there is at least _some_ structure to the possible Tailwind classes used in a file. E.g.: `abc#def` is definitely not a valid class, `hover:flex` definitely is. In a perfect world we limit the characters that can be used and defined a formal grammar that each candidate must follow, but that's not really an option right now (maybe this is something we can implement in future major versions).

The current candidate extractor we have has grown organically over time and required patching things here and there to make it work in various scenarios (and edge cases due to the different languages Tailwind is used in).

While there is definitely some structure, we essentially work in 2 phases:

1. Try to extract `0..n` candidates. (This is the hard part)
2. Validate each candidate to make sure they are valid looking classes (by validating against the few rules we have)

Another reason the current extractor is hard to reason about is that we need it to be fast and that comes with some trade-offs to readability and maintainability.

Unfortunately there will always be a lot of false positives, but if we extract more classes than necessary then that's fine. It's only when we pass the candidates to the core engine that we will know for sure if they are valid or not. (we have some ideas to limit the amount of false positives but that's for another time)

### Solution

Since the introduction of Tailwind CSS v4, we re-worked the internals quite a bit and we have a dedicated internal AST structure for candidates. For example, if you take a look at this:

```html
<div class="[@media(pointer:fine)]:data-[state=pending]:hover:text-red-500/(--my-opacity)"></div>
```

<details>

<summary>This will be parsed into the following AST:</summary>

```json
[
  {
    "kind": "functional",
    "root": "text",
    "value": {
      "kind": "named",
      "value": "red-500",
      "fraction": null
    },
    "modifier": {
      "kind": "arbitrary",
      "value": "var(--my-opacity)"
    },
    "variants": [
      {
        "kind": "static",
        "root": "hover"
      },
      {
        "kind": "functional",
        "root": "data",
        "value": {
          "kind": "arbitrary",
          "value": "state=pending"
        },
        "modifier": null
      },
      {
        "kind": "arbitrary",
        "selector": "@media(pointer:fine)",
        "relative": false
      }
    ],
    "important": false,
    "raw": "[@media(pointer:fine)]:data-[state=pending]:hover:text-red-500/(--my-opacity)"
  }
]
```

</details>

We have a lot of information here and we gave these patterns a name internally. You'll see names like `functional`, `static`, `arbitrary`, `modifier`, `variant`, `compound`, ...

Some of these patterns will be important for the new candidate extractor as well:

| Name                       | Example           | Description                                                                                         |
| -------------------------- | ----------------- | --------------------------------------------------------------------------------------------------- |
| Static utility (named)     | `flex`            | A simple utility with no inputs whatsoever                                                          |
| Functional utility (named) | `bg-red-500`      | A utility `bg` with an input that is named `red-500`                                                |
| Arbitrary value            | `bg-[#0088cc]`    | A utility `bg` with an input that is arbitrary, denoted by `[…]`                                    |
| Arbitrary variable         | `bg-(--my-color)` | A utility `bg` with an input that is arbitrary and has a CSS variable shorthand, denoted by `(--…)` |
| Arbitrary property         | `[color:red]`     | A utility that sets a property to a value on the fly                                                |

A similar structure exist for modifiers, where each modifier must start with `/`:

| Name               | Example                     | Description                              |
| ------------------ | --------------------------- | ---------------------------------------- |
| Named modifier     | bg-red-500`/20`             | A named modifier                         |
| Arbitrary value    | bg-red-500`/[20%]`          | An arbitrary value, denoted by `/[…]`    |
| Arbitrary variable | bg-red-500`/(--my-opacity)` | An arbitrary variable, denoted by `/(…)` |

Last but not least, we have variants. They have a very similar pattern but they _must_ end in a `:`.

| Name               | Example                     | Description                                                              |
| ------------------ | --------------------------- | ------------------------------------------------------------------------ |
| Named variant      | `hover:`                    | A named variant                                                          |
| Arbitrary value    | `data-[state=pending]:`     | An arbitrary value, denoted by `[…]`                                     |
| Arbitrary variable | `supports-(--my-variable):` | An arbitrary variable, denoted by `(…)`                                  |
| Arbitrary variant  | `[@media(pointer:fine)]:`   | Similar to arbitrary properties, this will generate a variant on the fly |

The goal with the new extractor is to encode these separate patterns in dedicated pieces of code (we called them "machines" because they are mostly state machine based and because I've been watching Person of Interest but I digress).

This will allow us to focus on each pattern separately, so if there is a bug or some new syntax we want to support we can add it to those machines.

One nice benefit of this is that we can encode the rules and handle validation as we go. The moment we know that some pattern is invalid, we can bail out early.

At the time of writing this, there are a bunch of machines:

<details>

<summary>Overview of the machines</summary>

- `ArbitraryPropertyMachine`

  Extracts candidates such as `[color:red]`. Some of the rules are:

  1. There must be a property name
  2. There must be a `:`
  3. There must ba a value

  There cannot be any spaces, the brackets are included, if the property is a CSS variable, it must be a valid CSS variable (uses the `CssVariableMachine`).

  ```
  [color:red]
  ^^^^^^^^^^^

  [--my-color:red]
  ^^^^^^^^^^^^^^^^
  ```

  Depends on the `StringMachine` and `CssVariableMachine`.

- `ArbitraryValueMachine`

  Extracts arbitrary values for utilities and modifiers including the brackets:

  ```
  bg-[#0088cc]
     ^^^^^^^^^

  bg-red-500/[20%]
             ^^^^^
  ```

  Depends on the `StringMachine`.

- `ArbitraryVariableMachine`

  Extracts arbitrary variables including the parentheses. The first argument must be a valid CSS variable, the other arguments are optional fallback arguments.

  ```
  (--my-value)
  ^^^^^^^^^^^^

  bg-red-500/(--my-opacity)
             ^^^^^^^^^^^^^^
  ```

  Depends on the `StringMachine` and `CssVariableMachine`.

- `CandidateMachine`

  Uses the variant machine and utility machine. It will make sure that 0 or more variants are directly touching and followed by a utility.

  ```
  hover:focus:flex
  ^^^^^^^^^^^^^^^^

  aria-invalid:bg-red-500/(--my-opacity)
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ```

  Depends on the `VariantMachine` and `UtilityMachine`.

- `CssVariableMachine`

  Extracts CSS variables, they must start with `--` and must contain at least one alphanumeric character or, `-`, `_` and can contain any escaped character (except for whitespace).

  ```
  bg-(--my-color)
      ^^^^^^^^^^

  bg-red-500/(--my-opacity)
              ^^^^^^^^^^^^

  bg-(--my-color)/(--my-opacity)
      ^^^^^^^^^^   ^^^^^^^^^^^^
  ```

- `ModifierMachine`

  Extracts modifiers including the `/`

  - `/[` will delegate to the `ArbitraryValueMachine`
  - `/(` will delegate to the `ArbitraryVariableMachine`

  ```
  bg-red-500/20
            ^^^

  bg-red-500/[20%]
            ^^^^^^

  bg-red-500/(--my-opacity)
            ^^^^^^^^^^^^^^^
  ```

  Depends on the `ArbitraryValueMachine` and `ArbitraryVariableMachine`.

- `NamedUtilityMachine`

  Extracts named utilities regardless of whether they are functional or static.

  ```
  flex
  ^^^^

  px-2.5
  ^^^^^^
  ```

  This includes rules like: A `.` must be surrounded by digits.

  Depends on the `ArbitraryValueMachine` and `ArbitraryVariableMachine`.

- `NamedVariantMachine`

  Extracts named variants regardless of whether they are functional or static. This is very similar to the `NamedUtilityMachine` but with different rules. We could combine them, but splitting things up makes it easier to reason about.

  Another rule is that the `:` must be included.

  ```
  hover:flex
  ^^^^^^

  data-[state=pending]:flex
  ^^^^^^^^^^^^^^^^^^^^^

  supports-(--my-variable):flex
  ^^^^^^^^^^^^^^^^^^^^^^^^^
  ```

  Depends on the `ArbitraryVariableMachine`, `ArbitraryValueMachine`, and `ModifierMachine`.

- `StringMachine`

  This is a low-level machine that is used by various other machines. The only job this has is to extract strings that start with double quotes, single quotes or backticks.

  We have this because once you are in a string, we don't have to make sure that brackets, parens and curlies are properly balanced. We have to make sure that balancing brackets are properly handled in other machines.

  ```
  content-["Hello_World!"]
           ^^^^^^^^^^^^^^

  bg-[url("https://example.com")]
          ^^^^^^^^^^^^^^^^^^^^^
  ```

- `UtilityMachine`

  Extracts utilities, it will use the lower level `NamedUtilityMachine`, `ArbitraryPropertyMachine` and `ModifierMachine` to extract the utility.

  It will also handle important markers (including the legacy important marker).

  ```
  flex
  ^^^^

  bg-red-500/20
  ^^^^^^^^^^^^^

  !bg-red-500/20      Legacy important marker
  ^^^^^^^^^^^^^^

  bg-red-500/20!      New important marker
  ^^^^^^^^^^^^^^

  !bg-red-500/20!     Both, but this is considered invalid
  ^^^^^^^^^^^^^^^
  ```

  Depends on the `ArbitraryPropertyMachine`, `NamedUtilityMachine`, and `ModifierMachine`.

- `VariantMachine`

  Extracts variants, it will use the lower level `NamedVariantMachine` and `ArbitraryValueMachine` to extract the variant.

  ```
  hover:focus:flex
  ^^^^^^
        ^^^^^^
  ```

  Depends on the `NamedVariantMachine` and `ArbitraryValueMachine`.

</details>

One important thing to know here is that each machine runs to completion. They all implement a `Machine` trait that has a `next(cursor)` method and returns a `MachineState`.

The `MachineState` looks like this:

```rs
enum MachineState {
  Idle,
  Done(Span)
}
```

Where a `Span` is just the location in the input where the candidate was found.

```rs
struct Span {
  pub start: usize,
  pub end: usize,
}
```

#### Complexities

**Boundary characters:**

When running these machines to completion, they don't typically check for boundary characters, the wrapping `CandidateMachine` will check for boundary characters.

A boundary character is where we know that even though the character is touching the candidate it will not be part of the candidate.

```html
<div class="flex"></div>
<!--       ^    ^    -->
```

The quotes are touching the candidate `flex`, but they will not be part of the candidate itself, so this is considered a valid candidate.

**What to pick?**

Let's imagine you are parsing this input:

```html
<div class="hover:flex"></div>
```

The `UtilityMachine` will find `hover` and `flex`. The `VariantMachine` will find `hover:`. This means that at a certain point in the `CandidateMachine` you will see something like this:

```rs
let variant_machine_state = variant_machine.next(cursor);
//  MachineState::Done(Span { start: 12, end: 17 })        // `hover:`

let utility_machine_state = utility_machine.next(cursor);
//  MachineState::Done(Span { start: 12, end: 16 })        // `hover`
```

They are both done, but which one do we pick? In this scenario we will always pick the variant because its range will always be 1 character longer than the utility.

Of course there is an exception to this rule and it has to do with the fact that Tailwind CSS can be used in different languages and frameworks. A lot of people use `clsx` for dynamically applying classes to their React components. E.g.:

```tsx
<div
  class={clsx({
    underline: someCondition(),
  })}
></div>
```

In this scenario, we will see `underline:` as a variant, and `underline` as a utility. We will pick the utility in this scenario because the next character is whitespace so this will never be a valid candidate otherwise (variants and utilities must be touching). Another reason this is valid, is because there wasn't a variant present prior to this candidate.

E.g.:

```tsx
<div
  class={clsx({
    hover:underline: someCondition(),
  })}
></div>
```

This will be considered invalid, if you do want this, you should use quotes.

E.g.:

```tsx
<div
  class={clsx({
    'hover:underline': someCondition(),
  })}
></div>
```

**Overlapping/covered spans:**

Another complexity is that the extracted spans for candidates can and will overlap. Let's take a look at this C# example:

```csharp
public enum StackSpacing
{
  [CssClass("gap-y-4")]
  Small,

  [CssClass("gap-y-6")]
  Medium,

  [CssClass("gap-y-8")]
  Large
}
```

In this scenario, `[CssClass("gap-y-4")]` starts with a `[` so we have a few options here:

1. It is an arbitrary property, e.g.: `[color:red]`
2. It is an arbitrary variant, e.g.: `[@media(pointer:fine)]:`

When running the parsers, both the `VariantMachine` and the `UtilityMachine` will run to completion but end up in a `MachineState::Idle` state.

- This is because it is not a valid variant because it didn't end with a `:`.
- It's also not a valid arbitrary property, because it didn't include a `:` to separate the property from the value.

Looking at the code as a human it's very clear what this is supposed to be, but not from the individual machines perspective.

Obviously we want to extract the `gap-y-*` classes here.

To solve this problem, we will run over an additional slice of the input, starting at the position before the machines started parsing until the position where the machines stopped parsing.

That slice will be this one: `[CssClass("gap-y-6")]` (we already skipped over the whitespace). Now, for every `[` character we see, will start a new `CandidateMachine` right after the `[`'s position and run the machines over that slice. This will now eventually extract the `gap-y-6` class.

The next question is, what if there was a `:` (e.g.: `[CssClass("gap-y-6")]:`), then the `VariantMachine` would complete, but the `UtilityMachine` will not because not exists after it. We will apply the same idea in this case.

Another issue is if we _do_ have actual overlapping ranges. E.g.: `let classes = ['[color:red]'];`. This will extract both the `[color:red]` and `color:red` classes. You have to use your imagination, but the last one has the exact same structure as `hover:flex` (variant + utility).

In this case we will make sure to drop spans that are covered by other spans.

The extracted `Span`s will be valid candidates therefore if the outer most candidate is valid, we can throw away the inner candidate.

```
Position:                       11112222222
                                67890123456
                                ↓↓↓↓↓↓↓↓↓↓↓

Span { start: 17, end: 25 }  //  color:red
Span { start: 16, end: 26 }  // [color:red]
```

#### Exceptions

**JavaScript keys as candidates:**

We already talked about the `clsx` scenario, but there are a few more exceptions and that has to do with different syntaxes.

**CSS class shorthand in certain templating languages:**

In Pug and Slim, you can have a syntax like this:

```pug
.flex.underline
  div Hello World
```

<details>

<summary>Generated HTML</summary>

```html
<div class="flex underline">
  <div>Hello World</div>
</div>
```

</details>

We have to make sure that in these scenarios the `.` is a valid boundary character. For this, we introduce a pre-processing step to massage the input a little bit to improve the extraction of the data. We have to make sure we don't make the input smaller or longer otherwise the positions might be off.

In this scenario, we could simply replace the `.` with a space. But of course, there are scenarios in these languages where it's not safe to do that.

If you want to use `px-2.5` with this syntax, then you'd write:

```pug
.flex.px-2.5
  div Hello World
```

But that's invalid because that technically means `flex`, `px-2`, and `5` as classes.

You can use this syntax to get around that:

```pug
div(class="px-2.5")
  div Hello World
```

<details>

<summary>Generated HTML</summary>

```html
<div class="px-2.5">
  <div>Hello World</div>
</div>
```

</details>

Which means that we can't simply replace `.` with a space, but have to parse the input. Luckily we only care about strings (and we have a `StringMachine` for that) and ignore replacing `.` inside of strings.

**Ruby's weird string syntax:**

```ruby
%w[flex underline]
```

This is valid syntax and is shorthand for:

```ruby
["flex", "underline"]
```

Luckily this problem is solved by the running the sub-machines after each `[` character.

### Performance

**Testing:**

Each machine has a `test_…_performance` test (that is ignored by default) that allows you to test the throughput of that machine. If you want to run them, you can use the following command:

```sh
cargo test test_variant_machine_performance --release -- --ignored
```

This will run the test in release mode and allows you to run the ignored test.

> [!CAUTION]
> This test **_will_** fail, but it will print some output. E.g.:

```
tailwindcss_oxide::extractor::variant_machine::VariantMachine: Throughput: 737.75 MB/s over 0.02s
tailwindcss_oxide::extractor::variant_machine::VariantMachine:   Duration: 500ns
```

**Readability:**

One thing to note when looking at the code is that it's not always written in the cleanest way but we had to make some sacrifices for performance reasons.

The `input` is of type `&[u8]`, so we are already dealing with bytes. Luckily, Rust has some nice ergonomics to easily write `b'['` instead of `0x5b`.

A concrete example where we had to sacrifice readability is the state machines where we check the `previous`, `current` and `next` character to make decisions. For a named utility one of the rules is that a `.` must be preceded by and followed by a digit. This can be written as:

```rs
match (cursor.prev, cursor.curr, cursor.next) {
  (b'0'..=b'9', b'.', b'0'..=b'9') => { /* … */ }
  _ => { /* … */ }
}
```

But this is not very fast because Rust can't optimize the match statement very well, especially because we are dealing with tuples containing 3 values and each value is a `u8`.

To solve this we use some nesting, once we reach `b'.'` only then will we check for the previous and next characters. We will also early return in most places. If the previous character is not a digit, there is no need to check the next character.

**Classification and jump tables:**

Another optimization we did is to classify the characters into a much smaller `enum` such that Rust _can_ optimize all `match` arms and create some jump tables behind the scenes.

E.g.:

```rs
#[derive(Debug, Clone, Copy, PartialEq)]
enum Class {
    /// ', ", or `
    Quote,

    /// \
    Escape,

    /// Whitespace characters
    Whitespace,

    Other,
}

const CLASS_TABLE: [Class; 256] = {
    let mut table = [Class::Other; 256];

    macro_rules! set {
        ($class:expr, $($byte:expr),+ $(,)?) => {
            $(table[$byte as usize] = $class;)+
        };
    }

    set!(Class::Quote, b'"', b'\'', b'`');
    set!(Class::Escape, b'\\');
    set!(Class::Whitespace, b' ', b'\t', b'\n', b'\r', b'\x0C');

    table
};
```

There are only 4 values in this enum, so Rust can optimize this very well. The `CLASS_TABLE` is generated at compile time and must be exactly 256 elements long to fit all `u8` values.

**Inlining**:

Last but not least, sometimes we use functions to abstract some logic. Luckily Rust will optimize and inline most of the functions automatically. In some scenarios, explicitly adding a `#[inline(always)]` improves performance, sometimes it doesn't improve it at all.

You might notice that in some functions the annotation is added and in some it's not. Every state machine was tested on its own and whenever the performance was better with the annotation, it was added.

### Test Plan

1. Each machine has a dedicated set of tests to try and extract the relevant part for that machine. Most machines don't even check boundary characters or try to extract nested candidates. So keep that in mind when adding new tests. Extracting inside of nested `[…]` is only handled by the outer most `extractor/mod.rs`.
2. The main `extractor/mod.rs` has dedicated tests for recent bug reports related to missing candidates.
3. You can test each machine's performance if you want to.

There is a chance that this new parser is missing candidates even though a lot of tests are added and existing tests have been ported.

To double check, we ran the new extractor on our own projects to make sure we didn't miss anything obvious.

#### Tailwind UI

On Tailwind UI the diff looks like this:

<details>

<summary>diff</summary>

```diff
diff --git a/./main.css b/./pr.css
index d83b0a506..b3dd94a1d 100644
--- a/./main.css
+++ b/./pr.css
@@ -5576,9 +5576,6 @@ @layer utilities {
     --tw-saturate: saturate(0%);
     filter: var(--tw-blur,) var(--tw-brightness,) var(--tw-contrast,) var(--tw-grayscale,) var(--tw-hue-rotate,) var(--tw-invert,) var(--tw-saturate,) var(--tw-sepia,) var(--tw-drop-shadow,);
   }
-  .\!filter {
-    filter: var(--tw-blur,) var(--tw-brightness,) var(--tw-contrast,) var(--tw-grayscale,) var(--tw-hue-rotate,) var(--tw-invert,) var(--tw-saturate,) var(--tw-sepia,) var(--tw-drop-shadow,) !important;
-  }
   .filter {
     filter: var(--tw-blur,) var(--tw-brightness,) var(--tw-contrast,) var(--tw-grayscale,) var(--tw-hue-rotate,) var(--tw-invert,) var(--tw-saturate,) var(--tw-sepia,) var(--tw-drop-shadow,);
   }
```

</details>

The reason `!filter` is gone, is because it was used like this:

```js
getProducts.js
23:          if (!filter) return true
```

And right now `(` and `)` are not considered valid boundary characters for a candidate.

#### Catalyst

On Catalyst, the diff looks like this:

<details>

<summary>diff</summary>

```diff
diff --git a/./main.css b/./pr.css
index 9f8ed129..4aec992e 100644
--- a/./main.css
+++ b/./pr.css
@@ -2105,9 +2105,6 @@
   .outline-transparent {
     outline-color: transparent;
   }
-  .filter {
-    filter: var(--tw-blur,) var(--tw-brightness,) var(--tw-contrast,) var(--tw-grayscale,) var(--tw-hue-rotate,) var(--tw-invert,) var(--tw-saturate,) var(--tw-sepia,) var(--tw-drop-shadow,);
-  }
   .backdrop-blur-\[6px\] {
     --tw-backdrop-blur: blur(6px);
     -webkit-backdrop-filter: var(--tw-backdrop-blur,) var(--tw-backdrop-brightness,) var(--tw-backdrop-contrast,) var(--tw-backdrop-grayscale,) var(--tw-backdrop-hue-rotate,) var(--tw-backdrop-invert,) var(--tw-backdrop-opacity,) var(--tw-backdrop-saturate,) var(--tw-backdrop-sepia,);
@@ -7141,46 +7138,6 @@
   inherits: false;
   initial-value: solid;
 }
-@property --tw-blur {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-brightness {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-contrast {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-grayscale {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-hue-rotate {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-invert {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-opacity {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-saturate {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-sepia {
-  syntax: "*";
-  inherits: false;
-}
-@property --tw-drop-shadow {
-  syntax: "*";
-  inherits: false;
-}
 @property --tw-backdrop-blur {
   syntax: "*";
   inherits: false;
```

</details>

The reason for this is that `filter` was only used as a function call:

```tsx
src/app/docs/Code.tsx
31:    .filter((x) => x !== null)
```

This was tested on all templates and they all remove a very small amount of classes that aren't used.

The script to test this looks like this:

```sh
bun --bun ~/github.com/tailwindlabs/tailwindcss/packages/@tailwindcss-cli/src/index.t -- -i ./src/styles/tailwind.css -o pr.css
bun --bun ~/github.com/tailwindlabs/tailwindcss--main/packages/@tailwindcss-cli/src/index.t -- -i ./src/styles/tailwind.css -o main.css

git diff --no-index --patch ./{main,pr}.css
```

This is using git worktrees, so the `pr` branch lives in a `tailwindcss` folder, and the `main` branch lives in a `tailwindcss--main` folder.

---

### Fixes:

- Fixes: https://github.com/tailwindlabs/tailwindcss/issues/15616
- Fixes: https://github.com/tailwindlabs/tailwindcss/issues/16750
- Fixes: https://github.com/tailwindlabs/tailwindcss/issues/16790
- Fixes: https://github.com/tailwindlabs/tailwindcss/issues/16801
- Fixes: https://github.com/tailwindlabs/tailwindcss/issues/16880 (due to validating the arbitrary property)

---

### Ideas for in the future

1. Right now each machine takes in a `Cursor` object. One potential improvement we can make is to rely on the `input` on its own instead of going via the wrapping `Cursor` object.
2. If you take a look at the AST, you'll notice that utilities and variants have a "root", these are basically prefixes of each available utility and/or variant. We can use this information to filter out candidates and bail out early if we know that a certain candidate will never produce a valid class.
3. Passthrough the `prefix` information. Everything that doesn't start with `tw:` can be skipped.

### Design decisions that didn't make it

Once you reach this part, you can stop reading if you want to, but this is more like a brain dump of the things we tried and didn't work out. Wanted to include them as a reference in case we want to look back at this issue and know _why_ certain things are implemented the way they are.

#### One character at a time

In an earlier implementation, the state machines were pure state machines where the `next()` function was called on every single character of the input. This had a lot of overhead because for every character we had to:

1. Ask the `CandidateMachine` which state it was in.
2. Check the `cursor.curr` (and potentially the `cursor.prev` and `cursor.next`) character.
3. If we were in a state where a nested state machine was running, we had to check its current state as well and so on.
4. Once we did all of that we could go to the next character.

In this approach, the `MachineState` looked like this instead:

```rs
enum MachineState {
  Idle,
  Parsing,
  Done(Span)
}
```

This had its own set of problems because now it's very hard to know whether we are done or not.

```html
<div class="hover:flex"></div>
<!--            ^          -->
```

Let's look at the current position in the example above. At this point, it's both a valid variant and valid utility, so there was a lot of additional state we had to track to know whether we were done or not.

#### `Span` stitching

Another approach we tried was to just collect all valid variants and utilities and throw them in a big `Vec<Span>`. This reduced the amount of additional state to track and we could track a span the moment we saw a `MachineState::Done(span)`.

The next thing we had to do was to make sure that:

1. Covered spans were removed. We still do this part in the current implementation.
2. Combine all touching variant spans (where `span_a.end + 1 == span_b.start`).
3. For every combined variant span, find a corresponding utility span.
   - If there is no utility span, the candidate is invalid.
   - If there are multiple candidate spans (this is in theory not possible because we dropped covered spans)
   - If there is a candidate _but_ it is attached to another set of spans, then the candidate is invalid. E.g.: `flex!block`
4. All left-over utility spans are candidates without variants.

This approach was slow, and still a bit hard to reason about.

#### Matching on tuples

While matching against the `prev`, `curr` and `next` characters was very readable and easy to reason about. It was not very fast. Unfortunately had to abandon this approach in favor of a more optimized approach.

In a perfect world, we would still write it this way, but have some compile time macro that would optimize this for us.

#### Matching against `b'…'` instead of classification and jump tables

Similar to the previous point, while this is better for readability, it's not fast enough. The jump tables are much faster.

Luckily for us, each machine has it's own set of rules and context, so it's much easier to reason about a single problem and optimize a single machine.

[^candidate]: A candidate is what a potential Tailwind CSS class _could_ be. It's a candidate because at this stage we don't know if it will actually produce something but it looks like it could be a valid class. E.g.: `hover:bg-red-500` is a candidate, but it will only produce something if `--color-red-500` is defined in your theme.


crates/oxide/src/lib.rs
```diff
@@ -40,9 +44,9 @@ fn init_tracing() {
 }
 
 #[derive(Debug, Clone)]
-pub struct ChangedContent {
-    pub file: Option<PathBuf>,
-    pub content: Option<String>,
+pub enum ChangedContent<'a> {
+    File(PathBuf, Cow<'a, str>),
+    Content(String, Cow<'a, str>),
 }
```
RobinMalfait, line 50: Updated the `ChangedContent` internally, we leaked the public implementation that was necessary to map values to the Node world (via `napi-rs`), but internally we can use a proper enum.

One reason why this is required is that we need access to the `extension`. This was only available by looking at the `file`, but we didn't have one for `content`.

crates/oxide/src/lib.rs
```diff
@@ -427,35 +451,59 @@ impl Scanner {
...
+                return None;
+            }
+        },
 
-    let Ok(content) = std::fs::read(&file).map_err(|e| {
-        event!(tracing::Level::ERROR, "Failed to read file: {:?}", e);
-        e
-    }) else {
-        return Default::default();
+        ChangedContent::Content(contents, extension) => (contents.into_bytes(), extension),
```
RobinMalfait, line 463: Refactored this such that we can have access to the content _and_ the extension. 
Before this we weren't applying the preprocessing if the `ChangedContent` was just a `Content` instead of a file.

Now both versions will be preprocessed.

crates/oxide/src/lib.rs
```diff
@@ -427,35 +451,59 @@ impl Scanner {
...
+
+                    // Consume everything else
+                    _ => {}
+                };
+
+                cursor.advance();
+            }
+
+            new_content.into()
+        }
```
RobinMalfait, line 499: Didn't want to encode this logic in the state machines themselves. Because state machines run to completion which might include `.` characters already.

Preprocessing this also means that we don't pay an additional cost _if_ you are dealing with other types of extensions.

crates/oxide/src/lib.rs
```diff
@@ -135,18 +140,39 @@ impl Scanner {
...
 
         let content = read_changed_content(changed_content).unwrap_or_default();
-        let extractor = Extractor::with_positions(&content[..], Default::default());
+        let original_content = &content;
+
+        // Workaround for legacy upgrades:
+        //
+        // `-[]` won't parse in the new parser (`[…]` must contain _something_), but we do need it
+        // for people using `group-[]` (which we will later replace with `in-[.group]` instead).
+        let content = content.replace("-[]", "XYZ");
```
RobinMalfait, line 149: This is definitely a hack, but it's sort of required for the `@tailwindcss/upgrade` tool to allow us to migrate `group-[]:flex` to `in-[.group]:flex`. The new parser won't even find `group-[]` because there must be _something_ inside.

Allowing empty `[]` in the parser just for this reason seemed a bit silly so opted to apply the workaround here.

crates/oxide/src/lib.rs
```diff
@@ -477,21 +525,78 @@ fn read_all_files(changed_content: Vec<ChangedContent>) -> Vec<Vec<u8>> {
 fn parse_all_blobs(blobs: Vec<Vec<u8>>) -> Vec<String> {
     let mut result: Vec<_> = blobs
         .par_iter()
-        .flat_map(|blob| blob.par_split(|x| matches!(x, b'\n')))
-        .map(|blob| Extractor::unique(blob, Default::default()))
+        .flat_map(|blob| blob.par_split(|x| *x == b'\n'))
+        .filter_map(|blob| {
+            if blob.is_empty() {
+                return None;
+            }
```
RobinMalfait, line 532: This happens quite a lot which allows us to not create a `Extractor` at all.

crates/oxide/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    // _run_machine::<tailwindcss_oxide::extractor::arbitrary_value_machine::ArbitraryValueMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::arbitrary_variable_machine::ArbitraryVariableMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::candidate_machine::CandidateMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::css_variable_machine::CssVariableMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::modifier_machine::ModifierMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::named_utility_machine::NamedUtilityMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::named_variant_machine::NamedVariantMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::string_machine::StringMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::utility_machine::UtilityMachine>(input)
+    // _run_machine::<tailwindcss_oxide::extractor::variant_machine::VariantMachine>(input)
```
RobinMalfait, line 46: We can delete these if we want, but it allowed me to test an individual machine in isolation so I could look at Instruments to figure out bottlenecks.

We don't even publish this anywhere so I figured to just keep it here for future use.

crates/oxide/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+fn main() {
+    let iterations = 10_000;
+    let input = include_bytes!("./fixtures/example.html");
+
+    let throughput = Throughput::compute(iterations, input.len(), || {
+        _ = black_box(
+            input
+                .split(|x| *x == b'\n')
```
RobinMalfait, line 58: This mimics how we do the real parsing on a line-by-line basis, but without the parallelism from Rayon. Including rayon here makes it much harder to reason about when you look at Instruments.

crates/oxide/src/parser.rs
```diff
@@ -1,1757 +0,0 @@
-use crate::{cursor::Cursor, fast_skip::fast_skip};
```
RobinMalfait, line 1: I dropped this, but maybe we can keep it around somewhere for additional benchmarks?

crates/oxide/src/extractor/mod.rs
```diff
@@ -0,0 +1,902 @@
...
+        let len = self.cursor.input.len();
+
+        // CSS Variable extractor
+        {
+            let cursor = &mut self.cursor.clone();
+            while cursor.pos < len {
+                if cursor.curr.is_ascii_whitespace() {
+                    cursor.advance();
+                    continue;
+                }
```
RobinMalfait, line 90: Maybe I could use the `fast_skip` here. Didn't even think of it.

crates/oxide/src/extractor/mod.rs
```diff
@@ -0,0 +1,902 @@
...
+            ),
+            // Not a valid arbitrary property or variant, array syntax
+            // Extract inner candidates
+            (
+                r#"["flex",["italic",["underline"]]]"#,
+                vec!["flex", "italic", "underline"],
+            ),
+            // Not a valid arbitrary variant (not followed by a candidate)
+            // Inner classes `is`, `italic` and `underline` are not valid in this context
+            (r#"[:is(italic):is(underline)]"#, vec![]),
```
RobinMalfait, line 696: If `(` and `)` were considered valid boundary characters then this should extract `is`, `italic` and `underline` but don't think that's necessary 🤔

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.80.1"
+channel = "1.85.0"
```
RobinMalfait, line 2: Bumped Rust 😎

crates/oxide/src/extractor/mod.rs
thecrypticace: Yeah, possibly! I say we do this in a followup PR tho

rust-toolchain.toml
thecrypticace: Seems fine to me

crates/oxide/src/extractor/mod.rs
thecrypticace: I think this is okay. I can't think of a situation where `(…)` marks a class list. We can always update things if someone opens an issue about it.

crates/oxide/src/lib.rs
thecrypticace: Yeah, I like this implementation. I'm cool with having some languages pay a small up-front cost to keep everything else fast and maintainable.

crates/oxide/src/extractor/css_variable_machine.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    set!(Class::Dash, b'-');
+    set!(Class::Escape, b'\\');
+    set!(Class::Whitespace, b' ', b'\t', b'\n', b'\r', b'\x0C');
+
+    set!(Class::AllowedCharacter, b'_');
+    set_range!(Class::AllowedCharacter, b'a'..=b'z');
+    set_range!(Class::AllowedCharacter, b'A'..=b'Z');
+    set_range!(Class::AllowedCharacter, b'0'..=b'9');
+
```
philipp-spiess, line 137: Might need to include extended unicode ranges here too: https://play.tailwindcss.com/8nUkRqhgvb

philipp-spiess: Okay I just checked and this does work v4 right now: https://play.tailwindcss.com/H7MlnsHqpy 😅

crates/oxide/src/extractor/mod.rs
RobinMalfait: Only thing I can think of is Ruby, where you can also do `%w(flex underline)`, but if we want to support that I feel like it should be a preprocessing step as well.

crates/oxide/src/extractor/arbitrary_value_machine.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    /// Parsing an arbitrary value
+    Parsing,
+}
+
+impl Machine for ArbitraryValueMachine {
+    #[inline(always)]
+    fn reset(&mut self) {
+        self.start_pos = 0;
+        self.state = State::Idle;
```
philipp-spiess, line 44: Don't think we need this state, we can do what the previous machines do and just return MachineState::Idle unless the cursor is an `OpenBracket` and then we already know we're in `Parsing` state and we can continue to consume the cursor with `while cursor.pos < len`

philipp-spiess (commented):
This is a partial review, going to finish this later. Really love reviewing the individual machines so far. One high level thought I have is about the `MachineState` type as it's not really holding any _state_ information right now anymore (as we moved most of the state onto the stack into function closures which I really like actually). Do with that information what you want so far, I need to besser formalize my thoughts by tomorrow :P 

Memo to myself: continue here https://github.com/tailwindlabs/tailwindcss/pull/16306/commits/38748b49b044acf2efb95a891c0b03418d7ddaf3

crates/oxide/src/extractor/mod.rs
thecrypticace: oh… I didn't realize that worked. Yeah I agree, make that a pre-processing step.

crates/oxide/src/extractor/string_machine.rs
```diff
@@ -0,0 +1,151 @@
...
+
+            cursor.advance()
+        }
+
+        MachineState::Idle
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum Class {
```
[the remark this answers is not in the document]
RobinMalfait: @thecrypticace had examples like this, but the idea is that you classify 0 or more "tokens" into a single class. So for example `, ', and " are 3 separate tokens, but will be classified into a single `Class::Quote`.

I don't think Token is conceptually correct here. If we want to go with the name `Token` we could call this a `TokenGroup` (which is a bit long though).

crates/oxide/src/extractor/string_machine.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    macro_rules! set {
+        ($class:expr, $($byte:expr),+ $(,)?) => {
+            $(table[$byte as usize] = $class;)+
+        };
+    }
+
+    set!(Class::Quote, b'"', b'\'', b'`');
+    set!(Class::Escape, b'\\');
+    set!(Class::Whitespace, b' ', b'\t', b'\n', b'\r', b'\x0C');
```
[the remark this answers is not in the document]
RobinMalfait: Haha yep, I just copied it over and over again. In some situations it makes more sense especially when using ranges. E.g.: `set_range!(Class::Alpha, b'a'..=b'z')`.

crates/oxide/src/extractor/string_machine.rs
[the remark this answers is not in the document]
philipp-spiess: Fiiine, I see your point. I don't feel strongly about it it was just my naive non-Rust-expert mind that thought this has a deeper meaning (like class in other languages often does lol)

crates/oxide/src/extractor/css_variable_machine.rs
```diff
@@ -0,0 +1,230 @@
...
+        if CLASS_TABLE[cursor.curr as usize] != Class::Dash
+            || CLASS_TABLE[cursor.next as usize] != Class::Dash
+        {
+            return MachineState::Idle;
+        }
+
+        let start_pos = cursor.pos;
+        let len = cursor.input.len();
+
+        cursor.advance();
```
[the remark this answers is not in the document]
RobinMalfait: You are right, we should be skipping 2 steps indeed. Didn't notice it because `-` is a valid character in the CSS variable.

However, here are some fun numbers lol

Current with `cursor.advance()`:
```
CssVariableMachine: Throughput: 1.38 GB/s
```

After with `cursor.advance_by(2)`:
```
CssVariableMachine: Throughput: 1.34 GB/s
```

But I think this has to do with how `advance_by` is implemented. Calling `cursor.advance()` twice results in:
```
CssVariableMachine: Throughput: 1.36 GB/s
```

And if I add a dedicated `cursor.advance_twice()`:
```
CssVariableMachine: Throughput: 1.39 GB/s
```
Implementation looks like this:
```rs
#[inline(always)]
pub fn advance_twice(&mut self) {
    self.pos += 2;

    self.prev = self.next;
    self.curr = *self.input.get(self.pos).unwrap_or(&0x00u8);
    self.next = *self
        .input
        .get(self.pos.saturating_add(1))
        .unwrap_or(&0x00u8);
}
```

crates/oxide/src/extractor/css_variable_machine.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        while cursor.pos < len {
+            match CLASS_TABLE[cursor.curr as usize] {
+                // https://drafts.csswg.org/css-syntax-3/#ident-token-diagram
+                //
+                Class::AllowedCharacter | Class::Dash => match CLASS_TABLE[cursor.next as usize] {
```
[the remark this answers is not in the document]
RobinMalfait: It does, but the second time it will also look at its next character.

```
--my-variable#
           ^   This is valid
            ^  Because this is valid
```
```
--my-variable#
            ^   This is valid
             ^  But this is not, so now we should stop, and can't continue
                if we immediately skipped 2 spots then we could be in a weird state.
```

Maybe we can model this in a better way though, but since this already runs at 1.37 GB/s I didn't really try to optimize this particular machine anymore.

crates/oxide/src/extractor/css_variable_machine.rs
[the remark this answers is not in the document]
RobinMalfait: There is another spot in this file where we use `advance_by(2)`, replacing that with `advance_twice()` as well results in:
```
CssVariableMachine: Throughput: 1.41 GB/s
```
😎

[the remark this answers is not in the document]
RobinMalfait: Applying this to other machines:
```diff
- ArbitraryValueMachine: Throughput: 652.71 MB/s
+ ArbitraryValueMachine: Throughput: 654.80 MB/s
```
```diff
- ArbitraryPropertyMachine: Throughput: 792.56 MB/s
+ ArbitraryPropertyMachine: Throughput: 802.05 MB/s
```
```diff
- ArbitraryVariableMachine: Throughput: 744.05 MB/s
+ ArbitraryVariableMachine: Throughput: 744.69 MB/s
```

So definitely not nothing. For some not a big deal, for others pretty interesting results.

crates/oxide/src/extractor/arbitrary_value_machine.rs
RobinMalfait: Yep you are right! Some left-overs from the previous implementation(**s**).

```diff
- ArbitraryValueMachine: Throughput: 654.80 MB/s
+ ArbitraryValueMachine: Throughput: 718.51 MB/s
```

Sweet!

crates/oxide/src/extractor/css_variable_machine.rs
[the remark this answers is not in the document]
thecrypticace: I was looking at cursor and I was really wanting to get rid of move_to, rewind_by, and advance_by. If you wanna find a way to delete those I won't be opposed. 😄

crates/oxide/src/extractor/css_variable_machine.rs
RobinMalfait: https://github.com/tailwindlabs/tailwindcss/commit/21fe01922248c6e5d8cef15de35559aae8d3ccd0 😎

crates/oxide/src/extractor/css_variable_machine.rs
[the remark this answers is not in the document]
philipp-spiess: Haha wow I was just thinking literally duplicating this line 🙈 

```rs
cursor.advance();
cursor.advance();
```

crates/oxide/src/extractor/css_variable_machine.rs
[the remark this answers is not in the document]
philipp-spiess: Fair! 👍

crates/oxide/src/extractor/candidate_machine.rs
```diff
@@ -0,0 +1,378 @@
...
+        b'\t' | b'\n' | b'\x0C' | b'\r' | b' ' | b'"' | b'\'' | b'`' | b'\0'
+    )
+}
+
+/// A candidate must be preceded by any of these characters.
+/// E.g.: `%w[flex]`
+///          ^
+#[inline(always)]
+fn is_valid_before_boundary(c: &u8) -> bool {
+    is_valid_common_boundary(c) || matches!(c, b'{' | b'[' | b'.')
```
philipp-spiess, line 197: This should probably include `(` for `%w(…)` Ruby syntax

crates/oxide/src/extractor/candidate_machine.rs
```diff
@@ -0,0 +1,378 @@
...
+                // Inside an object (no spaces, key)
+                (r#"let classes = {'{}':true};"#, vec!["let", "classes"]),
+                // Inside an object (value)
+                (
+                    r#"let classes = { primary: '{}' };"#,
+                    vec!["let", "classes", "primary"],
+                ),
+                // Inside an object (no spaces, value)
+                (r#"let classes = {primary:'{}'};"#, vec!["let", "classes"]),
+            ] {
```
philipp-spiess, line 347: let's add examples here for `%w(…)` and `%w[…]` Ruby wrappers too, wdyt?

crates/oxide/src/parser.rs
philipp-spiess, line 1: wont be missed

crates/oxide/src/lib.rs
philipp-spiess: Hm I wonder if this should be a flag though, otherwise we now need to sequentially scan every file we ever scan for a potential `-[]` when we only need that in the upgrade tool. Does this effect your benchmarks in any meaningful way?

crates/oxide/src/lib.rs
philipp-spiess: Yep that's definitely the way to go for these

crates/oxide/src/extractor/css_variable_machine.rs
```diff
@@ -135,6 +135,13 @@ const CLASS_TABLE: [Class; 256] = {
     set_range!(Class::AllowedCharacter, b'A'..=b'Z');
     set_range!(Class::AllowedCharacter, b'0'..=b'9');
 
+    // non-ASCII (such as Emoji): https://drafts.csswg.org/css-syntax-3/#non-ascii-ident-code-point
+    let mut i = 0x80;
+    while i <= 0xff {
+        table[i as usize] = Class::AllowedCharacter;
+        i += 1;
+    }
```
philipp-spiess, line 143: Can this also use the `set_range!` macro? 

```rs
    set_range!(Class:: AllowedCharacter, 0x80..=0xff)
```

crates/oxide/src/cursor.rs
```diff
@@ -41,10 +41,6 @@ impl<'a> Cursor<'a> {
         cursor
     }
 
-    pub fn rewind_by(&mut self, amount: usize) {
```
philipp-spiess, line 44: NICE!

philipp-spiess (commented):
Awesome stuff. One thing to consider is if we want to ship this as-is (which might include some bugs we need to fix up fast afterwards) or if we want to either have an opt-in or opt-out based approach. I'm curious what your thoughts are about this?

crates/oxide/src/lib.rs
RobinMalfait: This is code that only lives in the `get_candidates_with_positions` code which is only ever used in the upgrade tool not in any other hot path when extracting candidates during normal builds.

philipp-spiess: Oh gotcha I missed this, awesome!

crates/oxide/src/extractor/arbitrary_variable_machine.rs
```diff
@@ -0,0 +1,356 @@
...
+            // Arbitrary variables must be valid CSS variables
+            (r"(--my-\ color)", vec![]),
+            (r"(--my#color)", vec![]),
+            // Fallbacks cannot have spaces
+            (r"(--my-color, red)", vec![]),
+            // Fallbacks cannot have escaped spaces
+            (r"(--my-color,\ red)", vec![]),
+            // Variables must have at least one character after the `--`
+            (r"(--)", vec![]),
+            (r"(--,red)", vec![]),
```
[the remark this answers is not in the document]
RobinMalfait: Added it, but was already covered. Once we see a dash we go straight to the CSS variable machine which requires 2 dashes.

crates/oxide/src/extractor/arbitrary_property_machine.rs
```diff
@@ -0,0 +1,427 @@
...
+
+            State::ParsingProperty => {
+                while cursor.pos < len {
+                    match CLASS_TABLE[cursor.curr as usize] {
+                        Class::Dash => match CLASS_TABLE[cursor.next as usize] {
+                            // Start of a CSS variable
+                            //
+                            // E.g.: `[--my-color:red]`
+                            //         ^^
+                            Class::Dash => return self.parse_property_variable(cursor),
```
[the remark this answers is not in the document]
RobinMalfait: The CSS Variable machine _does_ require 2 dashes. In this spot we check the for the current and next values to be dashes and then we go ask the CSS variable machine to complete.

One important note here is that we do not advance the cursor so the CSS variable machine will still see the `--` (which technically means that we read the cursor.curr and cursor.next variables twice and look em up in a CLASS_TABLE but those are different class tables.)

crates/oxide/src/extractor/named_utility_machine.rs
```diff
@@ -0,0 +1,545 @@
...
+                Class::AlphaLower => match CLASS_TABLE[cursor.next as usize] {
+                    // Valid single character utility in between quotes
+                    //
+                    // E.g.: `<div class="a"></div>`
+                    //                    ^
+                    // E.g.: `<div class="a "></div>`
+                    //                    ^
+                    // E.g.: `<div class=" a"></div>`
+                    //                     ^
+                    Class::Whitespace | Class::Quote | Class::End => self.done(cursor.pos, cursor),
```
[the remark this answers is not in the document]
RobinMalfait: So in this case I think whitespaces and quotes (and end of the input) are valid because they represent 99% of the programming languages.

Now that we added a pre-processor for Ruby, this will be translated to `let a = %w a ` and therefore will be extracted properly 👍

crates/oxide/src/extractor/named_utility_machine.rs
```diff
@@ -0,0 +1,545 @@
...
+                    self.next(cursor)
+                }
+
+                // Valid start of a negative utility, if followed by another set of valid
+                // characters. `@` as a second character is invalid.
+                //
+                // E.g.: `-mx-2.5`
+                //        ^^
+                Class::Dash => match CLASS_TABLE[cursor.next as usize] {
+                    Class::AlphaLower | Class::AlphaUpper => {
```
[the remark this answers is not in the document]
RobinMalfait: Ah I agree, the `Class::AlphaUpper` can be dropped.

I don't think we should add `Class::At` because you can't create a `@utility` that starts with an `@` in userland. But we still want `@` for container queries. I'm not sure if `-@container` even makes sense in this case 🤔

crates/oxide/src/extractor/named_utility_machine.rs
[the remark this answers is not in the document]
philipp-spiess: Nice change!

crates/oxide/src/extractor/named_utility_machine.rs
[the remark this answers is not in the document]
RobinMalfait: So `@` is a bit of a reserved symbol. E.g.: https://play.tailwindcss.com/FEa1ysdyFY?file=css

crates/oxide/src/extractor/named_utility_machine.rs
```diff
@@ -0,0 +1,545 @@
...
+                        }
+
+                        Class::Dash => match CLASS_TABLE[cursor.next as usize] {
+                            // Start of an arbitrary value
+                            //
+                            // E.g.: `bg-[#0088cc]`
+                            //          ^^
+                            Class::OpenBracket => {
+                                cursor.advance();
+                                return match self.arbitrary_value_machine.next(cursor) {
```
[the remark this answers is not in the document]
RobinMalfait: Yeah love that too!

crates/oxide/src/extractor/candidate_machine.rs
RobinMalfait: Not necessary anymore because of the pre-processing step. 🥳

crates/oxide/src/extractor/css_variable_machine.rs
RobinMalfait: Ah yep! I was dealing with `\u{00B7}` values prior to this which could have different types (depending on the value you put in). But this can indeed be refactorred 👍

crates/oxide/src/extractor/named_utility_machine.rs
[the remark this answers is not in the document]
philipp-spiess: I think not for the JS API I'm afraid, because of: https://github.com/tailwindlabs/tailwindcss-container-queries

crates/oxide/src/extractor/string_machine.rs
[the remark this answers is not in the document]
RobinMalfait: Improved the DX here in a separate PR: https://github.com/tailwindlabs/tailwindcss/pull/16864

philipp-spiess (approved):

joshhanley:
@RobinMalfait hey we are having an issue with Livewire Flux since Tailwind v4.0.10 and I suspect it might be related to this PR. I tried compiling Tailwind locally to see if I could bisect exactly what PR/ commit is causing it, but I'm not smart enough to get it all running in a local repo 😆 so this PR is my best guess at what's causing the issue.

The problem we are having is we have a blade component for the sidebar `<flux:sidebar />` and it looks something like this (simplified)

```blade
@php
if ($sidebarIsStashable) {
    $attributes = $attributes->merge([
        'x-init' => '$el.classList.add(\'-translate-x-full\'); $el.classList.add(\'transition-transform\')',
    ]);
}
@endphp

<div x-data {{ $attributes->class('border-r w-64 p-4 min-h-dvh max-h-dvh top-0 fixed left-0') }}>
    {{ $slot }}
</div>
```

Since the changes in Tailwind v4.0.10, the styles `-translate-x-full` and `transition-transform` which are inside the conditionally added `x-init` are not being compiled by Tailwind and our sidebar is not being hidden.

If I add those classes in another file though, they get compiles correctly and our sidebar is being hidden as expected.

I also experimented removing the backslashes from before the single quotes and everything works and compiles as expected.

```php
'x-init' => "\$el.classList.add('-translate-x-full'); \$el.classList.add('transition-transform')",
```

Would you have any thoughts as to what might be causing that? It's causing issues because the new Laravel Livewire starter kit is using Flux and people's sidebars aren't hiding 😅

Thanks in advance!

RobinMalfait:
Hey @joshhanley, created a new issue for this here https://github.com/tailwindlabs/tailwindcss/issues/17023 so it's easier to track and provide a PR for. Thanks!

joshhanley:
@RobinMalfait no worries! Thanks for that

PR #16307, opened 2025-02-06
Remove useless `console.log` in `@tailwind/vite` package

martinschneider01:
This PR removes a (seemingly) unwanted `console.log` in the `@tailwind/vite` package.

I discovered this when I updated tailwind in my `astro` project, each `.astro` component log the full path of an imported `.css` file.

It seems that this was the commit that introduced it: https://github.com/tailwindlabs/tailwindcss/commit/e1a85ac260e065c069307deb7f908bc46035a3e5


RobinMalfait (approved):
Oops, good catch. Thanks!

divine:
> So I'm curious does anyone here even use TailwindCSS with Vite?

Well, a lot of people are using but console logs are just tiny iceberg IMO. See this for example https://github.com/nuxt/ui/issues/3275 :)

Thanks!

PR #16382, opened 2025-02-09
Bump and pin `prettier`

RobinMalfait:
This PR bumps the Prettier dependencies, and also pins the version.

Noticed that a PR with a single empty commit started failing at the time of writing this (https://github.com/tailwindlabs/tailwindcss/pull/16306). This is because prettier released a new minor version which results in slightly different output.

Let's bump prettier and handle the differences, but also pin the version to avoid this in the future.


philipp-spiess (approved):
Yes chef

PR #16414, opened 2025-02-10
Fix sorting numeric utilities when they have different magnitudes

thecrypticace:
When sorting utilities it was possible for a utility like `duration-700` to come before `duration-75` or `duration-1000` to come before `duration-150`.

This PR fixes this behavior by reading the full "magnitude" of each numbers before we compare them.


RobinMalfait (approved):
🤦‍♂️

Issue #16389, opened 2025-02-09
Inconsistent CSS assets between client and SSR builds with @tailwindcss/vite


mkdynamic:
**What version of Tailwind CSS are you using?**

4.0.5

**What build tool (or framework if it abstracts the build tool) are you using?**

Vite 6.1.0

**What version of Node.js are you using?**

Bun 1.2.2

**What browser are you using?**

N/A

**What operating system are you using?**

macOS

**Reproduction URL**

- Using the `@tailwindcss/vite` plugin: https://github.com/mkdynamic/vite-tailwind-ssr
- Using the `@tailwindcss/postcss` plugin: https://github.com/mkdynamic/vite-tailwind-postcss-ssr

**Describe your issue**

When using `@tailwindcss/vite`, and creating an SSR build with Vite, the CSS asset paths differ between the SSR and client builds. This is not true with `@tailwindcss/postcss`.

The above repos contain a minimal reproduction case, starting from a fresh React Vite project. Here are the reproduction steps below showing the bug:

With the `@tailwindcss/vite` plugin:

```bash
$ bun build:client
$ vite build --outDir build/client --manifest
vite v6.1.0 building for production...
✓ 29 modules transformed.
build/client/.vite/manifest.json         0.38 kB │ gzip:  0.21 kB
build/client/index.html                  0.39 kB │ gzip:  0.27 kB
build/client/assets/App-DWNkNxeI.css     6.10 kB │ gzip:  1.79 kB
build/client/assets/index-DO9bOB2u.js  185.88 kB │ gzip: 58.73 kB
✓ built in 463ms

$ bun build:server
$ vite build --outDir build/server --ssr=src/entry-ssr.tsx
vite v6.1.0 building SSR bundle for production...
✓ 4 modules transformed.
build/server/entry-ssr.js  0.46 kB
✓ built in 55ms

$ grep App- build/server/entry-ssr.js
const stylesheet = "/assets/App-DX11wV_z.css";
```

Note the asset path of the built `App.css` differs between the SSR and client builds (i.e. `App-BYq-i2yi.css !== App-DWNkNxeI.css`). This means that SSR rendered responses have the incorrect asset URL (which leads to a hydration error, if you are hydrating – or a missing CSS asset if you are only SSR rendering).

With the `@tailwindcss/postcss` based plugin:

```bash
$ bun build:client
$ vite build --outDir build/client --manifest
vite v6.1.0 building for production...
✓ 29 modules transformed.
[...]

Issue #16252, opened 2025-02-04
All the stylesheets CSS file loading twice on build


khanakia:
**What version of Tailwind CSS are you using?**

v4.0.2

**What build tool (or framework if it abstracts the build tool) are you using?**

react-router framework 7.1.3

**What version of Node.js are you using?**

v22.12.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

- Clone the Repo https://github.com/khanakia/reactrouter_css_loading_twice
- `pnpm build`
- Run `http-server-spa build/client index.html 8090` to preview the build

**Describe your issue**

Issue is described here 

https://github.com/remix-run/react-router/issues/12940#event-16182494636

philipp-spiess:
Hey! Thanks for the repro.

I looked into this from the Tailwind CSS extension side. What's happening is that React Router runs two separate builds, one for the server and one for the client. The Tailwind CSS extension progressively scans your components for class names and by the time the first build is finished (client build in this case), we haven't scanned through the server modules yet. This causes the stylesheet to become larger in the server build later which causes Vite or React Router to inline both CSS files since they're different... 

A repro for this that doesn't use Tailwind CSS btw is this minimal Vite plugin:

```js
{
  name: "transform-css",
  transform(code, id) {
    if (id.includes(".css?transform-only")) {
      return `/*! ${Date.now()} */\n${code}`;
    }
  },
},
```

Let me ask in your original React Router issue as well, trying to understand where to fix this since I'm not sure how Vite and React Router play together in that case.

khanakia:
This also causes the hydration error on react-router client bundle`build/client` 

as same tailwind CSS file `assets/app-Bwb2LP5W.css`  bundled with different name on server

![Image](https://github.com/user-attachments/assets/84c1c342-3866-4773-bcda-cd5f71f7630a)

```js
Uncaught Error: Hydration failed because the server rendered HTML didn't match the client. As a result this tree will be regenerated on the client. This can happen if a SSR-ed Client Component used:

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

https://react.dev/link/hydration-mismatch

  ...
[...]

Issue #15794, opened 2025-01-24
Astro client: directive will cause tailwindcss to be invalid


ckvv:
related https://github.com/withastro/astro/issues/13055

### Astro Info

```block
Astro                    v5.1.9
Node                     v22.11.0
System                   macOS (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/vue
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

**What version of Tailwind CSS are you using?**

4.0

**What build tool (or framework if it abstracts the build tool) are you using?**

astro

**What version of Node.js are you using?**

For example: v22

**Reproduction URL**

https://github.com/ckvv/astro-tailwindcss

**Describe your issue**

client: directive will cause tailwindcss to be invalid, The policy of the dev module is that CSS is not bundled after build.

### What's the expected result?

 use tailwindcss v4 in astro

### Link to Minimal Reproducible Example

https://github.com/ckvv/astro-tailwindcss

### Participation

- [x] I am willing to submit a pull request for this issue.

jamesharding:
I am having the same issue. After a full build of the astro site (using Tailwind 4.0 and Vite), all classes for components that use the `client:only="react"` are missing from the final build file. The react components are in .tsx files that are not gitignored, so I would have expected the classes to be picked up from the className property during a file scan?

Not a problem in dev obviously with all classes present, it is just in the final build.

daugvinasr:
Kinda similar problem with classes used in `.astro` files `<script>` tags as well. They work fine in dev mode, but are missing from the final bundle. 

![Image](https://github.com/user-attachments/assets/d44a84c9-1356-47fe-a2f0-2eb2da0ae572)

CosmicNIO:
Also had the same issue with a react component in my Astro project, with `client:only="react"`.

Was able to solve by adding my components directory to the global.css:

`@source "../components/";`

PhilMataz:
I'm facing this issue in Vue components too. 

Adding `@source "../components/"` to `global.css` fixed it, as suggested by @CosmicNIO. 

I checked the documentation but couldn't find any mention of this approach - is this the intended solution or a temporary fix?

philipp-spiess:
[...]

PR #16425, opened 2025-02-11
Vite: Add option to disable module-graph based scanning and turn it on for Astro

philipp-spiess:
This PR adds an option to the Vite plugin to control wether or not the module-graph based dependency scanner is used. 

The main motivation for doing this is a number of SSR specific build issues where differences in the dependency graphs for SSR and client builds would cause unexpected production builds. Some specific examples:

- #16389
- #16252
- #15794

The new option has a way to force `module-graph` and `file-system` based candidate scanning but will default to an `automatic` mode that will detect Astro and change the default behavior there to immediately fix the most prominent issue (#15794). 

It is configured like this:

```js
import { defineConfig } from 'vite'
import tailwindcss from '@tailwindcss/vite'

export default defineConfig({
  plugins: [tailwindcss({ scanner: 'file-system' })],
})
```

## Test plan

- I updated the Astro integration test to include an example using the `client:only` directive that was able to reproduce the issue #15794
- I also added an example that was able to reproduce the issue with React Router as outlined in #16252
- The existing SSR tests were also able to reproduce the issues form #16389 by having both entry points use different class names.

I ensured that the new option turns-on automatically for Astro and that enabling it manually fixes the other cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - _Experimental_: Add `user-valid` and `user-invalid` variants ([#12370](https://github.com/tailwindlabs/tailwindcss/pull/12370))
+- Vite: Add a new `scanner` option to disable module-graph based scanning ([#16425](https://github.com/tailwindlabs/tailwindcss/pull/16425))
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


justingolden21:
If true that [this would allow tailwind 4 to once again purge unused utility classes](https://stackoverflow.com/a/79438630/4907950), then that would be an ENORMOUS win for developers and the web as a whole!

Thank you for working on this!

philipp-spiess:
We're going with an option-less approach for now since it's less confusing and won't require debugging if things go wrong: #16631

Issue #16287, opened 2025-02-05
[v4] Source glob pattern `./**/*.{js,ts,jsx,tsx,mdx}` does not match Next.js-style catchall route segments like `[...slug]`.


nicksergeant:
**What version of Tailwind CSS are you using?**

v4.0.3

**What build tool (or framework if it abstracts the build tool) are you using?**

- Next.js 15.1.6
- PostCSS 8

**What version of Node.js are you using?**

23.5.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

https://github.com/nicksergeant/tailwind-v4-dynamic-params/

**Describe your issue**

If you are [disabling automatic detection](https://tailwindcss.com/docs/detecting-classes-in-source-files#disabling-automatic-detection) with:

```
@import 'tailwindcss' source(none);
```

and then configuring a glob pattern like [this](https://github.com/nicksergeant/tailwind-v4-dynamic-params/blob/main/app/globals.css#L1-L3):

```
@source './**/*.{js,ts,jsx,tsx,mdx}';
```

Source files in directories such as `[...slug]` will not be picked up: https://github.com/nicksergeant/tailwind-v4-dynamic-params/blob/main/app/%5B...slug%5D/page.tsx

I'm not sure if this is a bug per se, and I have found this workaround:

```
@source './**/*.{js,ts,jsx,tsx,mdx}';
@source './[[]**[]]/*.{js,ts,jsx,tsx,mdx}';
```

Example page showing missing style: https://github.com/nicksergeant/tailwind-v4-dynamic-params/blob/main/app/%5B...slug%5D/page.tsx#L4

But I was surprised that `./**/*` was not working to pick those up.

neoplastic:
Wow. That's the problem I was having, I was wondering why some of the custom classes weren't working in my code.

nicksergeant:
I'm also not sure if my posted workaround above is actually sufficient. For some reason it seems to be working in my isolated test case above, but not in our larger repo using the same versions.

nicksergeant:
> If its not working in your larger repo is there some complexity in the path names to your js/ts/etc files?

Yeah I'm not sure - we actually only have a single catchall route in the whole repo (`[...slug]`). We do have some dynamic routes like `[slug]` but we've only used the catchall once.

philipp-spiess:
[...]

PR #16457, opened 2025-02-12
Scan Next.js dynamic route segments with manual `@source` rules

philipp-spiess:
Part of #16287

## Test plan

Added unit and integration tests


philipp-spiess:
Going to close this as it's an upstream issue. We can reopen if we decide to work around it from our side but if we can fix the upstream resource I'd expect this test to also live in the upstream tests.

RobinMalfait (approved):
Pretty cool that this is almost a drop-in replacement 💪

PR #16507, opened 2025-02-13
Upgrade: Handle `darkMode` value with block syntax

### LINKED ISSUE #16171: [v4] upgrade does not work for dark mode variant with media query

v4.0.3

v23.7.0

Arch Linux

**Reproduction URL**

https://github.com/fiws/tailwind-issue-f1

**Describe your issue**

After running `npx @tailwindcss/upgrade@next`, my "dark" theme config in the old tailwind.config.js was transformed into `@custom-variant dark (@media not print { .dark & });` this does not work and appears to be invalid syntax. See below for details:

### before

https://github.com/fiws/tailwind-issue-f1/blob/676604340b5ad3ef2384ea86ab79d755a82dfb40/tailwind.config.js#L7

```ts
// tailwind.config.js
darkMode: ["variant", "@media not print { .dark & }"],
```

Image

### after

https://github.com/fiws/tailwind-issue-f1/blob/main/src/index.css#L3

```css
@custom-variant dark (@media not print { .dark & });
```

Image

Image

--- 

I also have no idea what to correct syntax is..

philipp-spiess:
Closes #16171

This PR handles `darkMode` variant configs containing braces (so creating sub-rules) the same way we handle it in the interop layer. Since the interop layer runs inside the `addVariant` API that we do not run here, I instead oped to copy the one liner.

## Test plan

Updated one of the migration tests to include a rule that wasn't working before. Ensured the new output works via https://play.tailwindcss.com/nR99uhKtv3

### DIFF
--- a/packages/@tailwindcss-upgrade/src/migrate-js-config.ts
+++ b/packages/@tailwindcss-upgrade/src/migrate-js-config.ts
@@ -196,7 +196,7 @@ async function migrateTheme(
 }
 
 function migrateDarkMode(unresolvedConfig: Config & { darkMode: any }): string {
-  let variant: string = ''
+  let variant: string | string[] = ''
   let addVariant = (_name: string, _variant: string) => (variant = _variant)
   let config = () => unresolvedConfig.darkMode
   darkModePlugin({ config, addVariant })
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-upgrade/src/migrate-js-config.ts
```diff
@@ -196,15 +196,36 @@ async function migrateTheme(
...
+    // Convert to the block syntax if a block is used
+    if (variantName.includes('{')) {
+      variantName = variantName.replace('}', '{ @slot }}')
+      output += `\n@tw-bucket custom-variant {\n@custom-variant dark {${variantName}}}\n`
+    } else {
+      output += `\n@tw-bucket custom-variant {\n@custom-variant dark (${variantName});\n}\n`
+    }
+  }
+
+  return output
```
thecrypticace, line 228: Will this actually work with an array of format strings? `@custom-variant` overrides

philipp-spiess: ok i will be less lazy

thecrypticace (approved):

packages/@tailwindcss-upgrade/src/migrate-js-config.ts
```diff
@@ -196,15 +196,38 @@ async function migrateTheme(
...
+  }
+
+  let customVariant = ''
+  for (let variantName of variant) {
+    // Convert to the block syntax if a block is used
+    if (variantName.includes('{')) {
+      customVariant += variantName.replace('}', '{ @slot }}') + '\n'
+    } else {
+      customVariant += variantName + '{ @slot }\n'
+    }
```
RobinMalfait, line 223: Not 100% if this is enough, can we add a few more test cases:

Nested media queries:
```
darkMode: ['variant', '@media not print { @media screen { .dark & } }'],
```
```css
/* dark:flex */
@media not print {
  @media screen {
    .dark .dark\:flex {
      display: flex;
    }
  }
}
```
Play: https://play.tailwindcss.com/n3k65J6HJF?file=config

RobinMalfait (approved):

PR #16513, opened 2025-02-13
Allow `border-*` to match both color and size namesaces at the same time

### LINKED ISSUE #16343: Generated CSS for border-width is not consistent with v3 when importing a config with `@config` directive

4.0.4

vite 6.1.0, @tailwindcss/vite 4.0.4, @tailwindcss/postcss 4.0.0

v22.11.0

N/A

macOS

**Reproduction URL**

Refer to this [repo](https://github.com/KarthikeyanRanasthala/tailwind-config-debug) in the following branches,

with v3: `tw-v3-with-preset`
with v4 & vite plugin: `tw-v4-with-preset-and-vite-plugin`

**Describe your issue**

Inconsistent output between v3 & v4

The generated css is incorrect when importing a tailwind config using `@config` directive. 

```js
// preset.js
export default {
  theme: {
    colors: {
      default: "#0000ff",
    },
    borderWidth: {
      default: "1px",
    },
  },
};
```

```js
// tailwind.config.js
import preset from './preset'

export default {
  content: [
    "./index.html",
    "./src/**/*.{js,ts,jsx,tsx}",
  ],
  presets: [preset],
}
```

```css
/* output in v3 */
.border-b-default {
    border-bottom-width: 1px
}
```

```css
/* output in v4 */
.border-b-default {
    border-bottom-color: #00f
}
```

For the above example, a custom theme has the same key `default` defined for properties borderWidth and color. Then for a class `border-b-default`, it should've generated `border-bottom-width: 1px` (as it did in v3) but it generates `border-bottom-color: #00f` (in v4)

v4 still seems to respect that explicitly defined properties should have higher priority ie `textColor > colors`, `fill > colors` etc. So it looks like border is the only one having this issue as `border-` is used for multiple properties like border-width, border-style and border-color.

philipp-spiess:
Closes #16343

In v3, if you had a `colors` and `borderWidth` key that resolved to the same value, using the `border-*` utility would actually match create both rules: https://play.tailwindcss.com/Loe4Z1eKky?file=config

In v4, however, it currently only looks at the color and then ends. This PR changes it to be compatible with v3.

My main motivation for fixing this is that this is actually more like the `@utility` API works right now already where this would generate both rules: https://play.tailwindcss.com/bQNe7wXOJb?file=css

```css
@utility border-* {
  color: --value(--color-*);
  border-width: --value(--border-width-*);
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don’t replace `_` in suggested theme keys ([#16433](https://github.com/tailwindlabs/tailwindcss/pull/16433))
 - Ensure `--default-outline-width` can be used to change the `outline-width` value of the `outline` utility
 - Ensure drop shadow utilities don't inherit unexpectedly ([#16471](https://github.com/tailwindlabs/tailwindcss/pull/16471))
+- Ensure `border` utility creates `border-width` and `border-color` declarations if both namespaces have a match ([#16513](https://github.com/tailwindlabs/tailwindcss/pull/16513))
 
 ## [4.0.6] - 2025-02-10
 
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -9945,7 +9945,38 @@ test('border with custom default border width', async () => {
       initial-value: solid;
     }"
   `)
-  expect(await run(['-border', 'border/foo'])).toEqual('')
+})
+
+test(`border where \`--color-*\` and \`--border-width-*\` matches`, async () => {
+  expect(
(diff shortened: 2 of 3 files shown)


philipp-spiess:
I wonder if we should audit all other utilities as well? This should probably be consistent for other things that have more than one overload, e.g. `bg``stroke`, `from` etc...

philipp-spiess:
I'm going to close this issue for now since we aren't sure wether we want to make this change yet. It's unfortunate for backward-compatibility but we also think that it's not super useful in real life and kind of unexpected since you're expecting mose of these utilities to only set one value and if you are overloading the namespaces to have "side effects" (e.g. by having a color named default and a width named default) it's going to cause confusing in the stylesheet. You can still do that with `@utility` if you want to in user space.

PR #16514, opened 2025-02-13
Allow `theme(…)` options when using `@import`

RobinMalfait:
This PR improves the developer experience when trying to use `theme(…)` options on an import.

Today, if you want to use Tailwind CSS, you can import it as:

```css
@import "tailwindcss";
```

But if you want to use any of the `theme(…)` options, like the `static` theme option, then you had to use this instead:

```css

@import 'tailwindcss/theme' layer(theme) theme(static);
```
In this scenario you have to be careful, because the `layer(…)` _must_ be the first option after an import (according to the spec). So if you use `@import 'tailwindcss/theme' theme(static) layer(theme);` then that's not going to work either.

This PR solves that by allowing you to use the `theme(…)` options directly on the `@import` statement:

```css
@import 'tailwindcss' theme(static);
```

The only edge case is when you want to use `theme(reference)`. A typical use case is for projects with `<style>` blocks where you want to _reference_ the CSS variables from the theme. 

If you use `@import 'tailwindcss' theme(reference);`, then all `@theme` blocks will be references and you can reference theme values. This is good. The bad part is that `@import 'tailwindcss';` also includes preflight CSS. This means that we will import the preflight CSS for every `<style>` block. This is probably not what you want.

The solution is to use `@reference 'tailwindcss';` instead which strips all of that information and only gives you access to CSS variables.

This PR also makes sure that if you use `theme(reference)` on an import that we still throw an error and suggest you use `@reference` instead. This is not a breaking change because right now if you use `@import` with `theme(…)` options it will already throw an error.

### Test plan:

1. Added dedicated tests to make sure we don't throw anymore.
2. Added test to make sure that we _do_ throw when using `theme(reference)` on an import.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Vite: Ensure setups with multiple Vite builds work as expected ([#16631](https://github.com/tailwindlabs/tailwindcss/pull/16631))
 - Vite: Ensure Astro production builds contain classes for client-only components ([#16631](https://github.com/tailwindlabs/tailwindcss/pull/16631))
 - Vite: Ensure utility classes are read without escaping special characters ([#16631](https://github.com/tailwindlabs/tailwindcss/pull/16631))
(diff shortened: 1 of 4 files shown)


philipp-spiess (approved):
Noice

PR #16542, opened 2025-02-14
Expose AST scope and parsing methods

sonofmagic:
In the current Tailwind CSS functionality, we may need to perform some static analysis on existing projects, including parsing the project's CSS, obtaining the AST, and performing transformations (similar to PostCSS's API). 

However, the currently exposed API does not provide such capabilities. Therefore, the purpose of this PR is to introduce these changes.

```ts
import { CSS, parseCss, CompileOptions } from 'tailwindcss'

function getContext(css: string, opts: CompileOptions = {}) {
  return parseCss(CSS.parse(css), opts)
}

const ctx = await getContext('@import \'tailwindcss\';')
// do stuff
// ...
```

### DIFF
--- a/packages/tailwindcss/package.json
+++ b/packages/tailwindcss/package.json
@@ -55,6 +55,14 @@
       "require": "./src/plugin.cts",
       "import": "./src/plugin.ts"
     },
+    "./ast": {
+      "require": "./src/ast.cts",
+      "import": "./src/ast.ts"
+    },
+    "./ast.js": {
+      "require": "./src/ast.cts",
+      "import": "./src/ast.ts"
+    },
     "./package.json": "./package.json",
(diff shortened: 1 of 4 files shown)


philipp-spiess:
@sonofmagic Hey! This is internal API that we will probably break in the future so I don't recommend anyone using it and exposing it in a new package invites exactly that use case so I don't think that's something we'd like to expose.

These APIs are also highly coupled towards Tailwind CSS, I suggest using a more generic CSS parser for your libraries like postcss.

What kind of static analysis do you want to do?

philipp-spiess:
Going to close this PR for now but I'd love to hear about your use case for this over in the [discussion you created](https://github.com/tailwindlabs/tailwindcss/discussions/16543).

PR #16579, opened 2025-02-16
minor: better type safety and best practice to escape function

soranoo:
Hello,

This PR aims to improve type safety and adhere to best practices for the selector escape function.
- Replaces `var` with `const` or `let` for better scoping and readability.


philipp-spiess (approved):
Hey! Thanks we don't usually take these kind of contributions since the risk of adding errors outweighs the style changes but it did seem unnecessary to me that we have `escape(…)` duplicated for our integration test setups so I decided to make some adjustments and land this. Thank you!

PR #16620, opened 2025-02-18
Upgrade: Use latest tag for packages

philipp-spiess:


RobinMalfait (approved):

RobinMalfait (commented):
Hah might need to update the tests since they check for an exact `4.0.0` match.

PR #16693, opened 2025-02-20
Fix integration tests for Windows

philipp-spiess:
The Nuxt preview server always starts on port 3000 even if that port is taken. With the added tests in #16631 there is now a higher chance these ports are already taken since e.g. react router prefers to start at port 3000 and so do other servers.
This PR changes this so that we assign a random port inside the test instead.

## Test plan

- Ensure Windows CI is green again


integrations/utils.ts
```diff
@@ -588,3 +589,11 @@ async function gracefullyRemove(dir: string) {
...
   }
 }
+
+export function getFreePort(): number {
+  let server = net.createServer()
+  server.listen(0)
+  let address = server.address() as net.AddressInfo
+  server.close()
+  return address.port
+}
```
RobinMalfait, line 599: NOOOOOO 😭

thecrypticace (approved):

PR #16715, opened 2025-02-21
Use amount of properties when sorting

RobinMalfait:
Right now we sort the nodes based on a pre-defined sort order based on the properties that are being used. The property sort order is defined in a list we maintain.

We also have to make sure that the property count is taken into account such that if all the "sorts" are the same, that we fallback to the property count. Most amount of properties should be first such that we can override it with more specific utilities that have fewer properties.

However, if a property doesn't exist, then it wouldn't be included in a list of properties therefore the total count was off. 

This PR fixes that by counting all the used properties. If a property already exists it is counted twice. E.g.:
```css
.foo {
  color: red;

  &:hover {
    color: blue;
  }
}
```

In this case, we have 2 properties, not 1 even though it's the same `color` property.

## Test plan:

1. Updated the tests that are now sorted correctly
2. Added an integration test to make sure that `prose-invert` is defined after the `prose-stone` classes when using the `@tailwindcss/typography` plugin where this problem originated from.

Note how in this play (https://play.tailwindcss.com/wt3LYDaljN) the `prose-invert` comes _before_ the `prose-stone` which means that you can't apply the `prose-invert` classes to invert `prose-stone`.

### DIFF
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -327,18 +327,19 @@ function getPropertySort(nodes: AstNode[]) {
     if (node.kind === 'declaration') {
       count++
 
-      if (!seenTwSort) {
-        if (node.property === '--tw-sort') {
-          let idx = GLOBAL_PROPERTY_ORDER.indexOf(node.value ?? '')
-          if (idx !== -1) {
-            order.add(idx)
-            seenTwSort = true
-          }
-        }
+      if (seenTwSort) continue
 
-        let idx = GLOBAL_PROPERTY_ORDER.indexOf(node.property)
-        if (idx !== -1) order.add(idx)
+      if (node.property === '--tw-sort') {
+        let idx = GLOBAL_PROPERTY_ORDER.indexOf(node.value ?? '')
+        if (idx !== -1) {
+          order.add(idx)
+          seenTwSort = true
+          continue
+        }
       }
+
+      let idx = GLOBAL_PROPERTY_ORDER.indexOf(node.property)
+      if (idx !== -1) order.add(idx)
     } else if (node.kind === 'rule' || node.kind === 'at-rule') {
       for (let child of node.nodes) {
         q.push(child)
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -95,8 +95,8 @@ export function compileCandidates(
     // Find the first property that is different between the two rules
     let offset = 0
     while (
-      aSorting.properties.order.length < offset &&
-      zSorting.properties.order.length < offset &&
+      offset < aSorting.properties.order.length &&
+      offset < zSorting.properties.order.length &&
       aSorting.properties.order[offset] === zSorting.properties.order[offset]
     ) {
       offset += 1
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -325,7 +325,8 @@ function getPropertySort(nodes: AstNode[]) {
     // operator because we guarded against `q.length > 0` above.
     let node = q.shift()!
     if (node.kind === 'declaration') {
-      count++
+      // Empty strings should still be counted, e.g.: `--tw-foo:;` is valid
+      if (node.value !== undefined) count++
 
       if (seenTwSort) continue
 
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -23,7 +23,7 @@ export function compileCandidates(
 ) {
   let nodeSorting = new Map<
     AstNode,
-    { properties: number[]; variants: bigint; candidate: string }
+    { properties: { order: number[]; count: number }; variants: bigint; candidate: string }
   >()
   let astNodes: AstNode[] = []
   let matches = new Map<string, Candidate[]>()
@@ -95,18 +95,19 @@ export function compileCandidates(
     // Find the first property that is different between the two rules
     let offset = 0
     while (
-      aSorting.properties.length < offset &&
-      zSorting.properties.length < offset &&
-      aSorting.properties[offset] === zSorting.properties[offset]
+      aSorting.properties.order.length < offset &&
+      zSorting.properties.order.length < offset &&
+      aSorting.properties.order[offset] === zSorting.properties.order[offset]
     ) {
       offset += 1
     }
 
     return (
       // Sort by lowest property index first
-      (aSorting.properties[offset] ?? Infinity) - (zSorting.properties[offset] ?? Infinity) ||
+      (aSorting.properties.order[offset] ?? Infinity) -
+        (zSorting.properties.order[offset] ?? Infinity) ||
       // Sort by most properties first, then by least properties
-      zSorting.properties.length - aSorting.properties.length ||
+      zSorting.properties.count - aSorting.properties.count ||
       // Sort alphabetically
       compare(aSorting.candidate, zSorting.candidate)
     )
@@ -124,7 +125,10 @@ export function compileAstNodes(candidate: Candidate, designSystem: DesignSystem
 
   let rules: {
     node: AstNode
-    propertySort: number[]
+    propertySort: {
+      order: number[]
+      count: number
+    }
   }[] = []
 
   let selector = `.${escape(candidate.raw)}`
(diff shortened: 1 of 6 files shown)


packages/tailwindcss/src/compile.ts
```diff
@@ -310,30 +314,40 @@ function applyImportant(ast: AstNode[]): void {
...
     let node = q.shift()!
     if (node.kind === 'declaration') {
-      if (node.property === '--tw-sort') {
-        let idx = GLOBAL_PROPERTY_ORDER.indexOf(node.value ?? '')
-        if (idx !== -1) {
-          propertySort.add(idx)
-          break
+      count++
+
+      if (!seenTwSort) {
```
RobinMalfait, line 330: We were using a `break` before once we noticed a `--tw-sort`, but now we want to make sure we count _all_ properties. So once we notice a `--tw-sort` we don't want to collect new properties in the `order` set anymore, but we still want to keep walking to count other properties.

philipp-spiess: Could we invert this condition so we reduce the level of nesting here? 

```suggestion
      if (teenTwSort) continue;
```

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -4214,16 +4214,16 @@ test('translate-y', async () => {
...
+    ".-translate-z-\\[var\\(--value\\)\\] {
       --tw-translate-z: calc(var(--value) * -1);
       translate: var(--tw-translate-x) var(--tw-translate-y) var(--tw-translate-z);
     }
 
+    .translate-y-px {
+      --tw-translate-y: 1px;
+      translate: var(--tw-translate-x) var(--tw-translate-y);
+    }
+
```
philipp-spiess, line 4226: I'm confused why the order here changed, didn't both classes define the same number of properties before? 🤔

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -12585,10 +12589,6 @@ test('font', async () => {
       font-family: ui-sans-serif;
     }
 
-    .font-sans {
-      font-family: var(--font-sans);
-    }
```
philipp-spiess, line 12590: Also wouldn't expect this moving.

philipp-spiess (commented):
LGTM but I'm confused why two utilities got reordered in the tests even though they set the same number of properties

packages/tailwindcss/src/utilities.test.ts
RobinMalfait: This is related to the bug we had originally where only known properties are counted, but we have to think if this is what we want though 🤔

So, if you look at the the property-order file, you see this:
```
  'translate',
  '--tw-translate-x',
  '--tw-translate-y',
```

and `--tw-translate-z` doesn't exist in that list (we can add it though).

This means that even though we have 2 properties, in the previous code this would result in a sort order of `[53]` for the `translate-z-*` class (53 is the order of the `translate` property).

And for `translate-y-*` it would result in `[53, 55]` where `55` is the order of the `--tw-translate-y` property.

In the old code, that also means that the `translate-z-*` class counted 1 property, and the `translate-y-*` counted 2 properties, therefore it is sorted this way.

In the new code, this still results in `[53, 55]` and `[53]` but now we track the count which is `2` in both cases.

Another thing that already existed is that we only compare those numbers up to where they have the same length. So in both scenarios that would be comparing `53`. Then we compare the property count, and _then_ we compare alphabetically.

So previously we would compare the property count which was 1 vs 2, now they are both 2 and therefore we fallback to the alphabetical sort where the negative `-translate-z` is sorted before the other one.

The current code handling all of that:
```ts
astNodes.sort((a, z) => {
  // SAFETY: At this point it is safe to use TypeScript's non-null assertion
  // operator because if the ast nodes didn't exist, we introduced a bug
  // above, but there is no need to re-check just to be sure. If this relied
  // on pure user input, then we would need to check for its existence.
  let aSorting = nodeSorting.get(a)!
  let zSorting = nodeSorting.get(z)!

  // Sort by variant order first
  if (aSorting.variants - zSorting.variants !== 0n) {
    return Number(aSorting.variants - zSorting.variants)
  }

  // Find the first property that is different between the two rules
  let offset = 0
  while (
    aSorting.properties.order.length < offset &&
    zSorting.properties.order.length < offset &&
    aSorting.properties.order[offset] === zSorting.properties.order[offset]
  ) {
    offset += 1
  }

  return (
    // Sort by lowest property index first
    (aSorting.properties.order[offset] ?? Infinity) -
      (zSorting.properties.order[offset] ?? Infinity) ||
    // Sort by most properties first, then by least properties
    zSorting.properties.count - aSorting.properties.count ||
    // Sort alphabetically
    compare(aSorting.candidate, zSorting.candidate)
  )
})
```

Looking at the count/order result, if I add the `--tw-translate-z` to the property-order list, then this would be the result:
```
--tw-translate-x: 1px;
 2 [ 53, 54 ]

--tw-translate-y: 1px;
 2 [ 53, 55 ]

--tw-translate-z: calc(var(--value) * -1);
 2 [ 53, 56 ]
```
Where `2` is the count of properties, and the array is the sorted list based on the property order.

So uhm, I think this code is completely wrong:
```ts
  // Find the first property that is different between the two rules
  let offset = 0
  while (
    aSorting.properties.order.length < offset &&
    zSorting.properties.order.length < offset &&
    aSorting.properties.order[offset] === zSorting.properties.order[offset]
  ) {
    offset += 1
  }
```
Because the `offset` will start at `0`, so the `aSorting.properties.order.length` and `zSorting.properties.order.length`  will never be less than `0`.

Thanks for being my rubber ducky! 😅 

---

Complete sidenote: this test is wrong because this is the `translate-z-*` test and we use `translate-y-*` 😅

packages/tailwindcss/src/utilities.test.ts
RobinMalfait: Aha! I accidentally counted properties that were set to `undefined | null`. So we counted `3` for this guy (`font-family`, `font-feature-settings`, `font-variation-settings`)

packages/tailwindcss/src/compile.ts
```diff
@@ -95,18 +95,19 @@ export function compileCandidates(
     // Find the first property that is different between the two rules
     let offset = 0
     while (
-      aSorting.properties.length < offset &&
-      zSorting.properties.length < offset &&
-      aSorting.properties[offset] === zSorting.properties[offset]
+      offset < aSorting.properties.order.length &&
+      offset < zSorting.properties.order.length &&
```
RobinMalfait, line 99: This check was always wrong. `offset` started as `0` and no array has a length of `< 0` 😅

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -4213,14 +4213,15 @@ test('translate-y', async () => {
 })
 
 test('translate-z', async () => {
-  expect(await run(['translate-y-px', '-translate-z-[var(--value)]'])).toMatchInlineSnapshot(`
-    ".translate-y-px {
-      --tw-translate-y: 1px;
-      translate: var(--tw-translate-x) var(--tw-translate-y);
+  expect(await run(['translate-z-px', '-translate-z-[var(--value)]'])).toMatchInlineSnapshot(`
```
RobinMalfait, line 4216: Use the correct `translate-z-px` instead. (previous commits also show the fix for the order change)

packages/tailwindcss/src/property-order.ts
```diff
@@ -75,6 +75,7 @@ export default [
   'translate',
   '--tw-translate-x',
   '--tw-translate-y',
+  '--tw-translate-z',
```
RobinMalfait, line 78: This was missing, but we do have it for `--tw-scale-z` and `--tw-rotate-z` so I figured why don't we just add this.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -4213,14 +4213,15 @@ test('translate-y', async () => {
...
+      --tw-translate-z: calc(var(--value) * -1);
+      translate: var(--tw-translate-x) var(--tw-translate-y) var(--tw-translate-z);
     }
 
-    .-translate-z-\\[var\\(--value\\)\\] {
-      --tw-translate-z: calc(var(--value) * -1);
+    .translate-z-px {
+      --tw-translate-z: 1px;
+      translate: var(--tw-translate-x) var(--tw-translate-y) var(--tw-translate-z);
       translate: var(--tw-translate-x) var(--tw-translate-y) var(--tw-translate-z);
```
philipp-spiess, line 4225: Unrelated to this PR but we found `translate-z-px` emits two css classes right now. Needs a follow-up fix

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -16385,28 +16386,28 @@ test('@container', async () => {
...
-    .\\@container-normal\\/sidebar {
-      container: sidebar;
+    .\\@container-\\[size\\] {
+      container-type: size;
     }
 
-    .\\@container\\/sidebar {
-      container: sidebar / inline-size;
+    .\\@container-normal {
+      container-type: normal;
```
philipp-spiess, line 16410: Ah nice so this previously counted `container` and `container-type` as the same but isn't doing that anymore. Neat!

philipp-spiess (approved):

packages/tailwindcss/src/utilities.test.ts
RobinMalfait: PR: https://github.com/tailwindlabs/tailwindcss/pull/16718

PR #16721, opened 2025-02-21
Pin `c12` version in Nuxt integration tests

philipp-spiess:
Work around https://github.com/unjs/c12/issues/232


pi0:
upstream issue should have been fixed in 2.0.4

philipp-spiess:
Ah nice ty!

PR #16825, opened 2025-02-26
Ensure `not-*` does not remove `:is(…)`

### LINKED ISSUE #16821: `not-open` is generated but not applied

v4.0.7

`@tailwindcss/cli` v4.0.7

v22.11.0

Chrome Version 132.0.6834.160 (Official Build) (arm64)

macOS 15.2

**Reproduction URL**

https://play.tailwindcss.com/0ZD73uWu9I

**Describe your issue**

When using `not-open` variant on a popover, the following CSS is generated in `utilities` layer by Tailwind, but not applied in the browser:

```css
.not-open\:hidden {
  &:not([open], :popover-open, :open) {
    display: none;
  }
}
```

This is probably because browsers currently do not support `:open` yet, and `:not` does not take a forgiving selector list like `:is`.

While I could override it with `@variant open (&:is([open], :popover-open));`, I feel like this is a bug.

philipp-spiess:
Resolves #16821

This PR removes a special case in the `not-*` variant compound that removed `:is(…)` if it was the only part of the inversed selector list. While in-theory this makes sense, `:is(…)` accepts a _forgiving_ selector list where as `:not(…)` does not. See the [last point here](https://developer.mozilla.org/en-US/docs/Web/CSS/:not#description).

This is an issue specifically in combinations with variants that have selectors that are not supported by all browsers yet, for example `:open`.

It seems to be the most expected to simply keep the `:is(…)` here in any case.

## Test plan
- Ensured the repro form #16821 now also works in browsers that do not support `:open` (Safari and Firefox at the time of writing this): 
     Screenshot 2025-02-26 at 15 36 22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Experimental_: Add `user-valid` and `user-invalid` variants ([#12370](https://github.com/tailwindlabs/tailwindcss/pull/12370))
 - _Experimental_: Add `wrap-anywhere`, `wrap-break-word`, and `wrap-normal` utilities ([#12128](https://github.com/tailwindlabs/tailwindcss/pull/12128))
 
+### Fixed
+
+- Ensure `not-*` does not remove `:is(…)` from variants ([#16825](https://github.com/tailwindlabs/tailwindcss/pull/16825))
+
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
Another option is to check whether this was used in another forgiving wrapper or not.

But I think that's an optimization that can be added later.

PR #16946, opened 2025-03-04
Add `pointer-*` variants

philipp-spiess:
Adds new variants under a feature flag:

- `pointer-none`
- `pointer-coarse`
- `pointer-fine`
- `any-pointer-none`
- `any-pointer-coarse`
- `any-pointer-fine`


RobinMalfait (approved):
Looks good to me, we can always play with the exact API, but since it's feature flagged let's merge it as is 💪

PR #16947, opened 2025-03-04
doc(CHANGELOG): correct PR number for `scripting`

sntran:
Tiny correction of the CHANGLOG


philipp-spiess (approved):
Thank you! I really need to disable code completions in this file 😬

PR #16985, opened 2025-03-06
Fix Slim attributes

### LINKED ISSUE #16975: 4.0.10 - Regression with SLIM templates

v4.0.10

@tailwindcss/cli

v22.12.0

Brave 1.76.73

macOS 15

**Reproduction URL**

https://github.com/ledermann/tailwind-4.0.10-slim-issue

**Describe your issue**

Since 4.0.10, the parser fails to handle brackets in SLIM templates. See this example:

```slim
.text-xl.text-red-600[
  data-foo="bar"
]
  | This line should be red
```

Here, the class `text-red-600` is not processed.

Image

It works fine with 4.0.9

RobinMalfait:
This PR fixes an issue in Slim templates where the start of attributes causes some candidates to be missing.

```slim
.text-xl.text-red-600[
  data-foo="bar"
]
  | This line should be red
```

Because of the `[` attached to the `text-red-600`, the `text-red-600` was not extracted because `[` is not a valid boundary character.

To solve this, we copied the Pug pre processor and created a dedicated Slim pre processor. Next, we ensure that we replace `[` with ` ` in this scenario (by also making sure that we don't replace `[` where it's important).

Additionally, we noticed that `.` was also replaced inside of arbitrary values such as URLs. This has been fixed for both Pug and Slim.

Fixes: #16975

# Test plan

1. Added failing tests
2. Existing tests still pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure classes containing `--` are extracted correctly ([#16972](https://github.com/tailwindlabs/tailwindcss/pull/16972))
+- Ensure classes ending in `[` are extracted in Slim templating language ([#16985](https://github.com/tailwindlabs/tailwindcss/pull/16985))
 
 ## [4.0.10] - 2025-03-05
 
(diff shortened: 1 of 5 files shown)


philipp-spiess (approved):

mikker:
Wow! I used to throw a, `.x` on the end to work around this `.flex.items-center.x[value=1]` 😅
Happy to see this

PR #16995, opened 2025-03-06
Fix sorting of classes that have undefined declarations

### LINKED ISSUE #16973: Custom utilities sorted incorrectly
v4.0.10

Tested on:
Next.js 15.1.7
PostCSS 8.5.2

v23.8.0

Chrome

macOS

**Reproduction URL**
https://play.tailwindcss.com/RhJ7x3kQUN

**Describe your issue**
In the upgrade guide it is mentioned that custom utilities are sorted based on the number of properties they define. See [this section](https://tailwindcss.com/docs/upgrade-guide#adding-custom-utilities).

Now as shown in the reproduction URL, this does not seem to function properly. The `type-h1` utility specified 3 properties, but are still sorted last in the CSS output. This makes it's specificity the highest and you can't override the font size with e.g. the `text-sm` class. 

If we however instead use the `@apply` directive it seems to work as expected.

To me this seems to be a bug, but I might just misunderstand the docs, or there might be some alternative preferred solution.

I looked into using [component classes](https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes) instead, but this was not really feasible as it doesn't work with IntelliSense (see [this play link](https://play.tailwindcss.com/CLjzTW6JKq)). Maybe this is the most suitable solution, but due to the IntelliSense issue it's not really feasible for our team right now.

Really love v4! Thanks for the hard work 🙌

philipp-spiess:
Closes #16973

Declaration values of `undefined` are an implementation detail and skipped when printing the CSS. Thus, these should not count towards the sort order at a..

## Test plan

- See added regression test

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -15036,11 +15036,6 @@ test('text', async () => {
       --leading-snug: 1.375;
     }
 
-    .text-sm {
-      font-size: var(--text-sm);
-      line-height: var(--tw-leading, var(--text-sm--line-height));
-    }
-
     .text-\\[10px\\]\\/none {
       font-size: 10px;
       line-height: 1;
@@ -15069,6 +15064,11 @@ test('text', async () => {
     .text-\\[xx-large\\]\\/6 {
       font-size: xx-large;
       line-height: calc(var(--spacing) * 6);
+    }
+
+    .text-sm {
+      font-size: var(--text-sm);
+      line-height: var(--tw-leading, var(--text-sm--line-height));
     }
 
     .text-sm\\/6 {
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/utilities.test.ts
```diff
@@ -15071,6 +15066,11 @@ test('text', async () => {
       line-height: calc(var(--spacing) * 6);
     }
 
+    .text-sm {
+      font-size: var(--text-sm);
+      line-height: var(--tw-leading, var(--text-sm--line-height));
+    }
+
```
philipp-spiess, line 15073: This now uses alphabetical sorting becuase it has the same number of props as the arbitrary text classes

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -16748,11 +16748,11 @@ describe('custom utilities', () => {
     expect(optimizeCss(compiled).trim()).toMatchInlineSnapshot(`
       "@layer utilities {
         .text-sm {
-          font-size: var(--text-sm, .875rem);
-          line-height: var(--tw-leading, var(--text-sm--line-height, 1.25rem));
           font-size: var(--text-sm, .8755rem);
           line-height: var(--text-sm--line-height, 1.255rem);
           text-rendering: optimizeLegibility;
+          font-size: var(--text-sm, .875rem);
+          line-height: var(--tw-leading, var(--text-sm--line-height, 1.25rem));
```
philipp-spiess, line 16755: This diff is weird but that's because of lightning flatting. What happens ehre is the custom new `text-sm` utility is now rendered before the core `text-sm`. The new prop ` text-rendering: optimizeLegibility;` is still added, the rest is intended behavior here since you can't _overwrite_ core utilities.

RobinMalfait (approved):

PR #17001, opened 2025-03-06
Ensure `}` and `{` are valid boundary characters

### LINKED ISSUE #16999: Missing classes because of `{` boundary and new extractor

v4.0.11

**Reproduction**

With the new extractor (https://github.com/tailwindlabs/tailwindcss/pull/16306) the class `h-4` and `mx-4` is not picked up anymore when using the following Twig template content:

```
<div class="flex items-center mx-4{% if session.isValid %}{% else %} h-4{% endif %}"></div>
```

(See https://play.tailwindcss.com/w3SazyVehg for the example on Tailwind Play and the generated CSS)

I assume the `{` after `h-4` and `mx-4` is somehow not considered a (possible) boundary anymore. I think this is likely a bug, as many template languages use `{` and `}` as delimiters in some way, so the possibility of a boundary should be taken into account.

RobinMalfait:
In a lot of templating languages, `{…}` is used to escape the template language to write some code logic, this means that if you have input that looks like this:

```html
<div class="flex items-center mx-4{% if session.isValid %}{% else %} h-4{% endif %}"></div>
```

That the if condition will be gone once it's compiled to HTML, but it also means that the `}` and `{` characters next to `mx-4{` and `h-4` should be valid boundary characters because Tailwind CSS typically operates on the source code, not the compiled code.

I didn't create a pre processor for this because unfortunately this syntax is often embedded in `.html` files, and I would love to not have to create a pre processor for `.html` files.

You might notice that there are other changes in this PR, they are all related to boundary characters, let me explain:

In the named utility machine, we repeated the allowed boundary characters. We will now re-use the shared function so that it's all in one location.
We still check for a few additional characters such as `:` because of JS keys, `/` because it could be a start of a modifier and `!` because the named utility could be important.

This actually exposed a bug where the `class` attribute in `<div class="…">…</div>` was not extracted (which we typically want). However, in languages like Svelte, this is a real thing you can do: `<div class:px-4="condition"></div>`. The pre processor will make sure to convert this input to `<div class px-4="condition"></div>` but that still means that we need to extract the `px-4`.

if a utility ends in an number this worked as expected, but `<div class:flex="condition"></div>` did not work. This PR also fixes that by adding a dedicated test and it's also the reason why you see a bunch of `class` candidates in the tests.

Fixes: #16999

# Test plan

1. Added new test
4. Existing tests pass

### DIFF
--- a/crates/oxide/src/extractor/named_utility_machine.rs
+++ b/crates/oxide/src/extractor/named_utility_machine.rs
@@ -1,6 +1,7 @@
 use crate::cursor;
 use crate::extractor::arbitrary_value_machine::ArbitraryValueMachine;
 use crate::extractor::arbitrary_variable_machine::ArbitraryVariableMachine;
+use crate::extractor::candidate_machine::is_valid_after_boundary;
 use crate::extractor::machine::{Machine, MachineState};
 use classification_macros::ClassifyBytes;
 
@@ -120,19 +121,22 @@ impl Machine for NamedUtilityMachine {
                         // E.g.: `:div="{ flex: true }"` (JavaScript object syntax)
                         //                    ^
                         Class::AlphaLower | Class::AlphaUpper => {
-                            match cursor.next.into() {
-                                Class::Quote
-                                | Class::Whitespace
-                                | Class::CloseBracket
-                                | Class::Dot
-                                | Class::Colon
-                                | Class::End
-                                | Class::Slash
-                                | Class::Exclamation => return self.done(self.start_pos, cursor),
-
-                                // Still valid characters
-                                _ => cursor.advance(),
+                            if is_valid_after_boundary(&cursor.next) || {
+                                // Or any of these characters
+                                //
+                                // - `:`, because of JS object keys
+                                // - `/`, because of modifiers
+                                // - `!`, because of important
+                                matches!(
+                                    cursor.next.into(),
+                                    Class::Colon | Class::Slash | Class::Exclamation
+                                )
+                            } {
+                                return self.done(self.start_pos, cursor);
                             }
+
+                            // Still valid characters
+                            cursor.advance()
                         }
 
                         Class::Dash => match cursor.next.into() {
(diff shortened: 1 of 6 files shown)


crates/oxide/src/extractor/named_utility_machine.rs
```diff
@@ -120,19 +121,22 @@ impl Machine for NamedUtilityMachine {
...
+                                matches!(
+                                    cursor.next.into(),
+                                    Class::Colon | Class::Slash | Class::Exclamation
+                                )
+                            } {
+                                return self.done(self.start_pos, cursor);
                             }
+
+                            // Still valid characters
+                            cursor.advance()
```
thecrypticace, line 139: This is fine for now but we'll want to perf test this. I wouldn't be surprised if this implementation causes a bit of a hit

crates/oxide/src/extractor/named_utility_machine.rs
```diff
@@ -213,14 +217,20 @@ impl Machine for NamedUtilityMachine {
...
+                                // - `/`, because of modifiers
+                                // - `!`, because of important
+                                matches!(
+                                    cursor.next.into(),
+                                    Class::Colon | Class::Slash | Class::Exclamation
+                                )
+                            } =>
+                            {
+                                return self.done(self.start_pos, cursor)
+                            }
```
thecrypticace, line 233: ditto here too

thecrypticace (approved):

crates/oxide/src/extractor/named_utility_machine.rs
RobinMalfait: Yep will do some classification next, and I still have a WIP branch to get rid of the Cursor

nicomen:
Please fix the regression for `]…[` also

RobinMalfait:
@nicomen please open a new issue with a minimal reproduction repo attached so we can take a look.

PR #17005, opened 2025-03-06
Improve boundary classification

RobinMalfait:
This PR cleans up the boundary character checking by using similar classification techniques as we used for other classification problems.

For starters, this moves the boundary related items to its own file, next we setup the classification enum.

Last but not least, we removed `}` as an _after_ boundary character, and instead handle that situation in the Ruby pre processor where we need it. This means the `%w{flex}` will still work in Ruby files.

---

This PR is a followup for https://github.com/tailwindlabs/tailwindcss/pull/17001, the main goal is to clean up some of the boundary character checking code. The other big improvement is performance. Changing the boundary character checking to use a classification instead results in:

Took the best score of 10 runs each:
```diff
- CandidateMachine: Throughput: 311.96 MB/s
+ CandidateMachine: Throughput: 333.52 MB/s
```

So a ~20MB/s improvement.

# Test plan

1. Existing tests should pass. Due to the removal of `}` as an after boundary character, some tests are updated.
2. Added new tests to ensure the Ruby pre processor still works as expected.

### DIFF
--- a/crates/oxide/src/extractor/boundary.rs
+++ b/crates/oxide/src/extractor/boundary.rs
@@ -1,0 +1,51 @@
+use crate::extractor::Span;
+
+/// A candidate must be preceded or followed by any of these characters
+/// E.g.: `<div class="flex">`
+///                   ^ Valid for `flex`
+///        ^ Invalid for `div`
+#[inline(always)]
+fn is_valid_common_boundary(c: &u8) -> bool {
+    matches!(
+        c,
+        b'\t' | b'\n' | b'\x0C' | b'\r' | b' ' | b'"' | b'\'' | b'`' | b'\0'
+    )
+}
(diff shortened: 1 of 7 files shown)


crates/oxide/src/extractor/boundary.rs
```diff
@@ -0,0 +1,104 @@
...
+    Common,
+
+    // Angular like attributes, e.g.:
+    //
+    // ````
+    // [class.foo]
+    //       ^
+    // ```
+    #[bytes(b'.')]
+    // Twig like templating languages, e.g.:
```
thecrypticace, line 69:
```suggestion
    // Twig-like templating languages, e.g.:
```

thecrypticace (approved):

PR #17027, opened 2025-03-07
Add `razor`/`cshtml` pre processing

### LINKED ISSUE #17022: Content transform in css config
### Discussed in https://github.com/tailwindlabs/tailwindcss/discussions/15742

<sup>Originally posted by **skttl** January 23, 2025</sup>
To make the default syntax for container queries work, I've added this in my v3 config:

```js
module.exports = {
  content: {
    files: [
      // these are relative to package.json
      './UI/Scripts/**/*.js',
      './Views/**/*.cshtml'
    ],
    transform: {
      cshtml: (content) => {
        return content.replace('@@', '@');
      }
    }
  },
```

This replaces `@@` with `@` before looking for classnames, making me able to write eg. `@@sm:text-red-500` instead of `@sm:text-red-500`. The latter would interpret @sm as a razor variable.

Is it possible to do something similar in the new css config format?</div>

RobinMalfait:
This PR fixes an issue in Razor template files where `@sm:flex` doesn't work and `@@sm:flex` is required.

In Tailwind CSS v3, some people used a custom transform to replace `@@` with just `@`. But in Tailwind CSS v4 we don't have this.

However, we can add a pre processor for `.cshtml` and `.razor` files.

Fixes: #17022

### DIFF
--- a/crates/oxide/src/lib.rs
+++ b/crates/oxide/src/lib.rs
@@ -468,6 +468,7 @@ pub fn pre_process_input(content: &[u8], extension: &str) -> Vec<u8> {
     use crate::extractor::pre_processors::*;
 
     match extension {
+        "cshtml" | "razor" => Razor.process(content),
         "pug" => Pug.process(content),
         "rb" | "erb" => Ruby.process(content),
         "slim" => Slim.process(content),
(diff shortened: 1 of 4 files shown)


crates/oxide/src/lib.rs
```diff
@@ -468,6 +468,7 @@ pub fn pre_process_input(content: &[u8], extension: &str) -> Vec<u8> {
     use crate::extractor::pre_processors::*;
 
     match extension {
+        "cshtml" | "razor" => Razor.process(content),
```
RobinMalfait, line 471: Used both `cshtml` and `razor`. See: https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-9.0

philipp-spiess (approved):

senneclaeskens:
@RobinMalfait 
I've added additional comments in https://github.com/tailwindlabs/tailwindcss/issues/17022

PR #17036, opened 2025-03-07
Make `--theme(…)` return CSS variables

### LINKED ISSUE #16904: [v4.0.9+] Variables are ignored when referenced within CSS Modules

For example: v4.0.9

Vite 6.2.0

22.0.0

Chrome, Safari and Firefox

macOS and Windows

**Reproduction URL**

https://github.com/boylett/tailwind-4-0-9-cssmodule-var-bug

**Describe your issue**

When referencing the main theme file within CSS Modules, variables are no longer recognised and exported in the main CSS.

This works fine in 4.0.8 (see ['working' branch](https://github.com/boylett/tailwind-4-0-9-cssmodule-var-bug/tree/working)) so am inclined to believe it has something to do with the new `@reference` keyword.

Within reproduction, the background colour of the main "Welcome" component wrapper should be `--color-hot-pink` rather than transparent.

On the 'main' branch, tw 4.0.9, the variable is simply ignored and not included in the final CSS.

### LINKED ISSUE #16945: v4 preflight theme variables aren't prefixed
Versions:
- Tailwind: v4.0.9
- Vite: 5.2.11
- Node: 20.14.0
- Chrome, Mac OS

[Reproduction URL](https://play.tailwindcss.com/fPhQksGLVR)

If you look at the reproduction URL above (took awhile to prefix all those classes lol) you can see that the `--default-font-family` does not map to the prefixed `--font-sans` declared in the `@theme`. It seems that since `--default-font-family: var(--font-sans);` it isn't getting updated to map to the prefixed value later. This was not an issue in v3.

The workaround I have is to declare `--default-font-family` in my own `@layer base`:

```css
@layer base {
  html {
    --default-font-family: 'Salesforce Sans', Helvetica, Arial, sans-serif;
  }
}
```

What ought to happen is the preflight reference points to the prefixed font so that the theme can correctly override it: `--default-font-family: var(--tw-font-sans);`

philipp-spiess:
Closes #16945

This PR changes the `--theme(…)` function now return CSS `var(…)` definitions unless used in places where `var(…)` is not valid CSS (e.g. in `@media (width >= theme(--breakpoint-md))`):

```css
/* input */
@theme {
  --color-red: red;
}
.red {
  color: --theme(--color-red);
}

/* output */
:root, :host {
  --color-red: red;
}
.red {
  color: var(--color-red);
}
```

Furthermore, this adds an `--theme(… inline)` option to the `--theme(…)` function to force the resolution to be inline, e.g.:

```css
/* input */
@theme {
  --color-red: red;
}
.red {
  color: --theme(--color-red inline);
}

/* output */
.red {
  color: red;
}
```

This PR also changes preflight and the default theme to use this new `--theme(…)` function to ensure variables are prefixed correctly.

## Test plan

- Added unit tests and a test that pulls in the whole preflight under a prefix theme.


packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
```diff
@@ -37,9 +37,9 @@ exports[`\`@import 'tailwindcss'\` is replaced with the generated CSS 1`] = `
     -webkit-text-size-adjust: 100%;
     tab-size: 4;
     line-height: 1.5;
-    font-family: var(--default-font-family, ui-sans-serif, system-ui, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji");
-    font-feature-settings: var(--default-font-feature-settings, normal);
-    font-variation-settings: var(--default-font-variation-settings, normal);
+    font-family: var(--default-font-family);
+    font-feature-settings: var(--default-font-feature-settings);
+    font-variation-settings: var(--default-font-variation-settings);
```
philipp-spiess, line 42: This might be controversial but the way the `--theme(…)` function works is that it omits the fallback when the theme is found within the design system (this is necessary to work with `@reference` where we want to append a fallback to the theme `var(…)` declaration based on the configured theme value.

As a result, this will now omit the the fallback values here. But if you scroll down a bit, `--font-sans` and `--default-font-family` CSS variables will be emitted anyways so we _know_ the fallback is not necessary thus save some bytes in the resulting CSS file.

packages/tailwindcss/src/design-system.ts
```diff
@@ -29,7 +29,7 @@ export type DesignSystem = {
   compileAstNodes(candidate: Candidate): ReturnType<typeof compileAstNodes>
 
   getVariantOrder(): Map<Variant, number>
-  resolveThemeValue(path: string): string | undefined
+  resolveThemeValue(path: string, forceInline?: boolean): string | undefined
```
philipp-spiess, line 32: I have to make this optional because of IntelliSense unfortunately. Also decided to make a boolean here since don't leak the `ThemeOption` enum beyond internal APIs right now.

packages/tailwindcss/theme.css
```diff
@@ -434,12 +434,12 @@
...
   --default-transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
-  --default-font-family: var(--font-sans);
-  --default-font-feature-settings: var(--font-sans--font-feature-settings);
-  --default-font-variation-settings: var(--font-sans--font-variation-settings);
-  --default-mono-font-family: var(--font-mono);
-  --default-mono-font-feature-settings: var(--font-mono--font-feature-settings);
-  --default-mono-font-variation-settings: var(--font-mono--font-variation-settings);
+  --default-font-family: --theme(--font-sans);
+  --default-font-feature-settings: --theme(--font-sans--font-feature-settings, normal);
+  --default-font-variation-settings: --theme(--font-sans--font-variation-settings, normal);
```
philipp-spiess, line 439: Right now `--theme(…)`, just like `theme(…)`, errors when a value is not part of your theme and no fallback value is defined. 

We could potentially change this for non-inlined values but I think this is actually a good default, it's a feature that a CSS `var(…)` can not do but we can since we have access to the theme object. So, for now I decided to keep it even though it means we have to define a fallback for these variables (which makes sense since these are not defined within the default theme).

Hmm now that I think of this it probably means that `--default-font-family: --theme(--font-sans);` might also need a default value since otherwise if you unset the `--font-*` namespace, this will not compile. 🤔

packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
```diff
@@ -10,11 +10,7 @@ exports[`\`@import 'tailwindcss'\` is replaced with the generated CSS 1`] = `
     --text-2xl--line-height: calc(2 / 1.5);
     --font-weight-bold: 700;
     --default-font-family: var(--font-sans);
-    --default-font-feature-settings: var(--font-sans--font-feature-settings);
-    --default-font-variation-settings: var(--font-sans--font-variation-settings);
     --default-mono-font-family: var(--font-mono);
-    --default-mono-font-feature-settings: var(--font-mono--font-feature-settings);
-    --default-mono-font-variation-settings: var(--font-mono--font-variation-settings);
```
philipp-spiess, line 17: Now that the `--default-font-feature-settings` theme variable is defined as:

```css
@theme {
  --default-font-feature-settings: --theme(--font-sans--font-feature-settings, initial);
}
```

The value is resolved to `initial` (and thus omitted) because we know that `--font-sans--font-feature-settings` is not defined in the theme. Note that this implies that values referenced via `--theme(…)` can not be defined at runtime only and NEED to be in the `@theme` block. I think this is a reasonable thing to do since we also _error_ when it's not in the `@theme` block (unless a fallback is specified like it is here).

packages/tailwindcss/src/css-functions.ts
```diff
@@ -51,15 +64,70 @@ function spacing(designSystem: DesignSystem, value: string, ...rest: string[]) {
...
+  }
+
+  if (fallback.length === 0) {
+    return resolvedValue
+  }
+
+  // Inject the fallback…
+  //
+  // - …as the value if the value returned is `initial`
+  // - …expect any `initial` fallbacks on `var(…)`, `theme(…)`, or `--theme(…)`
```
thecrypticace, line 107: Is this supposed to say `except`?

This entire comment is kinda hard to parse for me

packages/tailwindcss/src/css-functions.ts
```diff
@@ -142,3 +218,22 @@ function eventuallyUnquote(value: string) {
...
+function injectFallbackForInitialFallback(ast: ValueParser.ValueAstNode[], fallback: string): void {
+  ValueParser.walk(ast, (node) => {
+    if (node.kind !== 'function') return
+    if (node.value !== 'var' && node.value !== 'theme' && node.value !== '--theme') return
+
+    if (node.nodes.length === 1) {
+      node.nodes.push({
+        kind: 'word',
+        value: `, ${fallback}`,
+      })
```
thecrypticace, line 231: This seems _fine_ but don't we technically treat the `,` as a separate node when parsing? Should we do that here too?

packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -152,14 +152,212 @@ describe('--theme(…)', () => {
...
+        }
+      `),
+    ).toMatchInlineSnapshot(`
+      ".red {
+        font-family: var(--tw-default-font-family, Potato Sans, sans-serif);
+      }"
+    `)
+  })
+
+  test('--theme(…) does not inject the fallback if the fallback is `initial`', async () => {
```
thecrypticace, line 289: This test name feels wrong. The different thing about this test is that the theme key exists and therefore the fallback isn't used.

thecrypticace (changes requested):
left some comments about test names, comments, and a thing about the AST

packages/tailwindcss/src/css-functions.ts
philipp-spiess: yeah we do, the fallback could also consist of other, nested, functions etc. It's technically more correct to parse the fallback as well but it felt super wasteful to parse it just so it's stringified in the next line. If we ever change stuff so that the stringification won't work anymore if the word node contains stuff it shouldn't, we also have unit tests that would let us know this breaks so I decided to just put it in as a `word` node.

packages/tailwindcss/src/css-functions.test.ts
philipp-spiess: good catch!

packages/tailwindcss/src/css-functions.ts
philipp-spiess: I reworked this part, thanks

packages/tailwindcss/src/css-functions.ts
thecrypticace: 👍

thecrypticace (approved):

PR #17085, opened 2025-03-09
Treat starting single quote as verbatim text in Slim

### LINKED ISSUE #17081: 4.0.12 - Regression with SLIM templates (single quotes)

v4.0.12

@tailwindcss/cli

v22.12.0

Brave 1.76.73

macOS 15

**Reproduction URL**

https://github.com/ledermann/tailwind-4.0.12-slim-issue

**Describe your issue**

Similar to #16975 and #16998, there is another issue with SLIM templates. With 4.0.12, the parser fails to handle [single quotes (used to enforce trailing white space)](https://github.com/slim-template/slim?tab=readme-ov-file#verbatim-text-with-trailing-white-space-). 

See this example:

```slim
div
  'A single quote enforces trailing white space
  = 1234

.text-red-500.text-3xl
  | This text should be red
```

Here, the class `text-red-500` is not processed.

Image

RobinMalfait:
This PR fixes an issue in Slim templates where a single quote `'` at the start of the line (excluding white space) is considered a line indicator for verbatim text. It is not considered a string in this scenario.

So something like this:
```slim
div
  'Foo'
```

Will compile to:
```html
<div>Foo'</div>
```

Fixes: #17081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix `haml` pre-processing ([#17051](https://github.com/tailwindlabs/tailwindcss/pull/17051))
+- Treat starting single quote as verbatim text in Slim ([#17085](https://github.com/tailwindlabs/tailwindcss/pull/17085))
 
 ## [4.0.12] - 2025-03-07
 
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
I wonder if our pre-processor logic should also only process stuff per-line so we avoid this kind of behavior? 🤔

RobinMalfait:
@philipp-spiess I was thinking about that as well, but it makes it potentially harder for other checks where `[` and `]` needs to be balanced but they might be spread across multiple lines 🤔

PR #17113, opened 2025-03-10
Improve candidate extraction when candidates contain `.` characters

RobinMalfait:
This PR fixes an issue where some classes weren't properly extracted due to some incorrect assumptions in the pre processors.

Templating languages such as Haml, Slim and Pug allow you to write classes in a shorter way that are not properly contained inside of strings. E.g.:
```slim
p.flex.px-2
```

These candidates are not properly extracted because there are no bounding characters like quotes. To solve this, we pre-process it and replace `.` with ` ` characters. This results in something like:
```
p flex px-2
```

However, this has some challenges on its own. Candidates like `px-2.5` cannot be written in this shorthand form, instead they need to be in strings. Now we _cannot_ replace the `.` because otherwise we would change `px-2.5` to `px-2 5` which is wrong.

The next problem is that we need to know when they are in a "string". This has another set of problems because these templating languages allow you to write normal text that will eventually be the contents of the HTML tags.

```haml
.text-red-500.text-3xl
  | This text can't should be red
                 ^ Wait, is this the start of a string now???
```

In this example, if we consider the `'` the start of a string, when it's clearly not, how would we know it's for _sure_ not a string?

This ended up as a bit of a rabbit hole, but we came up with another approach entirely if we think about the original problem we want to solve which is when do we change `.` to ` ` characters.

One of the rules in our current extractor is that a `.` has to be between 2 numbers. Which works great in a scenario like: `px-2.5`. However, if you look at Haml or Slim syntax, this is also allowed:

```slim
p.bg-red-500.2xl:flex
           ^^^ Uh oh...
```

In this scenario, a `.` is surrounded by numbers so we shouldn't replace it with a space. But as you can see, we clearly do... so we need another heuristic in this case.

Luckily, one of the rules in Tailwind CSS is that a utility cannot start with a number, but a variant _can_. This means that if we see a scenario like `<digit>.<digit>` then we can just check if the value after the `.` is a valid variant or not.

In this case it is a valid variant so we _do_ want to replace the `.` with a ` ` even though we do have the `<digit>.<digit>` format.

🥴

# Test plan

1. Added additional tests.
2. Existing tests still pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Treat starting single quote as verbatim text in Slim ([#17085](https://github.com/tailwindlabs/tailwindcss/pull/17085))
 - Ensure `.node` and `.wasm` files are not scanned for utilities ([#17123](https://github.com/tailwindlabs/tailwindcss/pull/17123))
 - Improve performance when scanning `JSON` files ([#17125](https://github.com/tailwindlabs/tailwindcss/pull/17125))
+- Fix detecting strings in Haml, Pug, and Slim pre processors ([#17113](https://github.com/tailwindlabs/tailwindcss/pull/17113))
 
 ## [4.0.12] - 2025-03-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Treat starting single quote as verbatim text in Slim ([#17085](https://github.com/tailwindlabs/tailwindcss/pull/17085))
 - Ensure `.node` and `.wasm` files are not scanned for utilities ([#17123](https://github.com/tailwindlabs/tailwindcss/pull/17123))
 - Improve performance when scanning JSON files ([#17125](https://github.com/tailwindlabs/tailwindcss/pull/17125))
-- Fix detecting strings in Haml, Pug, and Slim pre processors ([#17113](https://github.com/tailwindlabs/tailwindcss/pull/17113))
+- Fix extracting candidates containing dots in Haml, Pug, and Slim pre processors ([#17113](https://github.com/tailwindlabs/tailwindcss/pull/17113))
 
 ## [4.0.12] - 2025-03-07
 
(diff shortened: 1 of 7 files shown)


crates/oxide/src/extractor/bracket_stack.rs
```diff
@@ -25,6 +25,7 @@ impl BracketStack {
                 b'(' => b')',
                 b'[' => b']',
                 b'{' => b'}',
+                b'<' => b'>',
```
thecrypticace, line 28: Any chance we could do this for just the preprocessors and not for all uses of the bracket stack in general?

thecrypticace: wait nvm I misunderstood what was happening. ignore me 👍

thecrypticace (dismissed):

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Treat starting single quote as verbatim text in Slim ([#17085](https://github.com/tailwindlabs/tailwindcss/pull/17085))
 - Ensure `.node` and `.wasm` files are not scanned for utilities ([#17123](https://github.com/tailwindlabs/tailwindcss/pull/17123))
 - Improve performance when scanning `JSON` files ([#17125](https://github.com/tailwindlabs/tailwindcss/pull/17125))
```
philipp-spiess, line 26:
```suggestion
- Improve performance when scanning JSON files ([#17125](https://github.com/tailwindlabs/tailwindcss/pull/17125))
```

philipp-spiess:
Screenshot 2025-03-11 at 15 36 41

Unfortunately this still breaks if a double-quote is used in the text. :/

@RobinMalfait have been chatting about this and we'll going to revert the string-based handling of _dots inside utilities in pug_ and instead make it replace dots to space unless:

- They are surrounded by numbers. That's what we do in Oxide right now: https://github.com/tailwindlabs/tailwindcss/blob/main/crates/oxide/src/extractor/named_utility_machine.rs#L247-L261
- They are inside a bracket stack (so arbitrary stuff works)

This way we can avoid thinking about strings in this languages

crates/oxide/src/extractor/pre_processors/test-fixtures/haml/dst-1.haml
```diff
@@ -0,0 +1,155 @@
...
+      More Coming Soon!
+    %p mt-2 text-xl text-center
+      Keen an eye out as we'll be adding more components, guides, and
+      %br max-sm:hidden
+      suggested gems for you to build amazing Rails apps!
+     mt-4 text-center
+        daisy_button "😉 Get Started", css: "btn-primary text-xl",
+        right_icon: "arrow-right", target: "_blank",
+        class: "px-2.5"
+        href: "https://github com/profoundry-us/loco_motion locomotion-components"
```
RobinMalfait, line 155: A bit silly, but the `.` and `#` characters are replaced because:

1. The `.` is not surrounded by numbers
2. The `#` is replaced because it's not inside of a bracket stack

While this does break the string, it doesn't matter from Tailwind's perspective and since we're not updating the real code, it's not an issue.

philipp-spiess (approved):
I like this approach a lot more. Having to worry about what these template languages consider strings and not felt like a rabbit-hole where the solution is to have a full parser for those languages. This fix is more surgical for sure 👍

PR #17142, opened 2025-03-12
Do not extract candidates containing JS string interpolation pattern `${`

### LINKED ISSUE #15853: v4. build fails with dynamically created classnames + at-rules

4.0.0

vite-6.0.11

v23.4.0

Firefox, but it doesn't matter

macOS

**Reproduction URL**

https://github.com/homeboldham/vite-tailwind-unknown-word 

it's a minimal vite + tailwind setup (using vite react ts template) and adding tailwind v4 via `@tailwindcss/vite` plugin.

**Describe your issue**

Please check https://github.com/homeboldham/vite-tailwind-unknown-word/blob/main/src/App.tsx
If you have smth like `h-[${h}px]` in your source code, 
AND
you have `at-rule` in your css. For example, I add font import here:
https://github.com/homeboldham/vite-tailwind-unknown-word/blob/main/src/index.css

then the build fails:
```
> vite-project@0.0.0 build
> tsc -b && vite build

vite v6.0.11 building for production...
✓ 15 modules transformed.
x Build failed in 100ms
error during build:
[vite:css] [postcss] /Users/foo/github/vite-tailwind-atrule/src/index.css:524:15: Unknown word
file: /Users/foo/github/vite-tailwind-atrule/src/index.css:524:14
    at Input.error (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/input.js:109:16)
    at Parser.unknownWord (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/parser.js:593:22)
    at Parser.other (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/parser.js:435:12)
    at Parser.parse (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/parser.js:470:16)
    at parse (/Users/foo/foo/vite-tailwind-atrule/node_modules/postcss/lib/parse.js:11:12)
    at new LazyResult (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/lazy-result.js:133:16)
    at Processor.process (/Users/foo/github/vite-tailwind-atrule/node_modules/postcss/lib/processor.js:53:14)
    at compileCSS (file:///Users/foo/github/vite-tailwind-atrule/node_modules/vite/dist/node/chunks/dep-M1IYMR16.js:48784:59)
    at async Object.transform (file:///Users/foo/github/vite-tailwind-atrule/node_modules/vite/dist/node/chunks/dep-M1IYMR16.js:48039:11)
    at async transform (file:///Users/foo/github/vite-tailwind-atrule/node_modules/rollup/dist/es/shared/node-entry.js:19794:16)
```

Everything works if it's just `@import` or `h-[${h}px]` alone. I can reproduce it only if they're both there.

It's not about `h-`. I believe any dynamically created className can cause this.

(please disregard the rationale behind the component in the code – I simplified it to nonsensical to reproduce the issue).

### LINKED ISSUE #17054: detecting classes in source files causes crash

4.0.11

"next": "15.2.2-canary.3"

v20.18.0

Chrome

Windows

**Reproduction URL**

https://github.com/Autumnal-Joy/tailwindcss-bug-reproduce

**Describe your issue**

Crash

```typescript
const c = "#52d71e";
console.log(`bg-[${c}]`);
```

```shell
 ⨯ ./app/globals.css:485:25
Parsing css source code failed
  483 |   }
  484 |   .bg-\[\$\{c\}\] {
> 485 |     background-color: ${c};
      |                         ^
  486 |   }
  487 |   .bg-background {
  488 |     background-color: var(--background);
```

I saw the [documentation](https://tailwindcss.com/docs/detecting-classes-in-source-files) saying that "Don't use props to build class names dynamically", but ​It seems a bit unconventional that `console.log` could also cause a crash.

RobinMalfait:
This PR fixes an issue where often people run into issues where they try to use string interpolation and it doesn't work. Even worse, it could result in crashes because we will actually generate CSS. This fix only filters out candidates with a pattern like `${`. If this occurs in a string position it is fine.

Another solution would be to add a pre processor for JS/TS (and all thousand file extension combinations) but the problem is that you can also write JS in HTML files so we would have to pre process HTML as well which would not be ideal.

# Test plan

1. Added tests to prove this works in arbitrary values, arbitrary variables in both utilities and variants.
2. Existing tests pass.
3. Some screenshots with before / after situations:

Given this input:
```ts
let color = '#0088cc';
let opacity = 0.8;
let name = 'variable-name';
let classes = [
  // Arbitrary Properties
  `[color:${color}]`
  `[${property}:value]`,
  `[--img:url('https://example.com?q=${name}')]`, // WONT WORK BUT VALID CSS

  // Arbitrary Values
  `bg-[${color}]`,

  // Arbitrary Variables
  `bg-(--my-${color})`,
  `bg-(--my-color,${color})`,

  // Arbitrary Modifier
  `bg-red-500/[${opacity}]`,
  `bg-red-500/(--my-${name})`,
  `bg-red-500/(--my-opacity,${opacity})`,

  // Arbitrary Variant
  `data-[state=${name}]:flex`,
  `supports-(--my-${name}):flex`,
  `[@media(width>=${value})]:flex`,
];
```

This is the result:

| Before | After |
| --- | --- |
| image | image |

Fixes: #17054
Fixes: #15853

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - _Experimental_: Add `user-valid` and `user-invalid` variants ([#12370](https://github.com/tailwindlabs/tailwindcss/pull/12370))
 - _Experimental_: Add `wrap-anywhere`, `wrap-break-word`, and `wrap-normal` utilities ([#12128](https://github.com/tailwindlabs/tailwindcss/pull/12128))
 
+### Fixed
+
+- Do not extract candidates with JS string interpolation `${` ([#17142](https://github.com/tailwindlabs/tailwindcss/pull/17142))
+
 ## [4.0.13] - 2025-03-11
 
 ### Fixed
--- a/crates/oxide/src/extractor/arbitrary_property_machine.rs
+++ b/crates/oxide/src/extractor/arbitrary_property_machine.rs
@@ -226,6 +226,11 @@ impl Machine for ArbitraryPropertyMachine<ParsingValueState> {
                 // URLs are not allowed
                 Class::Slash if start_of_value_pos == cursor.pos => return self.restart(),
 
+                // String interpolation-like syntax is not allowed. E.g.: `[${x}]`
+                Class::Dollar if matches!(cursor.next.into(), Class::OpenCurly) => {
+                    return self.restart()
+                }
+
                 // Everything else is valid
                 _ => cursor.advance(),
             };
@@ -275,6 +280,9 @@ enum Class {
 
     #[bytes(b'-')]
     Dash,
+
+    #[bytes(b'$')]
+    Dollar,
 
     #[bytes_range(b'a'..=b'z')]
     AlphaLower,
--- a/crates/oxide/src/extractor/arbitrary_property_machine.rs
+++ b/crates/oxide/src/extractor/arbitrary_property_machine.rs
@@ -411,4 +411,26 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_exceptions() {
+        for (input, expected) in [
+            // JS string interpolation
+            // In key
+            ("[${x}:value]", vec![]),
+            // As part of the key
+            ("[background-${property}:value]", vec![]),
+            // In value
+            ("[key:${x}]", vec![]),
+            // As part of the value
+            ("[key:value-${x}]", vec![]),
+            // Allowed in strings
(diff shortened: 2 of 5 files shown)


philipp-spiess (approved):
Nice one!

philipp-spiess:
@RobinMalfait Here's another issue that is resolved by this I think: https://github.com/tailwindlabs/tailwindcss/issues/15853

thecrypticace (approved):

PR #17153, opened 2025-03-12
Fix variants with `<digit>.</digit>` are extracted correctly

### LINKED ISSUE #17148: Version > 4.0.11 do not recognize dots in the custom screens breakpoints
It seems that from version `4.0.12` (included), any custom screen breakpoint (for example `1.5md`) that contains a dot is not recognized.

For example: v4.0.13

For example: Vite 5.4.14

For example: v22

For example: Chrome

For example: macOS

**Describe your issue**

Using a `tailwind.config.ts` file, and using the config below it does not recognize the breakpoints with a dot. It was doing it for the previous version.

```typescript
export default {
        ...
	theme: {
		...
		screens: {
			...
			// custom viewports
			'1.5md': '896px',
			'1.5lg': '1152px',
			'1.5xl': '1408px',
			'2.5xl': '1728px',
		},
		...
	},
	...
} satisfies Config;

```

RobinMalfait:
This PR fixes an issue where if you use a number with a decimal in a variant then it wasn't picked up correctly.

E.g.:
```
<div class="2xl:flex 1.5xl:flex"></div>
            ^^^^^^^^                        Picked up
                     ^^^^^^^^^^             Not picket up
```

This PR fixes that behavior by applying the same rules for utilities where a `.` is valid if it is surrounded by numbers.

# Test plan

1. Added test to ensure this is picked up
2. Existing tests pass
3. Ran the extractor on a real example with the following results:

| Before | After |
| --- | --- |
| image | image |

They are crossed out just because it's not a default value we know in the system, but you can see that the `1.` part is also extracted now.

Fixes: #17148

### DIFF
--- a/crates/oxide/src/extractor/named_variant_machine.rs
+++ b/crates/oxide/src/extractor/named_variant_machine.rs
@@ -391,6 +391,8 @@ mod tests {
                 vec!["group-[data-state=pending]/name:"],
             ),
             ("supports-(--foo)/name:flex", vec!["supports-(--foo)/name:"]),
+            // Odd media queries
+            ("1.5xl:flex", vec!["1.5xl:"]),
             // Container queries
             ("@md:flex", vec!["@md:"]),
             ("@max-md:flex", vec!["@max-md:"]),
--- a/crates/oxide/src/extractor/named_variant_machine.rs
+++ b/crates/oxide/src/extractor/named_variant_machine.rs
@@ -237,8 +237,8 @@ impl Machine for NamedVariantMachine<ParsingState> {
 
                 // A dot must be surrounded by numbers
                 //
-                // E.g.: `px-2.5`
-                //           ^^^
+                // E.g.: `2.5xl:flex`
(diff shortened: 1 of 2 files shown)


crates/oxide/src/extractor/named_variant_machine.rs
```diff
@@ -235,6 +235,22 @@ impl Machine for NamedVariantMachine<ParsingState> {
                 //             ^
                 Class::Colon => return self.done(self.start_pos, cursor),
 
+                // A dot must be surrounded by numbers
+                //
+                // E.g.: `px-2.5`
```
adamwathan, line 240: Should we use a variant example here instead of a utility value example?

RobinMalfait: Copy paste before going to the gym oops. Will fix, thanks!

thecrypticace (approved):
lgtm (aside from the comment)

PR #17247, opened 2025-03-17
Don't use `color-mix(…)` on `currentColor`

### LINKED ISSUE #17194: Safari 16.4 crash on `<input placeholder>` or `text-current` + opacity

v4.0.14

iOS Safari 16.4

**Reproduction URL**

https://play.tailwindcss.com/K2JSaOKDJX

```html
<p class="text-current/50">text</p>
```

https://play.tailwindcss.com/V9jkBlLwU8

```html
<input placeholder="text" />
```

**Describe your issue**

Both of the examples above crash the webpage in Safari 16.4:

Image

I don't know the full range of affected versions.

The trigger seems to be passing `currentColor` to `color-mix()`, I found more information about it here:

https://stackoverflow.com/questions/76436497/the-color-mix-property-involving-currentcolor-causes-safari-to-crash

My second example above crashes because that exact combo is in `preflight.css`:

https://github.com/tailwindlabs/tailwindcss/blob/50562a9c94e16934bfa8608e471c87c0157f7731/packages/tailwindcss/preflight.css#L291-L294

Maybe something like `color(from currentColor srgb r g b / 50%)` would work instead.

philipp-spiess:
Closes #17194

This PR works around a crash when rendering opacity on `currentColor` (as used by the placeholder styles in preflight) on Safari 16.4 and Safari 16.5. Unfortunately it seems that the [`color-mix(…)` function is not compatible with `currentColor` for these versions of Safari](https://stackoverflow.com/questions/76436497/the-color-mix-property-involving-currentcolor-causes-safari-to-crash). We tried a few different ways to work around this without success:

- Using an `@supports` media query to target these Safari versions and overwriting the placeholder still makes these browsers crash.
- Changing the way we apply opacity to `currentColor` in core doesn't seem to work for non-placeholder values: https://github.com/tailwindlabs/tailwindcss/issues/17194#issuecomment-2728949181 However, a wrong opacity is still better than a complete browser crash.

The work-around of using the `oklab(…)` function does seem to work for `::placeholder` styles in preflight though according to our testing so this PR applies this change to preflight. 

## Test plan

- See https://play.tailwindcss.com/WSsSTLHu8h?file=css 
- Tested on Chrome/Safari 16.4/Safari 18.3/Firefox

Screenshot 2025-03-17 at 11 32 47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix incorrect angle in `-bg-conic-*` utilities ([#17174](https://github.com/tailwindlabs/tailwindcss/pull/17174))
 - Fix `border-[12px_4px]` being interpreted as a `border-color` instead of a `border-width` ([#17248](https://github.com/tailwindlabs/tailwindcss/pull/17248))
+- Use the `oklab(…)` function when applying opacity to `currentColor` to work around a crash in Safari 16.4 and 16.5 ([#17247](https://github.com/tailwindlabs/tailwindcss/pull/17247))
 
 ## [4.0.14] - 2025-03-13
 
--- a/integrations/vite/astro.test.ts
+++ b/integrations/vite/astro.test.ts
@@ -19,7 +19,10 @@ test(
         import { defineConfig } from 'astro/config'
 
         // https://astro.build/config
-        export default defineConfig({ vite: { plugins: [tailwindcss()] } })
+        export default defineConfig({
+          vite: { plugins: [tailwindcss()] },
+          build: { inlineStylesheets: 'never' },
+        })
       `,
       'src/pages/index.astro': html`
        <div class="underline">Hello, world!</div>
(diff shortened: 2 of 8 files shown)


RobinMalfait (commented):
Do we also want to apply this for other usages of `currentColor`?

E.g.: `bg-current/20` would still use `color-mix(…)` I think

We might be able to replace our `withAlpha` helper internally to use the same trick here if it's used with `currentColor`

philipp-spiess:
@RobinMalfait Yeah I'm not sure. I guess it's better than hard crashing but the fix does not appear to work as expected in those cases (see PR description). It also won't work with CSS or theme variables set to `currentColor`, only when the literal value is used. That said, prob should still do it because the alternative is literally broken CSS :/ WDYT?

RobinMalfait:
Good morning! Oops completely read past that part and just looked at the code 🙈.

That's really strange that it doesn't work in other spots. Really annoying that it also still crashes when using a `@supports` query. So if I understand correctly:

1. `color-mix(…)` with `currentColor` -> crash
1. `oklab(…)` workaround with `currentColor` in `::placeholder` -> works
1. `oklab(…)` workaround with `currentColor` in anywhere else -> no crash, but doesn't work visually either

If that's true, then I think we should update it to `oklab` to get rid of the crash. Just need to verify that it works in other browsers as expected because if `bg-current/20` stopped working in other browsers that would be really sad.

My initial testing says that that still works: https://play.tailwindcss.com/adWyxqL0Ns?file=css 🤞

philipp-spiess:
@RobinMalfait Yep, if you take a look at my test plan, I tested on Chrome/Safari 16.4/Safari 18.3/Firefox and also found it working everywhere (even on latest Safari) as expected. Just old Safari does not apply the opacity which IMO is, like you said, better than a hard-crash.

RobinMalfait (approved):
Once CI passes this is good 👍

PR #17147, opened 2025-03-12
Add `@source inline(…)`

philipp-spiess:
This PR adds a new experimental feature that can be used to force-inline utilities based on an input string. The idea is that all utilities matching the source string will be included in your CSS:

```css
/* input.css */
@source inline('underline');

/* output.css */
.underline {
   text-decoration: underline;
}
```

Additionally, the source input is brace-expanded, meaning you can use one line to inline a whole namespace easily:
```css
/* input.css */
@source inline('{hover:,}bg-red-{50,{100..900..100},950}');

/* output.css */
.bg-red-50 {
  background-color: var(--color-red-50);
}
.bg-red-100 {
  background-color: var(--color-red-100);
}
.bg-red-200 {
  background-color: var(--color-red-200);
}
.bg-red-300 {
  background-color: var(--color-red-300);
}
.bg-red-400 {
  background-color: var(--color-red-400);
}
.bg-red-500 {
  background-color: var(--color-red-500);
}
.bg-red-600 {
  background-color: var(--color-red-600);
}
.bg-red-700 {
  background-color: var(--color-red-700);
}
.bg-red-800 {
  background-color: var(--color-red-800);
}
.bg-red-900 {
  background-color: var(--color-red-900);
}
.bg-red-950 {
  background-color: var(--color-red-950);
}
@media (hover: hover) {
  .hover\\:bg-red-50:hover {
    background-color: var(--color-red-50);
  }
  .hover\\:bg-red-100:hover {
    background-color: var(--color-red-100);
  }
  .hover\\:bg-red-200:hover {
    background-color: var(--color-red-200);
  }
  .hover\\:bg-red-300:hover {
    background-color: var(--color-red-300);
  }
  .hover\\:bg-red-400:hover {
    background-color: var(--color-red-400);
  }
  .hover\\:bg-red-500:hover {
    background-color: var(--color-red-500);
  }
  .hover\\:bg-red-600:hover {
    background-color: var(--color-red-600);
  }
  .hover\\:bg-red-700:hover {
    background-color: var(--color-red-700);
  }
  .hover\\:bg-red-800:hover {
    background-color: var(--color-red-800);
  }
  .hover\\:bg-red-900:hover {
    background-color: var(--color-red-900);
  }
  .hover\\:bg-red-950:hover {
    background-color: var(--color-red-950);
  }
}

```

[...]

PR #17255, opened 2025-03-17
Add `@source not` support

### LINKED ISSUE #15684: [v4] Tailwind CLI ignores .gitignore in watch mode

v4.0.0-beta.9

@tailwindcss/cli@4.0.0-beta.9

$ node --version
v23.5.0

N/A

macOS

**Reproduction URL**

```sh
git clone https://gist.github.com/condemil/b9990c17d419a0d9d1f0cc0f637fdb3c tw4-watch-gitignore-repro
```

**Describe your issue**

When tailwindcss cli is running in watch mode **and** ignored file is modified the output will contain css from the ignored file:

To reproduce:

1. Run the tailwindcss cli in watch mode

```sh
$ DEBUG=1 npx @tailwindcss/cli@4.0.0-beta.9 -i input.css -o output.css -w
≈ tailwindcss v4.0.0-beta.9

Done in 35ms

[36.26ms] [@tailwindcss/cli] (initial build)
[ 9.24ms]   ↳ Setup compiler
[ 7.97ms]   ↳ Scan for candidates
[10.85ms]   ↳ Build CSS
[ 1.84ms]   ↳ Write output
```

2. Check that `output.css` doesn't have `h-100`

```sh
$ cat output.css | grep h-1
$
```

3. Modify `test.go` and observe tailwindcss cli log

```sh
Done in 11ms

[11.25ms] [@tailwindcss/cli] (watcher)
[ 1.66ms]   ↳ Scan for candidates
[ 7.72ms]   ↳ Build CSS
[ 1.64ms]   ↳ Write output
```

4. Check that `output.css` now have `h-100`

```sh
$ cat output.css | grep h-1
  .h-100 {
```

### LINKED ISSUE #15882: [v4] Tailwind running with Next v15 on Docker takes ton of time and resources for first load

v4.0.0

Next.js v15.1.4

v23.0.0

N/A

MacOS

[**Reproduction URL**](https://github.com/lentscode/tailwind-issue-replication)

I have an issue running Next v15 + Tailwind v4 on a Docker container: the first compiling takes a huge amount of time compared to running `yarn dev` on local. Furthermore, it takes a ton of CPU and memory just for the first rendering.

**Local `yarn dev`**
Image

**Docker container**
Image

**`docker stats`**
Image

### LINKED ISSUE #15943: [v4] Ignore certain paths from scanning no longer supported

v4.0.0

Vite 6

22.13

**Describe your issue**

In v3 one could add a line with an exclamation mark `!` prefix in the `tailwind.config.js` to exclude certain files or paths from being scanned:

```js
    content: [
        './resources/views/**/*.blade.php',
        '!./resources/views/somefoldername/*',
    ],
```

but that option seems to have vanished in v4. I know I can do `@import 'tailwindcss' source(none);` and then  `@source '../../../resources/views';` which works but there is no way to exclude a subfolder from that path.

One work around is to add all the subfolders by hand, but that's a bit of a PITA when the project grows and you forget you need to add every view subfolder to the config css now.

### LINKED ISSUE #16669: @source doesn't work if dist/ is in the gitignore.

v4.0.7

NextJS 15.1.7

Node 23

Any browser

MacOS (Linux with Docker)

**Reproduction URL**

https://github.com/[adambirds/tailwind-issue-reproduction](https://github.com/adambirds/tailwind-issue-reproduction)

The reproduction may not use those exact versions as was put together quickly, and my main app is dockerised.

**Describe your issue**

If `dist/` is in your `.gitignore` file, it causes the `@source` directive to have issues when trying to tell tailwind to look for source files in a package in `node_modules`. Using my reproduction repo, you can edit `globals.css` to cause the issue to appear.

If set to the below it doesn't work even though it should.

```css
@import 'tailwindcss';
@source '../../node_modules/tablewind';
```

If set to the below it does, but this isn't desired to have to append the `/dist`:

```css
@import 'tailwindcss';
@source '../../node_modules/tablewind/dist';
```

The table in the repro app won't actually work as it requires a complex setup, but you can if the styling works as it will either look like this:

Working:
Image

Not Working:
Image

If you remove `dist/` from the .gitignore it works, but it should be possible to have this in there.

It was discussed with @adamwathan here: https://discord.com/channels/486935104384532500/1341775942846255145

### LINKED ISSUE #16920: .gitignore is too powerful and cannot be stopped

v4.0.9

postcss: 8.5.3

Node: v22.11.0
Bun: 1.2.5

N/A

Linux

**Describe your issue**

I have a mulitple wildcard patterns like `special-*.hbs` in my `.gitignore` and I cannot find a way to tell tailwind v4 to ignore my gitignore or to be able to specify an `@source "../src/special-*.hbs"`

Let me explain my process. My toolset copies a bunch of templates into my users site. They match `special-*.hbs`. These are in the `.gitignore` because they are not intended to be checked into their repository.

I have tried `@import "tailwindcss" source(none);` then specify `@source "../src";` but the `.gitignore` is still in effect, which I thought would get turned off using the `source(none)`.

What can I do here?

### LINKED ISSUE #17058: Plugin in Standalone CLI is picked up entirely.
v4.0.12

For example: v4.0.6

Ruby on Rails
[tailwindcss-rails](https://github.com/rails/tailwindcss-rails) which uses the Standalone CLI

None

N/A

macOS 15.2

**Reproduction URL**

https://github.com/pinzonjulian/daisy-bug

**Describe your issue**

When installing DaysiUI in a brand new Rails app using the [Tailwind Standalone CLI](https://tailwindcss.com/blog/standalone-cli) via tailwind-rails and the [Bundle File instructions in the docs](https://daisyui.com/docs/install/rails/) is not having the desired effect. All of DaisyUI seems to be added to the app/assets/builds/tailwind.css output file.

My theory is that Tailwind is interpreting the daisyui.js file as a source file of the project because it doesn't fall into any of the rules of files [ignored by Tailwind's scanner](https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned).

The resulting file weighs 177kb. If I comment @plugin ./daisiui.js the file weighs 8kb.

I recently opened an issue in DaisyUI due to this same problem and it was determined that the source of it is most likely Tailwind.  https://github.com/saadeghi/daisyui/issues/3621

To fix the issue I had to ignore all sources and add them manually
https://github.com/pinzonjulian/daisy-bug/compare/main...set-explicit-sources

It seems to me that Tailwind could automatically ignore plugin's path to prevent this from happening. If not, users are forced to list all paths manually which kind of goes against the objective of the new engine.

Thanks!

### LINKED ISSUE #17318: Performance issue with versions greater than >4.0.9

v4.0.15

Next 15.2.3

Node 22

N/A

Vercel

**Reproduction URL**

N/A

**Describe your issue**

Recently I have updated tailwindcss from v4.0.12 in one of my private project, the project is built on vercel, after updating the project takes a lot of time to build

Image

just tailwindcss and tailwindcss/postcss are updated from 4.0.9 to 4.0.15. no other changes.
After downgrading tailwindcss/postcss to v4.0.9, the builds are back to normal. In some project with considerable size, the build never finishes on vercel. 

Image

Note: There are no code changes between the upgrades and downgrades.

Works fine locally though, maybe because of the compute is higher on local machine.

RobinMalfait:
This PR adds a new source detection feature: `@source not "…"`. It can be used to exclude files specifically from your source configuration without having to think about creating a rule that matches all but the requested file:

```css
@import "tailwindcss";
@source not "../src/my-tailwind-js-plugin.js";
```

While working on this feature, we noticed that there are multiple places with different heuristics we used to scan the file system. These are:

- Auto source detection (so the default configuration or an `@source "./my-dir"`)
- Custom sources ( e.g. `@source "./**/*.bin"` — these contain file extensions)
- The code to detect updates on the file system

Because of the different heuristics, we were able to construct failing cases (e.g. when you create a new file into `my-dir` that would be thrown out by auto-source detection, it'd would actually be scanned). We were also leaving a lot of performance on the table as the file system is traversed multiple times for certain problems.

To resolve these issues, we're now unifying all of these systems into one `ignore` crate walker setup. We also implemented features like auto-source-detection and the `not` flag as additional _gitignore_ rules only, avoid the need for a lot of custom code needed to make decisions. 

High level, this is what happens after the now:

- We collect all non-negative `@source` rules into a list of _roots_ (that is the source directory for this rule) and optional _globs_ (that is the actual rules for files in this file). For custom sources (i.e with a custom `glob`), we add an allowlist rule to the gitignore setup, so that we can be sure these files are always included.
- For every negative `@source` rule, we create respective ignore rules.
- Furthermore we have a custom filter that ensures files are only read if they have been changed since the last time they were read.

So, consider the following setup:

```css
/* packages/web/src/index.css */
@import "tailwindcss";
@source "../../lib/ui/**/*.bin";
@source not "../../lib/ui/expensive.bin";
```

This creates a git ignore file that (simplified) looks like this:

```gitignore
# Auto-source rules
*.{exe,node,bin,…}
*.{css,scss,sass,…}
{node_modules,git}/

# Custom sources can overwrite auto-source rules
!lib/ui/**/*.bin

# Negative rules
lib/ui/expensive.bin
```

We then use this information _on top of your existing `.gitignore` setup_ to resolve files (i.e so if your `.gitignore` contains rules e.g. `dist/` this line is going to be added _before_ any of the rules lined out in the example above. This allows negative rules to allow-list your `.gitignore` rules.

To implement this, we're rely on the `ignore` crate but we had to make various changes, very specific, to it so we decided to fork the crate. All changes are prefixed with a `// CHANGED:` block but here are the most-important ones:

- We added a way to add custom ignore rules that _extend_ (rather than overwrite) your existing `.gitignore` rules
- We updated the order in which files are resolved and made it so that more-specific files can allow-list more generic ignore rules.
- We resolved various issues related to adding more than one base path to the traversal and ensured it works consistent for Linux, macOS, and Windows. 

## Behavioral changes

1. Any custom glob defined via `@source` now wins over your `.gitignore` file and the auto-content rules.
   - Resolves #16920
3. The `node_modules` and `.git` folders as well as the `.gitignore` file are now ignored by default (but can be overridden by an explicit `@source` rule).
   - Resolves #17318
   - Resolves #15882
4. Source paths into ignored-by-default folders (like `node_modules`) now also win over your `.gitignore` configuration and auto-content rules. 
    -  Resolves #16669
5. Introduced `@source not "…"` to negate any previous rules.
   - Resolves #17058
6. Negative `content` rules in your legacy JavaScript configuration (e.g. `content: ['!./src']`) now work with v4.
   - Resolves #15943 
7. The order of `@source` definitions matter now, because you can technically include or negate previous rules. This is similar to your `.gitingore` file.
9. Rebuilds in watch mode now take the `@source` configuration into account
   - Resolves #15684

## Combining with other features

Note that the `not` flag is also already compatible with [`@source inline(…)`](https://github.com/tailwindlabs/tailwindcss/pull/17147) added in an earlier commit:

```css
@import "tailwindcss";
@source not inline("container");
```

## Test plan

- We added a bunch of oxide unit tests to ensure that the right files are scanned
- We updated the existing integration tests with new `@source not "…"` specific examples and updated the existing tests to match the subtle behavior changes
- We also added a new special tag `[ci-all]` that, when added to the description of a PR, causes the PR to run unit and integration tests on all operating systems.

[ci-all]


vaynevayne:
Supporting glob or regular expression would be better than adding not or multi-line source, right?

crates/ignore/src/dir.rs
```diff
@@ -496,24 +488,47 @@ impl Ignore {
...
+                    continue;
+                }
+            } else if remaining.iter().any(|other| other.is_ignore()) {
+                continue;
+            }
+
+            return check.clone();
+        }
+
+        m_explicit
```
thecrypticace, line 525: If you flip the list order the code becomes a lot simpler and a lot clearer:
```rust
// CHANGED: We added logic to configure an order in which the ignore files are respected and
// allowed a whitelist in a later file to overrule a block on an earlier file.
let order = [
    // Manually added ignores
    &m_explicit,
    // .custom-ignore
    &m_custom_ignore,
    // .ignore
    &m_ignore,
    // .gitignore
    &m_gi,
    // .git/info/exclude
    &m_gi_exclude,
    // Global gitignore
    &m_global,
];

for check in order.into_iter() {
  if check.is_none() {
    continue
  }

  return check.clone();
}

m_explicit
```

You can even tweak that last part to something like this if you wanted:
```rust
let check = order.into_iter()
  .filter(|check| !check.is_none())
  .next()
  .unwrap_or(&m_explicit);

check.clone()
```

But

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
         # Exclude windows and macos from being built on feature branches
         on-main-branch:
-          - ${{ github.ref == 'refs/heads/main' }}
+          - ${{ github.ref == 'refs/heads/main' || contains(github.event.pull_request.body, '[ci-all]') }}
```
thecrypticace, line 29: I kinda feel like `on-main-branch` is the wrong name now 🤷‍♂️

crates/oxide/src/glob.rs
```diff
@@ -1,11 +1,117 @@
...
+/// A folder will be moved to the `base` part, and the `pattern` will be set to `**/*`.
+///
+/// ```diff
+/// - { base: '/', pattern: 'src/examples'}
+/// + { base: '/src/examples', pattern: '**/*'}
+/// ```
+///
+/// In addition, we will canonicalize the base path so we always work with the correctly resolved
+/// path.
+pub fn optimize_public_source_entry(source: &mut PublicSourceEntry) {
```
thecrypticace, line 39: I think this should be a method in `impl PublicSourceEntry`

crates/oxide/src/glob.rs
```diff
@@ -1,11 +1,117 @@
...
+    // Resolve base path immediately
+    let Ok(base) = dunce::canonicalize(&source.base) else {
+        event!(Level::ERROR, "Failed to resolve base: {:?}", source.base);
+        return;
+    };
+    source.base = base.to_string_lossy().to_string();
```
thecrypticace, line 45: This might be kinda strange to do but should we add a test for the non-utf-8 file path case? (idk if that's even feasible)

crates/oxide/src/glob.rs
```diff
@@ -1,11 +1,117 @@
...
+    };
+
+    // TODO: If the base does not exist on disk, try removing the last slash and try again.
+    let base = match dunce::canonicalize(&base) {
+        Ok(base) => base,
+        Err(err) => {
+            event!(tracing::Level::ERROR, "Failed to resolve glob: {:?}", err);
+            return;
+        }
+    };
```
thecrypticace, line 97: nit: this runs canonicalize twice on what is likely the same string in the case of the pattern being `**/*` or `*.js` or whatever

Maybe something we can improve? Not a big deal tho. It's gonna hit the fs cache in the OS I'd imagine.

crates/oxide/src/glob.rs
```diff
@@ -132,10 +238,25 @@ pub fn optimize_patterns(entries: &Vec<GlobEntry>) -> Vec<GlobEntry> {
...
+//  Original input:
+//  - `../project-b/**/*.{html,js}`
+//
+//  Expanded input:
+//  - `../project-b/**/*.html`
+//  - `../project-b/**/*.js`
+//
+//  Split results in:
+//  - `("../project-b", "**/*.html")`
+//  - `("../project-b", "**/*.js")`
```
thecrypticace, line 251: This comment is misleading since `split_pattern` doesn't do brace expansion

crates/oxide/src/scanner/auto_source_detection.rs
```diff
@@ -0,0 +1,54 @@
...
+/// - Ignoring common content directories like `.git` and `node_modules`
+/// - Ignoring file extensions we definitely don't want to include like `.css` and `.scss`
+/// - Ignoring common binary file extensions like `.png` and `.jpg`
+/// - Ignoring common files like `yarn.lock` and `package-lock.json`
+///
+pub static RULES: sync::LazyLock<Gitignore> = sync::LazyLock::new(|| {
+    let mut builder = GitignoreBuilder::new("");
+
+    builder
+        .add_line(None, &format!("{{{}}}", IGNORED_CONTENT_DIRS.join(",")))
```
thecrypticace, line 17: Should this have a `/` at the end of the pattern?

crates/oxide/src/scanner/auto_source_detection.rs
```diff
@@ -0,0 +1,54 @@
...
+
+static IGNORED_FILES: sync::LazyLock<Vec<&'static str>> = sync::LazyLock::new(|| {
+    include_str!("fixtures/ignored-files.txt")
+        .trim()
+        .lines()
+        .collect()
+});
+
+static IGNORED_CONTENT_DIRS: sync::LazyLock<Vec<&'static str>> =
+    sync::LazyLock::new(|| vec![".git"]);
```
thecrypticace: I haven't looked at other commits (yet) but I think all of these `static …` can be inlined into the first closure since they're not `pub …`

crates/ignore/src/dir.rs
philipp-spiess: Oh, haha, yeah I think that'll do:

If manual ignores add an allowlist, it should take precedence over other ignores so the early returns work, if it defines an ignore list it will also take precedence so it also can early return, duh.

crates/oxide/tests/scanner.rs
```diff
@@ -38,101 +45,147 @@ mod scanner {
...
+            globs,
+            normalized_sources,
+            candidates,
+        }
     }
 
-    fn test(paths_with_content: &[(&str, &str)]) -> Vec<String> {
-        scan(paths_with_content).0
+    fn scan(paths_with_content: &[(&str, &str)]) -> ScanResult {
+        scan_with_globs(paths_with_content, vec!["@source '**/*'"])
```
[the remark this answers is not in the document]
philipp-spiess: yep only testing so we can copy-paste integration tests in there and test easily. Good point regarding `#[cfg(test)]`

crates/oxide/src/scanner/auto_source_detection.rs
philipp-spiess: Ah, yep I think so. We learned that the inclusion of a slash in the gitignore rule makes it relative to the location of `.gitignore` but a stray `/` at the end does not seem to do that, it just makes it clear that it's a dir 👍 

Screenshot 2025-03-25 at 11 56 55

thecrypticace: Yeah, the trailing `/` forces a directory match. So it doesn't match a _file_ named `node_modules` for instance but does match a directory named `node_modules`

thecrypticace (approved):

ur5us:
@RobinMalfait Curious, is there a way to inspect the generated ignore list and emit what file change lead to another rebuild? Even with that new `@source not "…";` directive and our `.gitnore` I’m seeing constant rebuilds in our Rails app whenever a page load happens. I assumed it’s happening due to log files being watched (which should be ignored as per `.gitignore` file) so I tried to add a `@source not` directive but no dice. Now I wonder whether there’s a potential bug with path resolution or yet another edge case not covered by the current detection mechanism. 

Our setup is more complex as we use 3 separate builds for the marketing site, the admin UI and the booking engine. It’s roughly the following structure:

```
.
├── app
│   └── assets
│       └── stylesheets
│           ├── admin
│           │   ├── application.tailwind.css
│           │   └── tailwind.config.js
│           ├── booking
│           │   ├── application.tailwind.css
│           │   └── tailwind.config.js
│           └── cms
│               ├── application.tailwind.css
│               └── tailwind.config.js
|…
├──log
```

The `.gitignore` file lives in the root folder `.`. The config files are a variation of 
```js
import base from "../../../../tailwind.config.js"

export default {
  presets: [base],
  content: [
    "./app/views/cms/**/*.erb",
    "./app/views/common/**/*.erb",
    "./app/views/layouts/cms*.erb",
    "./app/helpers/cms*.rb",
    "./app/helpers/cms/**/*.rb",
    "./app/assets/stylesheets/cms/application.tailwind.css",
    "./app/javascript/controllers/pages/**/*.{js,ts}",
  ],
};
```
with the `base` being 
```js
export default {
  theme: {
    …
  },
  prefix: "tw"
};
```

We’re migrating from v3 hence why there are so many `content: […]` configs. On that note, I assume those are ignored now entirely, or‽ In other words, we need to migrate to `@source …`

jclusso:
@ur5us did you figure anything out on this? Using `@source '../../'` or `@source not '../../log'` doesn't work for me at all. Doing `@import 'tailwindcss/utilities.css' layer(utilities) source('../../');` does though. I don't like that though.

ur5us:
@jclusso No, I have not. Due to time constraints I've had to stay on v3 and work on actual features. Might revisit in 1 – 2 months.

PR #17273, opened 2025-03-18
Ensure multiple `--value(…)` or `--modifier(…)` calls don't delete subsequent declarations

RobinMalfait:
This PR fixes a bug in the handling of `@utility`. Essentially if you had a declaration where you used a `--modifier(…)` _and_ a `--value(…)` and both caused the declaration to be removed, the declaration after the current one would be removed as well.

This happened because 2 reasons:

1. Once we removed the declaration when we handled `--modifier(…)`, we didn't stop the walk and kept going.
2. The `replaceWith(…)` code allows you to call the function multiple times but immediately mutates the AST. This means that if you call it multiple times that you are potentially removing / updating nodes followed by the current one.

E.g.:

```css
@utility mask-r-* {
  --mask-right: linear-gradient(to left, transparent, black --value(percentage));
  --mask-right: linear-gradient(
    to left,
    transparent calc(var(--spacing) * --modifier(integer)),
    black calc(var(--spacing) * --value(integer))
  );
  mask-image: var(--mask-linear), var(--mask-radial), var(--mask-conic);
}
```

If this is used as `mask-r-10%`, then the first definition of `--mask-right` is kept, but the second definition of `--mask-right` is deleted because both `--modifier(integer)` and `--value(integer)` do not result in a valid value.

However, the `mask-image` declaration was also removed because the `replaceWith(…)` function was called twice. Once for `--modifier(integer)` and once for `--value(integer)`.

# Test plan

1. Added a test to cover this case.
2. Existing tests pass.
3. Added a hard error if we call `replaceWith(…)` multiple times.

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -149,9 +149,7 @@ export function walk(
         context,
         path,
         replaceWith(newNode) {
-          if (replacedNode) {
-            throw new Error('Cannot replace a node more than once')
-          }
+          if (replacedNode) return
           replacedNode = true
 
           if (Array.isArray(newNode)) {
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -149,6 +149,9 @@ export function walk(
         context,
         path,
         replaceWith(newNode) {
+          if (replacedNode) {
+            throw new Error('Cannot replace a node more than once')
+          }
           replacedNode = true
 
           if (Array.isArray(newNode)) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pre-process `<template lang="…">` in Vue files ([#17252](https://github.com/tailwindlabs/tailwindcss/pull/17252))
 - Remove redundant `line-height: initial` from Preflight ([#15212](https://github.com/tailwindlabs/tailwindcss/pull/15212))
 - Prevent segfault when loaded in a worker thread on Linux ([#17276](https://github.com/tailwindlabs/tailwindcss/pull/17276))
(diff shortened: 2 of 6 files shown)


packages/tailwindcss/src/ast.ts
```diff
@@ -149,6 +149,9 @@ export function walk(
         context,
         path,
         replaceWith(newNode) {
+          if (replacedNode) {
+            throw new Error('Cannot replace a node more than once')
+          }
```
RobinMalfait, line 154: We could also silently ignore this but then we are just hiding a bug.

Maybe we could write a better error message here that this is a Tailwind internal bug? /cc @adamwathan

adamwathan: Can we just not throw here? Is there something the user can do that triggers this or does it only signal a bug in Tailwind?

RobinMalfait: Nothing the user can do and signals a bug in Tailwind for sure because we are the only ones using this API. 

Only issue about silently ignoring is that we wouldn't immediately know about an issue. And worse, maybe some APIs start relying on this odd behavior.

If we had thrown here than the bug you ran into would be obvious but would also break everything 🫣

thecrypticace: Just dropping in a suggestion here:
```
Cannot replace a node more than once. This is a bug in Tailwind CSS. Please open an issue with a reproduction at https://github.com/tailwindlabs/tailwindcss/issues
```

thecrypticace (approved):

packages/tailwindcss/src/ast.ts
adamwathan: I feel like there are probably a million places in the code where we could add conditionals like this if we really wanted right? Feels weird to me that this would be the first time we ever introduced a throw to tell us we made a mistake. I don’t think we’ve ever done this any other time we fixed a bug.

RobinMalfait: Yeah makes sense. I silently bailed so at least we are not mutating _other_ AST nodes.

PR #17278, opened 2025-03-18
Fix class extraction followed by `(` in Slim

### LINKED ISSUE #17277: fails to scan for class in slim file (ruby/rails) (p-8)

≈ tailwindcss v4.0.14

tailwindcss-rails (4.2.0)

v22.5.1

Chrome

macOS

**Reproduction URL**

Have an application.html.slim with:

```
  body.border-t-4.p-8(class="#{body_classes}" data-hotwire-native="#{hotwire_native_app?}" data-controller="update-time-zone")
```

**Describe your issue**

it doesn't create the p-8 class

RobinMalfait:
This PR fixes an issue where using the class shorthand in Slim templates, followed by an `(` results in the last class being ignored. 

E.g.:

```slim
body.border-t-4.p-8(class="#{body_classes}" data-hotwire-native="#{hotwire_native_app?}" data-controller="update-time-zone")
```

This is because we will eventually extract `p-8` but it's followed by an invalid boundary character `(`.

To solve this, we make sure to replace the `(` with a space. We already do a similar thing when the classes are followed by an `[`.

One caveat, we _can_ have `(` in our classes, like `bg-(--my-color)`. But in my testing this is not something that can be used in the shorthand version. 

E.g.:
```slim
div.bg-(--my-color)
```

Compiles to:
```html
<div --my-color="" class="bg-"></div>
```

So I didn't add any special handling for this. Even when trying to escape the `(`, `-` and `)` characters, it still doesn't work. E.g.:

```slim
div.bg-\(--my-color\)
```

Compiles to:
```html
<div class="bg-">\(--my-color\)</div>
```

# Test plan

1. Added test for the issue
2. Existing tests pass
3. Verified via the extractor tool:

| Before | After |
| --- | --- |
| image | image |

---

Fixes: #17277

### DIFF
--- a/crates/oxide/src/extractor/pre_processors/slim.rs
+++ b/crates/oxide/src/extractor/pre_processors/slim.rs
@@ -116,7 +116,7 @@ mod tests {
                 " bg-red-500 2xl:flex bg-green-200 3xl:flex",
             ),
             // Keep dots in strings
-            (r#"div(class="px-2.5")"#, r#"div(class="px-2.5")"#),
+            (r#"div(class="px-2.5")"#, r#"div class="px-2.5")"#),
             // Replace top-level `(a-z0-9)[` with `$1 `. E.g.: `.flex[x]` -> `.flex x]`
             (".text-xl.text-red-600[", " text-xl text-red-600 "),
             // But keep important brackets:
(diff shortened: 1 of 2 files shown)


crates/oxide/src/extractor/pre_processors/slim.rs
```diff
@@ -80,6 +80,23 @@ impl PreProcessor for Slim {
...
+                // This means that we need to replace all these `(` and `)` with spaces to make
+                // sure that we can extract the `p-8`.
+                //
+                // However, we also need to make sure that we keep the parens that are part of the
+                // utility class. E.g.: `bg-(--my-color)`.
+                b'(' if bracket_stack.is_empty() && !matches!(cursor.prev, b'-') => {
+                    result[cursor.pos] = b' ';
+                    bracket_stack.push(cursor.curr);
+                }
+
```
thecrypticace, line 99: idk if it matters but should we worry about `bg-red-500/(--my-var)` here too? If so we should do `!matches!(cursor.prev, b'-' | b'/')`

thecrypticace: Okay, answering my own question with a note: I think we can drop that matches check entirely based on the PR desc

This is because the check (after adding the `/`) makes these two tests I just wrote pass:
```rs
let input = r#"
    body.border-(--my-color).p-8(class="\#{body_classes}" data-hotwire-native="\#{hotwire_native_app?}" data-controller="update-time-zone")
"#;
Slim::test_extract_contains(input, vec!["border-(--my-color)", "p-8"]);

let input = r#"
    body.border-red-500/(--my-color).p-8(class="\#{body_classes}" data-hotwire-native="\#{hotwire_native_app?}" data-controller="update-time-zone")
"#;
Slim::test_extract_contains(input, vec!["border-red-500/(--my-color)", "p-8"]);
```

but as you mentioned this is invalid Slim syntax

RobinMalfait: Yep yep, added an additional test just so that this case is covered: https://github.com/tailwindlabs/tailwindcss/pull/17278/commits/1a9c242f4abb37dd15bcda6cb9b451bcb43c9665

RobinMalfait: Alright we _do_ need this for this situation: https://github.com/tailwindlabs/tailwindcss/pull/17278/commits/645db11673316e4bb2e173cca162c5d990c7562e

```slim
div class="bg-(--my-color) bg-(--my-color)/(--my-opacity)"
```

thecrypticace (approved):

PR #17304, opened 2025-03-20
Add support for literal values in `--value('…')` and `--modifier('…')`

RobinMalfait:
This PR adds support for literal values inside the `--value('…')` and `--modifier('…')` functions. This allows you to safelist some known values you want to use:

E.g.:

```css
@utility tab-* {
  tab-size: --value('revert', 'initial');
}
```

This allows you to use `tab-revert` and `tab-initial` for example.

### DIFF
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -4706,6 +4706,7 @@ export function createCssUtility(node: AtRule) {
   if (IS_VALID_FUNCTIONAL_UTILITY_NAME.test(name)) {
     // API:
     //
+    // - `--value('literal')`         resolves a literal named value
     // - `--value(number)`            resolves a bare value of type number
     // - `--value([number])`          resolves an arbitrary value of type number
     // - `--value(--color)`           resolves a theme value in the `color` namespace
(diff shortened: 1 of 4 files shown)


thecrypticace:
Can we either add these to suggest() or make a note to do so later?

RobinMalfait:
> Can we either add these to suggest() or make a note to do so later?

@thecrypticace good idea, added support for that 👍

packages/tailwindcss/src/utilities.ts
```diff
@@ -4937,16 +4956,23 @@ export function createCssUtility(node: AtRule) {
...
-              .keysInNamespaces(modifierThemeKeys)
-              .map((x) => x.replaceAll('_', '.')),
-          },
-        ] satisfies SuggestionGroup[]
+        let values = []
+        for (let value of valueLiterals) {
+          values.push(value)
+        }
+        for (let value of designSystem.theme.keysInNamespaces(valueThemeKeys)) {
+          values.push(value.replaceAll('_', '.'))
```
thecrypticace, line 4964: Didn't we drop the .replaceAll(…) thing elsewhere in the codebase or no? I forget… I just remember something about that changing at some point

RobinMalfait: Yep is covered in the `suggest` function itself.

thecrypticace (approved):

PR #17311, opened 2025-03-20
Add Input & Output check to CLI

rudiv:
Throw an error if the input and output file for the CLI are identical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-
 - Fix class extraction followed by `(` in Pug ([#17320](https://github.com/tailwindlabs/tailwindcss/pull/17320))
 - Vite: Ensure that updates to an imported CSS file are properly propagated after updating templates ([#17347](https://github.com/tailwindlabs/tailwindcss/pull/17347))
 - Pre process `Slim` templates embedded in Ruby files ([#17336](https://github.com/tailwindlabs/tailwindcss/pull/17336))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for literal values in `--value('…')` and `--modifier('…')` ([#17304](https://github.com/tailwindlabs/tailwindcss/pull/17304))
 - Add suggestions when `--spacing(--value(integer, number))` is used ([#17308](https://github.com/tailwindlabs/tailwindcss/pull/17308))
 
+### Fixed
+
+- Error when input and output files resolve to the same file when using the CLI ([#17311](https://github.com/tailwindlabs/tailwindcss/pull/17311))
+
 ### [4.0.15] - 2025-03-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -95,6 +95,15 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
...
   }
 
+  // Check if the input and output file paths are identical, otherwise return an
+  // error to the user.
+  if (args['--input'] === args['--output']) {
+    eprintln(header())
+    eprintln()
+    eprintln(`Specified input file ${highlight(relative(args['--input']))} and output file ${highlight(relative(args['--output']))} are identical.`)
+    process.exit(1)
+  }
```
philipp-spiess, line 105: Thanks for the contribution! Do you mind adding a test for it as well? Check out the `integrations/cli` folder for some examples 👍 

Also we might consider doing a `path.resolve(…)` on the input and output first so that it will warn even in these cases:

```
tailwindcss --input foo.js --output ./node_modules/../foo.js
```

rudiv: Didn't find any tests - will add them in.

On the second part, this runs after the `path.resolve()` evaluation so would already handle this case.

philipp-spiess: Oh, yep you're right, great!

rudiv: Added, and also added a test for input file not existing as that wasn't covered.

philipp-spiess (approved):

philipp-spiess:
Awesome, thank you so much 🙇

packages/@tailwindcss-cli/src/commands/build/index.ts
```diff
@@ -95,6 +95,15 @@ export async function handle(args: Result<ReturnType<typeof options>>) {
     }
   }
 
+  // Check if the input and output file paths are identical, otherwise return an
+  // error to the user.
+  if (args['--input'] === args['--output']) {
```
RobinMalfait, line 100: Before we merge this I want to check 1 small thing. Commenting here to prevent auto merge from happening.

RobinMalfait (changes requested):

packages/@tailwindcss-cli/src/commands/build/index.ts
RobinMalfait: So one edge case here is that `-` can be used by both the `--input` and `--output`, e.g.: `--input - --output -`. The `-` means "use stdin and stdout respectively". In this case we don't want the error to show up.

Added a failing integration test + a fix for this.

(I also ran prettier in a separate commit)

RobinMalfait (approved):

CHANGELOG.md
```diff
@@ -24,9 +24,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Vite: Ensure that updates to an imported CSS file are properly propagated after updating templates ([#17347](https://github.com/tailwindlabs/tailwindcss/pull/17347))
+
```
RobinMalfait, line 27:
```suggestion
```

RobinMalfait:
Thanks you! These changes will be available in the next release.

PR #17336, opened 2025-03-22
Pre process `Slim` templates embedded in Ruby files

### LINKED ISSUE #17334: Tailwind v4 is not seeing classes in slim templates

v4.0.14

I'm using Ruby on Rails with gem `tailwindcss-rails` v4.2.1

Chrome

macOS

**Reproduction URL**

Tell me if it's really needed, because it takes time to setup such example.

**Describe your issue**

I'm trying to upgrade to Tailwing V4, but new Tailwind doesn't see classes in my templates. Tailwind v3 was seeing classes successfully.

In my project I use [Slim Templates](https://slim-template.github.io/) with [view_component](https://github.com/ViewComponent/view_component) gem.

This works:

```ruby
class QweComponent < ApplicationComponent
  slim_template <<~SLIM
    button.rounded-full
      | Some text
  SLIM
end
```

after calling `bin/rails tailwindcss:build` class `rounded-full` was added to generated `tailwind.css` file.

But this doesn't work:

```ruby
class QweComponent < ApplicationComponent
  slim_template <<~SLIM
    button.rounded-full.qwe
      | Some text
    button.rounded-full(
      class="qwe"
    )
      | Some text
  SLIM
end
```

class `rounded-full` was not added to generated `tailwind.css` file. It happens because `rounded-full` is followed by `.` or `(`. I suppose the same happens with HAML templates, other similar templating language.

RobinMalfait:
This PR fixes an issue where embedded Slim templates inside of Ruby files are not pre processed because we pre process based on a file extension.

This PR also handles embedded SLIM templates using the following syntax:

```rb
slim_template <<~SLIM
  .flex
    .flex-1
      h1 Hello World
    .flex-1
      p This is a test
SLIM
```

~~As far as I can tell, this is only a Slim template thing and not a Haml template thing but I could be wrong here. See: https://viewcomponent.org/guide/templates.html#interpolations~~

The ViewComponent package handles anything that looks like `{lang}_template`, so the lang here will be used as the pre processing language for now.

Fixes: #17334

# Test plan

1. Added test for this
2. Existing tests pass
3. Made sure that the snippet from the issue works as expected:

Added an example where we have a `slim_template` and a `svelte_template` to prove that it embeds based on the language. I also added a `html_template` with Svelte syntax to really make sure that that _doesn't_ work.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Increase Standalone hardware compatibility on macOS x64 builds ([#17267](https://github.com/tailwindlabs/tailwindcss/pull/17267))
 - Ensure that the CSS file rebuilds if a new CSS variable is used from templates ([#17301](https://github.com/tailwindlabs/tailwindcss/pull/17301))
 - Fix class extraction followed by `(` in Pug ([#17320](https://github.com/tailwindlabs/tailwindcss/pull/17320))
+- Pre process `Slim` templates embedded in Ruby files ([#17336](https://github.com/tailwindlabs/tailwindcss/pull/17336))
 
 ### Changed
 
--- a/crates/oxide/Cargo.toml
+++ b/crates/oxide/Cargo.toml
@@ -19,6 +19,7 @@ bexpand = "1.2.0"
 fast-glob = "0.4.3"
 classification-macros = { path = "../classification-macros" }
 regex = "1.11.1"
+fancy-regex = "0.14.0"
 
 [dev-dependencies]
 tempfile = "3.13.0"
--- a/crates/oxide/src/extractor/pre_processors/ruby.rs
+++ b/crates/oxide/src/extractor/pre_processors/ruby.rs
@@ -5,11 +5,12 @@ use crate::extractor::bracket_stack;
 use crate::extractor::pre_processors::pre_processor::PreProcessor;
 use crate::pre_process_input;
 use bstr::ByteSlice;
-use regex::Regex;
+use fancy_regex::Regex;
 use std::sync;
 
-static SLIM_TEMPLATE_REGEX: sync::LazyLock<Regex> =
-    sync::LazyLock::new(|| Regex::new(r#"<<[-~]SLIM\n([\s\S]*?)SLIM"#).unwrap());
+static TEMPLATE_REGEX: sync::LazyLock<Regex> = sync::LazyLock::new(|| {
+    Regex::new(r#"\s*(.*)_template\s*<<[-~]?([A-Z]+)\n([\s\S]*?)\2"#).unwrap()
+});
 
 #[derive(Debug, Default)]
 pub struct Ruby;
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Increase Standalone hardware compatibility on macOS x64 builds ([#17267](https://github.com/tailwindlabs/tailwindcss/pull/17267))
 - Ensure that the CSS file rebuilds if a new CSS variable is used from templates ([#17301](https://github.com/tailwindlabs/tailwindcss/pull/17301))
 - Fix class extraction followed by `(` in Pug ([#17320](https://github.com/tailwindlabs/tailwindcss/pull/17320))
+- Pre process `Slim` templates embedded in Ruby files ([#17336](https://github.com/tailwindlabs/tailwindcss/pull/17336))
```
philipp-spiess, line 42: looks like this was added to the 4.0.15 section too 🙈

crates/oxide/Cargo.toml
```diff
@@ -19,6 +19,7 @@ bexpand = "1.2.0"
 fast-glob = "0.4.3"
 classification-macros = { path = "../classification-macros" }
 regex = "1.11.1"
+fancy-regex = "0.14.0"
```
philipp-spiess, line 22: I'm not sure why the `regex` library alone is not enough?

crates/oxide/src/extractor/pre_processors/ruby.rs
```diff
@@ -3,6 +3,14 @@
 use crate::cursor;
 use crate::extractor::bracket_stack;
 use crate::extractor::pre_processors::pre_processor::PreProcessor;
+use crate::pre_process_input;
+use bstr::ByteSlice;
+use fancy_regex::Regex;
+use std::sync;
+
+static TEMPLATE_REGEX: sync::LazyLock<Regex> = sync::LazyLock::new(|| {
+    Regex::new(r#"\s*(.*?)_template\s*<<[-~]?([A-Z]+?)\n([\s\S]*?)\2"#).unwrap()
```
philipp-spiess, line 12: TIL about `\2` 🤯

philipp-spiess (approved):

crates/oxide/Cargo.toml
RobinMalfait: It's because the `regex` crate doesn't support backreferences:
image


https://github.com/rust-lang/regex?tab=readme-ov-file#:~:text=This%20includes%2C%20but%20is%20not%20limited%20to%2C%20look%2Daround%20andbackreferences.%20In%20exchange%2C%20all%20regex%20searches%20in%20this%20crate%20have%20worst%20caseO(m%20*%20n)%20time%20complexity%2C%20where%20m%20is%20proportional%20to%20the%20size%20of%20the%20regexand%20n%20is%20proportional%20to%20the%20size%20of%20the%20string%20being%20searched.

But luckily `fancy-regex` does support it. But as the `regex` crate also shows is that it is technically much slower (in my testing it's negligible for our use case though)

CHANGELOG.md
RobinMalfait: Woopsie, yep, fixed

PR #17361, opened 2025-03-24
Improve arbitrary value validation when parsing candidates

### LINKED ISSUE #17357: Declaring multiple CSS variables doesn't work with Vite

4.0.15

Vite 6.2.0

v20.18.1

Chrome

MacOS

**Reproduction URL**

https://github.com/samrobbins85/tailwind-variables (Not working as intended)

https://play.tailwindcss.com/8tmNO1FjuA (Working as intended)

**Describe your issue**

When trying to declare two CSS variables in one set of square brackets, it works on Tailwind play but not with a vite dev server for the same code

e.g. `[--foo:1rem;--bar:2rem]` doesn't work but `[--variable:5rem]` does work locally but both do on tailwind play

thecrypticace:
Fixes #17357

This affects the CDN and Play

Now candidates like these no longer parse and emit CSS:
- `[--foo:1rem;--bar:2rem]`
- `[&{color:red}]:flex`
- `data-[a]{color:red}foo[a]:flex`

### DIFF
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -1,5 +1,6 @@
 import type { DesignSystem } from './design-system'
 import { decodeArbitraryValue } from './utils/decode-arbitrary-value'
+import { isValidArbitrary } from './utils/is-valid-arbitrary'
 import { segment } from './utils/segment'
 
 const COLON = 0x3a
@@ -326,6 +327,9 @@ export function* parseCandidate(input: string, designSystem: DesignSystem): Iter
     let property = baseWithoutModifier.slice(0, idx)
     let value = decodeArbitraryValue(baseWithoutModifier.slice(idx + 1))
 
+    // Values can't contain `;` or `}` characters at the top-level.
+    if (!isValidArbitrary(value)) return
+
     yield {
       kind: 'arbitrary',
       property,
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/candidate.ts
```diff
@@ -552,6 +564,9 @@ export function parseVariant(variant: string, designSystem: DesignSystem): Varia
 
     let selector = decodeArbitraryValue(variant.slice(1, -1))
```
philipp-spiess, line 565: kinda bothers me that this var is called `selector` and not `arbitraryValue` like in the 5 other places that look identical haha

philipp-spiess (approved):

philipp-spiess (changes requested):
`bg-[url(https://example.com?q=;)]` will be thrown out but works in postcss so we might allow this too

RobinMalfait:
Our parser doesn't handle it properly so needs a separate fix (same PR or different PR). But semicolon without quotes work:

https://jsfiddle.net/cg6v57mp/1/

In Play this works: https://play.tailwindcss.com/hCXYekpfy6
But this doesn't: https://play.tailwindcss.com/hIwEPHkssN

packages/tailwindcss/src/candidate.ts
RobinMalfait: I mean, in this case it _is_ a selector because of the variant :P 

```
[.foo]:…
```

RobinMalfait (approved):
Aha I see, inlined it. Makes sense 👍

philipp-spiess (approved):
LGTM!

PR #17362, opened 2025-03-24
Add `::-webkit-details-marker` pseudo to `marker` variant

thecrypticace:
See #17360

This PR updates the `marker` variant to also target the marker present in `<summary>` elements in WebKit browsers. Chromium uses `::marker` and is therefore already covered.

### DIFF
--- a/packages/tailwindcss/src/variants.ts
+++ b/packages/tailwindcss/src/variants.ts
@@ -622,7 +622,6 @@ export function createVariants(theme: Theme): Variants {
   staticVariant('first-letter', ['&::first-letter'])
   staticVariant('first-line', ['&::first-line'])
 
-  // TODO: Remove alpha vars or no?
   staticVariant('marker', [
     '& *::marker',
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/variants.ts
```diff
@@ -623,7 +623,12 @@ export function createVariants(theme: Theme): Variants {
   staticVariant('first-line', ['&::first-line'])
 
   // TODO: Remove alpha vars or no?
```
philipp-spiess, line 625: Is this still relevant? I don't even know what this comment is trying to say so we might wanna remove it lol

philipp-spiess (approved):
Nice!

packages/tailwindcss/src/variants.ts
philipp-spiess, line 625:
```suggestion
```

RobinMalfait (approved):

PR #17431, opened 2025-03-28
Rework scripting variants

thecrypticace:
This replaces the `scripting`, `scripting-none`, and `scripting-initial` variants with one `noscript` variant that matches the name of the HTML `<noscript>` element. The previous `scripting` can then be represented as `not-noscript` e.g. `not-noscript:flex`.


RobinMalfait (approved):
`noscript`, memories 🥺

PR #17433, opened 2025-03-28
Extract used CSS variables from `.css` files

### LINKED ISSUE #16904: [v4.0.9+] Variables are ignored when referenced within CSS Modules

### LINKED ISSUE #17429: Custom colors not working in CSS

v4.0.17

Vite 6.2.0

v22.13.1

Chrome

macOS

https://github.com/mirorauhala/tw-custom-properties

**Issue**

Referencing colors as `--color-*` properties from a CSS module file does not work unless the color utility is mentioned elsewhere. 

For example, if you define a color such as `--color-red-custom-1: red` and use it from a CSS module file, like this:
```css 
.red { background: var(--color-red-custom-1); }
```

The final build will not contain the custom property defined and the color won't show correctly unless `bg-red-custom-1` is used elsewhere, which will cause the property to be included in the build.

**Expected**

The colors to work. The docs outline the following:

> Colors are exposed as CSS variables in the --color-* namespace, so you can reference them in CSS with variables like --color-blue-500 and --color-pink-700:

https://tailwindcss.com/docs/colors#referencing-in-css

RobinMalfait:
This PR fixes an issue where CSS variables could be used in CSS modules, but where never emitted in your final CSS.

Some backstory, when Tailwind CSS v4 came out, we _always_ emitted all CSS variables whether they were used or not.

Later, we added an optimization where we only emit the CSS variables that were actually used. The definition of "used" in this case is:

1. Used in your CSS  file(s) — (we check the final CSS AST for this)
2. Used _somewhere_ in any of your source files (e.g.: a JavaScript file accessing a variable)

The issue this PR tries to solve is with the very first point. If you are using CSS modules, then every CSS file is processed separately. This is not a choice Tailwind CSS made, but how other build tooling works (like Vite for example).

To prevent emitting all of Tailwind's Preflight reset and all utilities per CSS file, you can use the `@reference` directive instead of repeating `@import "tailwindcss";`. This is explained here: https://tailwindcss.com/docs/compatibility#explicit-context-sharing 

But now we are just _referencing_ them, not emitting them. And since the CSS module is not connected in any way to the main `index.css` file that contains the `@import "tailwindcss";` directive, we don't even see the CSS variables while processing the `index.css` file. (or wherever your main CSS file is)

This is where point 2 from above comes in. This is a situation where we rely on the extractor to find the used CSS variables so we can internally mark them as used.

To finally get to the point of this PR, the extractor only scans `.html`, `.js`, ... files but not `.css` files. So all the CSS variables used inside of CSS modules will not be generated.

This PR changes that behavior to also scan `.css` files. But _only_ for CSS variables (not any other type of class candidate). This is important, otherwise all your custom `@utility foo {}` definitions would always mark `foo` as a used class and include it in the CSS which is not always the case.

On top extracting CSS variables, we will also make sure that the CSS variables we find are in usage positions (e.g.: `var(--color-red-500)`) and not in definition positions (e.g.: `--color-red-500: #ff0000;`). This is important because we only want to emit the variables that are actually used in the final CSS output.

One future improvement not implemented here, is that technically we will also extract CSS variables that might not be used if defined in a `@utility`.

```css
@utility never-used {
  color: var(--color-red-500); /* --color-red-500 will be emitted, even if it might not be used */
}
```

Fixes: #16904
Fixes: #17429

# Test plan

1. Added a test where CSS variables are defined in `.css` files (and ignored)
2. Added a test where CSS variables are used in `.css` files (and included)

Testing on the reproduction defined in #16904, the `.module.css` file contains a reference to `var(--color-hot-pink)`, but generating a build shows that the variable definition is not available:

image

When you run the build again with the changes from this PR, then we _do_ see the definition of the `--color-hot-pink` in the root CSS file:
image

### DIFF
--- a/crates/oxide/src/extractor/mod.rs
+++ b/crates/oxide/src/extractor/mod.rs
@@ -56,17 +56,15 @@ impl fmt::Display for Extracted<'_> {
 #[derive(Debug)]
 pub struct Extractor<'a> {
     cursor: cursor::Cursor<'a>,
-    extension: Option<&'a str>,
 
     css_variable_machine: CssVariableMachine,
     candidate_machine: CandidateMachine,
 }
 
 impl<'a> Extractor<'a> {
-    pub fn new(input: &'a [u8], extension: Option<&'a str>) -> Self {
+    pub fn new(input: &'a [u8]) -> Self {
         Self {
             cursor: cursor::Cursor::new(input),
-            extension,
 
             css_variable_machine: Default::default(),
             candidate_machine: Default::default(),
@@ -210,7 +208,7 @@ mod tests {
     }
 
     fn extract_sorted_candidates(input: &str) -> Vec<&str> {
-        let mut machine = Extractor::new(input.as_bytes(), None);
+        let mut machine = Extractor::new(input.as_bytes());
         let mut actual = machine
             .extract()
             .iter()
@@ -224,7 +222,7 @@ mod tests {
     }
 
     fn extract_sorted_css_variables(input: &str) -> Vec<&str> {
-        let mut machine = Extractor::new(input.as_bytes(), None);
+        let mut machine = Extractor::new(input.as_bytes());
         let mut actual = machine
             .extract()
             .iter()
@@ -289,12 +287,12 @@ mod tests {
             let input = include_bytes!("../fixtures/example.html");
 
             let throughput = Throughput::compute(iterations, input.len(), || {
-                let mut extractor = Extractor::new(input, None);
+                let mut extractor = Extractor::new(input);
                 _ = black_box(extractor.extract());
             });
             eprintln!("Extractor throughput: {:}", throughput);
 
-            let mut extractor = Extractor::new(input, None);
+            let mut extractor = Extractor::new(input);
             let start = std::time::Instant::now();
             _ = black_box(extractor.extract().len());
             let end = start.elapsed();
--- a/crates/oxide/src/extractor/mod.rs
+++ b/crates/oxide/src/extractor/mod.rs
@@ -56,15 +56,17 @@ impl fmt::Display for Extracted<'_> {
 #[derive(Debug)]
 pub struct Extractor<'a> {
     cursor: cursor::Cursor<'a>,
+    extension: Option<&'a str>,
 
     css_variable_machine: CssVariableMachine,
     candidate_machine: CandidateMachine,
 }
 
 impl<'a> Extractor<'a> {
-    pub fn new(input: &'a [u8]) -> Self {
+    pub fn new(input: &'a [u8], extension: Option<&'a str>) -> Self {
         Self {
             cursor: cursor::Cursor::new(input),
+            extension,
 
             css_variable_machine: Default::default(),
             candidate_machine: Default::default(),
@@ -208,7 +210,7 @@ mod tests {
     }
 
     fn extract_sorted_candidates(input: &str) -> Vec<&str> {
-        let mut machine = Extractor::new(input.as_bytes());
+        let mut machine = Extractor::new(input.as_bytes(), None);
         let mut actual = machine
             .extract()
             .iter()
@@ -222,7 +224,7 @@ mod tests {
     }
 
     fn extract_sorted_css_variables(input: &str) -> Vec<&str> {
-        let mut machine = Extractor::new(input.as_bytes());
+        let mut machine = Extractor::new(input.as_bytes(), None);
         let mut actual = machine
             .extract()
             .iter()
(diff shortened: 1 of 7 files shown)


crates/oxide/src/extractor/mod.rs
```diff
@@ -139,6 +140,41 @@ impl<'a> Extractor<'a> {
 
         extracted
     }
+
+    pub fn extract_css_variables_from_css_files(&mut self) -> Vec<Extracted<'a>> {
```
RobinMalfait, line 144: Bit of a mouthful, open to suggestions. Note: this is completely internal API.

thecrypticace: `extract_variables_from_css` if you want a shorter name but this is fine imo

crates/oxide/src/scanner/mod.rs
```diff
@@ -402,6 +425,44 @@ fn read_all_files(changed_content: Vec<ChangedContent>) -> Vec<Vec<u8>> {
         .collect()
 }
 
+#[tracing::instrument(skip_all)]
+fn extract_css_variables(blobs: Vec<Vec<u8>>) -> Vec<String> {
```
RobinMalfait, line 429: I started passing options to the other `parse_all_blobs` function but it looked a bit messy. So duplicated it instead (even though there is a good chunk of duplication going on).

thecrypticace (approved):

crates/oxide/src/extractor/mod.rs
```diff
@@ -139,6 +140,41 @@ impl<'a> Extractor<'a> {
 
         extracted
     }
+
+    pub fn extract_variables_from_css(&mut self) -> Vec<Extracted<'a>> {
```
philipp-spiess, line 144: Could this be done in a pre-processor where we already have logic to do that based on file extension?

crates/oxide/src/scanner/mod.rs
philipp-spiess: I feel like the only difference really should be that we no-op the `Candidate` machine though? Can you elaborate what you meant with messy?

philipp-spiess (commented):
I know this was already merged but still left some questions

crates/oxide/src/scanner/mod.rs
RobinMalfait: I started passing through the contents and the extension and it leaked everywhere. Can see that in this commit: https://github.com/tailwindlabs/tailwindcss/pull/17433/commits/7dc7878c037fc3909feb5666638c816606745263

Then I was thinking about an `ExtractorOptions` struct that we could pass in as well, but that also required checking the options in the `extract()` call at runtime. While that's not the end of the world, that's additional check we have to perform but it's a useless check for 99% of the files we scan. To make things a bit worse, the extract() function is called for every line in every file. So there could be a _lot_ of unnecessary checks.

So instead of checking it in this hot path, I created a separate function instead.

crates/oxide/src/extractor/mod.rs
RobinMalfait: Sort of but not really, we do have the contents and the extension so we do have all the information we need, but we return a new potentially transformed input file.

```rs
pub fn pre_process_input(content: &[u8], extension: &str) -> Vec<u8> { … }
```

But we don't emit extracted variables here. So we could update this logic to also emit `Vec<Extracted>`. 

Another solution is that we could transform the CSS file and replace everything that is _not_ a used CSS variable with whitespace. That way the normal extractor won't find anything, and the CSS extractor will only find used variables.

But both these solutions seem like a hack, and a bit of an abuse for this `pre_process_input` function 😅 unless you're thinking about something completely different?

crates/oxide/src/scanner/mod.rs
```diff
@@ -248,6 +265,12 @@ impl Scanner {
                     .and_then(|x| x.to_str())
                     .unwrap_or_default(); // In case the file has no extension
 
+                // Special handing for CSS files to extract CSS variables
+                if extension == "css" {
+                    self.css_files.push(path);
+                    continue;
+                }
```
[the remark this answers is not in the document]
philipp-spiess: Mind following up on that? Just so we don't forget

crates/oxide/src/extractor/mod.rs
philipp-spiess: Yeah I think this does make sense, agree that abusing pre-process for this does feel wrong too now that I think about it more.

crates/oxide/src/scanner/mod.rs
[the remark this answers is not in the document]
RobinMalfait: For future reference. Follow up PR: https://github.com/tailwindlabs/tailwindcss/pull/17467

tranvansang:
the bug remains for my project: an astro file includes an scss module.

The scss has `@reference "./global.css";`and uses css variables.

In `global.css`, if the css variables are defined in `:root{}` block => works.
In `global.css`, if the css variables are defined in `@theme{}` block => not work.

kilobyte2007:
In our project, we have historically used SASS, and the files have the `scss` extension. Is there any way to make this also handle them?

PR #17434, opened 2025-03-28
Add `drop-shadow-*` color support

thecrypticace:
This PR adds support for two things:

- New `drop-shadow-{color}` utilities which can be used to control the filder drop shadow color
- New `drop-shadow-*/{alpha}` utilities which can be used to control the intensity of the drop shadow 

Note that drop-shadow-* utilities without a modifier use variables from your theme.

If you opt into using drop shadow colors or drop-shadow "intensity" utilities we will inline these values into your CSS when using a drop shadow color or drop shadow intensity utility and as such can't be changed by changing the theme variables on a per-element/tree basis.


packages/tailwindcss/src/utilities.ts
```diff
@@ -4318,20 +4321,162 @@ export function createUtilities(theme: Theme) {
...
+              decl(
+                '--tw-drop-shadow-size',
+                segment(value, ',')
+                  .map(
+                    (v) =>
+                      `drop-shadow(${replaceShadowColors(
+                        v,
+                        alpha,
+                        (color) => `var(--tw-drop-shadow-color, ${color})`,
+                      )})`,
```
RobinMalfait, line 4392: I think the API changed here because this is passing 3 values instead of 2

RobinMalfait: Ah I think it's because I pressed the update button (because of conflicts in the CHANGELOG.md file)

RobinMalfait (approved):
Art.

image
image

PR #17435, opened 2025-03-28
Improve compatibility with Safari 15

philipp-spiess:
This PR improves the compatibility with Tailwind CSS v4 with unsupported browsers with the goal to greatly improve compatibility with Safari 15. 

To make this work, this PR makes the following changes to all code

- Change `oklab(…)` default theme values to use a percentage in the first place (so instead of `--color-red-500: oklch(0.637 0.237 25.331);` we now define it as `--color-red-500: oklch(63.7% 0.237 25.331);` since this syntax has much broader support on Safari).
- Polyfill `@property` with a `@supports` query targeting older versions of Safari and Firefox *
- Create fallbacks for the `color-mix(…)` function that use _inlined color values from your theme_ so that they can be computed a compile time by `lightningcss`. These fallbacks will convert to srgb to increase compatibility.
- Create fallbacks for the _relative color_ feature used in the new shadow utilities and using `color-mix(…)` in case _relative color_ is applied on `currentcolor` (due to limited browser support)
- Create fallbacks for gradient interpolation methods (e.g. to support `bg-linear-to-r/oklab`)
- Polyfill `@media` queries range syntax.

## A simplified example

Given this example CSS input:

```css
@import 'tailwindcss';
@source inline('from-cyan-500/50 bg-linear-45');
```

Here's the updated output CSS including the newly added polyfills and updated `oklab` values:

```css
.bg-linear-45 {
  --tw-gradient-position: 45deg;
  background-image: linear-gradient(var(--tw-gradient-stops));
}

@supports (background-image: linear-gradient(in lab, red, red)) {
  .bg-linear-45 {
    --tw-gradient-position: 45deg in oklab;
  }
}

.from-cyan-500\\/50 {
  --tw-gradient-from: oklab(71.5% -.11682 -.08247 / .5);
  --tw-gradient-stops: var(--tw-gradient-via-stops, var(--tw-gradient-position), var(--tw-gradient-from) var(--tw-gradient-from-position), var(--tw-gradient-to) var(--tw-gradient-to-position));
}

  .from-cyan-500\\/50 {
    --tw-gradient-from: color-mix(in oklab, var(--color-cyan-500) 50%, transparent);
  }
}

:root, :host {
  --color-cyan-500: oklch(71.5% .143 215.221);
}

@supports (((-webkit-hyphens: none)) and (not (margin-trim: 1lh))) or ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
  @layer base {
    *, :before, :after, ::backdrop {
      --tw-gradient-position: initial;
      --tw-gradient-from: #0000;
      --tw-gradient-via: #0000;
      --tw-gradient-to: #0000;
      --tw-gradient-stops: initial;
      --tw-gradient-via-stops: initial;
      --tw-gradient-from-position: 0%;
      --tw-gradient-via-position: 50%;
      --tw-gradient-to-position: 100%;
    }
  }
}

@property --tw-gradient-position {
  syntax: "*";
  inherits: false
}

@property --tw-gradient-from {
  syntax: "<color>";
  inherits: false;
  initial-value: #0000;
}

@property --tw-gradient-via {
  syntax: "<color>";
  inherits: false;
  initial-value: #0000;
}

@property --tw-gradient-to {
  syntax: "<color>";
  inherits: false;
  initial-value: #0000;
}

@property --tw-gradient-stops {
  syntax: "*";
  inherits: false
}

@property --tw-gradient-via-stops {
  syntax: "*";
  inherits: false
}

@property --tw-gradient-from-position {
  syntax: "<length-percentage>";
  inherits: false;
  initial-value: 0%;
}

@property --tw-gradient-via-position {
  syntax: "<length-percentage>";
  inherits: false;
  initial-value: 50%;
}

@property --tw-gradient-to-position {
  syntax: "<length-percentage>";
  inherits: false;
  initial-value: 100%;
}
```

## \* A note on `@property` polyfills and CSS modules

On Next.js, CSS module files are required to be _pure_, meaning that all selectors must either be scoped to a class or an ID. Fortunatnyl for us, this does not apply to `@property` rules which we've been using before to initialize CSS variables.

However, since we're now bringing back the `@property` polyfills, that would cause unexpected rules to be exported from the CSS file as this:

```css
@reference "tailwindcss";

.skew {
  @apply skew-7;
}
```

Would turn to the following file:

```css
.skew {
  /* … */
}
@supports (/*…*/) {
  @layer base {
    *, :before, :after, ::backdrop {
      --tw-gradient-position: initial;
    }
  }
}
@property /* … */ 
```

Notice that this adds a `*` selector which is not considered pure.

Unfortunately there is no way for us to silence this warning or work around it, as the dependency causing this errors ([`postcss-modules-local-by-default`](https://github.com/css-modules/postcss-modules-local-by-default)) is bundled into Next.js. To work around crashes, these polyfills will not apply to CSS modules processed by the PostCSS extension for now.

## Testing on tailwindcss.com

To see the changes in effect, take a look at this screencast that compares tailwindcss.com on iOS 15.5 with a version that has the patches of this PR applied:

https://github.com/user-attachments/assets/1279d6f5-3c63-4f30-839c-198a789f4292

## Open questions

- [x] Should we make the `@property` polyfill not require an `@layer base` and instead place the generated CSS on the top of the file?
- [x] Add fallbacks for newly introduced usages of the relative color syntax.
- [x] Decide if we're ok with how we actually implement the `@media` range query workaround

## Test plan

- [x] Tested on tailwindcss.com via a preview build: https://tailwindcss-com-git-legacy-browsers-tailwindlabs.vercel.app/
- [x] Updated tests
- [x] Ensure we also test on Chrome 111, Safari 16.4, Firefox 128 to make sure we have no regressions. Also tested on Safari 16.4, 15.5, 18.0


packages/tailwindcss/src/design-system.ts
```diff
@@ -94,7 +95,8 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
           },
         })
 
-        astNodes = optimizeAst(astNodes, designSystem)
+        // Disable all polyfills to not unnecessarily pollute IntelliSense output
+        astNodes = optimizeAst(astNodes, designSystem, Polyfills.None)
```
RobinMalfait, line 99: Nice, I like this, good thinking!

packages/tailwindcss/src/ast.ts
```diff
@@ -314,6 +323,52 @@ export function optimizeAst(ast: AstNode[], designSystem: DesignSystem) {
...
+
+          // Change the colorspace to `srgb` since the fallback values should not be represented as
+          // `oklab(…)` functions again as their support in Safari <16 is very limited.
+          let colorspace = node.nodes[2]
+          if (
+            (colorspace.kind === 'word' && colorspace.value === 'oklab') ||
+            colorspace.value === 'oklch' ||
+            colorspace.value === 'lab' ||
+            colorspace.value === 'lch'
+          ) {
```
RobinMalfait, line 355: While likely fine, we are only checking the type of the colorspace.kind to be `word` for the `oklab` value, not for all the other values.

I think we want this instead:
```suggestion
          if (
            colorspace.kind === 'word' &&
            (colorspace.value === 'oklab' ||
              colorspace.value === 'oklch' ||
              colorspace.value === 'lab' ||
              colorspace.value === 'lch')
          ) {
```

packages/tailwindcss/src/index.test.ts
```diff
@@ -4657,11 +4720,347 @@ describe('@variant', () => {
...
+      ),
+    ).resolves.toMatchInlineSnapshot(`
+      ":root, :host {
+        --my-half: 50%;
+        --color-red-500: oklch(63.7% .237 25.331);
+      }
+
+      .text-red-500\\/\\(--my-half\\) {
+        color: #fb2c3680;
+      }
```
RobinMalfait, line 4934: Sad that this means that you can't change the `--my-half` value at runtime, but if it's not supported what can you do haha.

packages/tailwindcss/src/index.ts
```diff
@@ -697,7 +708,7 @@ export async function compileAst(
       }
 
       if (!utilitiesNode) {
-        compiled ??= optimizeAst(ast, designSystem)
+        compiled ??= optimizeAst(ast, designSystem, opts.polyfills)
```
RobinMalfait, line 711: I wonder if we should store the `polyfills` on the `designSystem` itself 🤔

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -263,6 +274,7 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
               DEBUG && I.end('AST -> CSS')
 
               DEBUG && I.start('Lightning CSS')
+              // @ts-ignore
```
thecrypticace, line 277: Can we drop this now or nah?

playgrounds/nextjs/app/page.module.css
```diff
@@ -1,228 +1,5 @@
...
-@media (prefers-color-scheme: dark) {
-  .vercelLogo {
-    filter: invert(1);
-  }
-
-  .logo {
-    filter: invert(1) drop-shadow(0 0 0.3rem #ffffff70);
-  }
-}
+@reference "./globals.css";
```
RobinMalfait, line 1: Do we want to ship the changes to the playground here?

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -114,6 +121,10 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
               onDependency: (path) => {
                 context.fullRebuildPaths.push(path)
               },
+              // In CSS Module files, we have to disable the `@property` polyfill since these will
+              // emit global `*` rules which are considered to be non-pure and will cause builds
+              // to fail.
+              polyfills: isCSSModuleFile ? Polyfills.All ^ Polyfills.AtProperty : Polyfills.All,
```
thecrypticace, line 127: Do we need to do this in the Vite plugin too?

RobinMalfait (commented):
Haven't functionally tested yet, but wanted to leave these notes already.

packages/tailwindcss/src/ast.ts
```diff
@@ -349,6 +404,30 @@ export function optimizeAst(ast: AstNode[], designSystem: DesignSystem) {
         return
       }
 
+      // Collect fallbacks for `@property` rules for Firefox support
+      // We turn these into rules on `:root` or `*` and some pseudo-elements
+      // based on the value of `inherits``
```
thecrypticace, line 409:
```suggestion
      // based on the value of `inherits`
```

packages/tailwindcss/src/ast.ts
```diff
@@ -491,6 +570,37 @@ export function optimizeAst(ast: AstNode[], designSystem: DesignSystem) {
     }
   }
 
+  // Fallbacks
+  if (polyfills & Polyfills.AtProperty) {
+    {
```
thecrypticace, line 575: The extra`{…}` block is not necessary

packages/tailwindcss/src/design-system.ts
thecrypticace, line 99: 😍 thank you!

packages/tailwindcss/src/utilities.ts
```diff
@@ -185,7 +186,7 @@ export function withAlpha(value: string, alpha: string): string {
 /**
  * Apply opacity to a color using `color-mix`.
  */
-export function replaceAlpha(value: string, alpha: string | null): string {
+export function replaceAlpha(value: string, alpha: string): string {
```
thecrypticace, line 189: This type was there on purpose (see check immediately after)

packages/tailwindcss/src/utils/replace-shadow-colors.ts
```diff
@@ -33,21 +41,61 @@ export function replaceShadowColors(
...
+  if (requiresFallback) {
+    return [
+      decl(property, `${prefix}${fallbackShadows.join(', ')}`),
+      rule('@supports (color: lab(from red l a b))', [
+        decl(property, `${prefix}${shadows.join(', ')}`),
+      ]),
+    ]
+  } else {
+    return [decl(property, `${prefix}${shadows.join(', ')}`)]
+  }
```
thecrypticace, line 100: I'm not 100% sure the supports query is right. We use relative color syntax in `text-shadow/{alpha}` but *not* in `text-shadow-red-500/{alpha}`. Also, in general, I feel like making `replaceShadowColors` produce declarations isn't the right move.

I'd prefer we did something like:
- Keep `replaceShadowColors` doing it was doing before, taking in a value, and returning a value
- Detect the use of RCS and/or color-mix and adjust the supports query accordingly (def. open to other ideas tho)
- Produce the supports query & decls outside of `replaceShadowColors` — either duplicating the checks _or_ in a separate function so it can still be reused?

Thoughts?

packages/tailwindcss/src/index.test.ts
```diff
@@ -4657,11 +4720,347 @@ describe('@variant', () => {
...
+
+      @property --inherit-value {
+        syntax: "*";
+        inherits: true;
+        initial-value: red;
+      }"
+    `)
+  })
+
+  it('can be disabled to not emit fallbacks (necessary for CSS modules)', async () => {
```
thecrypticace, line 5012:
```suggestion
  it('emitting fallbacks can be disabled (necessary for CSS modules)', async () => {
```

packages/@tailwindcss-postcss/src/index.ts
philipp-spiess: Pretty sure this is a leftover from a wrong build, my bad!

packages/@tailwindcss-postcss/src/index.ts
philipp-spiess: Nope we don’t. Vote doesn’t do any sort of linting like this with CSS modules. I need to test with vite + lightnincss tomorrow though

packages/tailwindcss/src/ast.ts
philipp-spiess: Oh yep huh

playgrounds/nextjs/app/page.module.css
philipp-spiess: This wanted to make this work, the example was having contents that weren’t used in the next example and it also had a lot of unused css. Kept the module css file in since we need it from time to time but imo it’s simpler as before 👍

packages/tailwindcss/src/utilities.ts
philipp-spiess: Ah yep but im doing the test in the caller now. Can remove the check afterwards now!

packages/@tailwindcss-postcss/src/index.ts
philipp-spiess: Also no issue when the CSS transformer is set to lightningcss 🫡

packages/tailwindcss/src/utils/replace-shadow-colors.ts
philipp-spiess: Yeah hmm I think you're right I also found this way a bit sketchy. I think ideally this `replaceShadowColor` function does only that: replaces colors for a shadow. It will _not_ directly call `replaceAlpha` but a callback instead. Then, a caller can do that and use replaceAlpha if necessary and then also insert the fallback. let me try that

philipp-spiess: https://github.com/tailwindlabs/tailwindcss/pull/17435/commits/941f71314551dd9e4da711aaf7ae663f29466c31

thecrypticace (approved):

silvenon:
Sorry for the late comment, but it is my understanding that with this PR merged the general Safari support for basic Tailwind goes from 16.4 to 16.2 (`color-mix` is supported down to 16.2). Is that true?

sehmbimanvir:
I am facing issue with gradient if I defined a color in rgb format

e.g

```css
// index.css

@theme inline {
    --color-custom: rgba(111, 160, 0);
}
```

Now If, I am using the class for gradient with opacity e.g

```html
<div class="bg-linear-to-tr from-0% to-40% from-custom/20" />
```

It's not working. Do, I need to change color to hex?

sehmbimanvir:
@philipp-spiess , any help regarding this?

https://github.com/tailwindlabs/tailwindcss/pull/17435#issuecomment-2888558243

philipp-spiess:
@sehmbimanvir Do you have a repro please? I suppose this _should_ work: https://play.tailwindcss.com/aqmzubxchz The play generates a `color-mix(…)` function BUT in a production build, [lightningcss is able to transpile it away](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.bg-linear-to-tr%20%7B%5Cn%20%20--tw-gradient-position%3A%20to%20top%20right%3B%5Cn%20%20%40supports%20(background-image%3A%20linear-gradient(in%20lab%2C%20red%2C%20red))%20%7B%5Cn%20%20%20%20--tw-gradient-position%3A%20to%20top%20right%20in%20oklab%3B%5Cn%20%20%7D%5Cn%20%20background-image%3A%20linear-gradient(var(--tw-gradient-stops))%3B%5Cn%7D%5Cn.from-custom%5C%5C%2F20%20%7B%5Cn%20%20--tw-gradient-from%3A%20color-mix(in%20oklab%2C%20rgba(111%2C%20160%2C%200)%2020%25%2C%20transparent)%3B%5Cn%20%20--tw-gradient-stops%3A%20var(--tw-gradient-via-stops%2C%20var(--tw-gradient-position)%2C%20var(--tw-gradient-from)%20var(--tw-gradient-from-position)%2C%20var(--tw-gradient-to)%20var(--tw-gradient-to-position))%3B%5Cn%7D%5Cn.from-0%5C%5C%25%20%7B%5Cn%20%20--tw-gradient-from-position%3A%200%25%3B%5Cn%7D%5Cn.to-40%5C%5C%25%20%7B%5Cn%20%20--tw-gradient-to-position%3A%2040%25%3B%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D).

sehmbimanvir:
Gradient, blur isn't working for a production build. I tried to extract using tailwind cli tool.

Here are the results.

```css
/* This is at line number 3000 */
@layer properties {
  @supports ((-webkit-hyphens: none) and (not (margin-trim: inline))) or ((-moz-orient: inline) and (not (color:rgb(from red r g b)))) {
    *, ::before, ::after, ::backdrop {
      --tw-translate-x: 0;
      --tw-translate-y: 0;
      --tw-translate-z: 0;
      --tw-scale-x: 1;
      --tw-scale-y: 1;
      --tw-scale-z: 1;
      --tw-rotate-x: initial;
      --tw-rotate-y: initial;
      --tw-rotate-z: initial;
      --tw-skew-x: initial;
      --tw-skew-y: initial;
       // more variables
```

```css
/* This is at line number 1600 */
.from-background-primary {
    --tw-gradient-from: var(--bg-primary);
    --tw-gradient-stops: var(--tw-gradient-via-stops, var(--tw-gradient-position), var(--tw-gradient-from) var(--tw-gradient-from-position), var(--tw-gradient-to) var(--tw-gradient-to-position));
  }
```

All variables are getting generated after the actual class. So, these initial values are replacing the property being used in class. @philipp-spiess, Any help or fix for this?

philipp-spiess:
@sehmbimanvir Please attach a full reproduction here, the variables are put in an `@layer properties` so they should still come first.

philipp-spiess:
> You can see, one https://github.com/layer properties is at top (which is empty) and other one is at bottom.

Yeah this is intentional. `@layer properties` with no children at the start will declare the layer as the first in the cascade, see:  https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Cascade_layers#creating_cascade_layers

Your CSS file and HTML seems to work fine for me in iOS 16.4 Safari, see:

Screenshot 2025-05-21 at 10 31 41

PR #17436, opened 2025-03-28
Add test coverage for property-specific colors

scottbedard:
There are several property-specific color variables available to maintain compatibility with v3. [My team and I would like to use them](https://github.com/tailwindlabs/tailwindcss/discussions/17400), but would feel more comfortable if they were formally supported by v4. This PR adds test coverage for those features.

@RobinMalfait has confirmed that these features are not intended to be removed, [see conversation here &rarr;](https://discord.com/channels/486935104384532500/546706299010678784/1355213322995110171)

@crswll Has opened a PR here for the related documentation
- https://github.com/tailwindlabs/tailwindcss.com/pull/2178


packages/tailwindcss/src/utilities.test.ts
```diff
@@ -21239,6 +21301,11 @@ test('shadow', async () => {
       box-shadow: var(--tw-inset-shadow), var(--tw-inset-ring-shadow), var(--tw-ring-offset-shadow), var(--tw-ring-shadow), var(--tw-shadow);
     }
 
+    .shadow-blue-500 {
+      --tw-shadow: #3b82f6;
+      box-shadow: var(--tw-inset-shadow), var(--tw-inset-ring-shadow), var(--tw-ring-offset-shadow), var(--tw-ring-shadow), var(--tw-shadow);
+    }
```
philipp-spiess, line 21307: This seems wrong, this ins setting the shadow property and not the shadow-color here. Looks like you want to define this as `--box-shadow-color-blue-500`.

philipp-spiess (commented):
Hey there! Thanks for your contribution. Adding more test coverage here makes sense but there seems to be an issue with box shadows. Mind looking into it?

packages/tailwindcss/src/utilities.test.ts
scottbedard: Interesting, now my tests are passing locally, but are failing on CI 🤔

philipp-spiess (approved):
Thanks for increasing the test coverage here! Highly appreciated!

packages/tailwindcss/src/utilities.test.ts
philipp-spiess: @scottbedard Haha yep sorry!

scottbedard: Thanks! Our team is excited to really polish our design system and modernize with Tailwind ❤️

PR #17464, opened 2025-03-31
Show warning when using unsupported bare value data type

RobinMalfait:
This PR will show a warning if you are using a bare value data type that is not supported.

Let's say you want to create a new utility that allows `color` to be a bare value data type like this:
```css
@utility paint-* {
  paint: --value([color], color);
}
```

This means that this would enable new syntax that we don't support yet. E.g.: `paint-#0088cc`.

The only supported data types for bare values are:

- `number` — `2.5`
- `integer` — `2`
- `ratio` — `1/2`
- `percentage` — `50%`

All other data types are not supported in this position. This PR will now show a warning:
~~~
Unsupported bare value data type: "color".
Only valid data types are: "number", "integer", "ratio", "percentage".

```css
--value([color],color)
                ^^^^^
```
~~~
Once we have better sourcemap / location tracking support, this warning will point to the exact spot, but for now, only a re-print of the AST can be used.

If you _do_ want to use other data types, then you will have to use arbitrary value syntax with `[…]` instead.

```css
@utility paint-* {
  paint: --value([color]);
}
```

This will allow for `paint-[#0088cc]` for example.

Note: this is not a behavioral change, we already didn't support other data types, but we silently ignored them. This means that we have to do more parsing at runtime when evaluating the utility. 

With this change, a warning is shown when registering the `@utility`, not when using it.

### DIFF
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -5692,6 +5692,16 @@ export function createUtilities(theme: Theme) {
   return utilities
 }
 
+// Only allowed bare value data types, to prevent creating new syntax that we
+// typically don't support right now. E.g.: `--value(color)` would allow you to
+// use `text-#0088cc` as a valid utility, which is not what we want.
+export const BARE_VALUE_DATA_TYPES = [
+  'number', // 2.5
+  'integer', // 8
+  'ratio', // 2/3
+  'percentage', // 25%
+]
+
 export function createCssUtility(node: AtRule) {
   let name = node.params
 
@@ -6084,12 +6094,7 @@ function resolveValueFunction(
       // Limit the bare value types, to prevent new syntax that we
       // don't want to support. E.g.: `text-#000` is something we
       // don't want to support, but could be built this way.
-      if (
-        arg.value !== 'number' &&
-        arg.value !== 'integer' &&
-        arg.value !== 'ratio' &&
-        arg.value !== 'percentage'
-      ) {
+      if (!BARE_VALUE_DATA_TYPES.includes(arg.value)) {
         continue
       }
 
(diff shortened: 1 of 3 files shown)


thecrypticace (approved):

packages/tailwindcss/src/utilities.ts
```diff
@@ -5744,6 +5753,36 @@ export function createCssUtility(node: AtRule) {
...
             }
+
+            // Validate bare value data types
+            else if (
+              node.kind === 'word' &&
+              !(node.value[0] === '[' && node.value[node.value.length - 1] === ']') && // Ignore arbitrary values
+              !BARE_VALUE_DATA_TYPES.includes(node.value)
+            ) {
+              console.warn(
+                `Unsupported bare value data type: "${node.value}".\nOnly valid data types are: ${BARE_VALUE_DATA_TYPES.map((x) => `"${x}"`).join(', ')}\n`,
```
RobinMalfait, line 5764: @adamwathan do you want to make any changes to this warning?

Output looks something like this right now:
~~~
Unsupported bare value data type: "color".
Only valid data types are: "number", "integer", "ratio", "percentage"

```css
--value([color],color)
                ^^^^^
```
~~~

philipp-spiess: My 5 ct: I don't think the term "bare value" is something that people of that API will be familiar with but maybe we just call it "value data type" (since it's the `--value(…)` function)? 

````
Unsupported value data type: "color".
Only valid data types are: "number", "integer", "ratio", "percentage".

```css
--value([color],color)
                ^^^^^
```
````

The rest I think is fine personally but I'd add a dot at the end of the second sentence

RobinMalfait: I think people writing plugins like that will be more familiar with the term (or at least should be because they will have to know the difference between `color` and `[color]`). 

Main reason I used it is because that's what we use in the docs for it: https://tailwindcss.com/docs/adding-custom-styles#bare-values

philipp-spiess (approved):

PR #17482, opened 2025-04-01
feat: add ignore option to tailwind plugin

AlemTuzlak:
Added an optional ignore flag for the tailwind plugin to ignore certain files and allow for easier configuration with other plugins

Needed for better interop with other plugins, eg:
https://github.com/tailwindlabs/tailwindcss/discussions/16712

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -135,7 +135,7 @@ function isPotentialCssRootFile(id: string, ignoredFiles: Exclude<TailwindPlugin
   if(isIgnoredPattern) return
 
   let extension = getExtension(id)
-  const isCssFile =
+  let isCssFile =
     (extension === 'css' || id.includes('&lang.css') || id.match(INLINE_STYLE_ID_RE)) &&
     // Don't intercept special static asset resources
     !SPECIAL_QUERY_RE.test(id) &&
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -10,10 +10,14 @@ const SPECIAL_QUERY_RE = /[?&](?:worker|sharedworker|raw|url)\b/
 const COMMON_JS_PROXY_RE = /\?commonjs-proxy/
 const INLINE_STYLE_ID_RE = /[?&]index\=\d+\.css$/
 
-export default function tailwindcss(): Plugin[] {
+export interface TailwindPluginOptions {
+  ignore?: (string | RegExp | ((id: string) => boolean))[]
+}
+
+export default function tailwindcss(pluginOptions?: TailwindPluginOptions): Plugin[] {
   let servers: ViteDevServer[] = []
   let config: ResolvedConfig | null = null
-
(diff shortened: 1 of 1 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -115,10 +119,23 @@ function getExtension(id: string) {
...
   if (id.includes('/.vite/')) return
+  let isIgnoredPattern = ignoredFiles.some(matcher => {
+    if(matcher instanceof RegExp){
+      return matcher.test(id);
+    }
+    if(matcher instanceof Function){
+      return matcher(id);
+    }
+
+    return matcher.includes(matcher)
```
AlemTuzlak, line 132: not sure what to do in this case if it's a string, or if you want to support this use-case at all, maybe it would make sense to offload it to the user with the matcher function above

AlemTuzlak:
Not sure if tests are needed? ready to write any if yes!

AlemTuzlak:
That would probably be the ideal outcome, although not sure if it's realistic as that's what vanilla extract tried to do as well and ran into interop issues with a few plugins and gave up on it, not saying it will happen to tailwind but it's a possibility, I'll create a repro for vanilla, the other one I'm not really familiar with and can't be of much help there, but I'm guessing you run into it as soon as you use polaris to generate css. For vanilla I know their vite plugin listens to `.css.ts` files then outputs `.vanilla.css` files in dev so then tailwind tries to parse them and results in the site crashing as it considers it invalid css.

AlemTuzlak:
@philipp-spiess sorry for the wait, here's a vanilla repro:
https://github.com/AlemTuzlak/underkill-stack
Not sure if it would solve all the issues with vanilla by fixing this one as the error i've seen is a different error message than this one but it's a great start.

philipp-spiess:
@AlemTuzlak The repro you sent does indeed error but that's because of a typo in the vanilla extract file:

```diff
diff --git a/app/reset.css.ts b/app/reset.css.ts
index 382fe19..4d4c446 100644
--- a/app/reset.css.ts
+++ b/app/reset.css.ts
@@ -22,7 +22,7 @@ const documentBannerHeadingParagraph = style({
 const documentBannerTextContainer = style({
   display: "grid",
   gridGap: 5,
-  gridTemplateRows: "repeat(max-content))",
+  gridTemplateRows: "repeat(max-content)",
 
   marginBlock: "auto",
   zIndex: 1,
```

When I apply this diff it builds just fine, seems like `vanilla.css` files are not the issue here but rather the fact that you can create invalid CSS easily without validation 🙈 

Going to close this PR for now as it doesn't seem necessary for Vanilla extract and we don't really have a use case for this feature otherwise but please feel free to proof me wrong.

lfantone:
@philipp-spiess I'm still having issues with this, I know that the css generated by vanilla-extract is valid because it works on v3 and it's also being compiled.
If I apply what was discussed over [here](https://github.com/tailwindlabs/tailwindcss/discussions/16712) then tailwind v4 works as expected.

philipp-spiess:
@lfantone Yes please set up a reproduction repository and create a new issue!

PR #17490, opened 2025-04-01
Ensure polyfills exist after external `@import` at-rules

RobinMalfait:
This PR fixes an issue where external `@import url(…)` at-rules were not at the top anymore due to the polyfills being pre-pended to the CSS output.

To solve this, we will inject a `@layer __tw-properties;` as a placeholder to at the top of the file. A body-less `@layer` is allowed to exist before any other rules in the file.

Later we inject all the polyfills at the bottom of the file inside the `@layer __tw-polyfills;` at-rule.

### DIFF
--- a/packages/tailwindcss/src/__snapshots__/utilities.test.ts.snap
+++ b/packages/tailwindcss/src/__snapshots__/utilities.test.ts.snap
@@ -1,10 +1,12 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`border-* 1`] = `
-"@supports (((-webkit-hyphens: none)) and (not (margin-trim: inline))) or ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
-  @layer base {
-    *, :before, :after, ::backdrop {
-      --tw-border-style: solid;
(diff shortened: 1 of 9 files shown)


packages/tailwindcss/src/__snapshots__/utilities.test.ts.snap
```diff
@@ -1,10 +1,12 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`border-* 1`] = `
-"@supports (((-webkit-hyphens: none)) and (not (margin-trim: inline))) or ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
-  @layer base {
-    *, :before, :after, ::backdrop {
-      --tw-border-style: solid;
+"@layer __tw-properties {
+  @supports (((-webkit-hyphens: none)) and (not (margin-trim: inline))) or ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
+    @layer base {
```
philipp-spiess, line 6: can remove layer base if we go with this approach

RobinMalfait:
Closing in favor of: #17506

PR #17502, opened 2025-04-02
Use `@tailwindcss/node` for import in CLI

### LINKED ISSUE #17501: `@tailwindcss/cli@4.1.0` misses `jiti` dependency
https://unpkg.com/@tailwindcss/cli@4.1.0/dist/index.mjs contains

```js
import{createJiti as Xl}from"jiti"
```

However, `jiti` is not in `dependencies` of `@tailwindcss/cli`:

```
  "dependencies": {
    "@parcel/watcher": "^2.5.1",
    "enhanced-resolve": "^5.18.1",
    "mri": "^1.2.0",
    "picocolors": "^1.1.1",
    "@tailwindcss/node": "4.1.0",
    "@tailwindcss/oxide": "4.1.0",
    "tailwindcss": "4.1.0"
  },
```

Granted, `@tailwindcss/node` does have `jiti` as a dependency.
But my monorepo does have a different version of `jiti` which is on version 1.x, which means the unintended version was resolved.

My suggestion is to add it as an explicit dependency and pin the version for robustness.

philipp-spiess:
Closes #17501

Seems like an oversight. The CLI does have a dependency on `@tailwindcss/node` so it should use it from the public import like the other stuff.

## Test plan

Looked at the minified `dist/` output to ensure the import for `jiti` is gone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Disable padding in `@source inline(…)` brace expansion ([#17491](https://github.com/tailwindlabs/tailwindcss/pull/17491))
 - Inject polyfills after `@import` and body-less `@layer` ([#17493](https://github.com/tailwindlabs/tailwindcss/pull/17493))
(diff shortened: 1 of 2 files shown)


RobinMalfait (approved):
Running a local build of this branch does fix the problem. 💪 

No more `"jiti"` in the bundle after this change.

Before:
image

After:
image

PR #17510, opened 2025-04-02
Replace currentColor with currentcolor (lowercase)

teddybradford:
Replaces `currentColor` with `currentcolor` (lowercase) to match what's defined in [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#currentcolor-color) and [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) (see: https://github.com/mdn/content/pull/16592).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3522,7 +3522,7 @@ No release notes
 
 - Lists now have no margins by default
 - `.pin` no longer sets width and height to 100%
-- SVG `fill` no longer defaults to currentColor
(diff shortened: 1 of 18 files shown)


philipp-spiess (approved):
Thanks! I think this is a nice change. Note that I had to revert a bunch of changed places since some snapshot is including CSS that is transformed via `lightningcss` which always changes `currentcolor` to `currentColor`.

PR #17511, opened 2025-04-02
Fix slow incremental builds (especially on Windows)

### LINKED ISSUE #16911: "@tailwindcss/vite": "4.0.8"  and  "@tailwindcss/vite": "4.0.9"  have massively increased load times for the CSS in Vite.
So I have noticed drastically increased load times when using Vite and Tailwind when updating to "@tailwindcss/vite": "4.0.8"  and  "@tailwindcss/vite": "4.0.9".

"@tailwindcss/vite": "4.0.8"  and  "@tailwindcss/vite": "4.0.9"
Image

"@tailwindcss/vite": "4.0.7"  and  lower.
Image

Not sure what is causing this. 
Happy to share more information about my setup or debug further with instructions.

Thanks

### LINKED ISSUE #17522: [4.1.1] Tailwindcss v4.1.x too slower，twenty times slower than 4.0.x

v4.1.x

Vite 6.2.4

v22.14.0

Chrome,

Windows

**Reproduction URL**

https://github.com/zyycn/tailwindcss-v4.1.x-speed

**Describe your issue**

Tailwindcss v4.1.x ，hot update in 4.0.x is between 100ms-200ms，but in 4.1.x is 2.3s-3s !

Image
Image

philipp-spiess:
Fixes: #16911
Fixes: #17522

When debugging this issue, we noticed that the `resolve_glob()` function from `detect_sources.rs` was really slow, specifically on Windows. It turns out that we're now computing the list of globs that we use to attach the file system watchers _repeatedly on every change_ instead of doing it only _once per compiler setup_. This changed with the updates we added to support `@source not "…";` in 4.1.

More specifically, we noticed that the way we compute these sources is causing a recursive scan through _all_ folders inside your working directory (including `node_modules/` and `.git/`). This happens because that logic uses a vanilla file-system walker and not our `ignore` crate for this.

When we started investigating it, our first reaction was to change both Vite and PostCSS to use the new `normalized_sources` list instead. This is a list of globs to watch that is computed solely on the input sources and does not require any file-system access and it's what we use in the CLI as well. Think of it as being a `**/*` rule for the current working-directory in the basic auto-content case.

When we made this change, to our surprise, nothing in our existing integration test broke. From memory, though, we remembered that it's important for PostCSS though to _not register a blanket watcher for the current working-directory_. It took us quite a bit of tinkering until we were able to craft an example that would cause an endless re-render loop with this patch applied. We've now [added a new integration test](https://github.com/tailwindlabs/tailwindcss/pull/17511/files#diff-7bc765548a52132ede460508e58dca709c4a00c9c56669bf9954cc4bc3d93c88) to have this context in the repository for the future but also it meant that we have to go back to creating a "smarter" list of globs for the file system watcher again.

To fix the regressions, this PR now does two things:

1. We now cache the list of calculated globs again since it's based only on input form the CSS file. This means that rebuilds are instant as we don't have to compute these globs anymore. This was the same behavior that we had in previous 4.0 releases.
2. Furthermore we also updated the file system walk to include our new globally ignored content directories. By bailing on any subdirectories in `node_modules` and `.git`, that walk is now _much_ faster. We also made sure that you can still add a custom source rule to point _inside_ a `node_modules` folder and it will still pick up changes like you would suggest (e.g. via `@source "../node_modules/my-ui-lib";`). 
 
The results? This PR fixes slow rebuilds on Windows where rebuilds go from ~2s to ~20ms.

We suspect that this issue is more pronounced on Windows because of a slower file system in general.

## Test plan

Tested it on a reproduction with the following results:

**Before:**

https://github.com/user-attachments/assets/10c5e9e0-3c41-4e1d-95f6-ee8d856577ef

**After:**

https://github.com/user-attachments/assets/2c7597e9-3fff-4922-a2da-a8d06eab9047

Zooming in on the times, it looks like this:

| Before    | After |
| -------- | ------- |
| image | image |

But with these changes:

We also tested this on Windows with the following results:

| Before    | After |
| -------- | ------- |
| image  | image |

[ci-all]

### DIFF
--- a/crates/oxide/src/scanner/mod.rs
+++ b/crates/oxide/src/scanner/mod.rs
@@ -93,7 +93,7 @@ pub struct Scanner {
     dirs: Vec<PathBuf>,
 
     /// All generated globs, used for setting up watchers
-    globs: Vec<GlobEntry>,
+    globs: Option<Vec<GlobEntry>>,
 
     /// Track unique set of candidates
     candidates: FxHashSet<String>,
@@ -296,16 +296,24 @@ impl Scanner {
 
     #[tracing::instrument(skip_all)]
     pub fn get_globs(&mut self) -> Vec<GlobEntry> {
+        if let Some(globs) = &self.globs {
+            return globs.clone();
+        }
+
         self.scan_sources();
 
+        let mut globs = vec![];
         for source in self.sources.iter() {
             match source {
                 SourceEntry::Auto { base } | SourceEntry::External { base } => {
-                    let globs = resolve_globs((base).to_path_buf(), &self.dirs, &self.extensions);
-                    self.globs.extend(globs);
+                    globs.extend(resolve_globs(
+                        (base).to_path_buf(),
+                        &self.dirs,
+                        &self.extensions,
+                    ));
                 }
                 SourceEntry::Pattern { base, pattern } => {
-                    self.globs.push(GlobEntry {
+                    globs.push(GlobEntry {
                         base: base.to_string_lossy().to_string(),
                         pattern: pattern.to_string(),
                     });
@@ -315,9 +323,12 @@ impl Scanner {
         }
 
         // Re-optimize the globs to reduce the number of patterns we have to scan.
-        self.globs = optimize_patterns(&self.globs);
+        globs = optimize_patterns(&globs);
 
-        self.globs.clone()
+        // Track the globs for subsequent calls
+        self.globs = Some(globs.clone());
+
+        globs
     }
 
     #[tracing::instrument(skip_all)]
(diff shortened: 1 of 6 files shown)


crates/oxide/src/scanner/mod.rs
```diff
@@ -296,16 +296,24 @@ impl Scanner {
...
         self.scan_sources();
 
+        let mut globs = vec![];
         for source in self.sources.iter() {
             match source {
                 SourceEntry::Auto { base } | SourceEntry::External { base } => {
-                    let globs = resolve_globs((base).to_path_buf(), &self.dirs, &self.extensions);
-                    self.globs.extend(globs);
+                    globs.extend(resolve_globs(
+                        (base).to_path_buf(),
```
thecrypticace, line 310:
```suggestion
                        base.to_path_buf(),
```

thecrypticace (approved):

FragsterAt:
What about the case when a new file appears after starting dev server? (for example new vue sfc component)

PR #17514, opened 2025-04-02
Fix race condition in Next.js with `--turbopack`

### LINKED ISSUE #17508: Tailwind CSS Not Updating Global Styles on Latest Version

v4.1.1

Next.js 15.2.4

v22.12.0

Chrome

macOS

After updating to the latest version of Tailwind CSS, global styles defined in global.css are not updating properly. The issue persists even after clearing the Next.js cache and restarting the development server. This was not an issue in the previous versions.

https://github.com/user-attachments/assets/69c7097e-9d47-48ab-8352-b90c0791c194

RobinMalfait:
This PR fixes an issue where if you use Next.js with `--turbopack` a race condition happens because the `@tailwindcss/postcss` plugin is called twice in rapid succession. 

The first call sees an update and does a partial update with the new classes. Next some internal `mtimes` are updated. The second call therefore doesn't see any changes anymore because the `mtimes` are the same, therefore it's serving its stale data.

Fixes: #17508

## Processing CSS files with only `@tailwind utilities;`

While working on a unit test for this, we also noticed that the PostCSS client did nothing at all when the input css contained only the `@tailwind utilities;` directive. It turns out that this was a directive that was not being checked for in the fast-bail code so in this PR we also changed it to account for that.

## Test plan

- Tested with the repro provided in #17508
- Added a new unit test that calls into the PostCSS plugin directly for the same change from the same JavaScript run-loop.


packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -138,9 +139,9 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
 
           // Setup the compiler if it doesn't exist yet. This way we can
           // guarantee a `build()` function is available.
-          context.compiler ??= await createCompiler()
```
thecrypticace, line 141: I'd probably immediately follow this up with `let compiler = await context.compiler` and shorten everything after this. Assuming that doesn't cause issues (it… shouldn't… I hope)

thecrypticace (approved):

PR #17515, opened 2025-04-02
Fix drop shadow filters with multiple shadows

thecrypticace:
It seems that I broke support for multiple drop-shadow filters when `@theme inline` was used in v4.1. This PR fixes that by segmenting the drop shadow value on top-level commas and wrapping each segment with `drop-shadow(…)` like we did in v4.0.


philipp-spiess (commented):
Change LGTM but let's make sure to add an example for this to the UI specs for shadows I think

philipp-spiess (approved):

PR #17523, opened 2025-04-03
Fix multi-value arbitrary inset shadow

### LINKED ISSUE #17520: inset-shadow-[spec1,spec2] renders all but the first shadow spec outset, not inset

v4.0.17 in my app, v4.11 in "Tailwind Play"

Next.js 15.2.4 but n/a

v20.10.0 but n/a

Firefox but affects all

Linux but affects all

**Reproduction URL**

https://play.tailwindcss.com/gI32O7VLv6

**Describe your issue**

I would expect `inset-shadow-[shadowspec1,shadowspec2]` to render 2 inset shadows, but in fact all but the first are rendered as outer shadows.

In my reproduction the second shadow is red, and should appear inside the element on the right side, but it appears outside the element on the left side.

I can work around it by adding `inset_` at the start of the 2nd (and further) shadow specs, such as with `inset-shadow-[shadowspec1,inset_shadowspec2]`

wongjn:
Fixes #17520

Fixes multi-value inset shadows to apply the `inset` prefix to each component instead of only the first.

Feel free to make the code nicer 😄

## Test plan

Ensure that a multi-value inset shadow now applies each shadow _inset_: 
Screenshot 2025-04-03 at 10 50 29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Don't rely on `@layer base` for the `@property` polyfills ([#17506](https://github.com/tailwindlabs/tailwindcss/pull/17506))
-- Fix multi-value arbitrary inset shadow ([#17523](https://github.com/tailwindlabs/tailwindcss/pull/17523))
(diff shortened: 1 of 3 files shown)


philipp-spiess (commented):
Thanks! I think this makes sense. Let me just check this out locally and see if the repro now works as expected 🙂

philipp-spiess (approved):

PR #17533, opened 2025-04-03
Vite: Don't register the current CSS file as a dependency on itself

### LINKED ISSUE #17512: Missing global.css file when using Tailwindcss Vite plugin

v4.1.1

Qwik 1.13.0, Vite 5.3.5

v23.8.0

Edge

Windows

**Reproduction URL**

https://github.com/[rasmus-dybmose/qwik-tailwindcss](https://github.com/rasmus-dybmose/qwik-tailwindcss)

**Describe your issue**

Tailwindcss is not working when using the latest version (v4.1.1) if I change the version to v4.0.17 (fixed) it works again.

It-looks like the browser can't fetch the ./src/global.css file, it dissaperes if I add the tailwindcss plugin to vite

https://github.com/user-attachments/assets/e008d5eb-f92b-4b96-9fc1-27e8b15235ee

philipp-spiess:
Closes #17512

One of the changes of the Oxide API in 4.1 is that it now emits the input CSS file itself as a dependency. This was fine in most of our testing but it turns out that certain integrations (in this case a Qwik project) don't like this and will silently crash with no CSS file being added anymore.

This PR fixes this by making sure we don't add the input file as a dependency on itself and also adds an integration test to ensure this won't regress again.

## Test plan

- Tested with the repro provided in #17512
- Added a minimal integration test based on that reproduction that I also validated will _fail_, if the fix is reverted.


thecrypticace (approved):

RobinMalfait (approved):
Verified that the test fails without the fix and passes with the fix.

PR #17554, opened 2025-04-04
PostCSS: Fix Turbopack 'one-revision-behind' bug

### LINKED ISSUE #17508: Tailwind CSS Not Updating Global Styles on Latest Version

philipp-spiess:
Closes #17508

This PR fixes another issue we found that caused dev builds with Next.js and Turbopack to resolve the CSS file that was saved one revision before the latest update.

When debugging this we noticed that the PostCSS entry is called twice for every one update when changing the input CSS file directly. That was caused by the input file itself being added as a _dependency_ so you would first get the callback that a _dependency_ has updated (at which point we look at the file system and figure out we need a full-rebuild because the input.css file has changed) and then another callback for when the _input file_ has updated. The problem with the second callback was that the file-system was already scanned for updates and since this includes the `mtimes` for the input file, we seemingly thought that the input file did not change. However, the issue is that the first callback actually came with an outdated PostCSS input AST...

We found that this problem arises when you register the input CSS as a dependency of itself. This is not expected and we actually guard against this in the PostCSS client. However, we found that the input `from` argument is _a relative path when using Next.js with Turbopack_ so that check was not working as expected. 

## Test plan

Added the change to the repro from #17508 and it seems to work fine now.

https://github.com/user-attachments/assets/2acb0078-f961-4498-be1a-b1c72d5ceda1

Also added a unit test to ensure we document that the input file path can be a relative path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- PostCSS: Resolve an issue where changes to the input CSS file showed outdated content when using Turbopack ([#17554](https://github.com/tailwindlabs/tailwindcss/pull/17554))
 
 ## [4.1.2] - 2025-04-03
 
--- a/packages/@tailwindcss-postcss/src/fixtures/example-project/input.css
+++ b/packages/@tailwindcss-postcss/src/fixtures/example-project/input.css
@@ -1,0 +1,1 @@
+/* the content for this file is set in the tests */
--- a/packages/@tailwindcss-postcss/src/index.test.ts
+++ b/packages/@tailwindcss-postcss/src/index.test.ts
@@ -420,3 +420,25 @@ describe('concurrent builds', () => {
     expect(await promise2).toContain('.red')
   })
 })
+
+test('does not register the input file as a dependency, even if it is passed in as relative path', async () => {
+  let processor = postcss([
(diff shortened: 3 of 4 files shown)


RobinMalfait (approved):

andymerskin:
I'm not sure if this is directly related to this issue, but after updating to Tailwind v4.1.3 which includes this change, I'm still not seeing changes reflected on each hot reload (but sometimes it works!), forcing me to restart my dev server for simple className changes in any context. For some reason Tailwind isn't picking up on new utilities being used. I haven't fully pinned down the cases where it's happening, but it's difficult to reproduce reliably.

Most recently, I changed the ordering inside a `clsx` + `tailwind-merge` call (`cn`), to move a `className` prop value at the end:
```tsx
// before
<div className={cn(className, getMyClassnames(props))} />

// after
<div className={cn(getMyClassnames(props), className)} />
```

No hot reload and no update after a hard refresh, but after restarting the dev server, my change looked correct.

### Environment:
- turbo 2.0.4
- vite 4.5.6
  - @vitejs/plugin-react-swc
  - @swc/plugin-styled-components
- react 17.x
- tailwind 4.1.3
- typescript 4.9.3

Overall things are working fine. If upgrading Vite or Turbo is necessary to squash some of this behavior, happy to do it, but it may be quite a lift for us depending on how smoothly that goes.

PR #17557, opened 2025-04-04
Handle Ruby `%w` syntax in Slim pre processing

### LINKED ISSUE #17542: Class not extracted when only occurrence is first element of Slim attribute array

v4.1.2

@tailwindcss/cli

v22.10.0

Chrome

macOS 15

**Reproduction URL**

https://github.com/markzalar/tailwind-4.0.10-slim-first-in-array

**Describe your issue**

This issue started happening in tailwindcss 4.0.10

If the only occurrence of a class is the first element of a Slim attribute array, the class doesn't get extracted. 

For example, if I have the following slim code

```slim
div[
  class=%w[bg-blue-500 w-10 h-10]
]
div[
  class=%w[w-10 bg-green-500 h-10]
]                
```

I would expect it to render as

Image

but if that's the only instance of `bg-blue-500` in the codebase it doesn't get extracted and the slim snippet above renders as

Image

RobinMalfait:
This PR fixes an issue where the Ruby `%w[…]` syntax causes utilities to not be properly extracted.

This PR will now ensure that it does get extracted correctly.

Given this input:
```slim
div[
  class=%w[bg-blue-500 w-10 h-10]
]
div[
  class=%w[w-10 bg-green-500 h-10]
]
```

Before this PR, we extracted everything but the `bg-blue-500`. The `w-10` was extracted but not because of the second div, but because of the first one.

Fixes: #17542

## Test plan

1. Added a test to ensure it's working correctly.

Looking at the extractor tool, you can see that it now gets extracted correctly. Top is before, bottom is with this change.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Show warning when using unsupported bare value data type in `--value(…)` ([#17464](https://github.com/tailwindlabs/tailwindcss/pull/17464))
 - PostCSS: Resolve an issue where changes to the input CSS file showed outdated content when using Turbopack ([#17554](https://github.com/tailwindlabs/tailwindcss/pull/17554))
+- Handle Ruby `%w` syntax in Slim pre processing ([#17557](https://github.com/tailwindlabs/tailwindcss/pull/17557))
 
 ## [4.1.2] - 2025-04-03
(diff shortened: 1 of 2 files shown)


thecrypticace (approved):

RobinMalfait:
For completeness, I don't think this syntax is valid in Pug (but I could be proven wrong). In Haml you can do it, but we already extract it properly:

image

PR #17566, opened 2025-04-05
refactor[integrations/test]: change `let` to `const` for those references doesn't change

samuel871211:
Some variables were declared with the `let` keyword but never reassigned. For the reasons below, it would be a better practice to change to `const`.

1. Prevents accidental reassignment
2. Signals intent clearly to other developers
3. Improves code readability
4. Enables potential compiler optimizations
5. Simplifies debugging
6. Promotes immutability
7. Reduces bugs from mutable state

### DIFF
--- a/integrations/cli/config.test.ts
+++ b/integrations/cli/config.test.ts
@@ -161,7 +161,7 @@ test(
     },
   },
   async ({ fs, spawn }) => {
-    let process = await spawn(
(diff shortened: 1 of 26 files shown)


RobinMalfait:
More info: https://github.com/tailwindlabs/tailwindcss/pull/17567#issuecomment-2780649699

PR #17591, opened 2025-04-07
Fix brace expansion with range going down

### LINKED ISSUE #17576: Unused test data in brace-expansion.test.ts
I found an issue in the `brace-expansion.test.ts` file where there's an unused test case at line 20.

**File**: [packages/tailwindcss/src/utils/brace-expansion.test.ts](https://github.com/tailwindlabs/tailwindcss/blob/5a77c9dfc4bcc1f83fe3269fbbcc27386d5e8572/packages/tailwindcss/src/utils/brace-expansion.test.ts#L20)

**Issue**: The test case for `'a/{0..10..5}/b'` is incorrectly structured. The third array element `['a/{10..0..5}/b', ['a/10/b', 'a/5/b', 'a/0/b']]` has no effect on the test execution.

**Current code**:
```ts
...
[
  'a/{0..10..5}/b',
  ['a/0/b', 'a/5/b', 'a/10/b'],
  ['a/{10..0..5}/b', ['a/10/b', 'a/5/b', 'a/0/b']],
],
...
```

**Expected structure**:
```ts
...
['a/{0..10..5}/b', ['a/0/b', 'a/5/b', 'a/10/b']],
['a/{10..0..5}/b', ['a/10/b', 'a/5/b', 'a/0/b']],
...
```

I'm not sure what the original intention of this array was. If I move this array to the parent level, the test fails, meaning `expand('a/{10..0..5}/b')` does not return `['a/10/b', 'a/5/b', 'a/0/b']`. The actual output is `['a/0/b', 'a/10/b', 'a/5/b']`.

RobinMalfait:
This PR fixes an issue where a test was setup wrong. While fixing the test, I also noticed that the actual computation was wrong. There are a few cases we have to make sure they work when working with ranges:

- `0..10..5` — range going from low to high, with a positive step. This should result in `0 5 10`. Already worked.
- `0..10..-5` — range going from low to high, with a negative step. This should result in `10 5 0`. This one was broken.
- `10..0..5` — range going from high to low, with a positive step. This should result in  `10 5 0`. This one was broken.
- `10..0..-5` — range going from high to low, with a negative step. This should result in  `0 5 10`. Already worked.

This technically means that there are multiple ways to do the same thing. And in case of Tailwind CSS the order doesn't really matter because we apply custom sort rules for utilities. Luckily, the order here doesn't have any influence on the actual generated CSS so it's not the end of the world to support it.

Intellisense does use a more heavy brace expansion library, and if we don't support this, it also means that you can see a preview of generated classes without generating the actual classes. This also fixes that.

E.g.: Note how the preview shows the correct values, but nothing has been generated (this is before the change of this PR).

image

Fixes: #17576

## Test plan

1. Hoisted the test (`a/{10..0..5}/b`), which failed
2. Fixed the implementation to make the test pass
3. Added an additional test (`a/{10..0..-5}/b`)


thecrypticace (approved):
I really should unify the brace expansion code in IntelliSense with the one we've got here.

Looks good 👍

PR #17628, opened 2025-04-09
Fix container names with hyphens

### LINKED ISSUE #17614: Hyphenated custom container sizes aren't handled properly

v4.1.3

Tailwind Play

Tailwind Play

n/a

n/a

**Reproduction URL**

https://play.tailwindcss.com/OFXbKDyZ5G

**Describe your issue**

It seems hyphenated custom theme properties are supposed to be supported, like in the example the `--color-nice-blue` works just fine with a `bg-nice-blue`.

But if I try the same with a custom container size I get broken behaviour.

If I set `--container-custom`, I can apply `@custom:font-bold` and I see it render bold (since we're wider than the specified 2rem). In Tailwind Play's autocomplete, when I type `@cus` I see `@custom:` and `@max-custom:` and `@min-custom:`. This is all as expected.

But if I set `--container-hyphenated-custom`, the same doesn't work, and I don't see it render bold. In the autocomplete, interestingly, I see `@max-hyphenated-custom:` and `@min-hyphenated-custom:`, but no `@hyphenated-custom:`.

wongjn:
Fixes #17614.

Candidate parsing for variants only account for the root `@` if there no hyphens. It seems like the current logic assumes if it *does* have a hyphen, then it would be one of `@min` or `@max`. However, with:

```css
@theme {
  --container-foo-bar: 1440px;
}
```
Then `@foo-bar` should be valid. However, we only check for `@foo-bar` and `@foo` as roots, but never `@`. This PR adds a check for `@` at the very end after iterating through root permutations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `color-mix(…)` polyfills do not cause used CSS variables to be removed ([#17555](https://github.com/tailwindlabs/tailwindcss/pull/17555))
 - Ensure the `color-mix(…)` polyfill creates fallbacks for theme variables that reference other theme variables ([#17562](https://github.com/tailwindlabs/tailwindcss/pull/17562))
 - Fix brace expansion in `@source inline('z-{10..0}')` with range going down ([#17591](https://github.com/tailwindlabs/tailwindcss/pull/17591))
+- Fix container query variant names couldn't container hyphens ([#17628](https://github.com/tailwindlabs/tailwindcss/pull/17628))
 
 ## [4.1.3] - 2025-04-04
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `color-mix(…)` polyfills do not cause used CSS variables to be removed ([#17555](https://github.com/tailwindlabs/tailwindcss/pull/17555))
 - Ensure the `color-mix(…)` polyfill creates fallbacks for theme variables that reference other theme variables ([#17562](https://github.com/tailwindlabs/tailwindcss/pull/17562))
 - Fix brace expansion in `@source inline('z-{10..0}')` with range going down ([#17591](https://github.com/tailwindlabs/tailwindcss/pull/17591))
+- Fix container query variant names couldn't container hyphens ([#17628](https://github.com/tailwindlabs/tailwindcss/pull/17628))
```
philipp-spiess, line 15:
```suggestion
- Ensure container query variant names can contain hyphens ([#17628](https://github.com/tailwindlabs/tailwindcss/pull/17628))
```

I'd remove the negation

packages/tailwindcss/src/candidate.ts
```diff
@@ -777,4 +777,10 @@ function* findRoots(input: string, exists: (input: string) => boolean): Iterable
 
     idx = input.lastIndexOf('-', idx - 1)
   } while (idx > 0)
+
+  // Try '@' variant after permutations. This allows things like `@max` of `@max-foo-bar`
+  // to match before looking for `@`.
+  if (input[0] === '@' && exists('@')) {
+    yield ['@', input.slice(1)]
+  }
```
philipp-spiess, line 777: I've been thinking of restructuring this somehow so that there's only one special-case for `@`. Let me know what you think about this (I pushed onto your PR). Tests seem to pass so it should be fine? 🤞

philipp-spiess (approved):

philipp-spiess:
Great stuff, ty!

PR #17630, opened 2025-04-10
Fix `fontSize` array upgrade

### LINKED ISSUE #17622: Upgrade cli does not output line height separately from the fontsize array

v4.1.3

Next.js 15.2.5

For example: v22.14.0

Chrome

Ubuntu Linux (24.10)

**Reproduction URL**

/

**Describe your issue**

In tailwind config, we could do something like this:

```ts
extend: {
  fontSize: {
    "title-custom": ["50px", "62px"]
  }
}
```
Which would generate
```css
.title-custom {
  font-size: 50px;
  line-height: 62px;
}
```
Upgrading to v4, the cli outputs
```css
 --text-title-custom: 50px, 62px;
```
And it does not generate any classes.

**Expected Behaviour**

The cli should treat the second element of the array as line-height and add a new variable for it rather than adding it to one variable separated by a comma.

```css
--text-title-custom: 50px;
--text-title-custom--line-height: 62px;
```

wongjn:
Fixes #17622.

Adds a specific handling case in `themeableValues()` in `packages/tailwindcss/src/compat/apply-config-to-theme.ts`. It seems like this has unique handling in v3 for an array value, whereby the second value is treated as a `line-height`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure the `color-mix(…)` polyfill creates fallbacks for theme variables that reference other theme variables ([#17562](https://github.com/tailwindlabs/tailwindcss/pull/17562))
 - Fix brace expansion in `@source inline('z-{10..0}')` with range going down ([#17591](https://github.com/tailwindlabs/tailwindcss/pull/17591))
 - Ensure container query variant names can contain hyphens ([#17628](https://github.com/tailwindlabs/tailwindcss/pull/17628))
+- Fix `fontSize` plain array tuple upgrade ([#17630](https://github.com/tailwindlabs/tailwindcss/pull/17630))
 
 ## [4.1.3] - 2025-04-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix brace expansion in `@source inline('z-{10..0}')` with range going down ([#17591](https://github.com/tailwindlabs/tailwindcss/pull/17591))
 - Ensure container query variant names can contain hyphens ([#17628](https://github.com/tailwindlabs/tailwindcss/pull/17628))
 - Fix `fontSize` plain array tuple upgrade ([#17630](https://github.com/tailwindlabs/tailwindcss/pull/17630))
+- Ensure compatibility with array tuples used in `fontSize` JS theme keys ([#17630](https://github.com/tailwindlabs/tailwindcss/pull/17630))
 
 ## [4.1.3] - 2025-04-04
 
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/compat/apply-config-to-theme.ts
```diff
@@ -130,7 +130,13 @@ export function themeableValues(config: ResolvedConfig['theme']): [string[], unk
     }
 
     if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
-      toAdd.push([path, value.join(', ')])
+      if (path[0] === 'fontSize') {
```
philipp-spiess, line 133:
```suggestion
      if (path[0] === 'fontSize' && value.length === 2) {
```

I'd probably want to make sure this _only_ works for tuples.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `color-mix(…)` polyfills do not cause used CSS variables to be removed ([#17555](https://github.com/tailwindlabs/tailwindcss/pull/17555))
 - Ensure the `color-mix(…)` polyfill creates fallbacks for theme variables that reference other theme variables ([#17562](https://github.com/tailwindlabs/tailwindcss/pull/17562))
 - Fix brace expansion in `@source inline('z-{10..0}')` with range going down ([#17591](https://github.com/tailwindlabs/tailwindcss/pull/17591))
+- Fix `fontSize` plain array tuple upgrade ([#17630](https://github.com/tailwindlabs/tailwindcss/pull/17630))
```
philipp-spiess, line 15:
```suggestion
- Ensure compatibility with array tuples used in `fontSize` JS theme keys ([#17630](https://github.com/tailwindlabs/tailwindcss/pull/17630))
```
Not 100% sure about this change but I want to make sure this has to do with JS theme keys so it's clearer who this fix is for. WDYT?

Maybe this is a bit more clear

philipp-spiess (approved):
Splendid, thanks! I left two questions inline and am curious to know your thoughts about them.

packages/tailwindcss/src/compat/apply-config-to-theme.ts
wongjn: I think if we are to keep maximum compatibility with v3, we shouldn't restrict to 2 values. In v3, [it still emits `line-height`](https://play.tailwindcss.com/Twg4KzukfM) with > 2 values, though I'm not sure how likely this would have occurred in the real world.

Though thinking about this, perhaps we may need to check that there is **at least** two values for the `value[1]` reference?

CHANGELOG.md
wongjn: Sure! Perhaps even having two changelog entries? As some may not understand that this also fixes this in the upgrade tool. So have one changelog entry mentioning better compatibility with in JS compate layer and then another entry saying that the `fontSize` array tuple is fixed for the upgrade tool.

philipp-spiess: two sound good!

packages/tailwindcss/src/compat/apply-config-to-theme.ts
philipp-spiess: ah yep let's check for at least 2 entries then

wongjn: I've just tested again in v3 and an array value with one element, i.e. `2xl: ['2rem']` actually *unsets* the line-height. I'll assume we want to mirror this behavior as well in the compat layer.

wongjn: OK! Sure thing :)

PR #10481, opened 2023-02-03
Add word-break: break-word with break-word

loqimean:
Add new CSS style, because `break-words` isn't working properly (not break words to new line as I need)

Screenshot 2023-02-03 at 13 51 50
Should be "приватних" together, like next:
Screenshot 2023-02-03 at 13 55 56

Issue #10004, opened 2022-12-04
`break-words` class might need to include `word-break: break-word`


LeptoFlare:
**What version of Tailwind CSS are you using?** `v3.2.4`
**What build tool (or framework if it abstracts the build tool) are you using?** `Gatsby v5.2.0`
**What version of Node.js are you using?** `18`
**What browser are you using?** `Firefox`
**What operating system are you using?** `Windows`

**Reproduction URL**

https://play.tailwindcss.com/OVE2UF43o2

**Describe your issue**

I state "might" because this might be a browser bug, or another class that I should add, apologies if this is the case.
- The `flex` class is *required* on the parent for this to occur, if I remove that, both results become identical.
- Adding `flex-col` *also* causes both results to become identical, which is a bit odd to me.

Here is what it looks like on my screen:
![image](https://user-images.githubusercontent.com/47790183/205520340-1ab51c06-a29d-47a2-a669-950695d770f0.png)

I found it odd that in the docs the `break-words` class uses `overflow-wrap` exclusively, while all of the other classes at least include `word-break`.
![image](https://user-images.githubusercontent.com/47790183/205520420-52ed056b-8473-45f7-b9ab-5e6749a59b01.png)

Perhaps all of those classes should include both `overflow-wrap` and `word-break`.

tonyxiao:
<img width="1122" alt="CleanShot 2023-05-26 at 16 58 35@2x" src="https://github.com/tailwindlabs/tailwindcss/assets/696842/f3165de3-fd06-4213-b5c8-90dbbe333f47">

It seems that `min-w-0` doesn't fix it in all situations. I'm using shadcn/ui `AlertDialogTitle` component and this is what I'm getting when having a long title. `overflow-wrap: anywhere` fixes fwiw

serkodev:
https://github.com/tailwindlabs/tailwindcss/discussions/12127 Created a related research and disscustion for this issue.

PR #17633, opened 2025-04-10
Add word-break: break-word to Tailwindcss

NankoPrinzhorn:
# Add word-break: break-word to Tailwindcss
Hi, 

I've noticed that there is no class for word-break: break-word;. 

With this PR a new class called `break-word` will be added that allows text to be wrapped by word.
This is an improvement on the already existing `break-all` class.

The difference is that break-all will not take words into account when slicing content on multiple lines, while break-word does. See [Mozilla docs on work break](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)

# Break-all
image

# Break-word
image

### DIFF
--- a/packages/tailwindcss/src/__snapshots__/intellisense.test.ts.snap
+++ b/packages/tailwindcss/src/__snapshots__/intellisense.test.ts.snap
@@ -3591,6 +3591,7 @@ exports[`getClassList 1`] = `
   "break-after-left",
   "break-after-page",
   "break-after-right",
+  "break-word",
   "break-all",
   "break-before-all",
   "break-before-auto",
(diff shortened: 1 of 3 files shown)


wongjn:
You should be able to use [`wrap-break-word`](https://tailwindcss.com/docs/overflow-wrap) to achieve a similar behavior. As per [the spec](https://www.w3.org/TR/css-text-3/#word-break-property), `break-word` is considered deprecated:

> the [`word-break`](https://www.w3.org/TR/css-text-3/#propdef-word-break) property also supports a deprecated `break-word` keyword.

Also see:
- https://github.com/tailwindlabs/tailwindcss/issues/10004#issuecomment-1337464731
- https://github.com/tailwindlabs/tailwindcss/pull/10481#issuecomment-1416077930

NankoPrinzhorn:
Hi @wongjn,

I've totally looked past those... I'll close the PR 🙈

Thanks

PR #17675, opened 2025-04-14
Prevent duplicate suggestions when using `@theme` and `@utility` together

### LINKED ISSUE #1313: Duplicate suggestion appears when a class is declared using @theme and then extended with a weaker @utility

v1.99.2

v0.14.15

v4.1.3

pnpm

Windows 11 with WSL (CentOS Stream 9)

**Reproduction URL**

**index.css**
```css
@import "tailwindcss";

@theme {
  --text-subheader: 0.75rem;
  --text-subheader--line-height: 1rem;
  --text-subheader--letter-spacing: 0.25rem;
  --text-subheader--font-weight: 600;
}

@utility text-subheader {
  text-transform: uppercase;
}
```

**index.js**
```js
import './index.css'

console.log(`
  <div class="text-subhe">
`)
```

**Describe your issue**

I declared a text style called `subheader`. Since passing nested keys is limited, I declared the additional uppercase separately in a weaker utility as a 'new' class. In the end, these get merged because they share the same name, so it's pointless for IntelliSense to show a duplicate hint.

Image

**Related**

* https://github.com/tailwindlabs/tailwindcss/discussions/16317#discussioncomment-12817719

**Extra**

I know that the strange duplication is caused by the fact that I redeclare a separate utility named `.text-subheader` on top of the `.text-subheader` declared in `@theme`; this can also be seen in the playground hint.

Still, it doesn't make sense to show two identical suggestions, even if they're coming from two different sources.

* https://play.tailwindcss.com/Lwqss6zmTS

Image

thecrypticace:
Fixes https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1313

Right now given this CSS:
```css
@theme reference {
  --text-header: 1.5rem;
}

@utility text-header {
  text-transform: uppercase;
}
```

You'll see two entries for `text-header` in IntelliSense completions but we only want you to see one. This PR solves this by merging their modifier lists and de-duping by class name.

### DIFF
--- a/packages/tailwindcss/src/intellisense.ts
+++ b/packages/tailwindcss/src/intellisense.ts
@@ -20,16 +20,18 @@ export type ClassEntry = [string, ClassMetadata]
 const IS_FRACTION = /^\d+\/\d+$/
 
 export function getClassList(design: DesignSystem): ClassEntry[] {
-  let list: ClassItem[] = []
+  let items = new DefaultMap<string, ClassItem>((utility) => ({
+    name: utility,
+    utility,
+    fraction: false,
+    modifiers: [],
+  }))
 
   // Static utilities only work as-is
   for (let utility of design.utilities.keys('static')) {
-    list.push({
-      name: utility,
-      utility,
-      fraction: false,
-      modifiers: [],
-    })
+    let item = items.get(utility)
+    item.fraction = false
+    item.modifiers = []
   }
 
   // Functional utilities have their own list of completions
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/intellisense.ts
```diff
@@ -20,16 +20,18 @@ export type ClassEntry = [string, ClassMetadata]
...
   for (let utility of design.utilities.keys('static')) {
-    list.push({
-      name: utility,
-      utility,
-      fraction: false,
-      modifiers: [],
-    })
+    let item = items.get(utility)
+    item.fraction = false
+    item.modifiers = []
```
RobinMalfait, line 34: These are already the defaults, so I think we can just do this instead:

```suggestion
    items.get(utility)
```

RobinMalfait: This does look a little weird though 😅

thecrypticace: Yeah… that's why I didn't write it like that :D

RobinMalfait: We need a new method on the `DefaultMap` for this

```ts
items.touch(utility)
```

Or something

RobinMalfait (approved):

PR #17733, opened 2025-04-21
Handle legacy key behavior in theme-driven suggestions for `@utility`

### LINKED ISSUE #1328: v4 intellisense suggests underscore instead of dot in custom utilities

v1.98.2

v0.14.15

v4.1.0

pnpm

macOS

**Tailwind config**

```css
@theme static {
  --container-height-3_5: calc(theme(--spacing) * 3.5);
)

@utility container-h-* {
  height: --value(--container-height- *);
}
```

**Reproduction URL**

TBD

**Describe your issue**

On the CSS side, the variable needs avoid `.`,
thus the CSS variable is `--container-height-3_5` instead of `--container-height-3.5`.

The `@utility container-h-* {}` generates/detect `container-h-3.5` instead of `container-h-3_5`.

However the intellisense suggests `container-h-3_5` which does not work.

My guess this is a regression caused by: https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1184

thecrypticace:
Fixes https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1328

The alpha and beta releases used `_` in theme keys to represent a `.`. This meant we used `--leading-1_5` instead of `--leading-1\.5` to add utilities like `leading-1.5`. We prefer the use of the escaped dot now but still want to make sure suggestions for the legacy key format still works as expected when surrounded by numbers.

This is the same as #16433 but for `@utility` since we apparently missed this when emitting suggestions for it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Don't scan `.geojson` files for classes by default ([#17700](https://github.com/tailwindlabs/tailwindcss/pull/17700))
 - Hide default shadow suggestions when missing theme keys ([#17743](https://github.com/tailwindlabs/tailwindcss/pull/17743))
+- Replace `_` with `.` in theme suggestions for `@utility` if surrounded by digits ([#17743](https://github.com/tailwindlabs/tailwindcss/pull/17743))
 
 ## [4.1.4] - 2025-04-14
 
--- a/packages/tailwindcss/src/intellisense.test.ts
+++ b/packages/tailwindcss/src/intellisense.test.ts
@@ -579,6 +579,10 @@ test('Theme keys with underscores are suggested with underscores', async () => {
       /* This will get suggeted with an underscore */
       --spacing-logo_margin: 0.875rem;
     }
+
+    @utility ex-* {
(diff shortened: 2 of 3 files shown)


philipp-spiess (approved):
Makes sense to make this behave the same as the default spacing utilities but I wonder when we can remove that special-casing for underscore separators like `--spacing-1_5`   😬 

Do you think this (=the special handling for `_`) can eventually be moved into the IntelliSense extensions? 🤔

thecrypticace:
@philipp-spiess oh… hmmm. I don't know. The main problem I see with moving it into IntelliSense is that we really only want this behavior for theme keys.

If someone were to define a utility like this:
```css
@utility foo-2_5 { color: red; }
```

IntelliSense wouldn't know enough to ignore the replacement for this one utility so it would get suggested as `foo-2.5` which would not work.

thecrypticace:
It would be nice to remove this behavior though since it was really only for backwards compatibility with the alpha/beta releases. Maybe this is something we can discuss and figure out a timeline if/when we can remove it.

philipp-spiess:
Yep totally makes sense. I wonder if in the future it would be helpful if a suggestion derrived from a theme key can also emit the used theme key variable somehow - could be fun for go-to-source or so _or_ to find out when using a deprecated theme key 🤔

PR #17763, opened 2025-04-23
Bump all Tailwind CSS related dependencies during upgrade

RobinMalfait:
This PR bumps all Tailwind CSS related dependencies when running the upgrade tool _if_ the dependency exists in your package.json file.

E.g.: if you have `tailwindcss` and `@tailwindcss/vite` in your package.json, then both will be updated to the latest version.

This PR is not trying to be smart and skip updating if you are already on the latest version. It will just try and update the dependencies and your package manager will do nothing in case it was already the latest version.

## Test plan

Ran this on one of my personal projects and this was the output:
image

Notice that I don't have `@tailwindcss/vite` logs because I am using `@tailwindcss/postcss`.


packages/@tailwindcss-upgrade/src/utils/packages.ts
```diff
@@ -31,6 +31,14 @@ export function pkg(base: string) {
         throw e
       }
     },
+    async has(name: string) {
+      try {
+        let packageJsonPath = resolve(base, 'package.json')
+        let packageJsonContent = await fs.readFile(packageJsonPath, 'utf-8')
+        return packageJsonContent.includes(`"${name}":`)
```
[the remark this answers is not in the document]
RobinMalfait: Also, not even trying to parse the JSON file. Took this from the prettier upgrade script we had (which I replaced with this) which means that you don't have to list `dependencies`, `devDependenies`, `optionalDependencies`, ...

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -305,12 +315,6 @@ async function run() {
     await migratePostCSSConfig(base)
   }
 
-  info('Updating dependencies…')
-  {
-    // Migrate the prettier plugin to the latest version
-    await migratePrettierPlugin(base)
-  }
```
RobinMalfait, line 312: Added this to the array of dependencies to upgrade.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -229,11 +228,22 @@ async function run() {
...
-      success(`Updated package: ${highlight('tailwindcss')}`, { prefix: '↳ ' })
-    } catch {}
+    for (let dependency of [
+      'tailwindcss',
+      '@tailwindcss/cli',
+      '@tailwindcss/postcss',
+      '@tailwindcss/vite',
+      '@tailwindcss/node',
+      '@tailwindcss/oxide',
+      'prettier-plugin-tailwindcss',
```
RobinMalfait, line 238: Added `prettier-plugin-tailwindcss` here instead of the dedicated `migratePrettierPlugin` step.

packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -229,11 +228,22 @@ async function run() {
...
     info('Updating dependencies…')
-    try {
-      // Upgrade Tailwind CSS
-      await pkg(base).add(['tailwindcss@latest'])
-      success(`Updated package: ${highlight('tailwindcss')}`, { prefix: '↳ ' })
-    } catch {}
+    for (let dependency of [
+      'tailwindcss',
+      '@tailwindcss/cli',
+      '@tailwindcss/postcss',
```
RobinMalfait, line 234: The PostCSS config is migrated later, which means that on v3 projects at the time of running this will result in a no-op.

thecrypticace (approved):

PR #17790, opened 2025-04-25
Add `h-lh`/`min-h-lh`/`max-h-lh` utilities to match an elements line height

thecrypticace:
This PR adds the following utilities that can be used to match an elements line height:

- `h-lh`
- `min-h-lh`
- `max-h-lh`

These are all equivalent to providing `1lh` as an arbitrary value. e.g. `h-[1lh]`


packages/tailwindcss/src/utilities.ts
```diff
@@ -951,22 +951,39 @@ export function createUtilities(theme: Theme) {
...
+      }
+    }
+
+    if (forHeight) {
+      staticUtility(`h-${key}`, [['height', value]])
+      staticUtility(`min-h-${key}`, [['min-height', value]])
+
+      if (forMax) {
+        staticUtility(`max-h-${key}`, [['max-height', value]])
+      }
```
adamwathan, line 986: I think it might be nicer to just store the common stuff in a variable and `concat` the extra stuff for the utilities that use them. Means a couple of loops instead of one but I think the simplicity/clarity of the code would be worth it.

thecrypticace: cleaned all that up 👍

adamwathan: 🤩

adamwathan (approved):

PR #17815, opened 2025-04-28
Skip `color-mix(…)` when opacity is `100%`

RobinMalfait:
This PR improves colors with alpha values where the alpha value results in 100%.

Before this change, a class like `bg-red-500/100` would be generated as:

```css
.bg-red-500\/100 {
  background-color: #ef4444;
}

  .bg-red-500\/100 {
    background-color: color-mix(in oklab, var(--color-red-500) 100%, transparent);
  }
}
```

But we don't need the `color-mix`, or the fallback styles at all in case the alpha value is 100%. 

After this change the `bg-red-500/100` class will generate as:

```css
.bg-red-500\/100 {
  background-color: var(--color-red-500);
}
```

Which is essentially the same as `bg-red-500`, but we can migrate that away in the future. At least the generated CSS is smaller.

## Test plan

1. Added a test to ensure the generated value doesn't include color-mix at all.

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -10695,6 +10695,8 @@ test('bg', async () => {
         'bg-red-500/2.75',
         'bg-red-500/[0.5]',
         'bg-red-500/[50%]',
+        'bg-red-500/100',
+        'bg-red-500/[100%]',
         'bg-blue-500',
         'bg-current',
         'bg-current/50',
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/utilities.test.ts
```diff
@@ -26032,7 +26042,7 @@ describe('spacing utilities', () => {
 
 describe('custom utilities', () => {
   test('custom static utility', async () => {
-    let { build } = await compile(css`
```
thecrypticace, line 26035: These format changes seem weird

thecrypticace: Did Prettier update and change this or something?

RobinMalfait: I was too quick to quit vim and prettier didn't run in time 😂

thecrypticace (approved):

PR #17831, opened 2025-04-30
Design system driven upgrade migrations

RobinMalfait:
This PR introduces a vastly improved upgrade migrations system, to migrate your codebase and modernize your utilities to make use of the latest variants and utilities.

It all started when I saw this PR the other day: https://github.com/tailwindlabs/tailwindcss/pull/17790 

I was about to comment "Don't forget to add a migration". But I've been thinking about a system where we can automate this process away. This PR introduces this system.

This PR introduces upgrade migrations based on the internal Design System, and it mainly updates arbitrary variants, arbitrary properties and arbitrary values.

## The problem

Whenever we ship new utilities, or you make changes to your CSS file by introducing new `@theme` values, or adding new `@utility` rules. It could be that the rest of your codebase isn't aware of that, but you could be using these values.

For example, it could be that you have a lot of arbitrary properties in your codebase, they look something like this:

```html
<div class="[color-scheme:dark] [text-wrap:balance]"></div>
```

Whenever we introduce new features in Tailwind CSS, you probably don't keep an eye on the release notes and update all of these arbitrary properties to the newly introduced utilities.

But with this PR, we can run the upgrade tool:

```console
npx -y @tailwindcss/upgrade@latest
```

...and it will upgrade your project to use the new utilities:

```html
<div class="scheme-dark text-balance"></div>
```

It also works for arbitrary values, for example imagine you have classes like this:

```html
<!-- Arbitrary property -->
<div class="[max-height:1lh]"></div>

<!-- Arbitrary value -->
<div class="max-h-[1lh]"></div>
```

Running the upgrade tool again:

```console
npx -y @tailwindcss/upgrade@latest
```

... gives you the following output:

```html
<!-- Arbitrary property -->
<div class="max-h-lh"></div>

<!-- Arbitrary value -->
<div class="max-h-lh"></div>
```

This is because of the original PR I mentioned, which introduced the `max-h-lh` utilities.

A nice benefit is that this output only has 1 unique class instead of 2, which also potentially reduces the size of your CSS file.

It could also be that you are using arbitrary values where you (or a team member) didn't even know an alternative solution existed.

E.g.:

```html
<div class="w-[48rem]"></div>
```

After running the upgrade tool you will get this:

```html
<div class="w-3xl"></div>
```

We can go further though. Since the release of Tailwind CSS v4, we introduced the concept of "bare values". Essentially allowing you to type a number on utilities where it makes sense, and we produce a value based on that number.

So an input like this:

```html
<div class="border-[123px]"></div>
```

Will be optimized to just:

```html
<div class="border-123"></div>
```

This can be very useful for complex utilities, for example, how many times have you written something like this:

```html
<div class="grid-cols-[repeat(16,minmax(0,1fr))]"></div>
```

Because up until Tailwind CSS v4, we only generated 12 columns by default. But since v4, we can generate any number of columns automatically.

Running the migration tool will give you this:

```html
<div class="grid-cols-16"></div>
```

### User CSS

But, what if I told you that we can keep going...

In [Catalyst](https://tailwindcss.com/plus/ui-kit) we often use classes that look like this for accessibility reasons:

```html
<div class="text-[CanvasText] bg-[Highlight]"></div>
```

What if you want to move the `CanvasText` and `Highlight` colors to your CSS:

```css
@import "tailwincdss";

@theme {
  --color-canvas: CanvasText;
  --color-highlight: Highlight;
}
```

If you now run the upgrade tool again, this will be the result:

```html
<div class="text-canvas bg-highlight"></div>
```

We never shipped a `text-canvas` or `bg-highlight` utility, but the upgrade tool uses your own CSS configuration to migrate your codebase.

This will keep your codebase clean, consistent and modern and you are in control.

Let's look at one more example, what if you have this in a lot of places:

```html
<div class="[scrollbar-gutter:stable]"></div>
```

And you don't want to wait for the Tailwind CSS team to ship a `scrollbar-stable` (or similar) feature. You can add your own utility:

```css
@import "tailwincdss";

@utility scrollbar-stable {
  scrollbar-gutter: stable;
}
```

```html
<div class="scrollbar-stable"></div>
```

## The solution — how it works

There are 2 big things happening here:

1. Instead of us (the Tailwind CSS team) hardcoding certain migrations, we will make use of the internal `DesignSystem` which is the source of truth for all this information. This is also what Tailwind CSS itself uses to generate the CSS file.

   The internal `DesignSystem` is essentially a list of all:

   1. The internal utilities
   2. The internal variants
   3. The default theme we ship
   4. The user CSS
      1. With custom `@theme` values
      2. With custom `@custom-variant` implementations
      3. With custom `@utility` implementations
2. The upgrade tool now has a concept of `signatures`

The signatures part is the most interesting one, and it allows us to be 100% sure that we can migrate your codebase without breaking anything.

A signature is some unique identifier that represents a utility. But 2 utilities that do the exact same thing will have the same signature.

To make this work, we have to make sure that we normalize values. One such value is the selector. I think a little visualization will help here:

| UTILITY          | GENERATED SIGNATURE     |
| ---------------- | ----------------------- |
| `[display:flex]` | `.x { display: flex; }` |
| `flex`           | `.x { display: flex; }` |

They have the exact same signature and therefore the upgrade tool can safely migrate them to the same utility.

For this we will prefer the following order:

1. Static utilities — essentially no brackets. E.g.: `flex`, `grid-cols-2`
2. Arbitrary values — e.g.: `max-h-[1lh]`, `border-[2px]`
3. Arbitrary properties — e.g.: `[color-scheme:dark]`, `[display:flex]`

We also have to canonicalize utilities to there minimal form. Essentially making sure we increase the chance of finding a match.

```
[display:_flex_] → [display:flex] → flex
[display:_flex]  → [display:flex] → flex
[display:flex_]  → [display:flex] → flex
[display:flex]   → [display:flex] → flex
```

If we don't do this, then the signatures will be slightly different, due to the whitespace:

| UTILITY            | GENERATED SIGNATURE       |
| ------------------ | ------------------------- |
| `[display:_flex_]` | `.x { display:  flex ; }` |
| `[display:_flex]`  | `.x { display:  flex; }`  |
| `[display:flex_]`  | `.x { display: flex ; }`  |
| `[display:flex]`   | `.x { display: flex; }`   |

### Other small improvements

A few other improvements are for optimizing existing utilities:

1. Remove unnecessary data types. E.g.:

   - `bg-[color:red]` -> `bg-[red]`
   - `shadow-[shadow:inset_0_1px_--theme(--color-white/15%)]` -> `shadow-[inset_0_1px_--theme(--color-white/15%)]`

   This also makes use of these signatures and if dropping the data type results in the same signature then we can safely drop it.

   Additionally, if a more specific utility exists, we will prefer that one. This reduced ambiguity and the need for data types.

   - `bg-[position:123px]` → `bg-position-[123px]`
   - `bg-[123px]` → `bg-position-[123px]`
   - `bg-[size:123px]` → `bg-size-[123px]`

2. Optimizing modifiers. E.g.:
   - `bg-red-500/[25%]` → `bg-red-500/25`
   - `bg-red-500/[100%]` → `bg-red-500`
   - `bg-red-500/100` → `bg-red-500`

3. Hoist `not` in arbitrary variants

   - `[@media_not_(prefers-color-scheme:dark)]:flex` → `not-[@media_(prefers-color-scheme:dark)]:flex` → `not-dark:flex` (in case you are using the default `dark` mode implementation

4. Optimize raw values that could be converted to bare values. This uses the `--spacing` variable to ensure it is safe.

   - `w-[64rem]` → `w-256`


wongjn:
So for `bg-[position:123px]`, `bg-[123px]` is generally preferred over `bg-position-[123px]`? Same with `bg-size-[]`?

RobinMalfait:
> So for `bg-[position:123px]`, `bg-[123px]` is generally preferred over `bg-position-[123px]`? Same with `bg-size-[]`?

@wongjn no that was just a poor example on my part. Let me add some dedicated tests for this example because `bg-position-[123px]` is definitely preferred over `bg-[123px]` for sure! 

In the future we will add more specific utilities (similar to `bg-position` and `bg-size` we introduced recently in v4.1.0) to reduce and ideally even get rid of the those data type hints.

packages/@tailwindcss-upgrade/src/utils/dimension.ts
```diff
@@ -0,0 +1,18 @@
+import { DefaultMap } from '../../../tailwindcss/src/utils/default-map'
+
+// Parse a dimension such as `64rem` into `[64, 'rem']`.
+export const dimensions = new DefaultMap((input) => {
+  let match = /^(?<value>-?(\d*)?\.?\d+)(?<unit>.*)$/.exec(input)
```
thecrypticace, line 5: I think something like this might be a bit better

```suggestion
  let match = /^(?<value>-?(?:\d*\.)?\d+)(?<unit>[a-z]+|%)$/i.exec(input)
```

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,368 @@
...
+
+          if (supportsNegative) {
+            let negativeCandidateString = `-${candidateString}`
+            let signature = signatures.get(negativeCandidateString)
+            if (typeof signature === 'string') lookup.get(signature).push(negativeCandidateString)
+          }
+        }
+      }
+    }
+  }
```
thecrypticace, line 71: I think you can simplify a good bit of this by using `getClassList(…)`. It's computing much of the same stuff

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,368 @@
...
+    candidate: Extract<Candidate, { kind: 'functional' | 'arbitrary' }>,
+  ): Generator<Candidate> {
+    // Find a corresponding utility for the same signature
+    let replacements = utilities.get(targetSignature)
+
+    // Multiple utilities can map to the same signature. Not sure how to migrate
+    // this one so let's just skip it for now.
+    //
+    // TODO: Do we just migrate to the first one?
+    if (replacements.length > 1) return
```
thecrypticace, line 211: as a note: I think the replacement would only be valid in this case if _all_ of the signatures matched for a given candidate.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    // Find a corresponding functional utility for the same signature
+    else if (replacements.length === 0) {
+      // An arbitrary property will only set a single property, we can use that
+      // to find functional utilities that also set this property.
+      let value =
+        candidate.kind === 'arbitrary' ? candidate.value : (candidate.value?.value ?? null)
+
+      if (value !== null) {
+        let bareValue = spacing.get(designSystem)?.get(value)
```
thecrypticace, line 250: The comments below with this variable name is highly confusing. Can we name this `spacingValue` instead or something?

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,368 @@
...
+              `${root}-[${value}]${printModifier(candidate.modifier)}`,
+            )) {
+              yield replacementCandidate
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
thecrypticace, line 305: [insert street fighter indentation meme]

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
RobinMalfait: Yep the `replacements` array here is `signature => [class1, class2, ...]`

So this would happen in a scenario like this:

Input: 
```
bg-(--color-red-500)
```

Signature (roughly):
```
.x { background-color: oklch(63.7% 0.237 25.331) }
```

Replacements (pre computed and match the same signature):
```
['bg-red-500', 'bg-red-500/100']
```

And if you had:
```css
@utility foo { @apply bg-red-500; }
```

Then that one would also match. So right now we just skip, but we could also pick the first one, or the shortest one, or ...

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
RobinMalfait: Hmm yeah ~~but there are some issues with that due to the modifiers and what not.~~

So current situation:
```
Total pre-computed utilities: 19301
[798.96ms] pre-compute utilities
```

When using `getClassList()`:
```
Total pre-computed utilities:19301
[841.78ms] pre-compute utilities
```

It is just a tiny bit slower.

I was thinking that there was an issue with the modifiers because we don't want to pre-compute `bg-red-500/50` for example. But the modifiers are separate so can clean that up nicely:
```ts
const preComputedUtilities = new DefaultMap<DesignSystem, DefaultMap<string, string[]>>((ds) => {
  let signatures = computeUtilitySignature.get(ds)
  let lookup = new DefaultMap<string, string[]>(() => [])

  for (let [className, meta] of ds.getClassList()) {
    let signature = signatures.get(className)
    if (typeof signature !== 'string') continue
    lookup.get(signature).push(className)

    for (let modifier of meta.modifiers) {
      // Modifiers representing numbers can be computed and don't need to be
      // pre-computed. Doing the math and at the time of writing this, this
      // would save you 250k additionally pre-computed utilities...
      if (isValidSpacingMultiplier(modifier)) {
        continue
      }

      let classNameWithModifier = `${className}/${modifier}`
      let signature = signatures.get(classNameWithModifier)
      if (typeof signature !== 'string') continue
      lookup.get(signature).push(classNameWithModifier)
    }
  }

  return lookup
})
```

packages/tailwindcss/src/utilities.ts
```diff
@@ -1300,6 +1300,7 @@ export function createUtilities(theme: Theme) {
       let value
       if (candidate.value.kind === 'arbitrary') {
         value = candidate.value.value
+        value = negative ? `calc(${value} * -1)` : value
```
philipp-spiess, line 1303: Seems like this needs a changelog entry

packages/tailwindcss/src/design-system.ts
```diff
@@ -127,6 +137,14 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
     compileAstNodes(candidate: Candidate) {
       return compiledAstNodes.get(candidate)
     },
+
+    printCandidate(candidate: Candidate) {
+      return printCandidate(designSystem, candidate)
+    },
+    printVariant(variant: Variant) {
+      return printVariant(variant)
+    },
```
philipp-spiess, line 146: What's the value of adding this to the design system? I remember we moved it into the upgrade tool to avoid shipping this with everyone installing `tailwindcss`. Since this is part of a default exported object now, it can't be DCE-ed 🤔 
Are we planning on using this in other places?

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
philipp-spiess: Does that mean we can make `getClassList()` faster? 😇

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  let replacementAsCss = designSystem
+    .candidatesToCss([designSystem.printCandidate(replacement)])
+    .join('\n')
+
+  let isSafeMigration = true
+  ValueParser.walk(ValueParser.parse(value), (node) => {
+    if (node.kind === 'function' && node.value === 'var') {
+      let variable = node.nodes[0].value
+      let r = new RegExp(`var\\(${variable}[,)]\\s*`, 'g')
```
philipp-spiess, line 325: Does this need to handle eventual space around the variable name? e.g.

```css
.foo {
  color: var( --red);
  color: var(--red );
}
```

etc.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-modernize-arbitrary-values.ts
```diff
@@ -317,48 +257,6 @@ export function migrateModernizeArbitraryValues(
...
-            else if (value === ':read-only') return 'read-only'
-            // Content
-            else if (value === ':empty') return 'empty'
-            // Interactive
-            else if (value === ':focus-within') return 'focus-within'
-            else if (value === ':focus') return 'focus'
-            else if (value === ':focus-visible') return 'focus-visible'
-            else if (value === ':active') return 'active'
-            else if (value === ':enabled') return 'enabled'
-            else if (value === ':disabled') return 'disabled'
```
philipp-spiess, line 360: love that we can rm this

packages/@tailwindcss-upgrade/src/utils/memcpy.ts
```diff
@@ -0,0 +1,7 @@
+export function memcpy<T extends object, U extends object | null>(target: T, source: U): U {
```
philipp-spiess, line 1: Really don't think this name is explanatory, especially now that it's a "global" thing. What about `replaceEntries`?

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
```diff
@@ -0,0 +1,404 @@
...
+
+  // Restore the `@theme inline` to the original value
+  // @ts-expect-error We are monkey-patching a method that's private
+  designSystem.theme.values.get = originalGet
+
+  // Re-add the `inline` option, in case future lookups are done
+  for (let value of restorableInlineOptions) {
+    value.options |= ThemeOptions.INLINE
+  }
+
```
philipp-spiess, line 402: should this be wrapped in a `try { … } finally { … }` for safety?

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
```diff
@@ -0,0 +1,404 @@
...
+      temporarilyDisableThemeInline(designSystem, () => substituteAtApply(ast, designSystem))
+
+      // We will be mutating the AST, so we need to clone it first to not affect
+      // the original AST
+      ast = structuredClone(ast)
+
+      // Optimize the AST. This is needed such that any internal intermediate
+      // nodes are gone. This will also cleanup declaration nodes with undefined
+      // values or `--tw-sort` declarations.
+      walk(ast, (node, { replaceWith }) => {
```
philipp-spiess, line 51: Is there a way we can reuse the optimize method from core here? Feels weird to have to re-implement this and it'll likely mean we'll diverge eventually 🤔

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
```diff
@@ -0,0 +1,404 @@
...
+          replaceWith([])
+        }
+      })
+
+      // Resolve theme values to their inlined value.
+      //
+      // E.g.:
+      //
+      // `[color:var(--color-red-500)]` → `[color:oklch(63.7%_0.237_25.331)]`
+      // `[color:oklch(63.7%_0.237_25.331)]` → `[color:oklch(63.7%_0.237_25.331)]`
```
philipp-spiess, line 75: I'm not sure if we should do this, it feels like if someone hardcodes the color value like this, they explicitly _don't want to use the variable (e.g. because the variable might change at runtime for some reason?)

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
RobinMalfait: I originally did that, but it made things slower because it's creating a new AST and tracking CSS variables, removing it from the output, hoisting at-root, ... and all of that jazz. But you are right that this might result in diverged code. 🤔 

Maybe it's fine, but here are some numbers from projects I tested it on.

Running this on Tailwind Plus: (not a lot of arbitrary properties/values are used)
- With the current implementation: 48s
- With the `optimizeAst` step: 55s

Running this on all Tailwind UI Templates (ran the script in parallel for each template and this is the final result):
- With the current implementation: 1m21s
- With the `optimizeAst` step: 2m48s
 
Running this on a personal project:
- With the current implementation: 29s
- With the `optimizeAst` step: 1m8s

philipp-spiess: Oof that's quite a difference. Yikes. Maybe `optimizeAst` needs to be refactored somehow? Idk, let's keep thinking about this, nothing you need to change immediately I feel like.

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
RobinMalfait: Do you mean for all of the inlined values or just this particular case (arbitrary property + raw value)?

Because this is executed on all values, so if we don't do this then it would never be safe to do these migrations:
- `w-[64rem]` → `w-256`
- `shadow-[0_1px_rgb(0_0_0_/_0.05)]` → `shadow-2xs`

Because these use variables behind the scenes. Unless you are using `@theme inline`, then it would be a guaranteed safe migration.

It also means that if you introduce a custom shadow because you want to get rid of the duplication that it would also not work without a `@theme inline`.

Personally I feel like this is a migration that you _do_ want in 99% of the cases. We also require to run this on a clean repo, and after upgrading we show a "Verify the changes and commit them to your repository."

So even if you don't want this, there is still a way to undo this change.

I'm fine with whatever we decide to do here, but I personally would love to keep it.

packages/tailwindcss/src/design-system.ts
RobinMalfait: This is a bit for selfish reasons, but here are the reasons for it:

- We already have `parseCandidate` and `parseVariant` exposed
- I've been copying the `printCandidate` / `printVariant` for the internal debugger / extractor tools
- @thecrypticace mentioned he has a use case for it in intellisense
- The total size of these print functions is still very tiny
   - Without, the `dist/` folder we publish is 606 KB
   - With, the `dist/` folder we publish is 614 KB

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
RobinMalfait: This shouldn't be necessary because the values are already canonicalized to their minimal form. So the whitespace inconsistencies should be handled already.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
thecrypticace: Yeah getClassList _does_ need to sort classes because people generally expect the suggestions to be sorted. It still needs some work to sort them even better tho.

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
thecrypticace: Wow that's a massive difference. We should absolutely profile optimizeAst(…) then and see if we can do something to make it faster.

packages/@tailwindcss-upgrade/src/utils/memcpy.ts
RobinMalfait: I took this from the C++ world (https://en.cppreference.com/w/cpp/string/byte/memcpy)

It's basically `Object.assign(dst, src)` but also clears all values in `dst` that are not in `src`.

I'm open to suggestions for a better name, btu I feel like `replaceEntries` is less correct 🤔

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
RobinMalfait: Good idea!

packages/tailwindcss/src/utilities.ts
RobinMalfait: Ah yep! I noticed this when one of the migrations felt off:

image

Spot the bug 👀

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    // Find a corresponding functional utility for the same signature
+    else if (replacements.length === 0) {
+      // An arbitrary property will only set a single property, we can use that
+      // to find functional utilities that also set this property.
+      let value =
+        candidate.kind === 'arbitrary' ? candidate.value : (candidate.value?.value ?? null)
+
+      if (value !== null) {
```
thecrypticace, line 218: Could drop a level of indentation if you flipped this conditional w/ `return`

RobinMalfait: Ah yep, I had more code below it, and since we were `yield`-ing, I didn't want to return because then the code below wouldn't run. `continue` also didn't work because not in a loop.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,340 @@
...
+          // Try as arbitrary value
+          for (let replacementCandidate of parseCandidate(designSystem, `${root}-[${value}]`)) {
+            yield replacementCandidate
+          }
+
+          // Try as arbitrary value with modifier
+          if (candidate.modifier) {
+            for (let replacementCandidate of parseCandidate(
+              designSystem,
+              `${root}-[${value}]${printModifier(candidate.modifier)}`,
```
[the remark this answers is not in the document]
RobinMalfait: Excellent question!
A: Skill issue

Fixed.

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
RobinMalfait: It's the upgrade tool so we can get away with slower builds, but this was surprising to me as well haha

packages/@tailwindcss-upgrade/src/codemods/template/migrate-drop-unnecessary-data-types.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    ['bg-[color:#008cc]', 'bg-[#008cc]'],
+
+    // A position can be inferred from the value
+    ['bg-[position:123px]', 'bg-[123px]'],
+
+    // A color is the default for `bg-*`
+    ['bg-(color:--my-value)', 'bg-(--my-value)'],
+
+    // A position is not the default, so the `position` data type is kept
+    ['bg-(position:--my-value)', 'bg-(position:--my-value)'],
```
thecrypticace, line 43: Do you think there's a version of this in the future where we migrate this one to `bg-position-(--my-value)`?

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -0,0 +1,368 @@
...
+    // [display:_flex_] => [display:flex]
+    // [display:_flex]  => [display:flex]
+    // [display:flex_]  => [display:flex]
+    // [display:flex]   => [display:flex]
+    // ```
+    //
+    let canonicalizedCandidate = printCandidate(designSystem, readonlyCandidate)
+    if (canonicalizedCandidate !== rawCandidate) {
+      return migrateArbitraryUtilities(designSystem, _userConfig, canonicalizedCandidate)
+    }
```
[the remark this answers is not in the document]
RobinMalfait: 1. Only reason I didn't do something like you suggest is because then you are parsing and printing values that we don't even care about. Right now we only care about arbitrary properties and function utilities with arbitrary properties.

   That said, we do have this piece of code in the main `migrate` function. So maybe this should just be a migration step (near the beginning) and not worry about it at all here 🤔 

    https://github.com/tailwindlabs/tailwindcss/blob/45cd32eed7d328bd14caa9895077f5df441a6da7/packages/%40tailwindcss-upgrade/src/codemods/template/migrate.ts#L65-L74
    
2. Related to the above
3. Moved it into the DesignSystem already. But let's see if we want to keep it there or not. Another conversation related to this: https://github.com/tailwindlabs/tailwindcss/pull/17831#discussion_r2071673474

packages/@tailwindcss-upgrade/src/codemods/template/migrate-drop-unnecessary-data-types.test.ts
RobinMalfait: This is already happening! But this test is just about the removal of unnecessary data types (or well, this specific migration).

Let me add some dedicated tests that go over all the migrations:
https://github.com/tailwindlabs/tailwindcss/blob/987b9e78b5cfd18fe6c7e05ff3551b29686f1a42/packages/%40tailwindcss-upgrade/src/codemods/template/migrate.test.ts#L39-L47

Started with some basic tests, but we could make this much more extensive in the future but feels out of scope for this particular PR I think.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
RobinMalfait: I changed / reworded these things, so should be better now 🤞

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
philipp-spiess: Yeah makes sense

packages/tailwindcss/src/design-system.ts
philipp-spiess: FINE

packages/@tailwindcss-upgrade/src/utils/memcpy.ts
philipp-spiess: Yeah I see how mempcy makes sense but it doesn't feel JS-y at all haha. I don't know of a better name too, for me it does clear out the entries of the object and puts new ones there that's why I suggested this. maybe it's `replaceObject` or so haha. If we can't align on anything another idea is also to just inline this and don't try to abstract?

thecrypticace (approved):

philipp-spiess (approved):
Really excited to run this on my projects 🎉

severinlandolt:
Shesh! This is absolutely insane ✨

RobinMalfait:
@severinlandolt you can try it already with the insiders version: 👀 
```sh
npx -y @tailwindcss/upgrade@insiders
```
Try it on your projects, and let us know (in a new issue) if you run into issues!

PR #17836, opened 2025-05-01
Ignore `@tailwind utilities` inside `@reference`

### LINKED ISSUE #17693: @reference causing re-renders in non related component with HMR

v4.1.3

Vite 6.2.5

v22.14.0

Chrome

Ubuntu

**Reproduction URL**

There is none because it's related to the npm run dev process. But attached is a screenshot where I change only something in ProgressSection.vue, the style is scoped and it re-builds basically all components

Image

**Describe your issue**

When running `npm run dev`, and making a change to one component, instead of the HMR triggering change only for this component, it triggers changes to ALL components that have @reference '@/style.css';

We need the @reference, because we want to use @apply instead of writing the styles inline. Falling back to inline is not an option for us, because the project then will not be maintainable and the code not readable.

In tailwindcss 3, the apply did not create such issues, because we didn't have to use @reference to use it.

Now it takes around 5-10 seconds to show a letter change from a vue component over vite, it's like going back to webpack 😆 

Yes, I have read the documentation that we could use var(--bg-blue-500) or something, but then we can't use spacings, breakspoints etc, all the benefits of tailwind.

thecrypticace:
You can use `@reference "tailwindcss"` or `@reference "../path/to/your/css/file.css"` to reference your theme for use in `@apply`, `theme(…)`, etc…

Unfortunatley, because the imported file still contains `@tailwind utilities` it would trigger a re-scan of the filesystem — even though the use of `@reference` ensures that no CSS can actually be output by the import itself.

This PR does two things:
- Adds some explicit feature detection tests for what features we pick up in a stylesheet based on the CSS written and how things are imported
- Explicitly ignores `@tailwind utilities` inside of `@reference` so it isn't a trigger for file scanning

Because of how Vite itself handles dependencies editing files on disk will still trigger a rebuild of any file using `@reference`. This is because Vite rebuilds files when _any_ of its transitive dependencies change.

For example, given this Vue file:
```vue
<style>
@reference "./styles.css";
</style>
<template> <!-- ... --> </template>
```

And this stylesheet:
```css
@import "tailwindcss";
```

The dependency chain looks like this: `file.vue -> styles.css -> {all the sources in your project}`

Vite sees that a file (e.g. `index.html`) has changed, thus `styles.css` needs change, which means `file.vue` needs to be compiled again as well. Now in reality we depend on the _on disk_ version of styles.css not the compiled version but Vite itself doesn't know that (or have a way to indicate this afaik).

Coming up with a solution to that problem will have to be a separate PR — but there is a workaround:

### 1. Inline the imports from `@import "tailwindcss";`

Replace this in your main stylesheet:
```css
@import "tailwindcss";
```

with this (this is basically what `node_modules/tailwindcss/index.css` is):
```css

@import 'tailwindcss/theme' layer(theme);

/* the rest of your styles imports, styles, etc… */
```

### 2. Split your stylesheet into "main" and "theme" parts

Your "theme" is comprised of the `@import 'tailwindcss/theme' layer(theme);` import, any custom `@theme` blocks, any `@config` directives, and any `@plugin` directives. Move all of these into their own file.

For example, replace this with two files:

```css
@import 'tailwindcss/theme' layer(theme);

@theme {
  --color-primary: #c0ffee;
}

@plugin "./my-plugin.js";

/* the rest of your styles imports, styles, etc… */
```

with a theme file:
```css
@import 'tailwindcss/theme' layer(theme);

/* all your `@theme` stuff goes in this file */
@theme {
  --color-primary: #c0ffee;
}

/* additionally any @config or @plugin does too */
@plugin "./my-plugin.js";
```

and your main CSS file:
```css
@import './my-theme.css'; /* I replaced this import */

/* the rest of your styles imports, styles, etc… */
```

### 3. Import only your "theme" file in your Vue components / CSS modules / etc…

```vue
<style>
@reference "./my-theme.css";
</style>
<template> <!-- ... --> </template>
```

Fixes #17693

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ensure negative arbitrary `scale` values generate negative values ([#17831](https://github.com/tailwindlabs/tailwindcss/pull/17831))
 - Fix HAML extraction with embedded Ruby ([#17846](https://github.com/tailwindlabs/tailwindcss/pull/17846))
+- Don't scan files for utilities when using `@reference` ([#17836](https://github.com/tailwindlabs/tailwindcss/pull/17836))
 
 ## [4.1.5] - 2025-04-30
 
--- a/packages/tailwindcss/src/index.test.ts
+++ b/packages/tailwindcss/src/index.test.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs'
 import path from 'node:path'
 import { describe, expect, it, test } from 'vitest'
-import { compile, Polyfills } from '.'
+import { compile, Features, Polyfills } from '.'
 import type { PluginAPI } from './compat/plugin-api'
 import plugin from './plugin'
 import { compileCss, optimizeCss, run } from './test-utils/run'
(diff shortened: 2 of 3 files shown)


philipp-spiess (approved):

Fedox-die-Ente:
Thank you so much for this fix.

Fedox-die-Ente:
@philipp-spiess is this live?

philipp-spiess:
@Fedox-die-Ente Yeah that went out with [v4.1.6](https://github.com/tailwindlabs/tailwindcss/releases/tag/v4.1.6) last week

PR #17846, opened 2025-05-02
Fix HAML extraction with embedded Ruby

### LINKED ISSUE #17813: Some classes are not generated in Ruby/HAML since upgrading to Tailwind 4

4.1.4

Standalone CLI

20.19

N/A

N/A

**Reproduction URL**

https://github.com/jeremybdk/tailwindhaml

**Describe your issue**

We are using custom classes like "w-[16px] h-[16px]" sometimes in ruby variable used in partials.
But it seems that they are not parsed correctly anymore since switching to Tailwind 4. We never had that issue on Tailwind 3.

I've added a small sample of a haml file that in my opinion should be parsed correctly and as you can see in the output some of the used classes like "w-[16px] h-[16px]" and "w-[12px] h-[12px]" are not included in the output.css while w-[20px] h-[20px] are included in the output.css

thank you

RobinMalfait:
This PR fixes and improves the HAML extractor by ensuring that whenever we detect lines or blocks of Ruby code (`-`, `=` and `~`) characters at the beginning of the line, we treat them as Ruby code.

Fixes: #17813

## Test Plan

1. Existing tests pass
2. Changed 1 existing test which embedded Ruby syntax
3. Added a dedicated test to ensure the HAML file in the linked issue is parsed correctly

Running this in the internal extractor tool you can see that the `w-[12px]`, `w-[16px]`, `h-[12px]`, and `h-[16px]` are properly extracted.

Note: the `mr-12px` is also extracted, but not highlighted because this is not a valid Tailwind CSS class.

image


crates/oxide/src/extractor/pre_processors/haml.rs
```diff
@@ -173,10 +421,20 @@ mod tests {
...
     // https://github.com/tailwindlabs/tailwindcss/pull/17051#issuecomment-2711181352
     #[test]
-    fn test_haml_full_file() {
-        let processed = Haml.process(include_bytes!("./test-fixtures/haml/src-1.haml"));
+    fn test_haml_full_file_17051() {
+        let processed = Haml.process(include_bytes!("./test-fixtures/haml/src-17051.haml"));
+        let actual = std::str::from_utf8(&processed).unwrap();
+        let expected = include_str!("./test-fixtures/haml/dst-17051.haml");
+
+        assert_eq!(actual, expected);
```
philipp-spiess, line 429: I find this a bit confusing to review. Looking at the snapshot alone I am really not sure if the right classes are extract. Should we run the extractor here and assert on the actual candidates being emitted?

crates/oxide/src/extractor/pre_processors/test-fixtures/haml/dst-17051.haml
philipp-spiess, line 1: I don't feel like this snapshot is super helpful for reviewing code. I know that some stuff changed but it's really unclear which of the symbols are now considered candidates and which ones aren't. I'm fine with keeping the snapshot if we can somehow highlight candidates. High level it feels like these stuff should not be changed in a PR like this, expect if we have something that we under-classified or over-classified for some reason. WDYT?

thecrypticace: I agree with Philipp here. What if we added some kind of separator and then listed all the extracted candidates at the bottom of the fixture?

RobinMalfait: Yeah can do that! I typically push a "failing test" first (which is the output of what we expect).

But let me split things here so the diff is clear between commits.

thecrypticace (approved):

PR #17889, opened 2025-05-06
Fix bug replacing modifier variable shorthand syntax underscores

### LINKED ISSUE #17888: Invalid variable shorthand modifier replacement in modifiers
Invalid variable shorthand modifier replacement in modifiers (`_` -> ` `)

<details><summary> Technical Configuration</summary>
<br>
<table>
    <tr></tr>
    <tr>
        <th align="left">What version of Tailwind CSS are you using?</th>
    </tr>
    <tr>
        <td>v4.1.5</td>
    </tr>
    <tr>
        <th align="left">What build tool (or framework if it abstracts the build tool) are you using?</th>
    </tr>
    <tr>
        <td>Tailwind Play</td>
    </tr>
    <tr>
        <th align="left">What version of Node.js are you using?</th>
    </tr>
    <tr>
        <td>N/A</td>
    </tr>
    <tr>
        <th align="left">What browser are you using?</th>
    </tr>
    <tr>
        <th align="left">Chrome</td>
    </tr>
    <tr>
        <th align="left">What operating system are you using?</th>
    </tr>
    <tr>
        <td>macOS</td>
    </tr>
</table>

</details>

---

**Reproduction URL**

https://play.tailwindcss.com/YvIekuzVRd

**Description:**

Tailwind CSS v4 handles underscores inside the arbitrary variable shorthand syntax `(...)` inconsistently depending on whether it's used as a utility's value or as a modifier.

**Expected Behavior:**

When using the `(...)` shorthand for a CSS variable containing underscores, such as `(--my_variable)`, the underscores should be consistently preserved and wrapped in `var()`, regardless of whether it's used directly as a utility value or as a modifier following a `/`. The expected CSS output should be `var(--my_variable)`.

**Actual Behavior:**

1.  **Utility Value:** When used directly as part of a utility (e.g., `p-(--x_x)`), the underscores are correctly preserved, resulting in `var(--x_x)`.
2.  **Modifier Value:** When used as a modifier (e.g., `test/(--x_x)`), the underscores within the variable name are incorrectly replaced with spaces *before* being wrapped in `var()`, resulting in `var(--x x)`.

**Possible Cause (Analysis):**

This seems related to the order of operations during parsing.
*   For utilities `-(...)`, the variable context seems to be recognized *before* or *during* the call to the internal `decodeArbitraryValue` function, allowing its specific rule for preserving underscores within `var()` arguments or `--` names to take effect.
*   For modifiers `/(...)`, the internal `parseModifier` function appears to call `decodeArbitraryValue` on the *raw* extracted content (`--x_x`) *first*, causing default underscore-to-space replacement. Only *after* this decoding step is the (now space-containing) result wrapped in `var()`.

This inconsistency makes the shorthand syntax unreliable when used within modifiers.

brandonmcconnell:
Resolves #17888

**Reproduction URL:** https://play.tailwindcss.com/YvIekuzVRd

Changes:
* Don't use `decodeArbitraryValue` when parsing variable shorthand syntax in modifiers
  * replace `decodeArbitraryValue(modifier.slice(1, -1))` with `modifier.slice(1, -1)`
  * added test case, passing ✅

---

Edit by maintainer: 

This PR fixes an issue where the arbitrary modifier shorthand (e.g.: `bg-red-500/(--my_variable)`) was incorrectly replacing the `_` with a ` ` (space). 
It's important that we still decode the arbitrary value and replace `_` with ` ` in other situations.

For example, if you have this:
```html
<div class="font-(family-name:--my_font,comic_sans_ms)"></div>
```
You would expect the following CSS:
```css
.font-\(family-name\:--foo\,comic_sans_ms\) {
  font-family: var(--my_font,comic sans ms);
}
```
Where the `_` in the variable is **not** replaced, but the `_` in the fallback value **is** being replaced.

### DIFF
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -530,21 +530,13 @@ function parseModifier(modifier: string): CandidateModifier | null {
     // represents a CSS variable.
     if (modifier[0] !== '-' || modifier[1] !== '-') return null
 
-    // Trim the modifier to remove any leading or trailing whitespace
-    modifier = modifier.trim()
+    // Values can't contain `;` or `}` characters at the top-level.
+    if (!isValidArbitrary(modifier)) return null
 
-    // Empty arbitrary values are invalid. E.g.: `data-(--):`
-    //                                                  ^^
-    // Note: we already validated that the `modifier` starts with `--`
-    if (modifier.length === 2) return null
-
-    // Wrap the value in `var(…)` to ensure that it is a CSS variable.
+    // Wrap the value in `var(…)` to ensure that it is a valid CSS variable.
     modifier = `var(${modifier})`
 
     let arbitraryValue = decodeArbitraryValue(modifier)
-
-    // Values can't contain `;` or `}` characters at the top-level.
-    if (!isValidArbitrary(arbitraryValue)) return null
 
     return {
       kind: 'arbitrary',
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -413,6 +413,9 @@ export function* parseCandidate(input: string, designSystem: DesignSystem): Iter
     // represents a CSS variable.
     if (value[0] !== '-' || value[1] !== '-') return
 
+    // Values can't contain `;` or `}` characters at the top-level.
+    if (!isValidArbitrary(value)) return
+
     roots = [[root, dataType === null ? `[var(${value})]` : `[${dataType}:var(${value})]`]]
   }
 
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -530,6 +530,14 @@ function parseModifier(modifier: string): CandidateModifier | null {
     // represents a CSS variable.
     if (modifier[0] !== '-' && modifier[1] !== '-') return null
 
+    // Trim the modifier to remove any leading or trailing whitespace
+    modifier = modifier.trim()
+
+    // Empty arbitrary values are invalid. E.g.: `data-(--):`
+    //                                                  ^^
+    // Note: we already validated that the `modifier` starts with `--`
+    if (modifier.length === 2) return null
+
     // Wrap the value in `var(…)` to ensure that it is a CSS variable.
     modifier = `var(${modifier})`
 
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey!

~~We used to have some functionality like this, but we reverted it in https://github.com/tailwindlabs/tailwindcss/pull/16206 (/cc @philipp-spiess) and we have some dedicated tests where we ensure that if you write `--foo\\_bar` that it correctly parses as-if you wrote `--foo_bar`:~~
https://github.com/tailwindlabs/tailwindcss/blob/5c5ae04db6d228a5882ae211e07cbb8c1bafc501/packages/tailwindcss/src/candidate.test.ts#L938-L108

In the event that we want to re-introduce this (where the variable doesn't require underscores), then we have some more work to do because the fallback value can also contain underscores where we _do_ want to convert them to spaces. E.g.:

```html
<div class="bg-(--my_variable,rgb(0_2_3))"></div>
```

Which already parses properly as:
```css
.bg-\(--my_variable\,rgb\(0_2_3\)\) {
  background-color: var(--my_variable,rgb(0 2 3));
}
```

But looking at the reproduction, I do feel like there is a bug with the modifier.
```html
<div class="bg-(--my_variable,rgb(0_2_3))/(--my_opacity,some_fallback_value)"></div>
```
Compiles to:
```css
.bg-\(--my_variable\,rgb\(0_2_3\)\)\/\(--my_opacity\,some_fallback_value\) {
  background-color: color-mix(in oklab, var(--my_variable,rgb(0 2 3)) var(--my opacity,some fallback value), transparent);
}
```

So the spaces int he fallback should be kept, but not in the actual variable (first argument).

packages/tailwindcss/src/candidate.ts
```diff
@@ -523,21 +523,32 @@ function parseModifier(modifier: string): CandidateModifier | null {
...
-    if (!isValidArbitrary(arbitraryValue)) return null
+    // A modifier with `(…)` should always start with `--` since it
+    // represents a CSS variable.
+    if (modifier[0] !== '-' || modifier[1] !== '-') return null
 
-    // Empty arbitrary values are invalid. E.g.: `data-():`
-    //                                                 ^^
-    if (arbitraryValue.length === 0 || arbitraryValue.trim().length === 0) return null
+    // Trim the modifier to remove any leading or trailing whitespace
+    modifier = modifier.trim()
```
thecrypticace, line 534: Can this actually happen before decoding the value?

RobinMalfait: You know, that is a good point actually!

thecrypticace (approved):

brandonmcconnell:
I appreciate the quick turnaround on this, @RobinMalfait and @thecrypticace. Thanks!

PR #17898, opened 2025-05-06
Make upgrades faster

RobinMalfait:
This PR fixes an issue where the upgrade tests were taking too long. This PR fixes that.

Essentially when updating dependencies we did this:
```console
npm install tailwindcss@latest
npm install @tailwindcss/postcss@latest
npm install prettier-plugin-tailwindcss@latest
```

But this is not ideal, because we are calling out to `npm` and run each dependency in isolation. 

With this PR, we essentially do it all in one go:
```console
npm install tailwindcss@latest @tailwindcss/postcss@latest prettier-plugin-tailwindcss@latest
```

## Test plan

Testing this locally, the results look like this:

| Before | After |
|--------|-------|
| image | image |

In CI:

| Before (with a failure) | After |
| --- | --- |
| image | image |

[ci-all]

### DIFF
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -231,29 +231,18 @@ async function run() {
 
     info('Updating dependencies…')
     {
-      let dependencies = (
-        await Promise.all(
-          [
-            'tailwindcss',
-            '@tailwindcss/cli',
-            '@tailwindcss/postcss',
-            '@tailwindcss/vite',
-            '@tailwindcss/node',
-            '@tailwindcss/oxide',
-            'prettier-plugin-tailwindcss',
-          ].map(async (dependency) => {
-            if (dependency === 'tailwindcss') {
-              return dependency
-            } else if (await pkg(base).has(dependency)) {
-              return dependency
-            }
-            return []
-          }),
-        )
-      ).flat()
-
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -230,18 +230,31 @@ async function run() {
...
+            '@tailwindcss/cli',
+            '@tailwindcss/postcss',
+            '@tailwindcss/vite',
+            '@tailwindcss/node',
+            '@tailwindcss/oxide',
+            'prettier-plugin-tailwindcss',
+          ].map(async (dependency) => {
+            if (dependency === 'tailwindcss') {
+              return dependency
+            } else if (await pkg(base).has(dependency)) {
```
thecrypticace, line 247: could you hoist `await pkg(base)` outside of this call? Then you don't need the promise.all and things get simpler again?

RobinMalfait: Only if we change how `pkg` works. Right now `pkg(base)` itself just returns an object with `async add()`, `async has()` and `async remove()` methods.

RobinMalfait: We could also make it sync

thecrypticace (approved):

PR #17925, opened 2025-05-08
Fix bug with nested @apply rules in utility classes (#17924)

### LINKED ISSUE #17924: @apply referencing another @utility breaks @apply in a nested rule

v4.1.5

https://play.tailwindcss.com/

Not relevant

Edge Latest Version

MacOs

**Reproduction URL**

https://play.tailwindcss.com/oBufuCc0yf

**Describe your issue**

@apply referencing another @utility breaks @apply in a nested rule. 
Problem is that :hover and :disabled inside `base-utility` stopped working when i referenced `base-utility` in a `@apply` of `parent-utility`.

It would be understandable for the styles of `base-utility` to not get applied completely if it were `@apply parent-utility` but the problem is that I'm using `base-utility` in the className and it should not change regardless of where `base-utility` is used.

:hover and :disabled are simple pseudo selectors. Tried with @variant, also does not work for the same reason.

Would not be a problem for me if @variant actually does work. If the @apply is NOT nested inside `base-utility`, it works fine. The part where its not working is only the @apply is inside a nested rule of `base-utility`.

philipp-spiess:
Fixes #17924

When an `@apply` pointed to utility that nested usages of `@apply`, these nested usages were not properly carried through the dependency chain. This was because we were only tracking dependencies on the immediate parent rather than all parents.

To fix this, this PR:

- Modifies the dependency resolution to track dependencies through the entire parent path
- Uses a `walk(…)` for the node replacement logic so that all nested `@apply` usages are also resolved (as these are now tracked in the dependency list anyways

## Test Plan

- Added a regression test for #17924 to the unit tests and ensure existing tests don't break

### DIFF
--- a/packages/tailwindcss/src/apply.ts
+++ b/packages/tailwindcss/src/apply.ts
@@ -23,7 +23,7 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
   let definitions = new DefaultMap(() => new Set<AstNode>())
 
   // Collect all new `@utility` definitions and all `@apply` rules first
-  walk([root], (node, { parent }) => {
+  walk([root], (node, { parent, path }) => {
     if (node.kind !== 'at-rule') return
 
     // Do not allow `@apply` rules inside `@keyframes` rules.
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/apply.ts
```diff
@@ -66,7 +66,12 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
       parents.add(parent)
 
       for (let dependency of resolveApplyDependencies(node, designSystem)) {
-        dependencies.get(parent).add(dependency)
+        // Mark every parent in the path as having a dependency to that utility.
+        for (let parent of path) {
+          if (parent === node) continue
+          if (!parents.has(parent)) continue
+          dependencies.get(parent).add(dependency)
+        }
```
RobinMalfait, line 74: Ah interesting, I knew we were performing a topological sort already so was confused why this _didn't_ work.

But this makes sense, and tracking the full path also makes sense.

RobinMalfait (approved):

PR #17929, opened 2025-05-08
Download platform specific package if `optionalDependencies` are skipped

### LINKED ISSUE #15806: Can't build on Linux ARM -> cannot find module '@tailwindcss/oxide-linux-arm64-gnu'

4.0.0

Vite 6

22.12.0

Occurs in CLI

Ubuntu Linux on arm64

**Reproduction URL**

/

**Describe your issue**

When running `tsc && vite build` the build fails and gives the following error:

```
error during build:
[vite:css] Failed to load PostCSS config (searchPath: [...]): [Error] Loading PostCSS Plugin failed: Cannot find module '@tailwindcss/oxide-linux-arm64-gnu'

```

philipp-spiess:
Closes #15806

This PR adds a new `postinstall` script to `@tailwindcss/oxide` that will attempt to download the platform-specific optional dependency to avoid issues when the package manager does not do that automatically (see #15806). The implementation for this is fairly simple: The npm package is downloaded from the official npm servers and extracted into the `node_modules` directory of the `@tailwidncss/oxide` installation.

## Test plan 

Since we still lack a solid repro of #15806, the way I tested this change was to manually remove all automatically-installed optional dependencies and then running the postinstall script manually. The script then downloads the right version package which makes the import to `@tailwidncss/oxide` work. An appropriate integration test was added too. 

I furthermore also validated that:

- This works across CI platforms [ci-all]
- The postinstall script bails out when running `pnpm install` in the dev setup. This is necessary since doing the initial install won't have any binary dependencies yet so it would download invalid versions from npm (as the version numbers locally refer to the last released version). We can safely bail out here though since this was never an issue with local development.
- The postinstall script does not do anything when the `@tailwindcss/oxide` library is added _unless_ the issue is detected.

[ci-all]

### DIFF
--- a/crates/node/scripts/install.js
+++ b/crates/node/scripts/install.js
@@ -1,0 +1,134 @@
+#!/usr/bin/env node
+
+/**
+ * @tailwindcss/oxide postinstall script
+ *
+ * This script ensures that the correct binary for the current platform and
+ * architecture is downloaded and available.
+ */
+
+const fs = require('fs')
+const path = require('path')
+const https = require('https')
+const { extract } = require('tar')
+const packageJson = require('../package.json')
+const detectLibc = require('detect-libc')
+
+const version = packageJson.version
+
+function getPlatformPackageName() {
+  const platform = process.platform
+  const arch = process.arch
+
+  let libc = ''
+  if (platform === 'linux') {
+    libc = detectLibc.isNonGlibcLinuxSync() ? 'musl' : 'gnu'
+  }
+
+  // Map to our package naming conventions
+  switch (platform) {
+    case 'darwin':
+      return arch === 'arm64' ? '@tailwindcss/oxide-darwin-arm64' : '@tailwindcss/oxide-darwin-x64'
+    case 'win32':
+      if (arch === 'arm64') return '@tailwindcss/oxide-win32-arm64-msvc'
+      if (arch === 'ia32') return '@tailwindcss/oxide-win32-ia32-msvc'
+      return '@tailwindcss/oxide-win32-x64-msvc'
+    case 'linux':
+      if (arch === 'x64') {
+        return libc === 'musl'
+          ? '@tailwindcss/oxide-linux-x64-musl'
+          : '@tailwindcss/oxide-linux-x64-gnu'
+      } else if (arch === 'arm64') {
+        return libc === 'musl'
+          ? '@tailwindcss/oxide-linux-arm64-musl'
+          : '@tailwindcss/oxide-linux-arm64-gnu'
+      } else if (arch === 'arm') {
+        return '@tailwindcss/oxide-linux-arm-gnueabihf'
+      }
+      break
+    case 'freebsd':
+      return '@tailwindcss/oxide-freebsd-x64'
+
+    case 'android':
+      return '@tailwindcss/oxide-android-arm64'
+    default:
+      return '@tailwindcss/oxide-wasm32-wasi'
+  }
+}
+
+function isPackageAvailable(packageName) {
+  try {
+    require.resolve(packageName)
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
+// Extract all files from a tarball to a destination directory
+async function extractTarball(tarballStream, destDir) {
+  if (!fs.existsSync(destDir)) {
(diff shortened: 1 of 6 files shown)


integrations/oxide/postinstall.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    expect(await exec('node test.js', opts)).toMatch(/SUCCESS/)
+  },
+)
+
+async function removePlatformSpecificExtensions(directory: string) {
+  const entries = await fs.readdir(directory, { withFileTypes: true })
+
+  for (const entry of entries) {
+    const fullPath = path.join(directory, entry.name)
```
RobinMalfait, line 59:
```suggestion
  let entries = await fs.readdir(directory, { withFileTypes: true })

  for (let entry of entries) {
    let fullPath = path.join(directory, entry.name)
```

crates/node/scripts/install.js
```diff
@@ -0,0 +1,139 @@
...
+        resolve()
+      })
+  })
+}
+
+async function main() {
+  // Don't run this script in the package source
+  if (fs.existsSync(path.join(__dirname, '..', 'build.rs'))) {
+    return
+  }
```
RobinMalfait, line 130: I think we should wrap this in a try/catch. Some people run stuff from Docker in the `/` path, so `../build.rs` will very likely crash.

Also let's hope nobody has a `build.rs` file in the parent directory 😅 

```suggestion
  // Don't run this script in the package source
  try {
    if (fs.existsSync(path.join(__dirname, '..', 'build.rs'))) {
      return
    }
  } catch {
    return
  }
```

crates/node/scripts/install.js
```diff
@@ -0,0 +1,139 @@
...
+
+  const packageName = getPlatformPackageName()
```
RobinMalfait, line 132:
```suggestion
  let packageName = getPlatformPackageName()
```

crates/node/scripts/install.js
```diff
@@ -0,0 +1,139 @@
...
+const https = require('https')
+const { extract } = require('tar')
+const packageJson = require('../package.json')
+const detectLibc = require('detect-libc')
+
+const version = packageJson.version
+
+function getPlatformPackageName() {
+  const platform = process.platform
+  const arch = process.arch
```
RobinMalfait, line 21:
```suggestion
  let platform = process.platform
  let arch = process.arch
```

crates/node/scripts/install.js
```diff
@@ -0,0 +1,139 @@
...
+const fs = require('fs')
+const path = require('path')
+const https = require('https')
+const { extract } = require('tar')
+const packageJson = require('../package.json')
+const detectLibc = require('detect-libc')
+
+const version = packageJson.version
+
+function getPlatformPackageName() {
```
RobinMalfait, line 19: I wonder if we can somehow re-use this logic from what NAPI generated for us. Since that is generated and using require already it might not be as easy.

Just have to make sure that _if_ we make changes, that we keep this in sync

crates/node/scripts/install.js
philipp-spiess: wait but the script is run within the npm package in the `scripts/` folder so the parent is always there and should not be on the root or am I misunderstanding something?

crates/node/scripts/install.js
```diff
@@ -0,0 +1,139 @@
...
+        return libc === 'musl'
+          ? '@tailwindcss/oxide-linux-arm64-musl'
+          : '@tailwindcss/oxide-linux-arm64-gnu'
+      } else if (arch === 'arm') {
+        return '@tailwindcss/oxide-linux-arm-gnueabihf'
+      }
+      break
+    case 'freebsd':
+      return '@tailwindcss/oxide-freebsd-x64'
+
```
philipp-spiess, line 51:
```suggestion
```

crates/node/scripts/install.js
philipp-spiess: still doesn't hurt to wrap in try catch I assume

RobinMalfait: Ah right, executed from the `scripts` folder, I interpreted it as root of the project. Makes sense!

RobinMalfait (approved):

PR #17941, opened 2025-05-08
Update template-extensions.txt

iliakan:
Fixes #17940


philipp-spiess:
Hey! Thanks for the PR but as we've mentioned in your issue https://github.com/tailwindlabs/tailwindcss/issues/17940, this file is not an allow list. So your `.ejs` files should already be scanned regardless (we scan all unknown file extensions). This list is used for another reason and while it _might_ make sense to add `.ejs` here, I really think we should figure out why your use case doesn't work first. So going to close this for now and we'll await a repro on the issue! Thanks!

PR #17969, opened 2025-05-10
Add ignore pattern for node_modules in globby search in upgrade cli

remorses:
When passing `gitignore: true` to globby it will start a search for all .gitignore files, this initial search includes node_modules making it hang forever for large monorepos with many files inside node_modules


RobinMalfait (approved):
This makes sense, thank you!

PR #18000, opened 2025-05-13
Migrate bare values to named values

RobinMalfait:
This PR improves the upgrade tool by also migrating bare values to named values defined in the `@theme`. It's similar to the existing migration we already had, but specialized for just bare values. This makes it a bit faster as well because we don't have to try and replace a bare value with a bare value.

Recently we shipped some updates dat allowed us to migrate arbitrary values (with square brackets), but we didn't migrate bare values yet.

That means that in this example:
```html
<div class="aspect-[16/9]"></div>
<div class="aspect-16/9"></div>
```

We migrated this to:
```html
<div class="aspect-video"></div>
<div class="aspect-16/9"></div>
```

With this change, we will also try and migrate the bare value to a named value. So this example:
```html
<div class="aspect-[16/9]"></div>
<div class="aspect-16/9"></div>
```

Now becomes:
```html
<div class="aspect-video"></div>
<div class="aspect-video"></div>
```

## Test plan

1. Added unit tests for the new functionality.
2. Ran this on a local project

Before:
image

After:
image

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate-bare-utilities.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate-bare-utilities.ts
@@ -1,0 +1,130 @@
+import { type Candidate } from '../../../../tailwindcss/src/candidate'
+import type { Config } from '../../../../tailwindcss/src/compat/plugin-api'
+import type { DesignSystem } from '../../../../tailwindcss/src/design-system'
+import { DefaultMap } from '../../../../tailwindcss/src/utils/default-map'
+import type { Writable } from '../../utils/types'
+import { baseCandidate } from './candidates'
+import { computeUtilitySignature, preComputedUtilities } from './signatures'
+
+const baseReplacementsCache = new DefaultMap<DesignSystem, Map<string, Candidate>>(
+  () => new Map<string, Candidate>(),
+)
+
+export function migrateBareValueUtilities(
+  designSystem: DesignSystem,
+  _userConfig: Config | null,
+  rawCandidate: string,
+): string {
+  let utilities = preComputedUtilities.get(designSystem)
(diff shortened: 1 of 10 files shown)


packages/@tailwindcss-upgrade/src/codemods/template/migrate-bare-utilities.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+function parseCandidate(designSystem: DesignSystem, input: string) {
+  return designSystem.parseCandidate(
+    designSystem.theme.prefix && !input.startsWith(`${designSystem.theme.prefix}:`)
+      ? `${designSystem.theme.prefix}:${input}`
+      : input,
+  )
+}
```
thecrypticace, line 130: Don't we have exactly this implementation of parseCandidate in some other migration? I feel like I've seen this before.

RobinMalfait: Ah yeah, I copied this whole file and deleted what I didn't need. Can move this to the `candidates.ts` file but should we _just_ call it `parseCandidate` or something like prefixAwareParseCandidate`

thecrypticace: Just `parseCandidate` is fine for now. Can improve the name if we feel it's necessary later.

thecrypticace (approved):

sumansuhag:
Hi,

 If 16/9 is defined as video in the theme's aspect ratio scale:
class="aspect-[16/9]" becomes class="aspect-video" (as before)
class="aspect-16/9" now also becomes class="aspect-video" (new behavior)

Good luck!,
SUMAN SUHAG

PR #18020, opened 2025-05-14
Ignore custom variants with `:merge(…)` selectors

### LINKED ISSUE #15617: [V4] :merge selector broken

v4.0.0-beta.9

Next.js 15.1.4

v22.13.0

Chrome

macOS

**Reproduction URL**

https://github.com/mhornbacher/tailwind-repro

**Describe your issue**

At a high level:
1. The following issue is triggered when adding any `peer-` class with hot reloading
2. All `peer-` classes are always active after a build

```
Parsing css source code failed
  530 |   }
  531 |   .peer-disabled\:text-purple-500 {
> 532 |     :merge(.peer):where([data-rac])[data-disabled] ~ & {
      |             ^
  533 |       color: var(--color-purple-500);
  534 |     }
  535 |     :merge(.peer):where(:not([data-rac])):disabled ~ & {

'merge' is not recognized as a valid pseudo-class. Did you mean '::merge' (pseudo-element) or is this a typo? at [project]/src/app/globals.css:531:12
```

The cause for this, as far as I can tell is as follows:

In the [documentation for plugins](https://tailwindcss.com/docs/plugins#parent-and-sibling-states) there is a special `:merge` selector designed for plugins (like react-aria-components in this example) to extend the `focus`, `disabled`... selectors, in this case to enable them to be triggered in more complex accessibility scenarios. (see the [react-aria implementation](https://github.com/adobe/react-spectrum/blob/main/packages/tailwindcss-react-aria-components/src/index.js#L121-L130) for example.

However in Tailwind 4.0 it seems that this selector is not handled by the new lighting CSS compiler. I know its not that common, but many projects rely on this to simplify accessibility development for their team. Which is something I know that the tailwind team cares about as well :)

philipp-spiess:
Closes #15617

## Summary

This PR ignores `addVariant(…)` legacy JS plugin calls for variants that are using the [`:merge(…)` selector](https://v3.tailwindcss.com/docs/plugins#parent-and-sibling-states) for parent and sibling states. We can ignore these now because in v4, `group-*` and `peer-*` variants _compound automatically_ and you don't have to define them anymore.

## Test plan

Added a unit test to ensure that the `optional` variant example from the v3 docs work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure that running the Standalone build does not leave temporary files behind ([#17981](https://github.com/tailwindlabs/tailwindcss/pull/17981))
 - Fix `-rotate-*` utilities with arbitrary values ([#18014](https://github.com/tailwindlabs/tailwindcss/pull/18014))
 - Upgrade: Change casing of utilities with named values to kebab-case to match updated theme variables ([#18017](https://github.com/tailwindlabs/tailwindcss/pull/18017))
-- Ignore custom variants using `:merge(…)` selectors in legacy JS plugins as they would create invalid CSS in v4 ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
+- Ignore custom variants using `:merge(…)` selectors in legacy JS plugins ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure that running the Standalone build does not leave temporary files behind ([#17981](https://github.com/tailwindlabs/tailwindcss/pull/17981))
 - Fix `-rotate-*` utilities with arbitrary values ([#18014](https://github.com/tailwindlabs/tailwindcss/pull/18014))
 - Upgrade: Change casing of utilities with named values to kebab-case to match updated theme variables ([#18017](https://github.com/tailwindlabs/tailwindcss/pull/18017))
+- Ignore custom variants using `:merge(…)` selectors in legacy JS plugins as they would create invalid CSS in v4 ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/compat/plugin-api.ts
```diff
@@ -115,6 +115,16 @@ export function buildPluginApi({
...
 
+      // Ignore variants emitting v3 `:merge(…)` rules. In v4, the `group-*` and `peer-*` variants
+      // compound automatically.
+      if (typeof variant === 'string') {
+        if (variant.includes(':merge(')) return
+      } else if (Array.isArray(variant)) {
+        if (variant.some((v) => v.includes(':merge('))) return
+      } else if (typeof variant === 'object') {
+        if (Object.keys(variant).some((v) => v.includes(':merge('))) return
+      }
```
thecrypticace, line 126: 1. We should detect this in matchVariant as well.
2. Should the object check happen recursively or no? I don't remember what v3 did there

philipp-spiess: 1. Ah yep forgot about `matchVariant`.
2. Hmm yeah probably you're right.

philipp-spiess: @thecrypticace The `matchVariant(…)` implementation is a bit annoying since we have to call the matcher function to sample the output. I tried throwing the variant out later but once we register it with the design system, the `peer-*` and `group-*` compounding doesn't work anymore. 

The only other idea I have is to check for wether the variant starts with `peer-` or `group-`. What do you think?

thecrypticace (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure that running the Standalone build does not leave temporary files behind ([#17981](https://github.com/tailwindlabs/tailwindcss/pull/17981))
 - Fix `-rotate-*` utilities with arbitrary values ([#18014](https://github.com/tailwindlabs/tailwindcss/pull/18014))
 - Upgrade: Change casing of utilities with named values to kebab-case to match updated theme variables ([#18017](https://github.com/tailwindlabs/tailwindcss/pull/18017))
+- Ignore custom variants using `:merge(…)` selectors in legacy JS plugins as they would create invalid CSS in v4 ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
```
philipp-spiess, line 20:
```suggestion
- Ignore custom variants using `:merge(…)` selectors in legacy JS plugins ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
```

PR #18025, opened 2025-05-14
Make candidate template migrations faster

### LINKED ISSUE #17974: @tailwindcss/upgrade - brakes boolean operators in Vue files

Upgrading to tailwindcss v4.1.6

Vite 5.2.3

For example: v22.14.0

macOS

**Describe your issue**

Running `npx @tailwindcss/upgrade@latest` broke logical operators in Vue files.
Examples:

Image
<br>
Image

---
Maybe related to 
https://github.com/tailwindlabs/tailwindcss/pull/14737
https://github.com/tailwindlabs/tailwindcss/pull/14774

RobinMalfait:
This PR makes the migrations for templates much faster. To make this work, I also had to move things around a bit (so you might want to check this PR commit by commit). I also solved an issue by restructuring the code.

### Performance

For starters, we barely applied any caching when migrating candidates from α to β. The problem with this is that in big projects the same candidates will appear _everywhere_, so caching is going to be useful here.

One of the reasons why we didn't do any caching is that some migrations were checking if a migration is actually safe to do. To do this, we were checking the `location` (the location of the candidate in the template). Since this location is unique for each template, caching was not possible.

So the first order of business was to hoist the `isSafeMigration` check up as the very first thing we do in the migration.

If we do this first, then the only remaining code relies on the `DesignSystem`, `UserConfig` and `rawCandidate`. 

In a project, the `DesignSystem` and `UserConfig` will be the same during the migration, only the `rawCandidate` will be different which means that we can move all this logic in a good old `DefaultMap` and cache the heck out of it. 

Running the numbers on our Tailwind Plus repo, this results in:
```
    Total seen candidates: 2 211 844
Total migrated candidates: 7 775 
               Cache hits: 1 575 700
```

That's a lot of work we _don't_ have to do. Looking at the timings, the template migration step goes from ~45s to ~10s because of this.

Another big benefit of this is that this makes migrations _actually_ safe. Before we were checking if a migration was safe to do in specific migrations. But other migrations were still printing the candidate which could still result in an unsafe migration.

For example when migrating the `blur` and the `shadow` classes, the `isSafeMigration` was used. But if the input was `!flex` then the safety check wasn't even checked in this specific migration.

### Safe migrations

Also made some changes to the `isSafeMigration` logic itself. We used to start by checking the location, but thinking about the problem again, the actual big problem we were running into is classes that are short like `blur`, and `shadow` because they could be used in other contexts than a Tailwind CSS class.

Inverting this logic means that more specific Tailwind CSS classes will very likely _not_ cause any issues at all. 

For example:
- If you have variants: `hover:focus:flex`
- If you have arbitrary properties: `[color:red]`
- If you have arbitrary values: `bg-[red]`
- If you have a modifier: `bg-red-500/50`
- If you have a `-` in the name: `bg-red-500`

Even better if we can't parse a candidate at all, we can skip the migrations all together.

This brings us to the issue in #17974, one of the issues was already solved by just hoisting the `isSafeMigration`. But to make the issue was completely solved I also made sure that in Vue attributes like `:active="…"` are also considered unsafe (note: `:class` is allowed).

Last but not least, in case of the `!duration` that got replaced with `duration!` was solved by verifying that the candidate actually produces valid CSS. We can compute the signature for this class.

The reason this wasn't thrown away earlier is because we can correctly parse `duration` but `duration` on its own doesn't exist, `duration-<number>` does exist as a functional utility which is why it parsed in the first place.

Fixes: #17974

## Test plan

1. Ran the tool on our Tailwind UI Templates repo to compare the new output with the "old" behavior and there were no differences in output.
2. Ran the tool on our Tailwind Plus repo, and the template migration step went from ~45s to ~10s.
3. Added additional tests to verify the issues in #17974 are fixed.

[ci-all] let's run this on all CI platforms...


packages/@tailwindcss-upgrade/src/codemods/template/migrate-legacy-classes.ts
```diff
@@ -111,42 +105,14 @@ export async function migrateLegacyClasses(
...
-        return fromCandidate.value === null || !fromCandidate.root.includes('-')
-      }
-
-      return false
-    })()
-
-    if (location && isPotentialProblematicClass && !isSafeMigration(location)) {
-      continue
-    }
-
```
RobinMalfait, line 149: This is now handled ahead of time before attempting to even migrate

packages/@tailwindcss-upgrade/src/codemods/template/migrate-legacy-classes.ts
```diff
@@ -72,11 +71,6 @@ export async function migrateLegacyClasses(
   designSystem: DesignSystem,
   _userConfig: Config | null,
   rawCandidate: string,
-  location?: {
-    contents: string
-    start: number
-    end: number
-  },
```
RobinMalfait, line 79: The main `migrate` function will still receive the location, but each individual migration will not

philipp-spiess (approved):
Nice! I think this makes sense. The only thought I had while going through it commit-by-commit was the fact that we know some utilities from v3 wouldn't work in v4 ootb (we do extend the design system somewhere for that IIRC) but it seems like you addressed this by opting out of that check if the migration is 3 => 4. 👍

PR #18038, opened 2025-05-15
Fix missing extracted classes containing `.` in Clojure

### LINKED ISSUE #17760: Candidate extraction from Clojure/Script files sometimes produce false negatives

v4.1.4

postcss-cli 11.0.1

v23.11.0

Chrome, Safari, Firefox

macOS

**Reproduction URL**

https://github.com/tailwindlabs/tailwindcss/pull/17087

Image

**Describe your issue**

First of all, a huge thank you for including Clojure.

The issue stems from classes written as keywords that contain fractions.  For example,

```clojure
($ :div {:class [:flex :gap-1.5 :p-1]} …)
```

In this case, the candidates extracted from `:gap-1.5` seem to be `:gap-1` and `5`, which will end up not matching any class present in the compiled HTML: `<div class="flex gap-1.5 p-1">…</div>`.

A suggested "good enough" solution is to, in addition to considering the keyword when broken down by period (`.`) separation, additionally consider the whole string as a candidate. I.e. such that it goes from considering the candidates,

```clojure
["gap-1" "5"]
```

To

```clojure
["gap-1" "5" "gap-1.5"]
```

This may end up producing some valid classes which are not actually present in the code (`gap-1` in this case), but is still preferable, as it trades precision for accuracy.

Another, more elaborate solution would be to join candidates pair-wise and accept them iff they produce a valid class. Eg., the candidates `["gap-1" "5" "p-1"]` are extracted. When the first pair is joined with `.`, they produce the candidate `gap-1.5`, which is therefore accepted in place of the previous two candidates, producing `["gap-1.5" "p-1"]`. In the next iteration, the join produces `["gap-1.5.p-1"]`, which is nonsense, and is therefore rejected.

More discussion can be found towards the end of https://github.com/tailwindlabs/tailwindcss/pull/17087.

RobinMalfait:
This PR fixes an issue in the Clojure pre-processor where candidates including `.` characters were not extracted correctly.

The solution here is to only replace the `.` with a ` ` when the `.` is not surrounded by numbers. This means that:

```
:.foo.bar
```
Becomes
```
: foo bar
```

But
```
:.gap-1.5.flex
```

Becomes
```
: gap-1.5 flex
```

This way the `gap-1.5` is correctly extracted.

## Test plan

1. Added a test for this case
2. Tested this in the extractor tool as well. Notice how the `gap-1.5` is correctly extracted here.

image

Fixes: #17760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade: Change casing of utilities with named values to kebab-case to match updated theme variables ([#18017](https://github.com/tailwindlabs/tailwindcss/pull/18017))
 - Upgrade: Fix unsafe migrations in Vue files ([#18025](https://github.com/tailwindlabs/tailwindcss/pull/18025))
 - Ignore custom variants using `:merge(…)` selectors in legacy JS plugins ([#18020](https://github.com/tailwindlabs/tailwindcss/pull/18020))
+- Fix missing extracted classes containing `.` in Clojure ([#18038](https://github.com/tailwindlabs/tailwindcss/pull/18038))
 
 ### Added
 
--- a/crates/oxide/src/extractor/pre_processors/clojure.rs
+++ b/crates/oxide/src/extractor/pre_processors/clojure.rs
@@ -42,6 +42,18 @@ impl PreProcessor for Clojure {
                     }
                 }
 
+                // A `.`  surrounded by digits is a decimal number, so we don't want to replace it.
+                //
+                // E.g.:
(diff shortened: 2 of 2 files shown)


philipp-spiess (approved):

eneroth:
Thanks @RobinMalfait, can confirm that these things work well now! 

I ran into another issue with pseudo classes though. Eg. `:first:lg:pl-8` doesn't seem to be picked up. I'll open an [issue](https://github.com/tailwindlabs/tailwindcss/issues/18336) for it.

By the way, the debugger that you're sharing screenshots from: is it accessible publicly? It would be great to try verify things like this before reporting them.

Issue #15115, opened 2024-11-22
[v4] --radius-full is not available as a variable


BeetrootShoulders:
**What version of Tailwind CSS are you using?**

4 beta.2

**What build tool (or framework if it abstracts the build tool) are you using?**

Tailwind Play

**What version of Node.js are you using?**

N/A

**What browser are you using?**

Firefox

**What operating system are you using?**

macOS

**Reproduction URL**

https://play.tailwindcss.com/bpvuR7VY5Z

**Describe your issue**

`--radius-full` does not appear to be available as a variable. As an obviously pointless example`rounded-(--radius-lg)` will result in an element with rounded corners, `rounded-(--radius-full)` will not.

philipp-spiess:
@BeetrootShoulders Out of curiosity: Do you mind sharing the more specific case you were running into that lead you to notice this variable not being defined? 🙂

tobimori:
I've noticed this as well, since I've tried to avoid `@apply` more and more and result to `theme()` or now `var()`. I agree there's no need in theory, but it's just consistency when writing CSS without utilities.

Related: https://discord.com/channels/486935104384532500/486935104384532502/1299322215094816870

BeetrootShoulders:
> @BeetrootShoulders Out of curiosity: Do you mind sharing the more specific case you were running into that lead you to notice this variable not being defined? 🙂

I actually opened the issue on behalf of a user on the TW Discord (where I try to help out as best I can): https://discord.com/channels/486935104384532500/486935104384532502/1309559484255047721

t0byman:
I would also like to use `var(--radius-full)` instead of `@apply rounded-full`

sumansuhag:
HI,
// tailwind.config.js (or default theme)
module.exports = {
  theme: {
    borderRadius: {
      'none': '0px',
      'sm': '0.125rem', // 2px
      DEFAULT: '0.25rem', // 4px (for `rounded`)
      'md': '0.375rem', // 6px
      'lg': '0.5rem',   // 8px
      'xl': '0.75rem',  // 12px
      '2xl': '1rem',   // 16px
      '3xl': '1.5rem',  // 24px
      'full': '9999px', // This is the key in question
    },
    // ...
  },
  // ...
};
[...]

PR #18056, opened 2025-05-16
Proposal: Allow overwriting static utilities that have a namespace

### LINKED ISSUE #16639: [v4] transition-property Theme CSS Variables

v4.0.6

Next.js: 15.1.3

v20.18.1

Safari

macOS

**Reproduction URL**

https://play.tailwindcss.com/pNcjYumxXW

**Describe your issue**

***Overview***

I am trying to migrate a project to v4, which has custom transition properties defined in the config.

Previously:
```jsx
// tailwind.config.ts
theme: { extend: { transitionProperty: { ring: 'box-shadow' } } }

// component.tsx
<input className="transition-[theme(transitionProperty.colors),theme(transitionProperty.ring)]" />
```

Expected migration:
```jsx
// globals.css
@theme {
  --transition-property-ring: box-shadow;
}

// component.tsx
<input className="transition-[theme(--transition-property-colors),theme(transition-property-ring)]" />
```

***Specifics***

The included transition properties are not included as css variables (i.e. `theme(--transition-property-colors)` does not work), but it does work with the old syntax, i.e. `theme(transitionProperty.colors)`.

Similar to the previous report in #14479, using the default transition properties (i.e. transition-colors) does not output css variables.

Also, for example, `--transition-property-ring: box-shadow` does work in the `@theme` declaration but it is not documented (afaict). If you would like this raised as a separate documentation issue I can do that, but I think that it may change based on how the default transition properties are implemented (i.e. should they also be css variables in the documentation?)

philipp-spiess:
This PR attempts to move static utilities that are overwriteable by a theme value to be a fallback rather than a conflicting implementation. The idea is to allow a theme value to take presedence over that static utility _and cause it not to generate_.

For example, when overwriting the `--radius-full` variant, it should ensure that the default `rounded-full` no longer emits the `calc(infinity * 1px)` declaration:

```ts
expect(
  await compileCss(
    css`
      @theme {
        --radius-full: 99999px;
      }
      @tailwind utilities;
    `,
    ['rounded-full'],
  ),
).toMatchInlineSnapshot(`
  ":root, :host {
    --radius-full: 99999px;
  }

  .rounded-full {
    border-radius: var(--radius-full);
  }"
`)
```

This allows anyone who wants `--radius-full` to be a CSS variable to simply define it in their theme:

```css
@theme {
  /* Make `--radius-full` a CSS variable without the utility generating two CSS classes */
  --radius-full: calc(infinity * 1px);
}
```

The idea is to extend this pattern across all functional utilities that also have static utilities that can collide with the namespace. This gives users more control over what they want as CSS variables when the defaults don't work for them, allowing them to resolve #16639 and #15115 in user space.

You may now find yourself thinking "but Philipp, why would someone want to be able to overwrite `--animate-none`. `none` surely always will mean no animation" and I would agree [but it's already possible right now anyways so this is not a new behavior! This PR just cleans up the generated output.](https://play.tailwindcss.com/StnQqm4V2e)

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -1334,7 +1334,6 @@ test('col-start', async () => {
     }
 
     .col-start-auto {
-      grid-column: var(--grid-column-start-auto);
       grid-column-start: var(--grid-column-start-auto);
     }"
   `)
@@ -1409,7 +1408,6 @@ test('col-end', async () => {
     }
 
     .col-end-auto {
-      grid-column: var(--grid-column-end-auto);
       grid-column-end: var(--grid-column-end-auto);
     }"
   `)
@@ -1573,7 +1571,6 @@ test('row-start', async () => {
     }
 
     .row-start-auto {
-      grid-row: var(--grid-row-start-auto);
       grid-row-start: var(--grid-row-start-auto);
     }"
   `)
@@ -1648,7 +1645,6 @@ test('row-end', async () => {
     }
 
     .row-end-auto {
-      grid-row: var(--grid-row-end-auto);
       grid-row-end: var(--grid-row-end-auto);
     }"
   `)
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -9771,6 +9771,25 @@ test('rounded', async () => {
     }"
   `)
   expect(
+    await compileCss(
+      css`
+        @theme {
+          --radius-full: 99999px;
+        }
+        @tailwind utilities;
+      `,
+      ['rounded-full'],
+    ),
+  ).toMatchInlineSnapshot(`
+    ":root, :host {
+      --radius-full: 99999px;
+    }
+
+    .rounded-full {
+      border-radius: var(--radius-full);
+    }"
+  `)
+  expect(
     await run([
       '-rounded',
       '-rounded-full',
@@ -9945,15 +9964,11 @@ test('rounded-t', async () => {
     }
 
     .rounded-t-full {
-      border-top-left-radius: 3.40282e38px;
-      border-top-right-radius: 3.40282e38px;
       border-top-left-radius: var(--radius-full);
       border-top-right-radius: var(--radius-full);
     }
 
     .rounded-t-none {
-      border-top-left-radius: 0;
-      border-top-right-radius: 0;
       border-top-left-radius: var(--radius-none);
       border-top-right-radius: var(--radius-none);
     }
@@ -10012,15 +10027,11 @@ test('rounded-r', async () => {
     }
 
     .rounded-r-full {
-      border-top-right-radius: 3.40282e38px;
-      border-bottom-right-radius: 3.40282e38px;
       border-top-right-radius: var(--radius-full);
       border-bottom-right-radius: var(--radius-full);
     }
 
     .rounded-r-none {
-      border-top-right-radius: 0;
-      border-bottom-right-radius: 0;
       border-top-right-radius: var(--radius-none);
       border-bottom-right-radius: var(--radius-none);
     }
@@ -10079,15 +10090,11 @@ test('rounded-b', async () => {
     }
 
     .rounded-b-full {
-      border-bottom-right-radius: 3.40282e38px;
-      border-bottom-left-radius: 3.40282e38px;
       border-bottom-right-radius: var(--radius-full);
       border-bottom-left-radius: var(--radius-full);
     }
 
     .rounded-b-none {
-      border-bottom-right-radius: 0;
-      border-bottom-left-radius: 0;
       border-bottom-right-radius: var(--radius-none);
       border-bottom-left-radius: var(--radius-none);
     }
@@ -10146,15 +10153,11 @@ test('rounded-l', async () => {
     }
 
     .rounded-l-full {
-      border-top-left-radius: 3.40282e38px;
-      border-bottom-left-radius: 3.40282e38px;
       border-top-left-radius: var(--radius-full);
       border-bottom-left-radius: var(--radius-full);
     }
 
     .rounded-l-none {
-      border-top-left-radius: 0;
-      border-bottom-left-radius: 0;
       border-top-left-radius: var(--radius-none);
       border-bottom-left-radius: var(--radius-none);
     }
@@ -10443,12 +10446,10 @@ test('rounded-tl', async () => {
     }
 
     .rounded-tl-full {
-      border-top-left-radius: 3.40282e38px;
       border-top-left-radius: var(--radius-full);
     }
 
     .rounded-tl-none {
-      border-top-left-radius: 0;
       border-top-left-radius: var(--radius-none);
     }
 
@@ -10503,12 +10504,10 @@ test('rounded-tr', async () => {
     }
 
     .rounded-tr-full {
-      border-top-right-radius: 3.40282e38px;
       border-top-right-radius: var(--radius-full);
     }
 
     .rounded-tr-none {
-      border-top-right-radius: 0;
       border-top-right-radius: var(--radius-none);
     }
 
@@ -10563,12 +10562,10 @@ test('rounded-br', async () => {
     }
 
     .rounded-br-full {
-      border-bottom-right-radius: 3.40282e38px;
       border-bottom-right-radius: var(--radius-full);
     }
 
     .rounded-br-none {
-      border-bottom-right-radius: 0;
       border-bottom-right-radius: var(--radius-none);
     }
 
@@ -10623,12 +10620,10 @@ test('rounded-bl', async () => {
     }
 
     .rounded-bl-full {
-      border-bottom-left-radius: 3.40282e38px;
       border-bottom-left-radius: var(--radius-full);
     }
 
     .rounded-bl-none {
-      border-bottom-left-radius: 0;
       border-bottom-left-radius: var(--radius-none);
     }
 
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -1054,6 +1054,26 @@ test('z-index', async () => {
       '-z-[var(--value)]/foo',
     ]),
   ).toEqual('')
+
+  expect(
+    await compileCss(
+      css`
+        @theme {
+          --z-index-auto: 42;
+        }
+        @tailwind utilities;
+      `,
+      ['z-auto'],
+    ),
+  ).toMatchInlineSnapshot(`
+    ":root, :host {
+      --z-index-auto: 42;
+    }
+
+    .z-auto {
+      z-index: var(--z-index-auto);
+    }"
+  `)
 })
 
 test('order', async () => {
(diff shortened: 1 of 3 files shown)


rozsazoltan:
The PR sounds good. The class duplication is frustrating.

philipp-spiess:
Going to retire this for now

philipp-spiess:
Actually still like this better than the alternatives. Let me see if an LLM can figure out all utilities without too much hassle.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -1214,6 +1294,27 @@ test('col-start', async () => {
...
+      `,
+      ['col-start-auto'],
+    ),
+  ).toMatchInlineSnapshot(`
+    ":root, :host {
+      --grid-column-start-auto: 7;
+    }
+
+    .col-start-auto {
+      grid-column: var(--grid-column-start-auto);
```
RobinMalfait: Interesting, this should be solved by #18907

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -9865,15 +10292,11 @@ test('rounded-t', async () => {
     }
 
     .rounded-t-full {
-      border-top-left-radius: 3.40282e38px;
-      border-top-right-radius: 3.40282e38px;
       border-top-left-radius: var(--radius-full);
       border-top-right-radius: var(--radius-full);
```
RobinMalfait, line 9951: Nice!

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -22056,6 +22606,89 @@ test('backdrop-filter', async () => {
...
+
+    :root, :host {
+      --backdrop-blur-none: 2px;
+    }
+
+    .backdrop-blur-none {
+      --tw-backdrop-blur: blur(var(--backdrop-blur-none));
+      -webkit-backdrop-filter: var(--tw-backdrop-blur, ) var(--tw-backdrop-brightness, ) var(--tw-backdrop-contrast, ) var(--tw-backdrop-grayscale, ) var(--tw-backdrop-hue-rotate, ) var(--tw-backdrop-invert, ) var(--tw-backdrop-opacity, ) var(--tw-backdrop-saturate, ) var(--tw-backdrop-sepia, );
+      backdrop-filter: var(--tw-backdrop-blur, ) var(--tw-backdrop-brightness, ) var(--tw-backdrop-contrast, ) var(--tw-backdrop-grayscale, ) var(--tw-backdrop-hue-rotate, ) var(--tw-backdrop-invert, ) var(--tw-backdrop-opacity, ) var(--tw-backdrop-saturate, ) var(--tw-backdrop-sepia, );
+    }
```
RobinMalfait, line 22721: Not related to this PR, but this output is crazy and made me think about something... I wonder if we can be a bit smarter here and optimize the output based on what you used.

So if you _only_ use `blur` that we get rid of all the other variables used here.

If you have a utility that sets some backdrop opacity, or if we detect that `--tw-backdrop-saturate` is used, that we start including it in the output.

Food for thought 🤔

packages/tailwindcss/src/utilities.ts
```diff
@@ -431,6 +432,14 @@ export function createUtilities(theme: Theme) {
...
             if (!value?.includes('/') && candidate.modifier) return
           }
+
+          if (value === null && !negative && desc.staticValues) {
+            let fallback = desc.staticValues[candidate.value.value]
+            if (fallback) {
+              if (candidate.modifier) return
+              return fallback
+            }
+          }
```
RobinMalfait:
```suggestion
          if (value === null && !negative && desc.staticValues && !candidate.modifier) {
            let fallback = desc.staticValues[candidate.value.value]
            if (fallback) return fallback
          }
```

RobinMalfait: Diff isn't super clear, but essentially move the `if (candidate.modifier) return` into the main `if`

packages/tailwindcss/src/utilities.ts
```diff
@@ -1171,43 +1225,93 @@ export function createUtilities(theme: Theme) {
...
+      center: [decl('transform-origin', 'center')],
+      top: [decl('transform-origin', 'top')],
+      'top-right': [decl('transform-origin', '100% 0')],
+      right: [decl('transform-origin', '100%')],
+      'bottom-right': [decl('transform-origin', '100% 100%')],
+      bottom: [decl('transform-origin', 'bottom')],
+      'bottom-left': [decl('transform-origin', '0 100%')],
+      left: [decl('transform-origin', '0')],
+      'top-left': [decl('transform-origin', '0 0')],
+    },
```
RobinMalfait, line 1229: I almost wish this was a Map so all the keys would have quotes around them 😂

packages/tailwindcss/src/utilities.ts
```diff
@@ -683,20 +700,26 @@ export function createUtilities(theme: Theme) {
     },
     themeKeys: ['--grid-column'],
     handle: (value) => [decl('grid-column', value)],
+    staticValues: {
+      auto: [decl('grid-column', 'auto')],
+    },
   })
-  staticUtility('col-span-full', [['grid-column', '1 / -1']])
+  // Ensure discoverability for `col-auto` after converting to fallback
+  suggest('col', () => [{ values: ['auto'], valueThemeKeys: ['--grid-column'] }])
```
RobinMalfait: I wonder if we can make these suggestions more automatic as part of the `functionalUtility` helper because we know the root, we know the values (keys of `staticValues`) and we know the `valueThemeKeys`.

Should simplify some of the additional code and ensure everything is in sync when we update some static values. Thoughts?

RobinMalfait: Can't provide a suggestion to code that didn't change (thanks GitHub!), but something like this should work I think:
```ts
    if (desc.staticValues) {
      let values = Object.keys(desc.staticValues)
      suggest(classRoot, () => [{ values, valueThemeKeys: desc.themeKeys ?? [] }])
    }
```
at the end of the `functionalUtility` function

RobinMalfait (commented):
I like it! Got a few questions and suggestions related to.. suggestions to simplify this a bit more.

packages/tailwindcss/src/utilities.ts
philipp-spiess: Hmmmm, right now though, the `functionalUtility` does not provide any suggestions on its own and I think that's a good thing: If we ever decide to double-down on the browser build, we'd want a convenient way to compile the suggestions away and then this current setup will become useful. That said for the previous `staticUtility()` helper, suggestions _are provided by default_ so it wasn't very consistent to begin with so I'm fine with either approach.

packages/tailwindcss/src/utilities.test.ts
philipp-spiess: Hmm, interesting, but unfortunately would be a breaking change if people used runtime to define these variables right? 🤔

RobinMalfait: It would, but we also consider those `--tw-*` variables as private-ish anyway 🤷‍♂️

packages/tailwindcss/src/utilities.ts
RobinMalfait: > right now though, the `functionalUtility` does not provide any suggestions on its own

What do you mean by this exactly? We do have this already:

https://github.com/tailwindlabs/tailwindcss/blob/340b59dcde2401f03f1edecf1c1332190adc9d44/packages/tailwindcss/src/utilities.ts#L449-L456

Which is part of the `functionalUtility` function when being called.

packages/tailwindcss/src/utilities.test.ts
philipp-spiess: Yeah fair I guess 👍

packages/tailwindcss/src/utilities.ts
philipp-spiess: Oh I see, interesting! Then I guess I am wrong and it should just work. no idea why namespaces keys aren't added automatically then too, though 🤔

packages/tailwindcss/src/utilities.test.ts
thecrypticace: I think this would be the first case of the presence of other utilities affecting the generated CSS for different utilities. Would also need to invalidate the internal AST caches for those utilities.

philipp-spiess: Yeah it does feel like a bit of a stretch, especially in regard to the changes in this PR.

RobinMalfait: @thecrypticace  Yep but I like to think about it in a different way. Similar to how we define CSS variables and remove them when they are not used, or similar to how we only emit CSS for used utilities today.

The CSS variables are also only emitted based on the presence of _other_ utilities (except for `@theme static`).

If you think about it like this, then it could be part of the optimization step where we remove certain `@property` at-rules or used `var(--tw-*)` in declarations. Might not be as easy to generalize this but it wouldn't touch the cached AST.

RobinMalfait: > Yeah it does feel like a bit of a stretch, especially in regard to the changes in this PR.

Hence the `Not related to this PR` part, just noticed it while looking over the diff 😅

So definitely something that should not be part of this PR. Please ignore all aforementioned comments 🙈.

packages/tailwindcss/src/utilities.ts
philipp-spiess: @RobinMalfait updated!

packages/tailwindcss/src/utilities.ts
```diff
@@ -638,9 +653,13 @@ export function createUtilities(theme: Theme) {
     },
     themeKeys: ['--z-index'],
     handle: (value) => [decl('z-index', value)],
+    staticValues: {
+      auto: [decl('z-index', 'auto')],
+    },
   })
 
   suggest('z', () => [
+    { values: ['auto'], valueThemeKeys: ['--z-index'] },
```
RobinMalfait: This can be dropped now:


```suggestion
```

packages/tailwindcss/src/utilities.ts
```diff
@@ -661,9 +678,14 @@ export function createUtilities(theme: Theme) {
...
     themeKeys: ['--order'],
     handle: (value) => [decl('order', value)],
+    staticValues: {
+      first: [decl('order', '-9999')],
+      last: [decl('order', '9999')],
+    },
   })
 
   suggest('order', () => [
+    { values: ['first', 'last'], valueThemeKeys: ['--order'] },
```
RobinMalfait: This can be dropped now:


```suggestion
```

packages/tailwindcss/src/utilities.ts
```diff
@@ -4515,8 +4546,46 @@ export function createUtilities(theme: Theme) {
...
+      },
     })
 
+    suggest('transition', () => [
+      {
+        values: ['none', 'all', 'colors', 'opacity', 'shadow', 'transform'],
+        valueThemeKeys: ['--transition-property'],
+      },
+    ])
+
```
RobinMalfait: This can be dropped now:


```suggestion
```

packages/tailwindcss/src/utilities.ts
```diff
@@ -4590,20 +4659,29 @@ export function createUtilities(theme: Theme) {
...
+        ],
+      },
     })
+
+    suggest('ease', () => [
+      {
+        values: ['initial', 'linear'],
+        valueThemeKeys: ['--ease'],
+      },
+    ])
```
RobinMalfait: This can be dropped now:


```suggestion
```

RobinMalfait (commented):
Few more things that can be deleted with that latest change.

A few things after this:

1. Need to update tests because of the fix in #18907 
2. Needs a changelog entry

RobinMalfait (approved):

PR #18057, opened 2025-05-16
Do not migrate declarations in `<style>` blocks

### LINKED ISSUE #17975: @tailwindcss/upgrade - brakes valid CSS

Upgrading to tailwindcss v4.1.6

Vite 5.2.3

For example: v22.14.0

macOS

**Describe your issue**

Running `npx @tailwindcss/upgrade@latest` broke valid CSS in Vue and HTML files.
Examples:

Image
<br>
Image

RobinMalfait:
This PR improves the upgrade tool by making sure that we don't migrate CSS declarations in `<style>…</style>` blocks.

We do this by making sure that:

1. We detect a declaration, the current heuristic is that the candidate is:
   - Preceded by whitespace
   - Followed by a colon and whitespace

   ```html
   <style>
   .foo {
       flex-shrink: 0;
      ^           ^^
   }
   </style>
   ```

2. We are in a `<style>…</style>` block

   ```html
   <style>
   ^^^^^^
   .foo {
       flex-shrink: 0;
   }
   </style>
   ^^^^^^^^
   ```

The reason we have these 2 checks is because just relying on the first heuristic alone, also means that we will not be migrating keys in JS objects, because they typically follow the same structure:

```js
let classes = {
 flex: 0,
^    ^^
}
```

Another important thing to note is that we can't just ignore anything in between `<style>…</style>` blocks, because you could still be using `@apply` that we _do_ want to migrate.

Last but not least, the first heuristics is not perfect either. If you are writing minified CSS then this will likely fail if there is no whitespace around the candidate.

But my current assumption is that nobody should be writing minified CSS, and minified CSS will very likely be generated and gitignored. In either situation, replacements in minified CSS will not be any worse than it is today.

I'm open to suggestions for better heuristics.

## Test plan

1. Added an integration test that verifies that we do migrate `@apply` and don't migrate the `flex-shrink: 0;` declaration.

Fixes: #17975

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
@@ -22,6 +22,37 @@ export function isSafeMigration(
   location: { contents: string; start: number; end: number },
   designSystem: DesignSystem,
 ): boolean {
+  // Ensure we are not migrating a candidate in a `<style>` block. The heuristic
+  // would be if the candidate is preceded by a whitespace and followed by a
+  // colon and whitespace.
+  //
+  // E.g.:
+  // ```vue
+  // <template>
+  //   <div class="foo"></div>
+  // </template>
+  //
+  //
+  // <style>
+  // .foo {
+  //   flex-shrink: 0;
+  //  ^           ^^
+  // }
+  // </style>
+  // ```
+  if (
+    // Whitespace before the candidate
+    location.contents[location.start - 1]?.match(/\s/) &&
+    // A colon followed by whitespace after the candidate
+    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/) &&
+    // A `<style` block is present before the candidate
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
```diff
@@ -22,6 +22,37 @@ export function isSafeMigration(
...
+  // }
+  // </style>
+  // ```
+  if (
+    // Whitespace before the candidate
+    location.contents[location.start - 1]?.match(/\s/) &&
+    // A colon followed by whitespace after the candidate
+    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/) &&
+    // A `<style` block is present before the candidate
+    location.contents.slice(0, location.start).includes('<style') &&
```
RobinMalfait, line 49: Note this is using `<style` not `<style>` because there could be additional attributes

philipp-spiess (approved):
Seems reasonable!

PR #18059, opened 2025-05-16
Improve error messages when `@apply` fails

RobinMalfait:
This PR improves error messages when `@apply` fails. Right now it gives you a generic error message that you cannot apply a certain utility.

```css
.foo {
  @apply bg-red-500;
}
```

Would result in:
```
Cannot apply unknown utility class: bg-red-500
```

However, there are some situations where we can give you more context about what's happening.

### Missing `@import "tailwindcss"` or `@reference`

If you are in a Vue file for example, and you have the following code:
```vue
<template>
  <div class="foo"></div>
</template>

<style>
.foo {
  @apply bg-red-500;
}
</style>
```

Then this will now result in:
```
Cannot apply unknown utility class `bg-white`. Are you using CSS modules or similar and missing `@reference`? https://tailwindcss.com/docs/functions-and-directives#reference-directive
```

We do this by checking if we found a `@tailwind utilities` or `@reference`. If not, we throw this more specific error.

### Explicitly excluded classes via `@source not inline('…')`

Or via the legacy `blocklist` from a JS config.

If you then have the following file:
```css
@import "tailwindcss";

@source not inline('bg-white');

.foo {
  @apply bg-white;
}
```

Then this will now result in:
```
Cannot apply utility class `bg-white` because it has been explicitly disabled: https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-excluding-classes
```

We do this by checking if the class was marked as invalid.

### Applying unprefixed class in prefix mode

If you have the prefix option configured, but you are applying a non-prefixed class, then we will show the following error:

Given this input:
```css
@import "tailwindcss" prefix(tw);

.foo {
  @apply underline;
}
```

The following error is thrown:
```
Cannot apply unprefixed utility class `underline`. Did you mean `tw:underline`?
```

### Applying known utilities with unknown variants

If you have unknown variants, then we will list them as well if the base utility does compile correctly.

Given this input:

```css
@import "tailwindcss";

.foo {
  @apply hocus:hover:pocus:bg-red-500;
}
```

The following error is thrown:
```
Cannot apply utility class `hocus:hover:pocus:bg-red-500` because the `hocus` and `pocus` variants do not exist.
```

## Test plan

1. Everything behaves the same, but the error messages give more details.
2. Updated tests with new error messages
3. Added new unit tests to verify the various scenarios
4. Added a Vue specific integration test with a `<style>…</style>` block using `@apply`

[ci-all] There are some newlines here and there, let's verify that they work identically on all platforms.

### DIFF
--- a/packages/tailwindcss/src/apply.ts
+++ b/packages/tailwindcss/src/apply.ts
@@ -197,26 +197,10 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
               )
             }
 
-            // Verify `@tailwind utilities` or `@reference` is used
-            let hasUtilitiesOrReference = false
-            walk(ast, (node, { context }) => {
-              // Find `@reference`
-              if (context.reference) {
-                hasUtilitiesOrReference = true
-                return WalkAction.Stop
-              }
-
-              // Find `@tailwind utilities`
-              else if (
-                node.kind === 'at-rule' &&
-                node.name === '@tailwind' &&
-                (node.params === 'utilities' || node.params.startsWith('utilities'))
-              ) {
-                hasUtilitiesOrReference = true
-                return WalkAction.Stop
-              }
-            })
-            if (!hasUtilitiesOrReference) {
+            // When the theme is empty, it means that no theme was loaded and
+            // `@import "tailwindcss"`, `@reference "app.css"` or similar is
+            // very likely missing.
+            if (designSystem.theme.size === 0) {
               throw new Error(
                 `Cannot apply unknown utility class: \`${candidate}\`.\nIt looks like you are missing a \`@reference "app.css"\` or \`@import "tailwindcss";\``,
               )
--- a/packages/tailwindcss/src/apply.ts
+++ b/packages/tailwindcss/src/apply.ts
@@ -4,6 +4,7 @@ import { compileCandidates } from './compile'
 import type { DesignSystem } from './design-system'
 import type { SourceLocation } from './source-maps/source'
 import { DefaultMap } from './utils/default-map'
+import { segment } from './utils/segment'
 
 export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
   let features = Features.None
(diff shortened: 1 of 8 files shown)


packages/tailwindcss/src/apply.ts
```diff
@@ -176,7 +176,54 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
...
+            // template files. This means that the `invalidCandidates` set will
+            // only contain explicit classes via:
+            //
+            // - `blocklist` from a JS config
+            // - `@source not inline(…)`
+            if (designSystem.invalidCandidates.has(candidate)) {
+              throw new Error(
+                `Cannot apply \`${candidate}\`, it seems like the utility was explicitly excluded and cannot be applied.\n\nMore info: https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-excluding-classes`,
+              )
+            }
```
thecrypticace, line 198: This is a nice feature 💯

thecrypticace (approved):
Left a note about the stylesheet name but I'm gonna go ahead and approve this. We can tweak those details later if need be.

packages/tailwindcss/src/apply.ts
```diff
@@ -176,7 +176,54 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
...
+                node.name === '@tailwind' &&
+                (node.params === 'utilities' || node.params.startsWith('utilities'))
+              ) {
+                hasUtilitiesOrReference = true
+                return WalkAction.Stop
+              }
+            })
+            if (!hasUtilitiesOrReference) {
+              throw new Error(
+                `Cannot apply unknown utility class: \`${candidate}\`.\nIt looks like you are missing a \`@reference "app.css"\` or \`@import "tailwindcss";\``,
```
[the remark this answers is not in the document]
RobinMalfait: I used `app.css` because we use that a lot in the docs, e.g.: https://tailwindcss.com/docs/theme#what-are-theme-variables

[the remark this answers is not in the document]
thecrypticace: Ah that's fair. Seems fine them.

packages/tailwindcss/src/apply.ts
```diff
@@ -176,7 +176,54 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
...
+              if (context.reference) {
+                hasUtilitiesOrReference = true
+                return WalkAction.Stop
+              }
+
+              // Find `@tailwind utilities`
+              else if (
+                node.kind === 'at-rule' &&
+                node.name === '@tailwind' &&
+                (node.params === 'utilities' || node.params.startsWith('utilities'))
```
philipp-spiess, line 213: Let's check the `designSystem` itself so it:

- works when there's only a theme file inlined
- we can check it in constant time
- it will also work if the theme config is only coming from legacy plugins

philipp-spiess (dismissed):

packages/tailwindcss/src/apply.ts
```diff
@@ -176,7 +177,60 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
...
+                }
+              }
+            }
+
+            // When the theme is empty, it means that no theme was loaded and
+            // `@import "tailwindcss"`, `@reference "app.css"` or similar is
+            // very likely missing.
+            if (designSystem.theme.size === 0) {
+              throw new Error(
+                `Cannot apply unknown utility class: \`${candidate}\`.\nIt looks like you are missing a \`@reference "app.css"\` or \`@import "tailwindcss";\``,
```
reinink, line 228:
```suggestion
                `Cannot apply \`${candidate}\`, as this class is unknown. It looks like you are missing \`@import "tailwindcss";\` or \`@reference "app.css"\``,
```

```txt
Cannot apply `bg-white`, as this class is unknown. It looks like you are missing either `@import "tailwindcss";` or `@reference "app.css";` in your project.
```

packages/tailwindcss/src/apply.ts
```diff
@@ -176,7 +177,60 @@ export function substituteAtApply(ast: AstNode[], designSystem: DesignSystem) {
...
+              )
+            }
+
+            // Fallback to most generic error message
+            throw new Error(`Cannot apply unknown utility class: \`${candidate}\``)
```
reinink, line 233:
```suggestion
            throw new Error(`Cannot apply \`${candidate}\`, as this class is unknown.`)
```

```txt
Cannot apply `bg-white`, as this class is unknown.
```

reinink (dismissed):
Some feedback — still needs to be reviewed by @adamwathan, so don't merge anything in yet.

thecrypticace (approved):

PR #18065, opened 2025-05-16
Improve `@tailwindcss/upgrade` and `pnpm` workspaces support

RobinMalfait:
This PR fixes an issue where an error such as:

image

Will be thrown during the upgrade process. This can happen when you are using `pnpm` and your CSS file includes a `@import "tailwindcss";`. In this scenario, `tailwindcss` will be loaded from a shared `.pnpm` folder outside of the current working directory.

In this case, we are also not interested in migrating _that_ file, but we also don't want the upgrade process to just crash.

I didn't see an option to ignore errors like this, so wrapped it in a try/catch instead.

It also fixes another issue where if you are using a pnpm workspace and run the upgrade tool from the root, then it throws you an error that you cannot add dependencies to the workspace root unless `-w` or `--workspace-root` flags are passed.

For this, we disable the check entirely using the `--ignore-workspace-root-check` flag. If we always used the `--workspace-root` flag, then the dependencies would always be added to the root, regardless of where you are running the script from which is not what we want. 

## Test plan

Before:

image

After:

image

Before:
 
image

After:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Upgrade: Do not migrate declarations that look like candidates in `<style>` blocks ([#18057](https://github.com/tailwindlabs/tailwindcss/pull/18057))
+- Upgrade: Improve `pnpm` workspaces support ([#18065](https://github.com/tailwindlabs/tailwindcss/pull/18065))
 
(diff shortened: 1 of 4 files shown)


thecrypticace:
Should we add tests for these two scenarios so we can make sure things keep working?

RobinMalfait:
> Should we add tests for these two scenarios so we can make sure things keep working?

Good idea, added some failing tests in between commits and they got fixed by the followup commits

philipp-spiess (approved):

Issue #17228, opened 2025-03-16
@tailwindcss/upgrade  - JavaScript heap out of memory


jcalfee:
**What version of Tailwind CSS are you using?**

v4.0.14

**What build tool (or framework if it abstracts the build tool) are you using?**

For example: solid-start 0.3.11

**What version of Node.js are you using?**

v20.18.1

**What operating system are you using?**

For example: Linux

**Describe your issue**

Could you please clarify the RAM requirements for the auto upgrade?  I'm up to 4GB and still running out of heap.  Is it expected that this would need so much heap or am I looking at a bug?  It takes about 5 min of high CPU before it crashes.  

The project is not that large (about 128 component files).  I certainly have css in ts code and jsx code.  I'm not sure what the updater supports.

I am unable to create a reproduction URL.  I'm just wondering if this much RAM and CPU is normal.  

```bash
$ NODE_OPTIONS="--max-old-space-size=4096" npx @tailwindcss/upgrade
⠋(node:808566) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(Use `node --trace-warnings ...` to show where the warning was created)
≈ tailwindcss v4.0.14

│ Searching for CSS files in the current directory and its subdirectories… 


<--- Last few GCs --->

[808605:0xca59510]   281087 ms: Mark-Compact 3777.8 (4128.5) -> 3764.7 (4128.0) MB, 3082.03 / 0.00 ms  (average mu = 0.139, current mu = 0.033) task; scavenge might not succeed
[808605:0xca59510]   281126 ms: Scavenge 3778.5 (4128.0) -> 3766.4 (4128.0) MB, 7.14 / 0.00 ms  (average mu = 0.139, current mu = 0.033) task; 
[808605:0xca59510]   281158 ms: Scavenge 3778.3 (4128.0) -> 3767.7 (4129.0) MB, 8.80 / 0.00 ms  (average mu = 0.139, current mu = 0.033) task; 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0xb8d0a3 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
[...]

PR #18068, opened 2025-05-16
Improve performance of upgrade tool

RobinMalfait:
This PR improves the performance of the upgrade tool due to a regression introduced by https://github.com/tailwindlabs/tailwindcss/pull/18057

Essentially, we had to make sure that we are not in `<style>…</style>` tags because we don't want to migrate declarations in there such as `flex-shrink: 0;`

The issue with this approach is that we checked _before_ the candidate if a `<style` cold be found and if we found an `</style>` tag after the candidate.

We would basically do this check for every candidate that matches.

Running this on our Tailwind UI codebase, this resulted in a bit of a slowdown:

```diff
- Before: ~13s
+  After: ~5m 39s
```

... quite the difference.

This is because we have a snapshot file that contains ~650k lines of code. Looking for `<style>` and `</style>` tags in a file that large is expensive, especially if we do it a lot.

I ran some numbers and that file contains ~1.8 million candidates.

Anyway, this PR fixes that by doing a few things:

1. We will compute the `<style>` and `</style>` tag positions only once per file and cache it. This allows us to re-use this work for every candidate that needs it.
2. We track the positions, which means that we can simply check if a candidate's location is within any of 2 start and end tags. If so, we skip it.

Running the numbers now gets us to:

```diff
- Before: ~5m 39s
+  After: ~9s
```

Much better!

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
@@ -45,13 +45,20 @@ export function isSafeMigration(
     // Whitespace before the candidate
     location.contents[location.start - 1]?.match(/\s/) &&
     // A colon followed by whitespace after the candidate
-    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/) &&
-    // A `<style` block is present before the candidate
-    location.contents.slice(0, location.start).includes('<style') &&
-    // `</style>` is present after the candidate
-    location.contents.slice(location.end).includes('</style>')
+    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/)
   ) {
-    return false
+    // Compute all `<style>` ranges once and cache it for the current files
+    let ranges = styleBlockRanges.get(location.contents)
+
+    for (let i = 0; i < ranges.length; i += 2) {
+      let start = ranges[i]
+      let end = ranges[i + 1]
+
+      // Check if the candidate is inside a `<style>` block
+      if (location.start >= start && location.end <= end) {
+        return false
+      }
+    }
   }
 
   let [candidate] = Array.from(parseCandidate(rawCandidate, designSystem))
--- a/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
@@ -1,5 +1,6 @@
 import { parseCandidate } from '../../../../tailwindcss/src/candidate'
 import type { DesignSystem } from '../../../../tailwindcss/src/design-system'
+import { DefaultMap } from '../../../../tailwindcss/src/utils/default-map'
 import * as version from '../../utils/version'
 
 const QUOTES = ['"', "'", '`']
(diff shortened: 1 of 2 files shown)


packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
```diff
@@ -44,16 +45,23 @@ export function isSafeMigration(
...
+
+      // Check if the candidate is inside a `<style>` block
+      if (location.start >= start && location.end <= end) {
+        return false
+      }
+    }
   }
 
-  let [candidate] = Array.from(parseCandidate(rawCandidate, designSystem))
+  let [candidate] = parseCandidate(rawCandidate, designSystem)
```
RobinMalfait, line 64: Small improvement, no need to convert to an array first...

packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
```diff
@@ -62,10 +70,10 @@ export function isSafeMigration(
   // So let's only skip if we couldn't parse and we are not in Tailwind CSS v3.
   //
   if (!candidate && version.isGreaterThan(3)) {
-    return true
+    return false
```
RobinMalfait, line 73: I'm a bit of a dumb dumb here. When we know that a candidate doesn't parse, and we know that we are in v4 and up, then we know that we can't migrate the candidate anyway so we can bail.

That makes it an unsafe migration (which skips all the migrations for this candidate). We were returning `true` which means that we were migrating the candidate. This still worked because eventually the migrations were no-ops because they all bailed.

But let's not do that work in the first place, yeah?

RobinMalfait: This change alone makes the upgrade tool on Tailwind UI's product folder go from ~12s to ~9s

RobinMalfait: Oh and that's because this branch was hit about 1 041 458 times in that codebase.

thecrypticace (changes requested):
Found a bug in `styleBlockRanges`

packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
```diff
@@ -168,3 +176,30 @@ export function isSafeMigration(
...
+  while (true) {
+    let startTag = source.indexOf('<style', offset)
+    if (startTag === -1) return ranges
+
+    // Ensure the style looks like:
+    // - `<style>`   (closed)
+    // - `<style …>` (with attributes)
+    if (!source[startTag + 6].match(/[>\s]/)) return ranges
+
+    offset = startTag + 1
```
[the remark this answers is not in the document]
RobinMalfait: Yep makes sense, thanks!

packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
```diff
@@ -44,16 +45,23 @@ export function isSafeMigration(
...
-    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/) &&
-    // A `<style` block is present before the candidate
-    location.contents.slice(0, location.start).includes('<style') &&
-    // `</style>` is present after the candidate
-    location.contents.slice(location.end).includes('</style>')
+    location.contents.slice(location.end, location.end + 2)?.match(/^:\s/)
   ) {
-    return false
+    // Compute all `<style>` ranges once and cache it for the current files
+    let ranges = styleBlockRanges.get(location.contents)
```
philipp-spiess, line 51: Will this map end up having the content of all your repo in there by the end of it? We've already seen people having memory issues with the upgrade tool (#17228) so I wonder if we should do something a bit smarter here. E.g. it's probably enough to only retain a list of up to NUM_CORES files EVER since files are always completed before we start a new one or so?

RobinMalfait: For future reference:

Tested this on our Tailwind UI codebase, and it indeed results in ~82mb of memory usage. However, the very last file we scan is an 650k line JSON file so even with an LRU cache in place it's still ~71mb of memory usage in our case.

I believe the biggest bottleneck is if you are scanning too many files (aka, node_modules is scanned for some reason) and that that's the root cause instead of this pre-computed cache.

Going to keep it as-is for now, but if people run into memory issues like this, then we know where to look.

thecrypticace (approved):

Issue #18067, opened 2025-05-16
[V4] Regression in ability to order custom variants with built-in ones?


Aurailus:
**What version of Tailwind CSS are you using?**

v4.1.7

**What build tool (or framework if it abstracts the build tool) are you using?**

Webpack 5.96.1 + PostCSS 8.4.48

**What version of Node.js are you using?**

v20.18.0

**What browser are you using?**

Chrome

**What operating system are you using?**

Arch Linux

**Reproduction URL**

https://play.tailwindcss.com/UBrKdpd6MD?file=css

**Describe your issue**

In Tailwind V3, it was possible to re-declare built-in variants (e.g. active:) so that they take priority over custom variants that should have lower precedence. In Tailwind V4, this doesn't seem to be possible, which severely messes with the ability to make custom "shorthand" variants for focus and hover states (hocus:) without losing the ability to also have active: or disabled: or any of the builtin variants take precedence. I posted about this in this [discussion](https://github.com/tailwindlabs/tailwindcss/discussions/17860) but I didn't get any response. Is there a way to work around this in V4, or is this just a regression?

Aurailus:
Pinned this down to the fact that the variants code will re-use an existing variant's `order` property if it's redefined, meaning that it's impossible to give `active` a higher precedence by redefining it using `custom-variant`. See #18070 for a PR which fixes this.

thecrypticace:
See my comment here https://github.com/tailwindlabs/tailwindcss/pull/18070#issuecomment-2891385954

PR #18070, opened 2025-05-17
Allow `@custom-variant` to redefine built-in variants with proper `order`.

Aurailus:
## Summary

Adds an "override" property to the object bag passed to `Variants.static` which forces a variant to get a new `order` even if it has already been defined. This property is set in `index.ts:399` for the `@custom-variant` rule to allow custom variants to redefine existing variants like `active:` and have them be properly ordered relative to other custom variants, a.l.a tailwind v3. See #18067 for the motivating regression.

## Test plan

Tested using the playground directory. This is a very insignificant (~17 LOC) change which I don't expect to cause many new issues. I can add more robust testing if necessary, just let me know.

### DIFF
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -396,6 +396,7 @@ async function parseCss(
             },
             {
               compounds: compoundsForSelectors([...styleRuleSelectors, ...atRuleParams]),
+              override: true,
             },
           )
         })
--- a/packages/tailwindcss/src/variants.ts
+++ b/packages/tailwindcss/src/variants.ts
@@ -69,7 +69,11 @@ export class Variants {
   static(
     name: string,
     applyFn: VariantFn<'static'>,
-    { compounds, order }: { compounds?: Compounds; order?: number } = {},
+    {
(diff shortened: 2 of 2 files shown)


thecrypticace:
The fact that a redefined variant reuses its order was an intentional design decision. This lets someone override `hover` in place without necessarily affecting how it works relative to the order of other variants. e.g. to tweak or remove the media query for it entirely. If we did not do that then tweaking the selector for a variant could result in things no longer matching the way they did relative to other variants. 

For example, if this moved `hover` to the end:
```
@custom-variant hover (&:hover);
```

Then this div would still be blue when clicked when all someone was doing was trying to remove the media query from `hover` that we ship with by default:

```
<div class="bg-red-500 hover:bg-blue-500 active:bg-green-500 size-32"></div>
```

Changing this *now* could break backwards compatibility with previous 4.x releases in subtle and unexpected ways so I don't see us doing this.

Aurailus:
@thecrypticace so then how do you propose resolving this in practice? Because this *is* a regression from the behaviour of Tailwind 3.x

PR #18126, opened 2025-05-23
Upgrade deprecated `order-none` to `order-0`

RobinMalfait:
This PR adds an initial version for deprecated utilities. Right now it's hardcoded to just the `order-none` utility.

This means that `order-0` and `order-[0]` will not be migrated to `order-none` anymore. We did that automatically because we prefer named utilities over bare values and arbitrary values.

With this PR, `order-none` is ignored.

Similarly, `order-none` will be migrated to `order-0` instead (defined in a separate migration for deprecated values). Made it a new migration instead of using the legacy migration because there all utilities still exist, but are defined differently (e.g.: `shadow`, `shadow-sm`, `shadow-xs`).

This PR is also an initial version, it doesn't add any form of `deprecated` flag or feature on a per-utility implementation basis. This therefor has the side effect that if you have a custom `order-none` defined, that it will also be ignored during migrations.

## Test plan

1. Added tests to ensure the `order-0` is not migrated to `order-none`
2. Added tests to ensure `order-none` is migrated to `order-0` (if it's safe to do so, the signature is still computed to ensure the output is the same).
3. Ran this on the Tailwind Plus codebase and ensured that `order-0` is not migrated to `order-none` and that `order-none` is migrated to `order-0`.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
@@ -76,8 +76,8 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s', 
     ['order-[0]', 'order-0'],
     ['order-0', 'order-0'],
 
-    // Do not migrate away `order-none` if it's customly defined and thus not
-    // safe to migrate to `order-0`
+    // Do not migrate `order-none` if defined as a custom utility as it is then
+    // not safe to migrate to `order-0`
     ['order-none', 'order-none'],
   ])(`${testName} with custom implementations`, async (candidate, result) => {
     if (strategy === 'with-variant') {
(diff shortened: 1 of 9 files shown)


thecrypticace:
Maybe move the definition of order-none from utilities.ts to the legacy utilities file too?

packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
```diff
@@ -63,4 +71,38 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s',
     let migrated = await migrate(designSystem, {}, candidate)
     expect(migrated).toEqual(result)
   })
+
+  test.each([
+    ['order-[0]', 'order-0'],
+    ['order-0', 'order-0'],
+
+    // Do not migrate away `order-none` if it's customly defined and thus not
+    // safe to migrate to `order-0`
```
thecrypticace, line 80:
```suggestion
    // Do not migrate `order-none` if defined as a custom utility as it is then
    // not safe to migrate to `order-0`
```

thecrypticace (approved):

PR #18133, opened 2025-05-23
Fix "Cannot read properties of undefined" crash on malformed arbitrary value

### LINKED ISSUE #17064: TypeError: Cannot read properties of undefined (reading 'nodes')
4.0.0

@tailwindcss/cli

Bun 1.2.x

Edge

Windows

**Reproduction URL**
N/A - error is in the build process and can't be replicated on play

**Describe your issue**
When running `bunx @tailwindcss/cli -i ./Blazor/Blazor/wwwroot/app.css -o ./Blazor/Blazor/wwwroot/app.min.css --watch` it intermittently outputs `TypeError: Cannot read properties of undefined (reading 'nodes')`

Image

RobinMalfait:
This PR fixes a crash when an arbitrary value was malformed and crashed the build.

If you have a utility like `[--btn-border:var(--color-maroon)/90)]` which is malformed, it will crash the build. It might not be easy to spot but the easy is the additional `)` after the `90`.

The reason this crashes is because we parse the value `var(--color-maroon)/90)` and when we see `)` we assume it's the end of a "function" which also assumes it was preceded by a `(`. This is not the case and we crash.

This PR fixes that by not assuming the parsed object is available and uses `?` to be safe and only access `nodes` if it's available.

I'm actually not 100% sure what the best solution is in this scenario because these candidates could (and will) be returned from Oxide so even if we throw a more descriptive error, it will still crash the build and you might not even have control over the candidate.

This candidate will now eventually generate the following CSS:
```css
.\[--btn-border\:var\(--color-maroon\)\/90\)\] {
  --btn-border: var(--color-maroon) / ;
}
```

Which still looks odd, but even Lightning CSS doesn't throw an error in this case (because it's a CSS variable definition), so I think it's the best we can do. If you open your devtools you will see the weird values, so it's still debug-able.

image

Fixes: #17064

## Test plan

Manually tested the candidate that crashed it, and after the change generated the above CSS. Then used it in JSFiddle to proof it's fixed now. https://jsfiddle.net/z850ykew/

Couldn't use Tailwind Play because the candidate will cause a crash there as well 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade: Improve `pnpm` workspaces support ([#18065](https://github.com/tailwindlabs/tailwindcss/pull/18065))
 - Upgrade: Migrate deprecated `order-none` to `order-0` ([#18126](https://github.com/tailwindlabs/tailwindcss/pull/18126))
 - Support Leptos `class:` attributes when extracting classes ([#18093](https://github.com/tailwindlabs/tailwindcss/pull/18093))
+- Fix "Cannot read properties of undefined" crash on malformed arbitrary value ([#18133](https://github.com/tailwindlabs/tailwindcss/pull/18133))
 
 ## [4.1.7] - 2025-05-15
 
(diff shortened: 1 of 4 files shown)


thecrypticace (approved):

oneyoung19:
Hi guys, thanks a lot. It turns out the issue was actually `*=" "`. Once I removed it, everything worked.

PR #18147, opened 2025-05-25
feat: allow `size` theme namespace in `width` and `height`

phaux:
I just think it would make sense:

```css
@theme {
  --size-button: 2.5rem; /* should be usable with square buttons and wide buttons */
  --size-bar: 3.5rem; /* should be usable with horizontal and vertical toolbars */
}
```

```html
<!-- vertical toolbar -->
<div class="flex flex-col w-bar">
  <!-- square button with icon -->
  <button class="size-button"><svg>...</svg></button>
</div>

<!-- horizontal toolbar -->
<div class="flex flex-row h-bar">
  <!-- wide button with text -->
  <button class="h-button">Text</button>
</div>
```

### DIFF
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -1005,12 +1005,12 @@ export function createUtilities(theme: Theme) {
   )
 
   for (let [name, namespaces, property] of [
-    ['w', ['--width', '--spacing', '--container'], 'width'],
-    ['min-w', ['--min-width', '--spacing', '--container'], 'min-width'],
-    ['max-w', ['--max-width', '--spacing', '--container'], 'max-width'],
-    ['h', ['--height', '--spacing'], 'height'],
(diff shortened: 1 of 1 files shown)


adamwathan:
Hey! Going to say no to this one for now, want to keep the theme namespaces as light as we can. Can always revisit in the future if there's a lot of demand or strong reasons to add. Thanks regardless!

rozsazoltan:
Until then, you can implement the custom namespace for your own projects:

* https://play.tailwindcss.com/3Jl0rUAKmt?file=css

```css
@utility w-* {
  width: --value(--size-*);
}

@utility h-* {
  height: --value(--size-*);
}

@utility min-w-* {
  min-width: --value(--size-*);
}

@utility min-h-* {
  min-height: --value(--size-*);
}

@utility max-w-* {
  max-width: --value(--size-*);
}

@utility max-h-* {
  max-height: --value(--size-*);
}
```

PR #18184, opened 2025-05-30
Improve upgrade migrations

RobinMalfait:
This PR fixes 2 issues with the migration tool where certain classes weren't migrated. This PR fixes those 2 scenarios:

### Scenario 1

When you have an arbitrary opacity modifier that doesn't use `%`, but is just a number typically between `0` and `1` then this was not converted to the bare value equivalent before.

E.g.:

```html
<div class="bg-[#f00]/[0.16]"></dv>
```

Will now be converted to:

```html
<div class="bg-[#f00]/16"></dv>
```

### Scenario 2

Fixes a bug when a CSS function was used in a fallback value in the CSS variable shorthand syntax. In that case we didn't migrate the class to the new syntax.

This was because we assumed that a `(` was found, that we are dealing with a CSS function. 

E.g.: 
```html
<div class="w-[--spacing(1)]"></div>
                        ^  This indicates a CSS function, we should not be 
                           converting this to `w-(--spacing(1))`
```

But if a function was used as a fallback value, for example:

```html
<div class="bg-[--my-color,theme(colors.red.500)]"></dv>
```

Then we also didn't migrate it, but since the function call is in the fallback, we can still migrate it.

Will now properly be converted to:

```html
<div class="bg-(--my-color,var(--color-red-500))"></dv>
```

## Test plan

1. Added a test for the first case
2. Added a test for the second case
3. Also added an integration-like test that runs all the migration steps to make sure that the `theme(…)` in the fallback also gets updated to `var(…)`. This one caught an issue because the `var(…)` wasn't handling prefixes correctly.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate-theme-to-var.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate-theme-to-var.ts
@@ -193,6 +193,10 @@ export function createConverter(designSystem: DesignSystem, { prettyPrint = fals
     let variable = `--${keyPathToCssProperty(toKeyPath(path))}` as const
     if (!designSystem.theme.get([variable])) return null
 
+    if (designSystem.theme.prefix) {
+      return `--${designSystem.theme.prefix}-${variable.slice(2)}`
+    }
+
     return variable
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Upgrade: migrate arbitrary modifiers with values between `0..=1` to bare values `/[0.16]` -> `/16` ([#18184](https://github.com/tailwindlabs/tailwindcss/pull/18184))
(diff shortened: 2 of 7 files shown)


packages/@tailwindcss-upgrade/src/codemods/template/migrate-theme-to-var.ts
```diff
@@ -193,6 +193,10 @@ export function createConverter(designSystem: DesignSystem, { prettyPrint = fals
     let variable = `--${keyPathToCssProperty(toKeyPath(path))}` as const
     if (!designSystem.theme.get([variable])) return null
 
+    if (designSystem.theme.prefix) {
+      return `--${designSystem.theme.prefix}-${variable.slice(2)}`
+    }
+
```
RobinMalfait, line 199: When adding the test to the `migrate.test.ts` file that runs all the migrations (with variants, prefix, ...) it caught an issue where we didn't handle the `--tw-` prefix properly when reading from theme values, which is actually necessary when you have a prefix.

See https://play.tailwindcss.com/BlXfDMDtZz

thecrypticace (approved):

PR #18212, opened 2025-06-03
Migrate negative arbitrary values to negative bare values

RobinMalfait:
This PR adds some improvements to the upgrade tool where it can now also migrate negative arbitrary values to negative bare values.

We already had support for the positive version of this:
```diff
- mb-[32rem]
+ mb-128
```

But now it can also handle negative values:
```diff
- mb-[-32rem]
+ -mb-128
```

The only tricky part here is that we had to hoist the `-` sign. Before this PR, we were actually generating `mb--128` and that is invalid so it was thrown out.

## Test plan

1. Added a test to ensure that the negative values are correctly transformed.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
@@ -159,7 +159,12 @@ export function migrateArbitraryUtilities(
         candidate.kind === 'arbitrary' ? candidate.value : (candidate.value?.value ?? null)
       if (value === null) return
 
-      let spacingMultiplier = spacing.get(designSystem)?.get(value)
+      let spacingMultiplier = spacing.get(designSystem)?.get(value) ?? null
+      let rootPrefix = ''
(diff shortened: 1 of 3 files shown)


philipp-spiess (approved):

packages/@tailwindcss-upgrade/src/codemods/template/migrate-arbitrary-utilities.ts
```diff
@@ -159,14 +159,21 @@ export function migrateArbitraryUtilities(
         candidate.kind === 'arbitrary' ? candidate.value : (candidate.value?.value ?? null)
       if (value === null) return
 
-      let spacingMultiplier = spacing.get(designSystem)?.get(value)
+      let spacingMultiplier = spacing.get(designSystem)?.get(value) ?? null
+      let rootPrefix = ''
+      if (spacingMultiplier !== null && spacingMultiplier < 0) {
+        rootPrefix = '-'
+        spacingMultiplier = Math.abs(spacingMultiplier)
```
RobinMalfait, line 166: This is also an option 🤔
```suggestion
        spacingMultiplier *= -1
```

PR #18220, opened 2025-06-04
fix: don't break CSS keywords when formatting math expressions

### LINKED ISSUE #18219: Incorrect formatting of valid hyphenated values placed inside a CSS function: `fit-content` becomes `fit - content`

For example: v4.1.8

Tailwind Play

**Reproduction URL**

https://play.tailwindcss.com/kbBgEMHTsM

**Describe your issue**

```html
<div class="min-h-[min(fit-content,calc(100dvh-4rem))]"></div>
```

The expected result would be:

```css
.min-h-\[min\(fit-content\,calc\(100dvh-4rem\)\)\] {
  min-height: min(fit-content, calc(100dvh - 4rem));
}
```

Instead - since it's inside a `min()` function - every hyphen gets surrounded by spaces:

```css
.min-h-\[min\(fit-content\,calc\(100dvh-4rem\)\)\] {
  min-height: min(fit - content, calc(100dvh - 4rem));
}
```

rozsazoltan:
Fixes #18219

## Summary

In an arbitrary value, if there's a non-numeric character both before and after a hyphen, there's no need for a space.

## Test plan

`decodeArbitraryValue` will correctly format special CSS values like `fit-content`. I believe spaces are only necessary if there's a digit either before or after the hyphen.

```js
decodeArbitraryValue('min(fit-content,calc(100dvh-4rem))')
```

This way, the result of the following arbitrary value will also be correct:

```html
<div class="min-h-[min(fit-content,calc(100dvh-4rem))]"></div>
```

```css
.min-h-\[min\(fit-content\,calc\(100dvh-4rem\)\)\] {
  min-height: min(fit-content, calc(100dvh - 4rem));
}
```

### DIFF
--- a/packages/tailwindcss/src/utils/math-operators.ts
+++ b/packages/tailwindcss/src/utils/math-operators.ts
@@ -111,7 +111,6 @@ export function addWhitespaceAroundMathOperators(input: string) {
     else if ((char === '+' || char === '*' || char === '/' || char === '-') && formattable[0]) {
       let trimmed = result.trimEnd()
       let prev = trimmed[trimmed.length - 1]
-      let rest = input.slice(i + 1) || ''
       let next = input[i + 1] || ''
 
       // If we're preceded by an operator don't add spaces
@@ -127,7 +126,7 @@ export function addWhitespaceAroundMathOperators(input: string) {
       }
 
       // If there's no number before or after it and it's not followed by a math function, don't add spaces.
-      else if (next.length > 0 && !/\d/.test(prev + next) && !MATH_FUNCTIONS.some(fn => rest.startsWith(`${fn}(`))) {
+      else if (next.length > 0 && !/\d/.test(prev + next) && !MATH_FUNCTIONS.some(fn => input.slice(i + 1).startsWith(`${fn}(`))) {
         result += char
         continue
       }
--- a/packages/tailwindcss/src/utils/math-operators.ts
+++ b/packages/tailwindcss/src/utils/math-operators.ts
@@ -1,3 +1,18 @@
+const LOWER_A = 0x61
+const LOWER_Z = 0x7a
+const LOWER_E = 0x65
+const UPPER_E = 0x45
+const ZERO = 0x30
+const NINE = 0x39
+const ADD = 0x2b
+const SUB = 0x2d
+const MUL = 0x2a
+const DIV = 0x2f
+const OPEN_PAREN = 0x28
+const CLOSE_PAREN = 0x29
+const COMMA = 0x2c
+const SPACE = 0x20
+
 const MATH_FUNCTIONS = [
   'calc',
   'min',
@@ -37,18 +52,17 @@ export function addWhitespaceAroundMathOperators(input: string) {
   let lastValuePos = null
 
   for (let i = 0; i < input.length; i++) {
-    let char = input[i]
-    let charCode = char.charCodeAt(0)
+    let char = input.charCodeAt(i)
 
     // Track if we see a number followed by a unit, then we know for sure that
     // this is not a function call.
-    if (charCode >= 48 && charCode <= 57) {
+    if (char >= ZERO && char <= NINE) {
       valuePos = i
     }
 
     // If we saw a number before, and we see normal a-z character, then we
     // assume this is a value such as `123px`
-    else if (valuePos !== null && charCode >= 97 && charCode <= 122) {
+    else if (valuePos !== null && char >= LOWER_A && char <= LOWER_Z) {
       valuePos = i
     }
 
@@ -59,8 +73,8 @@ export function addWhitespaceAroundMathOperators(input: string) {
     }
 
     // Determine if we're inside a math function
-    if (char === '(') {
-      result += char
+    if (char === OPEN_PAREN) {
+      result += String.fromCharCode(char)
 
       // Scan backwards to determine the function name. This assumes math
       // functions are named with lowercase alphanumeric characters.
@@ -69,9 +83,9 @@ export function addWhitespaceAroundMathOperators(input: string) {
       for (let j = i - 1; j >= 0; j--) {
         let inner = input.charCodeAt(j)
 
-        if (inner >= 48 && inner <= 57) {
+        if (inner >= ZERO && inner <= NINE) {
           start = j // 0-9
-        } else if (inner >= 97 && inner <= 122) {
+        } else if (inner >= LOWER_A && inner <= LOWER_Z) {
           start = j // a-z
         } else {
           break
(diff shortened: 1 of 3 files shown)


rozsazoltan:
Inside CSS functions, whitespace is added before and after every math operator; I believe this should only be necessary if there is a number before or after the operator.

A number or another operator or function?

Or maybe it's easier to just list the exceptions. I'll leave it as a draft for now.

rozsazoltan:
```js
decodeArbitraryValue('min(fit-content,calc(100dvh-4rem))')
decodeArbitraryValue('min(fit-content,calc(100dvh-4rem)-calc(50dvh-2px))')
```

Now:

```css
min(fit-content, calc(100dvh - 4rem))
min(fit-content, calc(100dvh - 4rem)-calc(50dvh - 2px))
```

And if, in addition to checking for numbers, we also check whether a math function follows it:

```js
let rest = (input.slice(i + 1) || '').trimEnd()
let next = input[i + 1] || ''

// ...

// If there's no number before or after it and it's not followed by a math function, don't add spaces.
else if (
  next.length > 0 &&
  !/\d/.test(prev + next) &&
  !MATH_FUNCTIONS.some(fn => rest.startsWith(fn + '('))
) {
  result += char
  continue
}
```

Will be:

```css
min(fit-content, calc(100dvh - 4rem))
min(fit-content, calc(100dvh - 4rem) - calc(50dvh - 2px))
```

rozsazoltan:
Another possible case is when there is text before it, and an operator follows it - for example: -1, or (2*3), etc.
```
1 - )    // illegal - not relevant
1 - (    // is possible
1 - -1   // is possible
1 - +1   // is possible
```
```js
next !== '(' && next !== '+' && next !== '-'
```

Also, when dealing with functions, we need to consider that formatting is not only required for standard functions, but actually for anything that looks like a function.
```
1 - calc (  // illegal - not relevant
1 - calc(   // is possible
1 - theme(  // is possible (not standard)
```
```js
// Original for only CSS functions
!MATH_FUNCTIONS.some(fn => rest.startsWith(fn + '('))

// New for standard function declaration
!/^[a-zA-Z_$][a-zA-Z0-9_$]*\(/.test(input.slice(i + 1))
```

I believe it's unnecessary to declare every single case, since introducing a new function should come with updating this accordingly. Also, this way the process can work better with custom syntax as well.

wongjn:
A few other edge cases that don't work which we should perhaps consider. Mainly scientific notation and `calc-size()` since they may include `-` in their notations:

```js
['min(-3.4e-2-var(--foo),calc-size(auto))', 'min(-3.4e-2 - var(--foo), calc-size(auto))'],
```
```
Expected: "min(-3.4e-2 - var(--foo), calc-size(auto))"
Received: "min(-3.4e - 2 - var(--foo), calc - size(auto))"
```

```js
[
  'clamp(-10e3-var(--foo),calc-size(max-content),var(--foo)+-10e3)',
  'clamp(-10e3 - var(--foo), calc-size(max-content), var(--foo) + -10e3)'
],
```
```
Expected: "clamp(-10e3 - var(--foo), calc-size(max-content), var(--foo) + -10e3)"
Received: "clamp(-10e3 - var(--foo), calc - size(max-content), var(--foo) + -10e3)"
```

RobinMalfait:
Thanks! I iterated a bit more on this because I didn't like the regex parts.

I started with a similar approach we used before, where we safelisted a few known idents. But then cleaned it up a bit more by making sure we don't have any hardcoded identifiers to make it more future proof.

In order to do this, we have to track a bit more information, but there are already some rules you identified. We want spaces if:
- There is a digit before the operator
- There is a digit after the operator
- There is a `(` after the operator (start of a parenthesized expression)
- There is a `)` before the operator (end of a parenthesized expression / function)
- The operator is followed by another operator. E.g.: `1px + -2px`
- The operator was not preceded by a value (+unit). This is important to handle scenarios like this: `calc(1rem-theme(…))`. In this case `rem-theme` could look like a valid function, but since `1rem` was used it definitely is _not_ part of the function.

Some of them are technically not _required_, but they will look better if it's consistent.

I also added better support for the scientific notation.

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -33,21 +30,33 @@ export function addWhitespaceAroundMathOperators(input: string) {
...
-      hasKnownFunctions = true
-      return `$${KNOWN_DASHED_FUNCTIONS.indexOf(fn)}$(`
-    })
-  }
-
   let result = ''
   let formattable: boolean[] = []
 
+  let valuePos = null
+  let lastValuePos = null
```
RobinMalfait, line 52: Not 100% sure what to call these. Essentially when you see the `1` in `1rem` we start tracking the position, then as long as you see `a-z` we will increment this position.

This basically allows us to track the last value + unit (if any) to know that the `1rem` in `calc(1rem-theme(…))` is definitely _not_ part of the `theme` function.

RobinMalfait: Technically if your value looks like `1.5rem` then it would reset at the `.`, but it will resume at the `5` again, so I think that's safe.

Similarly if you use scientific notation, e.g.: `-3.4e-2` it will eventually resume at the `2`.

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -33,21 +30,33 @@ export function addWhitespaceAroundMathOperators(input: string) {
...
+
+    // Track if we see a number followed by a unit, then we know for sure that
+    // this is not a function call.
+    if (charCode >= 48 && charCode <= 57) {
+      valuePos = i
+    }
+
+    // If we saw a number before, and we see normal a-z character, then we
+    // assume this is a value such as `123px`
+    else if (valuePos !== null && charCode >= 97 && charCode <= 122) {
```
RobinMalfait: This does assume `a-z` lowercase. But we made that assumption earlier already.

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -111,9 +120,20 @@ export function addWhitespaceAroundMathOperators(input: string) {
...
       let trimmed = result.trimEnd()
       let prev = trimmed[trimmed.length - 1]
+      let prevCode = prev.charCodeAt(0)
+      let prevprevCode = trimmed.charCodeAt(trimmed.length - 2)
+
+      let next = input[i + 1]
+      let nextCode = next?.charCodeAt(0)
+
+      // Do not add spaces for scientific notation, e.g.: `-3.4e-2`
+      if ((prev === 'e' || prev === 'E') && prevprevCode >= 48 && prevprevCode <= 57) {
```
RobinMalfait: I wish we had `b'e'` so we could compare char codes directly instead of strings.

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -129,27 +149,31 @@ export function addWhitespaceAroundMathOperators(input: string) {
...
-    // This is specifically to handle this value in the round(…) function:
-    //
-    // ```
-    // round(to-zero, 1px)
-    //       ^^^^^^^
-    // ```
-    //
-    // This is because the first argument is optionally a keyword and `to-zero`
-    // contains a hyphen and we want to avoid adding spaces inside it.
-    else if (formattable[0] && input.startsWith('to-zero', i)) {
```
RobinMalfait, line 149: Bye bye hard coded `to-zero`!

packages/tailwindcss/src/utils/math-operators.ts
thecrypticace: same

packages/tailwindcss/src/utils/math-operators.ts
thecrypticace: 😍

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -111,9 +120,20 @@ export function addWhitespaceAroundMathOperators(input: string) {
     else if ((char === '+' || char === '*' || char === '/' || char === '-') && formattable[0]) {
       let trimmed = result.trimEnd()
       let prev = trimmed[trimmed.length - 1]
+      let prevCode = prev.charCodeAt(0)
+      let prevprevCode = trimmed.charCodeAt(trimmed.length - 2)
```
thecrypticace: `prevPrevCode`

I feel like there's a better name but I can't think of one

RobinMalfait: Yeah it's a little silly isn't it lol

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -129,27 +149,31 @@ export function addWhitespaceAroundMathOperators(input: string) {
...
+        (nextCode >= 48 && nextCode <= 57) ||
+        // Previous is end of a function call (or parenthesized expression)
+        prev === ')' ||
+        // Next is start of a parenthesized expression
+        next === '(' ||
+        // Next is an operator
+        next === '+' ||
+        next === '*' ||
+        next === '/' ||
+        next === '-' ||
```
thecrypticace: 1. Can we drop `next` and just use char codes for these?
2. Can we put the char codes behind named constants (like we do in segment and other places)

packages/tailwindcss/src/utils/math-operators.ts
```diff
@@ -111,9 +120,20 @@ export function addWhitespaceAroundMathOperators(input: string) {
     else if ((char === '+' || char === '*' || char === '/' || char === '-') && formattable[0]) {
       let trimmed = result.trimEnd()
       let prev = trimmed[trimmed.length - 1]
+      let prevCode = prev.charCodeAt(0)
+      let prevprevCode = trimmed.charCodeAt(trimmed.length - 2)
+
+      let next = input[i + 1]
+      let nextCode = next?.charCodeAt(0)
```
thecrypticace: `let nextCode = input.charCodeAt(i+1)`

`NaN` and `undefined` have the same comparison semantics for the checks we're doing I think.

thecrypticace (approved):

PR #18235, opened 2025-06-05
Fix  conditional check for vite config css configuration

jamestalton:
## Summary

The tailwindcss vite implementation is not checking for the conditional existence of the optional `css` options in the vite config.

See: https://github.com/vitejs/vite/blob/0d60667e03d91cc0d48dd2cdbd8154d94e0aba74/packages/vite/src/node/config.ts#L355
```typescript
  css?: CSSOptions
```

In 99% of the cases this works because I think vite normally always fills this in.

> Updated description - orginally mentioned vite preview when meaning vitest-preview

But in the case of [vitest-preview](https://www.vitest-preview.com/) it is not setting this optional field. Which is valid.
The correct fix is for tailwindcss vite to handle this optional field since it is marked as optional.

This should fix vitest-preview with the latest tailwindcss.

## Test plan

There is no functional change as this just adds the extra needed optional chain check.

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -6,12 +6,12 @@ import {
   normalizePath,
   optimize,
   toSourceMap,
-} from '@tailwindcss/node';
-import { clearRequireCache } from '@tailwindcss/node/require-cache';
-import { Scanner } from '@tailwindcss/oxide';
-import fs from 'node:fs/promises';
-import path from 'node:path';
-import type { Plugin, ResolvedConfig, ViteDevServer } from 'vite';
+} from '@tailwindcss/node'
+import { clearRequireCache } from '@tailwindcss/node/require-cache'
+import { Scanner } from '@tailwindcss/oxide'
+import fs from 'node:fs/promises'
+import path from 'node:path'
+import type { Plugin, ResolvedConfig, ViteDevServer } from 'vite'
 
 const DEBUG = env.DEBUG
 const SPECIAL_QUERY_RE = /[?&](?:worker|sharedworker|raw|url)\b/
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -6,12 +6,12 @@ import {
   normalizePath,
   optimize,
   toSourceMap,
-} from '@tailwindcss/node'
-import { clearRequireCache } from '@tailwindcss/node/require-cache'
-import { Scanner } from '@tailwindcss/oxide'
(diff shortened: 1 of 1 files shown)


thecrypticace:
`config` here is the _resolved Vite config_ which means the correct types to reference are here where `css` is *not* optional: https://github.com/vitejs/vite/blob/0d60667e03d91cc0d48dd2cdbd8154d94e0aba74/packages/vite/src/node/config.ts#L587

Also not seeing any kind of crash with `vite preview`. Do you have a reproduction?

jamestalton:
This is the error I get without that fix. I'll work on a sandbox for reproduction.

```
TypeError: Cannot read properties of undefined (reading 'devSourcemap')
    at R.factory (/node_modules/@tailwindcss/vite/dist/index.mjs:1:1542)
    at R.get (/node_modules@tailwindcss/vite/dist/index.mjs:1:3011)
    at TransformContext.transform (/node_modules/@tailwindcss/vite/dist/index.mjs:1:1946)
    at Object.transform (/node_modules/vitest-preview/node_modules/vite/dist/node/chunks/dep-3e87c7b2.js:40612:44)
    at /node_modules/vitest-preview/node_modules/vite/dist/node/chunks/dep-3e87c7b2.js:60579:53
    at async Promise.all (index 0)
    at async devHtmlHook (/node_modules/vitest-preview/node_modules/vite/dist/node/chunks/dep-3e87c7b2.js:60574:5)
    at async applyHtmlTransforms (/node_modules/vitest-preview/node_modules/vite/dist/node/chunks/dep-3e87c7b2.js:43465:21)
    at async /node_modules/vitest-preview/dist/previewServer.mjs:41:18
```

The issue could be that tailwindcss expects a ResolvedConfig where vitest-preview is just providing a UserConfig.

jamestalton:
Ah, I didn't write up that description correctly.
It is `vitest-preview` vs `vite preview`.
https://www.vitest-preview.com/

I see that `vite-preview` has it's own dependency of vite which is `3.2.11`.
Meanwhile all my dependencies are up to date and my vite is 6.3.5.

In vitest-preview previewServer.mjs it is calling
```
  const vite = await createServer$1({
    server: {
      middlewareMode: true
    },
    appType: "custom"
  });
```
Where no `css` config is part of that.

Maybe the older vite `3.2.11` does not turn that into a compatible ResolvedConfig.

Suggestions or thoughts?

I could go down the route of trying to get vitest-preview updated, but it has not had much movement in it's other PRs.

I'm good with closing this issue and I'll try to address with vitest-preview.

thecrypticace:
Yeah, I would suggest trying to get this fixed in `vitest-preview` instead. 👍

PR #18245, opened 2025-06-06
Itsmefminsaf patch 1

itsmefminsaf:
## Summary

I have added some extra breakpoints for mobile screen like 20rem 26rem


RobinMalfait:
Hey!

Thanks for the PR, but going to close this for now since I don't think these values are necessary to be included by default.

If they are useful for your project then I would recommend to add them to your own theme instead. 

Thanks!

PR #18287, opened 2025-06-11
Address No "exports" main defined in package json for vite

### LINKED ISSUE #16751: Vite Plugin Does Not Work With Storybook Vite

4.0.7

Storybook Vite (8.5.8)

v22.12.0

N/A

macOS

**Describe your issue**

Tried to integrate the Vite plugin with tailwind V4 in Storybook Vite (8.5.8) and hit this error:

```
SB_CORE-SERVER_0007 (MainFileEvaluationError): Storybook couldn't evaluate your .storybook/main.ts file.

Original error:
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/marcuswood/moonshot/platform/node_modules/@tailwindcss/vite/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:314:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:605:13)
    at resolveExports (node:internal/modules/cjs/loader:640:36)
```

Storybook config
```ts
import { StorybookConfig } from '@storybook/react-vite';
import tailwindcss from '@tailwindcss/vite';

const config: StorybookConfig = {
  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
  addons: ['@storybook/addon-essentials'],
  framework: '@storybook/react-vite',
  staticDirs: ['./public'],
  viteFinal: (config) => {
    config.plugins?.push(tailwindcss());

    return config;
  },
  core: {
    builder: '@storybook/builder-vite',
  },
};

export default config;
```

If I add this patch, everything works no problem:
```patch
diff --git a/package.json b/package.json
index b2ec5209a1ffc89586176fb2c1fec2f0577112a2..c906c6bf8025ab5cf59b63cfdf4443117b96c62e 100644
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.mts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "default": "./dist/index.mjs"
     }
   },
   "dependencies": {

```

Not sure why this is needed, I keep getting mixed up on the ESM stuff. Looks pretty harmless to add? Filing the issue in case other people run into it!

moshie:
## Summary

Package JSON is missing the main property

Fixes: #16751 

Resolves:

Screenshot 2025-06-11 at 16 34 19

## Test plan

I applied this fix directly in my local node modules and this fixed the error

### DIFF
--- a/packages/@tailwindcss-vite/package.json
+++ b/packages/@tailwindcss-vite/package.json
@@ -21,6 +21,7 @@
     "provenance": true,
     "access": "public"
   },
+  "main": "./dist/index.mjs",
   "exports": {
     ".": {
       "types": "./dist/index.d.mts",
--- a/packages/@tailwindcss-vite/package.json
+++ b/packages/@tailwindcss-vite/package.json
@@ -21,11 +21,11 @@
     "provenance": true,
     "access": "public"
   },
-  "main": "./dist/index.mjs",
   "exports": {
     ".": {
       "types": "./dist/index.d.mts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "default": "./dist/index.mjs",
     }
   },
   "dependencies": {
--- a/packages/@tailwindcss-vite/package.json
+++ b/packages/@tailwindcss-vite/package.json
@@ -25,7 +25,7 @@
     ".": {
       "types": "./dist/index.d.mts",
       "import": "./dist/index.mjs",
-      "default": "./dist/index.mjs",
+      "default": "./dist/index.mjs"
     }
   },
   "dependencies": {


packages/@tailwindcss-vite/package.json
```diff
@@ -24,7 +24,8 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.mts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "default": "./dist/index.mjs",
```
RodEsp: Since this is a `.json` file you can't have a trailing comma ([JSON spec](https://www.json.org))

```suggestion
      "default": "./dist/index.mjs"
```

moshie:
@RodEsp Good catch thanks, updated

moshie:
@philipp-spiess Sorry to ping you but this bug is affecting us and was wondering if you wouldn't mind reviewing this PR for me please?

thecrypticace:
Doesn't this imply that you *have* to use a version of node where `require(esm)` is enabled? 

afaik the only reason this would fail is because `require(…)` is being used instead of `import ""` or `await import(…)` to load a file which feels weird in this case since Vite is basically all in on ESM and Vite's CJS build has been deprecated for quite a while. Afaik it's going away in Vite 7 too so I'm curious to know how/why this is happening / still a problem.

moshie:
Yeah I get the feeling something is wrong in playwright's component testing which uses [Vite](https://vitejs.dev/) under the hood to create the components bundle and serve it.

```ts
// playwright-ct.config.ts
import { default as tailwindcss } from "@tailwindcss/vite";
import { defineConfig } from "@playwright/experimental-ct-react";

export default defineConfig({
	use: {
		ctViteConfig: {
			plugins: [react(), tailwindcss()],
		},
	},
});
```

I can see from the error tract that they are using the CJS loader.

I'll try and make time to create a reproduction which I'll share in a new issue

moshie:
So turns out the fix is to add `"type": "module"` to your package json file this forces playwright to build using esm

PR #18292, opened 2025-06-12
Feat: Adds rust config options to detect threading support in oxide

jlukic:
This modifies Oxide to use helpers for parallel processing that allow Oxide to compile to `unknown-unknown`. 

This lets Oxide work in the browser without dealing with the rabbit hole of permissions around using `SharedArrayBuffer` which include custom HTTP header requirements like `Cross-Origin-Embedder-Policy` and `Cross-Origin-Opener-Policy`

This is used in [tailwindcss-iso](https://github.com/Semantic-Org/tailwindcss-iso) and you can see [the wasm in use in Semantic Playground here](https://next.semantic-ui.com/examples/tailwind)

#11610

### DIFF
--- a/crates/oxide/Cargo.toml
+++ b/crates/oxide/Cargo.toml
@@ -3,13 +3,14 @@ name = "tailwindcss-oxide"
 version = "0.1.0"
 edition = "2021"
 
+[lib]
+crate-type = ["lib", "cdylib"]
+
 [dependencies]
 bstr = "1.11.3"
 globwalk = "0.9.1"
 log = "0.4.22"
-rayon = "1.10.0"
 fxhash = { package = "rustc-hash", version = "2.1.1" }
-crossbeam = "0.8.4"
 tracing = { version = "0.1.40", features = [] }
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 walkdir = "2.5.0"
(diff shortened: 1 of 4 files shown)


thecrypticace:
I'm planning on taking a look at this soon btw. One thing I definitely want to figure out is a way to unify the Node + WASM APIs (so like no separate `WasmChangedContent` or `WasmScanner` types and what not) so there's less duplication if I can.

thecrypticace:
Reminder for me to investigate soon: NAPI-RS v3 is out now. There's a chance things can be simplified with that re: WASM

jlukic:
Really good point it seems like that would make everything a lot simpler

https://napi.rs/blog/announce-v3

> WebAssembly is the biggest update this time. In V3, you can compile your project into WebAssembly with almost no code changes. If the compilation target is wasm32-wasip1-threads or higher, you can directly run code that uses Rust features like std::thread and tokio in the browser without any additional modifications.

attilah:
@thecrypticace @jlukic it would be nice to have beside NAPI-RS a C FFI interface (if it's not there yet) for .NET C# (and other) consumption as well.

RobinMalfait:
@attilah what would be your use case in the .NET C# world?

Tailwind CSS is a TypeScript project it is not entirely written in Rust.
The only place we use Rust (and NAPI-RS) is to traverse the file system and extract classes from your source files. This part is a separate package (`@tailwindcss/oxide`) and used by `@tailwindcss/vite`, `@tailwindcss/postcss` and `@tailwindcss/cli`.

attilah:
@RobinMalfait I am aware, the use case is make tailwind  compiler available for Blazor WASM, Blazor Server and "regular" .NET server side as well. After that you can build anything on it, developer tooling, WYSIWYG editor similarly what @jlukic posted as an example, but with Blazor.

I actually created the C FFI interface Yesterday and will try to create a PR for it during the weekend.

attilah:
@thecrypticace I tried to have a stab at unifying the types between node and wasm as you requested, please check it [here](https://github.com/attilah/tailwindcss/commit/727ac9732d4d911284526a930aa2e7396b0014aa) it is based on top of this PR, so @jlukic please chime in. 

The reason for this is that if I create a C FFI binding you'd probably expect the same to share types as much as possible.

attilah:
@RobinMalfait sorry I was not clear in the comment above, how would it be useful for .NET is that the Tailwind compiler would be invoked from a JS engine on the backend, to get compiled CSS and then post processed as needed.

PR #18337, opened 2025-06-18
Document `--watch=always` in the CLI

### LINKED ISSUE #1835: Document the `--watch=always` flag on the CLI
While trying to run the CLI as a subcommand within a project (running from Rust code), I encountered a challenge where the `tailwindcss` command would exit rather than watching and rebuilding, even with the `-w`/`--watch` flag. It turns out by default that if `stdin` is closed, `tailwindcss` will close too to avoid becoming a zombie process, and that you can turn off this behavior with the un-documented flag `--watch=always`. The code for this is here: https://github.com/tailwindlabs/tailwindcss/blob/next/packages/%40tailwindcss-cli/src/commands/build/index.ts#L218-L224

It would be nice if this flag were documented somewhere.

thecrypticace:
Closes https://github.com/tailwindlabs/tailwindcss.com/issues/1835

We had this in v3 as an undocumented option. We still have it in v4 but it's not documented under the CLI usage but should be. This PR adds this.

**before**
Screenshot 2025-06-18 at 09 41 40

**after**
Screenshot 2025-06-18 at 09 41 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Add heuristic to skip candidate migrations inside `emit(…)` ([#18330](https://github.com/tailwindlabs/tailwindcss/pull/18330))
+- Document `--watch=always` in the CLI's usage ([#18337](https://github.com/tailwindlabs/tailwindcss/pull/18337))
 
 ## [4.1.10] - 2025-06-11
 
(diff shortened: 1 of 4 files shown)


RobinMalfait (approved):
I like it

Little bit sad that it's 2 sentences because for all the other help text we don't have a `.` at the end 😅

thecrypticace:
Should we split the two sentences on two separate lines and line them up while dropping the punctuation?

RobinMalfait:
Nah we already do some wrapping when your terminal is smaller and then it would just become a bit weird. We could make it a single sentence maybe?

```
Watch for changes and rebuild as needed, use `always` to keep watching when stdin is closed
```
🙈

RobinMalfait (approved):

PR #18338, opened 2025-06-18
Extract candidates with variants in Clojure/ClojureScript keywords

eneroth:
## Summary

Taking a shot at fixing my own complaint from #18336. 

Added failing test and fix for extracting classes containing a pseudo-class from Clojure keywords. Eg., `:hover:text` -> `"hover:text"`. This would previously produce `["hover", "text"]`.

## Test plan

- Add a failing test. 
- Verify that it fails with `cargo test`. 
- Add fix. 
- Verify that `cargo test` has no further complaints. 

ATT: @RobinMalfait

### DIFF
--- a/crates/oxide/src/extractor/pre_processors/clojure.rs
+++ b/crates/oxide/src/extractor/pre_processors/clojure.rs
@@ -54,7 +54,7 @@ impl PreProcessor for Clojure {
                     // Keep the `.` as-is
                 }
 
-                // A `:` surrounded by letters denotes a pseudo-class. Keep as is.
+                // A `:` surrounded by letters denotes a variant. Keep as is.
                 //
                 // E.g.:
                 // ```
--- a/crates/oxide/src/extractor/pre_processors/clojure.rs
+++ b/crates/oxide/src/extractor/pre_processors/clojure.rs
@@ -42,7 +42,7 @@ impl PreProcessor for Clojure {
                     }
                 }
 
-                // A `.`  surrounded by digits is a decimal number, so we don't want to replace it.
+                // A `.` surrounded by digits is a decimal number, so we don't want to replace it.
                 //
                 // E.g.:
                 // ```
@@ -52,6 +52,18 @@ impl PreProcessor for Clojure {
                 b'.' if cursor.prev.is_ascii_digit() && cursor.next.is_ascii_digit() => {
 
                     // Keep the `.` as-is
+                }
+
+                // A `:` surrounded by letters denotes a pseudo-class. Keep as is.
+                //
+                // E.g.:
+                // ```
+                // lg:pr-6"
+                //   ^
+                // ``
+                b':' if cursor.prev.is_ascii_alphabetic() && cursor.next.is_ascii_alphabetic() => {
+
+                    // Keep the `:` as-is
                 }
 
                 b':' | b'.' => {
@@ -177,5 +189,17 @@ mod tests {
         "#;
 
         Clojure::test_extract_contains(input, vec!["flex", "gap-1.5", "p-1"]);
+    }
+
+    // https://github.com/tailwindlabs/tailwindcss/issues/18336
+    #[test]
+    fn test_extraction_of_pseudoclasses_from_keywords() {
+        let input = r#"
+            ($ :div {:class [:flex :first:lg:pr-6]} …)
+
+            :.hover:bg-white
+        "#;
+
+        Clojure::test_extract_contains(input, vec!["flex", "first:lg:pr-6", "hover:bg-white"]);
     }
 }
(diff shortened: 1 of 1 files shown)


thecrypticace:
o/ I'm fairly unfamiliar with ClojureScript (so maybe there's an obvious answer) but is there a reason to use keyword literals for class lists rather than strings?

thecrypticace:
Like to my knowledge you can't write classes like `2xl:flex` with that since keywords can't start with a number (if I'm reading the docs right anyway…)

eneroth:
@thecrypticace No worries! 

There are a lot of different reasons why keywords are preferred in Clojure, but the most basic one comes down to types: precisely because it's _not_ a string, but captures something symbolic. It disambiguates, in your editor, and at runtime, from things you actually do want to capture in strings: names, text, etc. So, a combination of being both general enough, and specific enough, makes it useful.

Other than that, keywords are interned: the keyword `:hello` and the keyword `:hello` are identical in memory, which has all kinds of implications.

A common DSL for writing HTML and CSS within the Clojure community is called Hiccup, and looks like this:
```clojure
[:p
 [:span.bg-white "Hello"]
 [:span#unique.text-red-500 "World"]]
```

The semantics are `:<tag>#<id>.<class-1>.<class-2>….<class-n>`.

As for numbers, the intention was that eg. `:2` wasn't supported, but a bug in an early Clojure version made it possible to use it. And Clojure goes to extreme lengths to never backwards compatibility, so it's in fact supported.

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -54,6 +54,18 @@ impl PreProcessor for Clojure {
                     // Keep the `.` as-is
                 }
 
+                // A `:` surrounded by letters denotes a pseudo-class. Keep as is.
```
thecrypticace: Is pseudo-class the wrong word here? Did you mean `variant` maybe?

eneroth: I don't know my nomenclature, I'm sure you're right.

thecrypticace:
Ah okay that's helpful — thanks! I left a comment. I *think* you mean't variant and not pseudo-class (e.g. `sm:dark:…` are variants in Tailwind CSS terminology. Sometimes they map to pseudo class but not always. (media queries, container queries, other selectors, etc…)

crates/oxide/src/extractor/pre_processors/clojure.rs
thecrypticace: No worries. I was a bit confused at first until I realized what you meant.

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -54,6 +54,18 @@ impl PreProcessor for Clojure {
...
                 }
 
+                // A `:` surrounded by letters denotes a variant. Keep as is.
+                //
+                // E.g.:
+                // ```
+                // lg:pr-6"
+                //   ^
+                // ``
+                b':' if cursor.prev.is_ascii_alphabetic() && cursor.next.is_ascii_alphabetic() => {
```
thecrypticace: I'd update this myself but git doesn't seem to want me to push to your branch.

```suggestion
                b':' if cursor.prev.is_ascii_alphanumeric() && cursor.next.is_ascii_alphanumeric() => {
```

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -178,4 +190,16 @@ mod tests {
 
         Clojure::test_extract_contains(input, vec!["flex", "gap-1.5", "p-1"]);
     }
+
+    // https://github.com/tailwindlabs/tailwindcss/issues/18336
+    #[test]
+    fn test_extraction_of_pseudoclasses_from_keywords() {
+        let input = r#"
+            ($ :div {:class [:flex :first:lg:pr-6]} …)
```
thecrypticace:
```suggestion
            ($ :div {:class [:flex :first:lg:pr-6 :first:2xl:pl-6 :group-hover/2:2xs:pt-6]} …)
```

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -178,4 +190,16 @@ mod tests {
...
+
+            :.hover:bg-white
+        "#;
+
+        Clojure::test_extract_contains(input, vec!["flex", "first:lg:pr-6", "hover:bg-white"]);
```
thecrypticace:
```suggestion
        Clojure::test_extract_contains(input, vec!["flex", "first:lg:pr-6", "first:2xl:pl-6", "group-hover/2:2xs:pt-6", "hover:bg-white"]);
```

thecrypticace (approved):

thecrypticace:
Thanks! This will be in the next release. You can use our insiders build before then to test it:
```
```

It may take about 15–30m before it's built and available on NPM.

eneroth:
Many thanks @thecrypticace!

PR #18345, opened 2025-06-19
Fix trailing `)` from interfering with extraction in Clojure keywords

### LINKED ISSUE #18344: Trailing `)` interferes with candidate extraction from Clojure/Script keywords

v4.1.10

postcss-cli 11.0.1

v24.2.0

Chrome, Safari, Firefox

macOS

**Reproduction URL**

No reproduction URL for this one.

**Describe your issue**

In Clojure, it seems that surrounding brackets interfere with the extraction of classes from keywords. E.g.:

```clojure 
(when condition :bg-white :bg-black)
```

The presence of `)` at the end seems to prevent `:bg-black` from being captured. I have a suggestion for how to fix this in the pre-processor, but I'm afraid that my limited experience of Rust will show, so I will kindly ask for your good input on the matter.

eneroth:
## Summary

In a form like, 

```clojure 
(if condition :bg-white :bg-black)
```

`:bg-black` will fail to extract, while `:bg-white` is extracted as expected. This PR fixes this case, implements more comprehensive candidate filtering, and supersedes a previous PR.

Having recently submitted a PR for handling another special case with Clojure keywords (the presence of `:` inside of keywords), I thought it best to invert the previous strategy: Instead of handling special cases one by one, consume keywords according to the Clojure reader spec. Consume nothing else, other than strings.

Because of this, this PR is a tad more invasive rather than additive, for which I apologize. The strategy is this:
- Strings begin with a `"` and ends with an unescaped `"`. Consume everything between these delimiters (existing case).
- Keywords begin with `:`, and end with whitespace, or one out of a small set of specific reserved characters. Everything else is a valid character in a keyword. Consume everything between these delimiters, and apply the class splitting previously contained in the outer loop. My previous special case handling of `:` inside of keywords in #18338 is now redundant (and is removed), as this is a more general solution.
- Discard _everything else_. 

I'm hoping that a strategy that is based on Clojure's definition of strings and keywords will pre-empt any further issues with edge cases.

Closes #18344.

## Test plan
- Added failing tests.
- `cargo test` -> failure
- Added fix
- `cargo test` -> success


thecrypticace:
@eneroth Can you see if the checkbox to allow maintainers to edit the PR is checked? I've got a few tweaks I want to make.

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -26,52 +35,81 @@ impl PreProcessor for Clojure {
...
-                // ``
-                b'.' if cursor.prev.is_ascii_digit() && cursor.next.is_ascii_digit() => {
+                // Consume keyword until a terminating character is reached.
+                b':' => {
+                    result[cursor.pos] = b' ';
 
-                    // Keep the `.` as-is
-                }
+                    while cursor.pos < len {
+                        cursor.advance();
```
thecrypticace: Doing this first could raise a potential panic on line 91, move to the end of the loop.

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -26,52 +35,81 @@ impl PreProcessor for Clojure {
...
-                //
-                // E.g.:
-                // ```
-                // gap-1.5
-                //      ^
-                // ``
-                b'.' if cursor.prev.is_ascii_digit() && cursor.next.is_ascii_digit() => {
+                // Consume keyword until a terminating character is reached.
+                b':' => {
+                    result[cursor.pos] = b' ';
```
thecrypticace, line 69: add a cursor.advance here

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -5,6 +5,14 @@ use bstr::ByteSlice;
 #[derive(Debug, Default)]
 pub struct Clojure;
 
+#[inline]
+fn is_keyword_terminator(byte: u8) -> bool {
+    matches!(
+        byte,
+        b'"' | b';' | b'@' | b'^' | b'`' | b'~' | b'(' | b')' | b'[' | b']' | b'{' | b'}' | b'\\'
+    ) || byte.is_ascii_whitespace()
+}
```
thecrypticace:
```suggestion
fn is_keyword_character(byte: u8) -> bool {
    matches!(
        byte,
        b'+' | b'-' | b'/' | b'*' | b'_' | b'#' | b'.' | b':' | b'?'
    ) | byte.is_ascii_alphanumeric()
}
```

I'd go for a positive check versus a negative one. Using `|` instead of `||` produces better, branchless assembly too: https://godbolt.org/z/94WqeEz1n

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -26,52 +35,81 @@ impl PreProcessor for Clojure {
...
+                            // E.g.:
+                            // ```
+                            // flex.gap-1.5
+                            //     ^
+                            // ```
+                            b'.' => {
+                                result[cursor.pos] = b' ';
+                            }
+                            // End of keyword.
+                            _ if is_keyword_terminator(cursor.curr) => {
```
thecrypticace:
```suggestion
                            _ if !is_keyword_character(cursor.curr) => {
```

In conjunction with the earlier comment.

thecrypticace:
Also could use a changelog entry

crates/oxide/src/extractor/pre_processors/clojure.rs
eneroth: Alright, but this excludes a lot of characters that are valid in keywords. Are you explicitly selecting the ones valid in Tailwind?

eneroth: Btw, the assembly inspector is cool!

eneroth:
Sorry @thecrypticace, from what I can see on [Github docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), there's supposed to be a checkbox for me to check here:

Screenshot 2025-06-28 at 08 32 05

But I can't see any. In lieu of that, rebased on main and amended with your changes, and added a change log entry.

thecrypticace:
Huh that's super weird. I wonder why it's missing. Thanks for making the changes!

crates/oxide/src/extractor/pre_processors/clojure.rs
thecrypticace: > Alright, but this excludes a lot of characters that are valid in keywords. Are you explicitly selecting the ones valid in Tailwind?

Kinda, yes. I tried finding a grammar for ClojureScript but I can't find any single one that "agree"s on what characters are actually allowed in a keyword. Our grammar for what's allowed in a utility name is fairly restricted *until* you open up an arbitrary value and `[]` aren't allowed in keywords. Probably need to add `%` to this list though if that's a valid keyword char.

Is there a good source that can detail what's actually allowed?

> Btw, the assembly inspector is cool!

Compiler explorer is so useful — such an awesome tool.

eneroth: Yeah, `%` is allowed. Unfortunately, the docs don't contain an exhaustive list of allowed characters, and the authoritative list is the implementation.

I've looked at the code and verified against with Claude to get a list of valid characters other than alphanumerics. I've also evaluated the monstrosity `:!#$%&'*+-.d/:<=>?_|` locally in the REPL to verify this, and it works.

```
! (33) - exclamation
# (35) - hash/pound
$ (36) - dollar
% (37) - percent
& (38) - ampersand
' (39) - single quote
* (42) - asterisk
+ (43) - plus
- (45) - hyphen/minus
. (46) - period/dot
/ (47) - forward slash
: (58) - colon
< (60) - less than
= (61) - equals
> (62) - greater than
? (63) - question mark
_ (95) - underscore
| (124) - pipe/vertical bar
```

Nevertheless, there's a point to your idea of sticking with characters valid in Tailwind, as this eliminates keywords that happen to be valid in Clojure, but are definitely not valid Tailwind. It might also be somewhat less future proof, if Tailwind happens to start using a char in the future that isn't currently being used.

crates/oxide/src/extractor/pre_processors/clojure.rs
```diff
@@ -5,6 +5,14 @@ use bstr::ByteSlice;
 #[derive(Debug, Default)]
 pub struct Clojure;
 
+#[inline]
+fn is_keyword_character(byte: u8) -> bool {
+    matches!(
+        byte,
+        b'+' | b'-' | b'/' | b'*' | b'_' | b'#' | b'.' | b':' | b'?'
+    ) | byte.is_ascii_alphanumeric()
+}
```
thecrypticace:
```suggestion
/// This is meant to be a rough estimate of a valid ClojureScript keyword
///
/// This can be approximated by the following regex:
/// /::?[a-zA-Z0-9!#$%&*+./:<=>?_|-]+/
///
/// However, keywords are intended to be detected as utilities. Since the set
/// of valid characters in a utility (outside of arbitrary values) is smaller,
/// along with the fact that neither `[]` nor `()` are allowed in keywords we
/// can simplify this list quite a bit.
#[inline]
fn is_keyword_character(byte: u8) -> bool {
    return matches!(
        byte,
        b'!' | b'%' | b'*' | b'+' | b'-' | b'.' | b'/' | b':' | b'_'
    ) | byte.is_ascii_alphanumeric();
}
```

I updated this list to:
- Include `!` and `%` since they should be there
- Drop `#` and `?` as they can't be used in a named utility / variant today
- Order them in byte value order (not really needed just did it anyway)

crates/oxide/src/extractor/pre_processors/clojure.rs
thecrypticace: Ah yeah that's quite a list. Many of those can't be used in a named utility so can be omitted. I think it's reasonable to keep a list of these chars so I've added a suggestion with a regex, some explanation about how the list was chosen, and a tweaked list of characters.

crates/oxide/src/extractor/pre_processors/clojure.rs
thecrypticace: If this list looks good to you this should hopefully be the last change and can get this merged. I don't think I missed anything but lemme know if I did.

crates/oxide/src/extractor/pre_processors/clojure.rs
eneroth: Then to the best of my knowledge, we're complete.

thecrypticace:
Ah weird ` | b'#'` being gone makes the tests fail. I guess we can keep it in for the time being then. 🤔 

edit: I know why. It's a bit annoying. We need to read whole keywords *then* split them into utilities. I wonder if this means we need an expanded character set no matter what.

Can do that as a followup bug fix if we turn out to need that.

thecrypticace (approved):

thecrypticace:
I'm gonna go head and merge this so I can get it in. Will fix the failing `#` test in a followup commit.

eneroth:
@thecrypticace Ah yeah, because of the syntax `<elem>#<id>.<class-1>.<class-2>…<class-n>`. This is unfortunately a requirement  for Clojure for the [Hiccup DSL](https://github.com/weavejester/hiccup/wiki/Syntax#css-style-sugar) to work.

thecrypticace:
Yeah, it makes sense. We just can't be quite as selective about certain characters during the pre-processing stage which is fine.

yannvanhalewyn:
Ironically in v4.1.12 extraction of classes in symbol lists were broken. I'm not 💯 sure it's due to this PR but something to keep in mind.

```clojure
[:div {:class '[class-1 class-2
                class-3 class-4]}]
```

eneroth:
@yannvanhalewyn Goodness, I wasn't even aware that classes as symbols were a thing, sorry if that was due to my intervention.

Doesn't symbols in that manner make conditional logic etc. tricky?

yannvanhalewyn:
@eneroth no worries it happens :). We've pinned back to `4.1.11` until this works again.

It composes pretty well for conditional logic:

```clojure
(def hl-class-names '[ring ring-blue-500])

[:div
 {:class (cond-> '[input w-full]
           textarea? (conj 'textarea)
           (seq errors) (concat '[border-red-500 bg-red-100])
           highlight? (concat hl-class-names))}]
```

Another benefit is it allows splitting long classlists over multiple lines, and introducing logical grouping:

```clojure
[:div
 {:class '[h-100 lg:h-200 max-w-32 mx-auto py-60
           flex flex-col justify-end items-center
           lg:flex-row lg:justify-between
           bg-cover bg-center bg-no-repeat rounded-3xl overflow-hidden
           font-semibold text-gray-900]}]
```

Note: slashes make them invalid tokens, those need to be explicit strings: `'[p-32 "text-black/50"]`

Issue #17295, opened 2025-03-20
Angular 19 cannot automatically recover using @apply with Tailwind CSS 4


micobarac:
Tailwind CSS: 4.0.14
Angular: 19.2.4
Node: 23.6.0
Chrome: 134.0.6998.89 (Official Build) (arm64)
MacOS: Sequoia 15.3.2 (24D81)

**Reproduction URL**

[https://github.com/micobarac/angular19](https://github.com/micobarac/angular19)

**Describe your issue**

Reproducing the issue:

- run `yarn install && yarn start`
- open `src/tailwind.css`
- change `bg-white` class within `@apply` directive to `bg-unknown` at line 4
- observe the error: 

```
✘ [ERROR] Cannot apply unknown utility class: bg-unknown [plugin angular-css]

    angular:styles/global:styles:1:8:
      1 │ @import 'src/styles.css';
```

- change `bg-unknown` class within `@apply` directive back to `bg-white` at line 4
- observe the error (**Angular should have recovered by now**): 

```
✘ [ERROR] Cannot apply unknown utility class: bg-unknown [plugin angular-css]

    angular:styles/global:styles:1:8:
      1 │ @import 'src/styles.css';
```

I can't tell if this problem resides inside `Angular 19` or `Tailwind CSS 4`.

I opened issue here:

[https://github.com/angular/angular-cli/issues/29789](https://github.com/angular/angular-cli/issues/29789)

[https://github.com/angular/angular-cli/issues/29789#issuecomment-2737483148](https://github.com/angular/angular-cli/issues/29789#issuecomment-2737483148)

_In Tailwind CSS version 4, syntax errors trigger an exception instead of a PostCSS error. This causes PostCSS to crash, preventing it from tracking the files Tailwind processes. As a result, these files aren't added to the file watcher / dependency tree, leading to issues where cached compilation results aren't invalidated when a file changes, since it isn't recognized as a transitive dependency._

I opened discussion here, too:

[https://github.com/tailwindlabs/tailwindcss/discussions/17077](https://github.com/tailwindlabs/tailwindcss/discussions/17077)

Shuriken933:
Same problem with Nuxt (vue js)

micobarac:
What's worse, Angular team do not plan to fix this, stating:

_Closing as there isn’t much we can do here.
Tailwind is throwing an exception that causes the underlying PostCSS instance to crash and not return which files to watch._

DairoCantillo:
[...]

PR #18373, opened 2025-06-23
Re-throw errors from PostCSS nodes

### LINKED ISSUE #18370: Angular: Error in PostCSS + tailwind doesn't break the build
With Angular and Postcss + Tailwind, when I have an error in a css file, it displays a big red error in the logs, but the build continue and is done successfuly.

The expected behavior is it should break the build, so you don't ship bugged CSS. The error seems due to tailwind not reporting correctly the error to postcss.

**Reproduction:**

- create a new angular project with default options.
- npm install tailwindcss @tailwindcss/postcss postcss --force
- create the .postcssrc.json with 
```
{
  "plugins": {
    "@tailwindcss/postcss": {}
  }
}
```
- add the follwing in the css of a component `@reference "not-found.css";`
- npm run build

The error is displayed but the build continue and is successful.

**Version:**

"@tailwindcss/postcss": "4.1.10"
"postcss": "8.5.6"
Angular CLI: 20.0.3
Node: 20.19.2
Package Manager: npm 11.4.1
OS: linux x64

thecrypticace:
Fixes #18370


thecrypticace:
As far as I can tell from the original reproduction in #17295 this still doesn't help.

From this comment https://github.com/angular/angular-cli/issues/29789#issuecomment-2737483148

This issue was related to us not throwing a PostCSS error but even if we do that it doesn't work.

This will take some more investigation.

integrations/postcss/index.test.ts
meet4171, line 1: nice work

thecrypticace:
I found the Angular bug.

Leaving me a note with the reduced CSS plugin so I can file a bug report:

```js
/*
 * Commentary:
 * This plugin demonstrates a rebuild bug in Angular 19
 *
 * The setup consists of three files:
 *
 * File 1: styles.css:
 * ```css
 * @import "./imported.css";
 * ```
 *
 * File 2: imported.css:
 * ```css
 * @--throw-me;
 * ```
 *
 * File 3: This plugin
 *
 * Reproduction:
 *
 * # Build
 * - `npm run start`
 *
 * # Cause an error
 * - `gsed -i 's/@--throw-me/@throw-me/' src/imported.css`
 *
 * # Revert the error and see it's still broken
 * - `gsed -i 's/@throw-me/@--throw-me/' src/imported.css`
 *
 * # Try editing it again and see that it doesn't rebuild
 * - `gsed -i 's/@--throw-me/@--throw-me-2/' src/imported.css`
 *
 * # To recover you *must* edit the "root" stylesheet
 *
 * - `echo '\n.test{}' > src/styles.css`
 */

module.exports = function tailwindcss() {
  return {
    postcssPlugin: "tailwindcss",
    async Once(root, { result }) {
      console.log("postcss called for", result.opts.from);

      // Precondition:
      //
      // We only care about compiles to `styles.css` for this bug report
      if (!result.opts.from?.endsWith("styles.css")) return;

      // Note:
      //
      // Removing the `@import` prevents this plugin from being called for
      // `imported.css` too. This does not affect reproduction of the bug.
      root.removeAll();

      // Requirement 1:
      //
      // `styles.css` must emit a dependency on a file that it also `@import`s
      // during a successful build.
      let path = require("path");
      let importedPath = path.resolve(path.dirname(result.opts.from), "./imported.css");
      result.messages.push({
        type: "dependency",
        plugin: "my-plugin",
        parent: result.opts.from,
        file: importedPath,
      });

      // Requirement 2:
      //
      // `styles.css` must throw an error after a successful rebuild due to
      // a change in an `@import`-ed stylesheet.
      //
      // We'll do this by checking the content directly as we're not using
      // `postcss-import` or anything like that here.
      //
      // Trigger the error by changing `@--throw-me` to `@throw-me`.
      let fs = require("fs");
      let importedFile = fs.readFileSync(importedPath, "utf-8");
      if (importedFile.includes("@throw-me")) throw root.error("ohno");
    },
  };
};

module.exports.postcss = true;
```

packages/@tailwindcss-postcss/src/ast.test.ts
```diff
@@ -84,7 +84,7 @@ it('should convert a Tailwind CSS AST into a PostCSS AST', () => {
   `
 
   let ast = parse(input)
-  let transformedAst = cssAstToPostCssAst(ast)
+  let transformedAst = cssAstToPostCssAst(ast, undefined, postcss)
```
RobinMalfait:
```suggestion
  let transformedAst = cssAstToPostCssAst(postcss, ast)
```

packages/@tailwindcss-postcss/src/ast.ts
```diff
@@ -1,20 +1,18 @@
...
 
 const EXCLAMATION_MARK = 0x21
 
-export function cssAstToPostCssAst(ast: AstNode[], source: PostcssSource | undefined): PostCssRoot {
-  let inputMap = new DefaultMap<Source, Input>((src) => {
-    return new Input(src.code, {
+export function cssAstToPostCssAst(
+  ast: AstNode[],
+  source: postcss.Source | undefined,
+  postcss: postcss.Postcss,
```
RobinMalfait:
```suggestion
  postcss: postcss.Postcss,
  ast: AstNode[],
  source: postcss.Source | undefined,
```

packages/@tailwindcss-postcss/src/index.ts
```diff
@@ -310,7 +310,7 @@ function tailwindcss(opts: PluginOptions = {}): AcceptedPlugin {
               } else {
                 // Convert our AST to a PostCSS AST
                 DEBUG && I.start('Transform Tailwind CSS AST into PostCSS AST')
-                context.cachedPostCssAst = cssAstToPostCssAst(tailwindCssAst, root.source)
+                context.cachedPostCssAst = cssAstToPostCssAst(tailwindCssAst, root.source, postcss)
```
RobinMalfait:
```suggestion
                context.cachedPostCssAst = cssAstToPostCssAst(postcss, tailwindCssAst, root.source)
```

RobinMalfait (commented):
Looking good, but I think this small API would make it even better.

RobinMalfait (approved):

PR #18404, opened 2025-06-27
Don't consider the global important state in `@apply`

### LINKED ISSUE #18400: Using important automatically applies it to utilites set via @apply in v4
v4.1.11

Angular and NX, however it can be repro'd in the playground

v24.2.0

Zen (runs on a Firefox base)

macOS

**Reproduction URL**

https://play.tailwindcss.com/kM1swlss8C Minimal example, but shows the issue that padding utilities do not do anything anymore because the `@apply` is enforcing `!important` when it didn't use to in v3

**Describe your issue**

When importing tailwind and making use of important (i.e. `@import "tailwindcss" important`), v4 seems to automatically apply `!important` to any class that is created with `@apply`.

This did not happen in v3 which allowed me to decide whether I needed it or not when using `@apply`, and simultaneously ensuring that when a utility class is used in CSS (i.e. `p-4`) it overrode any CSS that came before it. This is especially important as I use Angular Material, which forces their styles into the `<head>` earlier than application styles causing them to always win in a priority competition, and this behaviour cannot be changed.

Important being forced on `@apply` rules messes with complex custom classes (if you've ever tried to customise the browser's built-in slider input, you'll understand my pain), wherein instead of having my class reference a CSS variable and have that variable have sensible defaults that can simply be overridden with an inline style, i.e. 
```css
// slider.css
// exerpt from a much large class that has to separately handle mozilla and webkit, because sliders SUCK
.slider {
    // Mozilla track gradient (simulates a "filling" bar)
        @apply [&::-moz-range-track]:bg-linear-to-r [&::-moz-range-track]:from-(--left-bar) [&::-moz-range-track]:via-(--right-bar) [&::-moz-range-track]:to-(--right-bar);
        // Mozilla track value setter (allows 1 variable to control both pats of the fill)
        @apply [&::-moz-range-track]:from-(percentage:--value) [&::-moz-range-track]:via-(percentage:--value) [&::-moz-range-track]:to-100%;
}
```

```html
// some-component.html
<input type="range" max="100" class="slider" [style.--value.%]="value"  [value]="value" />
```

With v4 now I am forced to set the variable with `!important` in order to override the default value I have set at 50%

thecrypticace:
Fixes #18400

In v3 when you used `important: true` it did not affect `@apply`. However, in v4 it does and there's no way to make it *not*. This is definitely a bug and would be unexpected for users coming from v3 who use `@apply` and `important` together.

Basically, the following code, along with the detected utility `flex` in source files…

```css
@import 'tailwindcss/utilities' important;
.flex-explicitly-important {
  @apply flex!;
}
.flex-not-important {
  @apply flex;
}
```

… would output this:
```css
.flex {
  display: flex !important;
}
.flex-explicitly-important  {
  display: flex !important;
}
.flex-not-important {
  display: flex !important;
}
```

But it's expected that `@apply` doesn't consider the "global" important state. This PR addresss this problem and now the output is this:

```css
.flex {
  display: flex !important;
}
.flex-explicitly-important  {
  display: flex !important;
}
.flex-not-important {
  display: flex; /* this line changed */
}
```

If you want to mark a utility as important in `@apply` you can still use `!` after the utility to do so as shown above.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
@@ -2,7 +2,7 @@ import { substituteAtApply } from '../../../../tailwindcss/src/apply'
 import { atRule, styleRule, toCss, walk, type AstNode } from '../../../../tailwindcss/src/ast'
 import { printArbitraryValue } from '../../../../tailwindcss/src/candidate'
 import * as SelectorParser from '../../../../tailwindcss/src/compat/selector-parser'
-import type { DesignSystem } from '../../../../tailwindcss/src/design-system'
+import { CompileAstFlags, type DesignSystem } from '../../../../tailwindcss/src/design-system'
 import { ThemeOptions } from '../../../../tailwindcss/src/theme'
 import { DefaultMap } from '../../../../tailwindcss/src/utils/default-map'
 import { isValidSpacingMultiplier } from '../../../../tailwindcss/src/utils/infer-data-type'
@@ -40,7 +40,16 @@ export const computeUtilitySignature = new DefaultMap<
       // Use `@apply` to normalize the selector to `.x`
       let ast: AstNode[] = [styleRule('.x', [atRule('@apply', utility)])]
 
-      temporarilyDisableThemeInline(designSystem, () => substituteAtApply(ast, designSystem))
+      temporarilyDisableThemeInline(designSystem, () => {
+        // There's separate utility caches for respect important vs not
+        // so we want to compile them both with `@theme inline` disabled
+        for (let candidate of designSystem.parseCandidate(utility)) {
+          designSystem.compileAstNodes(candidate, CompileAstFlags.None)
+          designSystem.compileAstNodes(candidate, CompileAstFlags.RespectImportant)
+        }
+
+        substituteAtApply(ast, designSystem)
+      })
 
       // We will be mutating the AST, so we need to clone it first to not affect
       // the original AST
--- a/packages/tailwindcss/src/compile.ts
+++ b/packages/tailwindcss/src/compile.ts
@@ -9,7 +9,7 @@ import {
   type StyleRule,
 } from './ast'
 import { type Candidate, type Variant } from './candidate'
-import { type DesignSystem } from './design-system'
+import { CompileAstFlags, type DesignSystem } from './design-system'
 import GLOBAL_PROPERTY_ORDER from './property-order'
 import { asColor, type Utility } from './utilities'
 import { compare } from './utils/compare'
@@ -19,7 +19,10 @@ import type { Variants } from './variants'
 export function compileCandidates(
   rawCandidates: Iterable<string>,
   designSystem: DesignSystem,
-  { onInvalidCandidate }: { onInvalidCandidate?: (candidate: string) => void } = {},
+  {
+    onInvalidCandidate,
+    respectImportant,
+  }: { onInvalidCandidate?: (candidate: string) => void; respectImportant?: boolean } = {},
 ) {
   let nodeSorting = new Map<
     AstNode,
@@ -44,6 +47,12 @@ export function compileCandidates(
     matches.set(rawCandidate, candidates)
   }
 
+  let flags = CompileAstFlags.None
+
+  if (respectImportant || respectImportant === undefined) {
+    flags |= CompileAstFlags.RespectImportant
+  }
+
   let variantOrderMap = designSystem.getVariantOrder()
 
   // Create the AST
@@ -51,7 +60,7 @@ export function compileCandidates(
     let found = false
 
     for (let candidate of candidates) {
-      let rules = designSystem.compileAstNodes(candidate)
+      let rules = designSystem.compileAstNodes(candidate, flags)
       if (rules.length === 0) continue
 
       found = true
@@ -119,9 +128,15 @@ export function compileCandidates(
   }
 }
 
-export function compileAstNodes(candidate: Candidate, designSystem: DesignSystem) {
+export function compileAstNodes(
+  candidate: Candidate,
+  designSystem: DesignSystem,
+  flags: CompileAstFlags,
+) {
   let asts = compileBaseUtility(candidate, designSystem)
   if (asts.length === 0) return []
+
+  let respectImportant = Boolean(flags & CompileAstFlags.RespectImportant)
 
   let rules: {
     node: AstNode
@@ -136,7 +151,10 @@ export function compileAstNodes(candidate: Candidate, designSystem: DesignSystem
   for (let nodes of asts) {
     let propertySort = getPropertySort(nodes)
 
-    if (candidate.important || designSystem.important) {
+    // If the candidate itself is important then we want to always mark
+    // the utility as important. However, at a design system level we want
+    // to be able to opt-out when using things like `@apply`
+    if (candidate.important || (designSystem.important && respectImportant)) {
       applyImportant(nodes)
     }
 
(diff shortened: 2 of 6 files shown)


packages/tailwindcss/src/design-system.ts
```diff
@@ -18,6 +18,11 @@ import { DefaultMap } from './utils/default-map'
 import { extractUsedVariables } from './utils/variables'
 import { Variants, createVariants } from './variants'
 
+export const enum CompileAstFlags {
+  None = 0,
+  RespectImportant = 1 << 0,
```
RobinMalfait, line 23: Blast from the past 🥲

packages/tailwindcss/src/index.test.ts
```diff
@@ -666,7 +666,7 @@ describe('@apply', () => {
   })
 
   // https://github.com/tailwindlabs/tailwindcss/issues/16935
-  it('should now swallow @utility declarations when @apply is used in nested rules', async () => {
+  it('should not swallow @utility declarations when @apply is used in nested rules', async () => {
```
RobinMalfait, line 669: 😂

packages/tailwindcss/src/compile.ts
```diff
@@ -44,14 +47,20 @@ export function compileCandidates(
     matches.set(rawCandidate, candidates)
   }
 
+  let flags = CompileAstFlags.None
+
+  if (respectImportant || respectImportant === undefined) {
```
RobinMalfait:
```suggestion
  if (respectImportant ?? true) {
```

Too cute? 😂

packages/tailwindcss/src/compile.ts
```diff
@@ -136,7 +151,10 @@ export function compileAstNodes(candidate: Candidate, designSystem: DesignSystem
   for (let nodes of asts) {
     let propertySort = getPropertySort(nodes)
 
-    if (candidate.important || designSystem.important) {
+    // If the candidate itself is important then we want to always mark
+    // the utility as important. However, at a design system level we want
+    // to be able to opt-out when using things like `@apply`
+    if (candidate.important || (designSystem.important && respectImportant)) {
```
RobinMalfait: What about hoisting this check?
```suggestion
    if (candidate.important || respectImportant) {
```

packages/tailwindcss/src/compile.ts
```diff
@@ -119,10 +128,16 @@ export function compileCandidates(
...
-export function compileAstNodes(candidate: Candidate, designSystem: DesignSystem) {
+export function compileAstNodes(
+  candidate: Candidate,
+  designSystem: DesignSystem,
+  flags: CompileAstFlags,
+) {
   let asts = compileBaseUtility(candidate, designSystem)
   if (asts.length === 0) return []
 
+  let respectImportant = Boolean(flags & CompileAstFlags.RespectImportant)
```
RobinMalfait: If we move the `designSystem.important` check here, then we don't need to recompute the `designSystem.import && respectImportant` in the loop below.

Bonus points: it would short-circuit and not even bother with the flags checking when not using the `important` flag on the import.

```suggestion
  let respectImportant = designSystem.important && Boolean(flags & CompileAstFlags.RespectImportant)
```

packages/tailwindcss/src/design-system.ts
```diff
@@ -57,26 +62,28 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
...
-        designSystem,
-      )
-    } catch (err) {
-      // If substitution fails then the candidate likely contains a call to
-      // `theme()` that is invalid which may be because of incorrect usage,
-      // invalid arguments, or a theme key that does not exist.
-      return []
-    }
+  let compiledAstNodes = new DefaultMap<number>((flags) => {
+    return new DefaultMap<Candidate>((candidate) => {
```
RobinMalfait, line 66: 😍

Having a simple number for the flags is very nice because we can use it as an efficient key of the map. Love it!

packages/tailwindcss/src/design-system.ts
```diff
@@ -134,8 +141,10 @@ export function buildDesignSystem(theme: Theme): DesignSystem {
     parseVariant(variant: string) {
       return parsedVariants.get(variant)
     },
-    compileAstNodes(candidate: Candidate) {
-      return compiledAstNodes.get(candidate)
+    compileAstNodes(candidate: Candidate, flags?: CompileAstFlags) {
+      flags ??= CompileAstFlags.RespectImportant
```
RobinMalfait:
```suggestion
    compileAstNodes(candidate: Candidate, flags = CompileAstFlags.RespectImportant) {
```

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
```diff
@@ -40,7 +40,16 @@ export const computeUtilitySignature = new DefaultMap<
...
       let ast: AstNode[] = [styleRule('.x', [atRule('@apply', utility)])]
 
-      temporarilyDisableThemeInline(designSystem, () => substituteAtApply(ast, designSystem))
+      temporarilyDisableThemeInline(designSystem, () => {
+        // There's separate utility caches for respect important vs not
+        // so we want to compile them both with `@theme inline` disabled
+        for (let candidate of designSystem.parseCandidate(utility)) {
+          designSystem.compileAstNodes(candidate, CompileAstFlags.None)
+          designSystem.compileAstNodes(candidate, CompileAstFlags.RespectImportant)
+        }
```
RobinMalfait, line 49: I'm trying to understand why this is needed 🤔 

Shouldn't the `substituteAtApply` from below pick from the "correct" one?

RobinMalfait: A test does fail without this so... 😅

RobinMalfait (commented):
I think this makes sense, just a few suggestions.

I tested this on Catalyst and Tailwind Plus to make sure the output is identical (especially because we don't use `@apply` there). And the output is the same 👍

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
thecrypticace: It's because the precomputing happens with `CompileAstFlags.None` (because of using `@apply`). Later when we do candidatesToCss it uses the default `CompileAstFlags.RespectImportant` which wasn't cached so the signatures are computed lazily then which ends up respecting `@theme inline` even when we've turned it off.

packages/tailwindcss/src/design-system.ts
thecrypticace: 😂

packages/tailwindcss/src/compile.ts
thecrypticace: nah, I like it

packages/tailwindcss/src/compile.ts
thecrypticace: Done

packages/tailwindcss/src/compile.ts
thecrypticace: Good idea

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
RobinMalfait: Ah fair yep makes sense

packages/tailwindcss/src/design-system.ts
thecrypticace: Yeah, I originally had a `boolean` here and was like… I kinda want this to be a *bit* more configurable if we really need to and it feels like it makes it clearer that this is an "options" / "flags" argument.

RobinMalfait (approved):

PR #18432, opened 2025-07-02
Detect classes inside Elixir charlist, word list, and string sigils

### LINKED ISSUE #18431: Tailwind 4 doesn't detect classes in parenthesis

v4.1.7

**What's the issue?**

In elixir, a list of strings can be generated by writing e.g. `~W(bg-red-500 text-blue-500)`. This is equivalent to writing `["bg-red-500", "text-blue-500"]`. An example of this behavior to define a list of predefined classes to go into HTML components can be seen [here](https://github.com/maxmarcon/live_select/blob/6bea2e0310414cfca5e22180fd614db358cc6e1c/lib/live_select/component.ex#L53). Had those lines been written using the `["bg-red-500", "text-blue-500"]` or even `~W( bg-red-500 text-blue-500 )`, the tailwind tool would've picked up the styles correctly. The first style (`~W(bg-red-500 text-blue-500)`) also used to work correctly with Tailwind 3, but after upgrading to Tailwind 4, it doesn't seem to work anymore.

In short, after upgrading to Tailwind 4, it seems like the tokenizer that searches for tailwind classes includes parenthesis as part of a token, meaning that writing `~W(bg-red-500 text-blue-500)` in my source code means that Tailwind will no longer pick up those classes. This behavior has changed from Tailwind 3.

Was this change on purpose? I don't see any mention in the release notes or upgrade-guide about the tokenization-process changing, but perhaps I'm missing something.

See also these issues: https://github.com/maxmarcon/live_select/issues/118, https://github.com/phoenixframework/phoenix/issues/6331

thecrypticace:
Fixes #18431

The `~W(…)` syntax in Elixir is a _sigil_. It's basically another way to write strings, arrays of strings, etc… The syntax lets you use a handful of surrounding brackets like `~W(…)`, `~W[…]`, `~W{…}`, `~W"…"`, etc… to let you write lists without necessarily having to escape characters.

In v3 our extractor was able to pick these up but in v4 Oxide does not. I've added a preprocessor for Elixir files so we can modify the code before our main extractor sees it.

Now things like this: `~W(text-white bg-gray-600)` will get turned into ` ~W text-white bg-gray-600 ` which can easily be processed by our extractor.

The sigils we support are:
- `~s` / `~S` (strings)
- `~w` / `~W` (word lists)
- `~c` / `~C` (charlists)

We're specifically detecting the use of `(…)`, `[…]`, and `{…}` as using quotes already works today.

### DIFF
--- a/crates/oxide/src/extractor/pre_processors/elixir.rs
+++ b/crates/oxide/src/extractor/pre_processors/elixir.rs
@@ -1,0 +1,150 @@
+use std::str::from_utf8;
+
+use crate::cursor;
+use crate::extractor::bracket_stack::BracketStack;
+use crate::extractor::pre_processors::pre_processor::PreProcessor;
+
+#[derive(Debug, Default)]
+pub struct Elixir;
+
+impl PreProcessor for Elixir {
+    fn process(&self, content: &[u8]) -> Vec<u8> {
+        let mut cursor = cursor::Cursor::new(&content);
+        let mut result = content.to_vec();
+        let mut bracket_stack = BracketStack::default();
+
+        while cursor.pos < content.len() {
+            // Look for a sigil marker
+            if cursor.curr != b'~' {
+                cursor.advance();
+                continue;
+            }
+
+            // Scan charlists, strings, and wordlists
+            if !matches!(cursor.next, b'c' | b'C' | b's' | b'S' | b'w' | b'W') {
+                cursor.advance();
+                continue;
+            }
+
+            cursor.advance_twice();
+
+            // Match the opening for a sigil
+            if !matches!(cursor.curr, b'(' | b'[' | b'{') {
+                continue;
+            }
+
+            // Replace the opening bracket with a space
+            result[cursor.pos] = b' ';
+
+            // Scan until we find a balanced closing one and replace it too
+            bracket_stack.push(cursor.curr);
+
+            while cursor.pos < content.len() {
+                cursor.advance();
+
+                match cursor.curr {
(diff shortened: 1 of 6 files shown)


thecrypticace:
@RobinMalfait if you could take a peek at this at some point. I think I did it right 😅

crates/oxide/src/extractor/pre_processors/elixir.rs
```diff
@@ -0,0 +1,150 @@
+use std::str::from_utf8;
+
```
RobinMalfait:
```suggestion
```

crates/oxide/src/extractor/pre_processors/elixir.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::cursor;
+use crate::extractor::bracket_stack::BracketStack;
+use crate::extractor::pre_processors::pre_processor::PreProcessor;
+
+#[derive(Debug, Default)]
+pub struct Elixir;
+
+impl PreProcessor for Elixir {
+    fn process(&self, content: &[u8]) -> Vec<u8> {
+        let mut cursor = cursor::Cursor::new(&content);
```
RobinMalfait:
```suggestion
        let mut cursor = cursor::Cursor::new(content);
```

RobinMalfait (approved):
Apart from these 2 suggestions, looks good!

The `Elixir::test` already makes sure that we keep the same length so that's already covered for us.

PR #18485, opened 2025-07-08
fix: allow `process.env.DEBUG` to be a boolean in `@tailwindcss/node`

okxiaoliang4:
TanStack Start build to `cloudflare-module`, `debug` value type is boolean.
reference: https://developers.cloudflare.com/workers/framework-guides/web-apps/tanstack/

## Summary

image


thecrypticace:
This seems like a bug in `cloudflare-module`? afaik `process.env.*` should always be strings. The types for Node seem to indicate this:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f8a95123ad42b8d1c06927c1bfac4fb9a13bf345/types/node/v22/process.d.ts#L346-L351

thecrypticace:
Discussed this a bit internally and while it feels like a bug that process.env.* isn't a string we're gonna pull this one in anyway.

thecrypticace (approved):

thecrypticace:
Aside: the screenshot is compiled v3 code — not v4. We could make this change for v3 as well if absolutely necessary.

PR #18511, opened 2025-07-11
Center the dropdown icon added to an input with a paired datalist in Chrome

### LINKED ISSUE #18499: datalist marker

^4.1.10

Next.js 15.3.4

v24.3.0

Chrome

Linux

**Describe your issue**

the marker is not centered - it depends on the font family

Image

Image

 Temporary solution:

```css
    input::-webkit-calendar-picker-indicator {
        font: initial;
    }
```

thecrypticace:
This PR tweaks the dropdown arrow added to an input by Chrome when it has a `list` attribute pointing to a `<datalist>`.

Right now the arrow isn't centered vertically:

Screenshot 2025-07-14 at 15 41 50

The cause of this is the line height being inherited into the pseudo element which controls how the marker is positioned. I *think* this is because it's being drawn with unicode symbols but I'm not sure. It could just be from the `list-item` display.

After this PR changes the line height its centered again:

Screenshot 2025-07-14 at 15 42 05

Some notes:

This only affects Chrome and also does not appear to cause issues for date/time inputs. While weird that this pseudo is the one used for a `<datalist>` marker it is indeed correct.

Fixes #18499

Can use this Play to test the change: https://play.tailwindcss.com/jzT35CRpr0


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix trailing `)` from interfering with extraction in Clojure keywords ([#18345](https://github.com/tailwindlabs/tailwindcss/pull/18345))
 - Detect classes inside Elixir charlist, word list, and string sigils ([#18432](https://github.com/tailwindlabs/tailwindcss/pull/18432))
 - Track source locations through `@plugin` and `@config` ([#18345](https://github.com/tailwindlabs/tailwindcss/pull/18345))
+- Center the `::marker` added to an input with a paired datalist ([#18511](https://github.com/tailwindlabs/tailwindcss/pull/18511))
```
reinink: Is it fair/correct to call this a `::marker`? When I see `::marker` I think of list markers, not an input indicator.

thecrypticace: Yes: Screenshot 2025-07-14 at 15 51 48

thecrypticace: can also change this to just say dropdown icon or something

reinink: Yeah might be more helpful, but if it's a `::marker`, I can't argue with it.

thecrypticace: I went ahead and updated it to say dropdown icon anyway. The more I think about it the more I'm like "yeah it's a `::marker` but that's probably an implementation detail and most people aren't gonna know that"

packages/tailwindcss/preflight.css
```diff
@@ -349,6 +349,14 @@ textarea {
   padding-block: 0;
 }
 
+/*
+  Center dropdown marker shown on inputs with paired `<datalist>`s in Chrome
```
reinink: I think we always end these with periods. Also, I think we sometimes include links to the original issues.

```suggestion
  Center dropdown marker shown on inputs with paired `<datalist>`s in Chrome. (https://github.com/tailwindlabs/tailwindcss/issues/18499)
```

reinink (approved):

PR #18665, opened 2025-08-04
Fix `require.cache` error when loaded through a typescript file in Node 22.18+

gaultier:
This fixes this bug:

```
  ✘ [ERROR] Cannot convert undefined or null to object [plugin postcss]
      
          ../../node_modules/tailwindcss/lib/lib/setupTrackingContext.js:69:33:
            69 │             delete require.cache[file];
               ╵                                  ^
      
          at getTailwindConfig (/Users/philippe.gaultier/company-code/cloud/elements/node_modules/tailwindcss/lib/lib/setupTrackingContext.js:69:34)
          at /Users/philippe.gaultier/company-code/cloud/elements/node_modules/tailwindcss/lib/lib/setupTrackingContext.js:100:92
          at /Users/philippe.gaultier/company-code/cloud/elements/node_modules/tailwindcss/lib/processTailwindFeatures.js:46:11
          at plugins (/Users/philippe.gaultier/company-code/cloud/elements/node_modules/tailwindcss/lib/plugin.js:38:69)
          at LazyResult.runOnRoot (/Users/philippe.gaultier/company-code/cloud/elements/packages/elements-react/node_modules/postcss/lib/lazy-result.js:329:16)
          at LazyResult.runAsync (/Users/philippe.gaultier/company-code/cloud/elements/packages/elements-react/node_modules/postcss/lib/lazy-result.js:258:26)
          at LazyResult.async (/Users/philippe.gaultier/company-code/cloud/elements/packages/elements-react/node_modules/postcss/lib/lazy-result.js:160:30)
          at LazyResult.then (/Users/philippe.gaultier/company-code/cloud/elements/packages/elements-react/node_modules/postcss/lib/lazy-result.js:404:17)
```

Not sure why we end up in this situation. It typically triggers when everything has been cleaned (`rm -rf node_modules && npm ci && npm run build`) in my [project](https://github.com/ory/elements).

EDIT: On our side we have found out that this issue popped up with node 22.18 and was not present in 22.17.


gaultier:
Sharing here my colleague's @pcaillaudm theory on the root cause:

> What I suspect being the root cause of the issue is that 22.18 is now able to execute TypeScript files without additional configuration. https://nodejs.org/en/blog/release/v22.18.0
It was not the case in 22.17 so when tailwind would require('./tailwind.config.ts'), it would insert the module into require.cache.
In 22.18, the module is now certainly executed through the new ESM type-stripping loader directly and never added to require.cache
So tailwind’s cache-busting loop tries to delete an entry that isn’t there.

And indeed, the bug does not trigger when running nodejs 22.18 with the option `--no-experimental-strip-types`.

thecrypticace:
Looks like this happens because Node.js loaded a Typescript postcss config file. Your original repo appeared to work fine if the postcss config was named `.js` instead of `.ts`.

Minimal repro:
```ts
// file.mjs
import postcss from "postcss";
import config from "./postcss.config.ts";
await postcss(config.plugins).process("");
```

```ts
// postcss.config.ts
module.exports = {
  plugins: [require("tailwindcss")],
};
```

I'm gonna do a bit more poking around to see what happens when things are required through the typescript loader but this fix/workaround seems reasonable.

thecrypticace:
Minimal no Tailwind CSS repro:

```js
// test.mjs
import "./file-1.ts";
```

```ts
// file-1.ts
require("./file-2.cjs");
```

```ts
// file-2.cjs
delete require.cache["foo"];
```

Only happens if the typescript file is loaded via `import` or `await import(…)`

thecrypticace (approved):

thecrypticace:
Thanks! I'll likely tag a v3 release sometime next week.

PR #18724, opened 2025-08-12
github/workflows: add riscv64-linux target

onny:
## Summary

Adding riscv64-linux target to the Github workflow release pipelines

## Test plan

Not sure how to test it yet :(

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -51,6 +51,11 @@ jobs:
             container:
               image: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian-aarch64
           - os: ubuntu-latest
+            target: riscv64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


thecrypticace:
Hey, I'm gonna close this one. A RISCV build isn't possible for us:

- Bun, which we use for our Standalone executable, does not support RISCV targets: https://bun.sh/docs/bundler/executables#supported-targets
- Nor does [Lightning CSS](https://github.com/parcel-bundler/lightningcss)
- Ditto with [Parcel Watcher](https://github.com/parcel-bundler/watcher)

The only thing here that *could* be compiled for a RISCV arch is a binary for Oxide. I'm not sure if NAPI-RS supports this or not though and we have no way of testing this currently anyway.

Aside: The Rust target is wrong. The right one is `riscv64gc-unknown-linux-gnu`.

PR #18732, opened 2025-08-14
Drop warning from browser build

zhe-he:
To ignore warnings when using the CDN, just add no-warn.
No src check was added since the CDN URL might vary (e.g., not always https://cdn.tailwindcss.com).

```html
<script src="https://cdn.tailwindcss.com" no-warn></script>
```

https://github.com/tailwindlabs/tailwindcss/issues/18731

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add configuration to ignore CDN warnings ([#18731](https://github.com/tailwindlabs/tailwindcss/issues/18731))
+- Drop warning from browser build ([#18731](https://github.com/tailwindlabs/tailwindcss/issues/18731))
 
 ## [4.1.12] - 2025-08-13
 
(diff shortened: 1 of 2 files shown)


wongjn:
FYI, https://cdn.tailwindcss.com is the v3 CDN, which is different to the `@tailwindcss/browser` package. This PR would only affect the latter, not the former.

thecrypticace:
We've decided to drop the warning entirely from the browser build for v4. Especially for people who are importing it into their JS through npm.

Though, I still maintain you probably *shouldn't* use it in production for the same reasons I outlined in https://github.com/tailwindlabs/tailwindcss/issues/18731.

For now the v3 CDN is unchanged.

thecrypticace (approved):

inoryxun:
> 请注意，https://cdn.tailwindcss.com是 v3 CDN，与`@tailwindcss/browser`软件包不同。此 PR 只会影响后者，不会影响前者。
   <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>这是V4 的cdn地址吗？为什么 px-8 py-4 这一种的不起作用了？V3 反而能用

PR #18734, opened 2025-08-14
Mark `.hdr` and `.exr` as binary extensions

RobinMalfait:
This PR fixes an issue where `.hdr` files were scanned for candidates even though it's a binary file.

As a workaround today, you could use:

```css
@source not "**/*.hdr";
```

To ignore `.hdr` files, but let's bake it in by default instead.

Fixes: #18733


thecrypticace (approved):

thecrypticace:
We should add `exr` as well. It's also a common HDR image format.

RobinMalfait:
@thecrypticace added `.exr` as well. Going to sneak it into the next release 👀

thecrypticace (approved):

PR #18766, opened 2025-08-19
Prototype loading `@plugin` and `@config` from non-default exports

thecrypticace:
This will make using `@config` less necessary for plugins which don't use `export default`. Added to `@config` for feature parity.


RobinMalfait:
Going to close this for now, and re-open it when there is a more concrete use case.

PR #18769, opened 2025-08-20
Replace deprecated `clip` with `clip-path` in `sr-only`

bronisMateusz:
## Summary

This PR replaces the deprecated `clip` property used in the `sr-only` utility with `clip-path`, and updates the corresponding reset in `not-sr-only`.

- Closes [tailwindlabs/tailwindcss#18768](https://github.com/tailwindlabs/tailwindcss/issues/18768)
- Replaces `clip: rect(0, 0, 0, 0);` with `clip-path: inset(50%);` in `sr-only`
- Replaces `clip: auto;` with `clip-path: none;` in `not-sr-only`
- Updates unit test snapshots to reflect the new CSS output

Rationale:

- `clip` is deprecated and flagged by modern linters; `clip-path` is the recommended modern alternative while preserving the intended visually-hidden behavior.

Before:

```css
.sr-only {
  clip: rect(0, 0, 0, 0);
}

.not-sr-only {
  clip: auto;
}
```

After:

```css
.sr-only {
  clip-path: inset(50%);
}

.not-sr-only {
  clip-path: none;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show suggestions for known `matchVariant` values ([#18798](https://github.com/tailwindlabs/tailwindcss/pull/18798))
 - Migrate `aria` theme keys to `@custom-variant` ([#18815](https://github.com/tailwindlabs/tailwindcss/pull/18815))
 - Migrate `data` theme keys to `@custom-variant` ([#18816](https://github.com/tailwindlabs/tailwindcss/pull/18816))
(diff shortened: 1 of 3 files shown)


thecrypticace:
At one time using `clip-path: inset(50%)` would cause massive scroll performance problems in Chrome. Though this was many years ago. Will definitely need testing to ensure this isn't a problem any longer before we do this.

RobinMalfait:
I tested this on an older Android phone in Chrome and it seems to work smoothly. Used play: https://play.tailwindcss.com/P8PkfV5N99

https://github.com/user-attachments/assets/60d61130-cd14-492d-b75d-b78444cd0727

thecrypticace (approved):

PR #18809, opened 2025-08-27
Drop exact duplicate declarations from output CSS within a style rule

### LINKED ISSUE #18178: Use a `content` utility with `before`/`after` variants adds the `content` CSS property twice

v4.1.8

Tailwind Play

**Reproduction URL**

https://play.tailwindcss.com/rO3bQlDyWh

**Describe your issue**

The content utility adds the content property. Since the content property is required to display `::before` and `::after` pseudo-elements, the CSS rule `content: var(--tw-content);` is automatically added to the class when using `before` or `after` variants.

However, if I apply a custom content utility together with the before or after variants, the content property is added to the class twice:
* First, due to the `before`/`after` variant (source: [./src/variants.ts L641 - L667](https://github.com/tailwindlabs/tailwindcss/blob/193eb84f6c4c23ef43166162bed99d0fb3485c8b/packages/tailwindcss/src/variants.ts#L641C5-L667C6))
* Second, due to the `content` utility

```css
/* Generated Utilities */

.content-\[\'custom\'\] {
  --tw-content: 'custom';
  content: var(--tw-content);
}
.before\:content-\[\'custom\'\] {
  &::before {
    content: var(--tw-content);
    --tw-content: 'custom';
    content: var(--tw-content);
  }
}
.after\:content-\[\'custom\'\] {
  &::after {
    content: var(--tw-content);
    --tw-content: 'custom';
    content: var(--tw-content);
  }
}
```

In this case, it would be sufficient to keep only one of them.

```css
/* Expected Generated Utilities */

.content-\[\'custom\'\] {
  --tw-content: 'custom';
  content: var(--tw-content);
}
.before\:content-\[\'custom\'\] {
  &::before {
    --tw-content: 'custom';
    content: var(--tw-content);
  }
}
.after\:content-\[\'custom\'\] {
  &::after {
    --tw-content: 'custom';
    content: var(--tw-content);
  }
}
```

thecrypticace:
Fixes #18178

When someone writes a utility like `after:content-['foo']` it'll produce duplicate `content: var(--tw-content)` declarations. I thought about special casing these but we already have an optimization pass where we perform a full walk of the AST, flattening some rules (with the `&` selector), analyzing declarations, etc… We can utilize that existing spot in core to analyze and remove duplicate declarations within rules across the AST.

The implementation does this by keeping track of declarations within a style rule and keeps the last one for any *exact duplicate* which is a tuple of `(property, value, important)`. This does require some additional loops but preseving the *last* declaration is important for correctness with regards to CSS nesting.

For example take this nested CSS:
```css
.foo {
  color: red;
  & .bar {
    color: green;
  }
  color: red;
}
```

It expands to this:
```css
.foo {
  color: red;
}
.foo.bar {
  color: green;
}
.foo {
  color: red;
}
```

If you remove the *last* rule then a `<div class="foo bar">…</div>` will have green text when its supposed to be red. Since that would affect behavior we have to always preserve the last declaration for a given property.

We could go further and eliminate multiple declarations for the same property *but* this presents a problem: every property and value must be understood and combined with browser targets to understand whether or not that property may act as a "fallback" or whether definitely overwrites its previous value in all cases. This is a much more complicated task that is much more suited to something light Lighting CSS.

### DIFF
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -357,6 +357,31 @@ export function optimizeAst(
         transform(child, nodes, context, depth + 1)
       }
 
+      // Keep the last decl when there are exact duplicates. Keeping the *first* one might
+      // not be correct when given nested rules where a rule sits between declarations.
+      let seen: Record<string, AstNode[]> = {}
+      let toRemove = new Set<AstNode>()
+
+      // Keep track of all nodes that produce a given declaration
+      for (let child of nodes) {
+        if (child.kind !== 'declaration') continue
+
+        let key = `${child.property}:${child.value}:${child.important}`
+        seen[key] ??= []
+        seen[key].push(child)
+      }
+
+      // And remove all but the last of each
+      for (let key in seen) {
+        for (let i = 0; i < seen[key].length - 1; ++i) {
+          toRemove.add(seen[key][i])
+        }
+      }
+
+      if (toRemove.size > 0) {
+        nodes = nodes.filter((node) => !toRemove.has(node))
+      }
+
       if (nodes.length === 0) return
 
       // Rules with `&` as the selector should be flattened
--- a/packages/tailwindcss/src/ast.ts
+++ b/packages/tailwindcss/src/ast.ts
@@ -351,26 +351,19 @@ export function optimizeAst(
 
     // Rule
     else if (node.kind === 'rule') {
-      // Rules with `&` as the selector should be flattened
-      if (node.selector === '&') {
-        for (let child of node.nodes) {
-          let nodes: AstNode[] = []
-          transform(child, nodes, context, depth + 1)
-          if (nodes.length > 0) {
-            parent.push(...nodes)
-          }
-        }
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/ast.ts
```diff
@@ -357,6 +357,31 @@ export function optimizeAst(
...
+      let seen: Record<string, AstNode[]> = {}
+      let toRemove = new Set<AstNode>()
+
+      // Keep track of all nodes that produce a given declaration
+      for (let child of nodes) {
+        if (child.kind !== 'declaration') continue
+
+        let key = `${child.property}:${child.value}:${child.important}`
+        seen[key] ??= []
+        seen[key].push(child)
```
RobinMalfait, line 371: jk 😂

RobinMalfait (approved):

packages/tailwindcss/src/ast.ts
thecrypticace: I mean basically lol

PR #18815, opened 2025-08-28
Migrate `aria` theme keys

RobinMalfait:
This PR migrates `aria` theme keys when migrating from Tailwind CSS v3 to v4.

While working on improving some of the error messages to get more insights into why migrating the JS file changed (https://github.com/tailwindlabs/tailwindcss/pull/18808), I ran into an issue where I couldn't think of a good comment to why `aria` theme keys were not being migrated. (Internally we have `aria` "blocked").

So instead of figuring out a good error message..., I just went ahead and added the migration for `aria` theme keys.

Let's imagine you have the following Tailwind CSS v3 configuration:
```ts
export default {
  content: ['./src/**/*.html'],
  theme: {
    extend: {
      aria: {
        // Built-in (not really, but visible because of intellisense)
        busy: 'busy="true"',

        // Automatically handled by bare values
        foo: 'foo="true"',
    //  ^^^   ^^^            ← same names

        // Not automatically handled by bare values because names differ
        bar: 'baz="true"',
    //  ^^^   ^^^            ← different names

        // Completely custom
        asc: 'sort="ascending"',
        desc: 'sort="descending"',
      },
    },
  },
}
```

Then we would generate the following Tailwind CSS v4 CSS:

```css
@custom-variant aria-bar (&[aria-baz="true"]);
@custom-variant aria-asc (&[aria-sort="ascending"]);
@custom-variant aria-desc (&[aria-sort="descending"]);
```

Notice how we didn't generate a custom variant for `aria-busy` or `aria-foo` because those are automatically handled by bare values.

We could also emit a comment near the CSS to warn about the fact that `@custom-variant` will always be sorted _after_ any other built-in variants.

This could result in slightly different behavior, or different order of classes when using `prettier-plugin-tailwindcss`.

I don't know how important this is, because before this PR we would just use `@config './tailwind.config.js';`.
Edit: when using the `@config` we override `aria` and extend it, which means that it would be in the expected order 🤔

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
@@ -145,7 +145,7 @@ async function migrateTheme(
       for (let [key, value] of Object.entries(resolvedConfig.theme.aria ?? {})) {
         // Will be handled by bare values if the names match.
         // E.g.: `aria-foo:flex` should produce `[aria-foo="true"]`
-        if (new RegExp(`^${key}=['"]true['"]$`).test(`${value}`)) continue
+        if (new RegExp(`^${key}=(['"]?)true\\1$`).test(`${value}`)) continue
 
         // Create custom variant
         variants.set(`aria-${key}`, `&[aria-${value}]`)
--- a/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
@@ -97,7 +97,7 @@ async function migrateTheme(
   designSystem: DesignSystem,
   unresolvedConfig: Config,
   base: string,
-): Promise<string | null> {
+): Promise<string> {
   // Resolve the config file without applying plugins and presets, as these are
   // migrated to CSS separately.
   let configToResolve: ConfigFile = {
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
```diff
@@ -151,51 +189,46 @@ async function migrateTheme(
...
   }
 
-  css += '}\n' // @theme
-
-  if ('container' in resolvedConfig.theme) {
-    let rules = buildCustomContainerUtilityRules(resolvedConfig.theme.container, designSystem)
-    if (rules.length > 0) {
-      css += '\n' + toCss([atRule('@utility', 'container', rules)])
+  if (variants.size > 0) {
+    css += '\n@tw-bucket custom-variant {\n'
```
RobinMalfait, line 225: We could emit a comment here as well for the different sort order. But I wonder if we should try and make the sort order consistent and group them near the real `aria` variants 🤔

thecrypticace (approved):

packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
RobinMalfait: Going to skip the comment for now; Since you are upgrading from v3 to v4 you have to check everything anyway.

We also call that out in the output of the upgrade tool:

Image

PR #18817, opened 2025-08-28
Migrate `supports` theme keys

RobinMalfait:
This PR is a follow up of #18815 and #18816, but this time let's migrate the `supports` theme keys.

Let's imagine you have the following Tailwind CSS v3 configuration:
```ts
export default {
  content: ['./src/**/*.html'],
  theme: {
    extend: {
      supports: {
        // Automatically handled by bare values (using CSS variable as the value)
        foo: 'foo: var(--foo)', // parentheses are optional
        bar: '(bar: var(--bar))',

        // Not automatically handled because names differ
        baz: 'qux: var(--foo)',
   //   ^^^   ^^^       ← different names

        // Custom
        grid: 'display: grid',
      },
    },
  },
}
```

Then we would generate the following Tailwind CSS v4 CSS:

```css
@custom-variant supports-baz {
  @supports (qux: var(--foo)) {
    @slot;
  }
}
@custom-variant supports-grid {
  @supports (display: grid) {
    @slot;
  }
}
```

Notice how we didn't generate a custom variant for `data-foo` or `data-bar` because those are automatically handled by bare values.

I also went with the longer form of `@custom-variant`, we could use the single selector approach, but that felt less clear to me.

```css
@custom-variant supports-baz (@supports (qux: var(--foo)));
@custom-variant supports-grid (@supports (display: grid));
```

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
@@ -24,6 +24,7 @@ import {
   isValidOpacityValue,
   isValidSpacingMultiplier,
 } from '../../../../tailwindcss/src/utils/infer-data-type'
+import * as ValueParser from '../../../../tailwindcss/src/value-parser'
 import { findStaticPlugins, type StaticPluginOptions } from '../../utils/extract-static-plugins'
 import { highlight, info, relative } from '../../utils/renderer'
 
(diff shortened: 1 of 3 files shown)


packages/@tailwindcss-upgrade/src/codemods/config/migrate-js-config.ts
```diff
@@ -406,15 +440,11 @@ const ALLOWED_THEME_KEYS = [
...
 ]
-const BLOCKED_THEME_KEYS = ['supports']
 function onlyAllowedThemeValues(theme: ThemeConfig): boolean {
   for (let key of Object.keys(theme)) {
     if (!ALLOWED_THEME_KEYS.includes(key)) {
       return false
     }
-    if (BLOCKED_THEME_KEYS.includes(key)) {
-      return false
-    }
```
RobinMalfait, line 418: 👋 bye bye blocked theme keys

We still have `aria`, `data` and `supports` handling in: https://github.com/tailwindlabs/tailwindcss/blob/feat%2Fupgrade-supports-theme/packages/tailwindcss/src/compat/theme-variants.ts

This is still necessary in case the config file could not be migrated due to other issues. So didn't get rid of that part.

thecrypticace (approved):

PR #18834, opened 2025-08-30
Add `ring-initial` utility

rollraw:
## Summary
There is no way to undo a `ring-inset` at the moment. This PR adds utility `ring-initial` to do that. Has been discussed at #3016

### DIFF
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -5486,6 +5486,7 @@ export function createUtilities(theme: Theme) {
     ])
 
     staticUtility('ring-inset', [boxShadowProperties, ['--tw-ring-inset', 'inset']])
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey!

Thanks for the PR! However, going to close it for now because I think _if_ we want to do it, that we will introduce these default keywords for more/all utilities.

For now, I would recommend to add a custom utility instead. You can achieve that using something like:
```css
@utility ring-initial {
  --tw-ring-color: initial;
}
```

You could also use `ring-[initial]` if you don't want to add a custom utility.

rollraw:
> For now, I would recommend to add a custom utility instead. You can achieve that using something like:
> 
> ```css
> @utility ring-initial {
>   --tw-ring-color: initial;
> }
> ```
> 
> You could also use `ring-[initial]` if you don't want to add a custom utility.

None of those would work with tw-merge.

RobinMalfait:
That seems like an issue with tw-merge then, since they do work in Tailwind CSS

PR #18855, opened 2025-09-02
Add '.vercel' to ignored content directories

mischnic:
Just like `.next` is ignored, ignore `.vercel` as well.

### DIFF
--- a/crates/oxide/src/scanner/fixtures/ignored-content-dirs.txt
+++ b/crates/oxide/src/scanner/fixtures/ignored-content-dirs.txt
@@ -11,3 +11,4 @@ venv
 __pycache__
 .svelte-kit
 .pnpm-store
+.vercel


RobinMalfait:
Hey! We don't have any .vercel files or folders so this shouldn't be needed.

timneutkens:
Hey @RobinMalfait this is about the folders that Tailwind tries to crawl/watch, we saw a customer hit this case where it tried to read through all of `.vercel` which is incorrect. Is this the wrong file to update to change Tailwind's default ignores?

RobinMalfait:
Oops, yep you are right. I for some reason interpreted this as if `.vercel` was added to `.gitignore` 🤦‍♂️

RobinMalfait (approved):

RobinMalfait:
Sorry for the miscommunication. This is indeed the correct file, thanks!

Will be available in the next release 👍

timneutkens:
No problem! Thought so 😄

PR #18869, opened 2025-09-03
Consider variants starting with `@-` to be invalid (e.g. `@-2xl:flex`)

RobinMalfait:
This PR fixes a small parsing issue where variants such as `@-2xl:flex` would parse, but were handled as-if they were `@2xl:flex` instead.

Noticed this while working on: #18867 

This is because when we parse normal variants like `data-foo` then we want to have a `data` root and a `foo` value, not a `-foo` value.

If you are now using `@-2xl:flex`, then no CSS will be generated for this anymore. If you were relying on this for some reason, you should use `@2xl:flex` instead.

## Test plan

Before:

image

After:

image


thecrypticace (approved):

mariansimecek:
@RobinMalfait 
Hi, this change breaks some styles on our website. I found out that the old migration script automatically generated `@-4xl` (the current version doesn’t do that). It took me some time to find the root cause of the issue. I just wanted to let you know that there may be people who used the migration script, and now their websites will break. You may want to consider marking this change as breaking.

PR #18885, opened 2025-09-04
Handle `@variant` inside `@custom-variant`

### LINKED ISSUE #18524: Nested `@variant` inside `@custom-variant`

v4.1.11

**Reproduction URL**

https://play.tailwindcss.com/BK1Bs5n1jp?file=css

**Describe your issue**

For the purpose of reusing, I declare a `@custom-variant` for `@support`. In another custom variant `@scroll-stuck-top`, I would like to use that `supports-scroll-state` variant to add fallback. I expected the nested `@variant` would be replaced with the definition of the corresponding `@custom-variant`, but instead, it was kept as-is.

```css
@custom-variant supports-scroll-state {
  @supports (container-type: scroll-state) {
    @slot;
  }
}

@custom-variant @scroll-stuck-top {
  @container scroll-state(stuck: top) { @slot; }

  @variant not-supports-scroll-state {
    \@scroll-stuck-top & { @slot }
  }
}
```

### Expected Output

```css
.\@scroll-stuck-top\:border-b-2 {
  @container scroll-state(stuck: top) {
    border-bottom-style: var(--tw-border-style);
    border-bottom-width: 2px;
  }
  @supports not (container-type: scroll-state) {
    \@scroll-stuck-top & {
      border-bottom-style: var(--tw-border-style);
      border-bottom-width: 2px;
    }
  }
}
```

### Actual Output

```css
.\@scroll-stuck-top\:border-b-2 {
  @container scroll-state(stuck: top) {
    border-bottom-style: var(--tw-border-style);
    border-bottom-width: 2px;
  }
  @variant not-supports-scroll-state {
    \@scroll-stuck-top & {
      border-bottom-style: var(--tw-border-style);
      border-bottom-width: 2px;
    }
  }
}
```

RobinMalfait:
This PR fixes an issue where you cannot use `@variant` inside a `@custom-variant`. While you can use `@variant` in normal CSS, you cannot inside of `@custom-variant`. Today this silently fails and emits invalid CSS.
```css
@custom-variant dark {
  @variant data-dark {
    @slot;
  }
}
```
```html
<div class="dark:flex"></div>
```

Would result in:
```css
.dark\:flex {
  @variant data-dark {
    display: flex;
  }
}
```

To solve it we have 3 potential solutions:

1. Consider it user error — but since it generates CSS and you don't really get an error you could be shipping broken CSS unknowingly.
1. We could try and detect this and not generate CSS for this and potentially show a warning.
1. We could make it work as expected — which is what this PR does.

Some important notes:

1. The evaluation of the `@custom-variant` only happens when you actually need it. That means that `@variant` inside `@custom-variant` will always have the implementation of the last definition of that variant.

   In other words, if you use `@variant hover` inside a `@custom-variant`, and later you override the `hover` variant, the `@custom-variant` will use the new implementation.
1. If you happen to introduce a circular dependency, then an error will be thrown during the build step.

You can consider it a bug fix or a new feature it's a bit of a gray area. But
one thing that is cool about this is that you can ship a plugin that looks like
this:
```css
@custom-variant hocus {
  @variant hover {
    @slot;
  }

  @variant focus {
    @slot;
  }
}
```

And it will use the implementation of `hover` and `focus` that the user has defined. So if they have a custom `hover` or `focus` variant it will just work.

By default `hocus:underline` would generate:
```css
@media (hover: hover) {
  .hocus\:underline:hover {
  }
}

.hocus\:underline:focus {
}
```

But if you have a custom `hover` variant like:
```css
@custom-variant hover (&:hover);
```

Then `hocus:underline` would generate:
```css
.hocus\:underline:hover, .hocus\:underline:focus {
}
```

### Test plan

1. Existing tests pass
2. Added tests with this new functionality handled
3. Made sure to add a test for circular dependencies + error message
4. Made sure that if you "fix" the circular dependency (by overriding a variant) that everything is generated as expected.

Fixes: https://github.com/tailwindlabs/tailwindcss/issues/18524

### DIFF
--- a/packages/tailwindcss/src/compat/plugin-api.ts
+++ b/packages/tailwindcss/src/compat/plugin-api.ts
@@ -154,7 +154,7 @@ export function buildPluginApi({
 
       // CSS-in-JS object
       else if (typeof variant === 'object') {
-        designSystem.variants.fromAst(name, objectToAst(variant))
+        designSystem.variants.fromAst(name, objectToAst(variant), designSystem)
       }
     },
     matchVariant(name, fn, options) {
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -22,7 +22,7 @@ import { substituteAtImports } from './at-import'
 import { applyCompatibilityHooks } from './compat/apply-compat-hooks'
 import type { UserConfig } from './compat/config/types'
 import { type Plugin } from './compat/plugin-api'
-import { applyVariant, compileCandidates } from './compile'
+import { compileCandidates } from './compile'
 import { substituteFunctions } from './css-functions'
 import * as CSS from './css-parser'
 import { buildDesignSystem, type DesignSystem } from './design-system'
@@ -32,7 +32,7 @@ import { createCssUtility } from './utilities'
 import { expand } from './utils/brace-expansion'
 import { escape, unescape } from './utils/escape'
 import { segment } from './utils/segment'
-import { compoundsForSelectors, IS_VALID_VARIANT_NAME } from './variants'
+import { compoundsForSelectors, IS_VALID_VARIANT_NAME, substituteAtVariant } from './variants'
 export type Config = UserConfig
 
 const IS_VALID_PREFIX = /^[a-z]+$/
@@ -636,30 +636,7 @@ async function parseCss(
     firstThemeRule.nodes = [context({ theme: true }, nodes)]
   }
 
-  // Replace the `@variant` at-rules with the actual variant rules.
-  if (variantNodes.length > 0) {
-    for (let variantNode of variantNodes) {
-      // Starting with the `&` rule node
-      let node = styleRule('&', variantNode.nodes)
-
-      let variant = variantNode.params
-
-      let variantAst = designSystem.parseVariant(variant)
-      if (variantAst === null) {
-        throw new Error(`Cannot use \`@variant\` with unknown variant: ${variant}`)
-      }
-
-      let result = applyVariant(node, variantAst, designSystem.variants)
-      if (result === null) {
-        throw new Error(`Cannot use \`@variant\` with variant: ${variant}`)
-      }
-
-      // Update the variant at-rule node, to be the `&` rule node
-      Object.assign(variantNode, node)
-    }
-    features |= Features.Variants
-  }
-
+  features |= substituteAtVariant(ast, designSystem)
   features |= substituteFunctions(ast, designSystem)
   features |= substituteAtApply(ast, designSystem)
 
--- a/packages/tailwindcss/src/index.ts
+++ b/packages/tailwindcss/src/index.ts
@@ -32,6 +32,7 @@ import { createCssUtility } from './utilities'
 import { expand } from './utils/brace-expansion'
 import { escape, unescape } from './utils/escape'
 import { segment } from './utils/segment'
+import { topologicalSort } from './utils/topological-sort'
 import { compoundsForSelectors, IS_VALID_VARIANT_NAME, substituteAtVariant } from './variants'
 export type Config = UserConfig
 
@@ -150,7 +151,8 @@ async function parseCss(
 
   let important = null as boolean | null
   let theme = new Theme()
-  let customVariants: ((designSystem: DesignSystem) => void)[] = []
+  let customVariants = new Map<string, (designSystem: DesignSystem) => void>()
+  let customVariantDependencies = new Map<string, Set<string>>()
   let customUtilities: ((designSystem: DesignSystem) => void)[] = []
   let firstThemeRule = null as StyleRule | null
   let utilitiesNode = null as AtRule | null
@@ -390,7 +392,7 @@ async function parseCss(
           }
         }
 
-        customVariants.push((designSystem) => {
+        customVariants.set(name, (designSystem) => {
           designSystem.variants.static(
             name,
             (r) => {
@@ -411,6 +413,7 @@ async function parseCss(
             },
           )
         })
+        customVariantDependencies.set(name, new Set<string>())
 
         return
       }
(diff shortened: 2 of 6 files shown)


packages/tailwindcss/src/variants.ts
```diff
@@ -95,12 +101,11 @@ export class Variants {
       name,
       (r) => {
         let body = structuredClone(ast)
+        if (usesAtVariant) substituteAtVariant(body, designSystem)
```
RobinMalfait, line 104: This is technically evaluated lazily and every time this function is called instead of ahead of time.

But once we compute a variant we cache it anyway so should be fine.

packages/tailwindcss/src/variants.ts
```diff
@@ -95,12 +101,11 @@ export class Variants {
...
         substituteAtSlot(body, r.nodes)
         r.nodes = body
       },
-      {
-        compounds: compoundsForSelectors(selectors),
-      },
+      { compounds: compoundsForSelectors(selectors) },
```
RobinMalfait, line 108: Custom variants that re-use existing `@variant`s are not compoundable. I had an implementation where the `@variant` substitution was happening ahead of time to get the selectors out but we have some checks in `in-*` variant for example that doesn't allow nesting and `@variant …` is implemented using nesting...

I think this is something we can solve in a separate PR and probably after handling flattening ourselves... can of worms.

packages/tailwindcss/src/index.ts
```diff
@@ -605,8 +616,27 @@ async function parseCss(
     sources,
   })
 
-  for (let customVariant of customVariants) {
-    customVariant(designSystem)
+  for (let name of customVariants.keys()) {
+    // Pre-register the variant to ensure its position in the variant list is
+    // based on the order we see them in the CSS.
+    designSystem.variants.static(name, () => {})
```
RobinMalfait, line 622: Open for suggestions here. Like maybe a `designSystem.variants.reserve(name)` but felt silly to introduce a new method just for this...

But my idea was to have the same behavior as-if you are overwriting internal variants that should maintain the sort order.

RobinMalfait: There's an argument for keeping the position of the lastly defined `@custom-variant` because otherwise if you were relying on a library that now introduces a `@custom-variant` with the same name, the sort order will be different.

packages/tailwindcss/src/utils/topological-sort.ts
```diff
@@ -0,0 +1,36 @@
+export function topologicalSort<Key>(
+  graph: Map<Key, Set<Key>>,
+  options: { onCircularDependency: (path: Key[], start: Key) => void },
+): Key[] {
+  let seen = new Set<Key>()
+  let wip = new Set<Key>()
+
+  let sorted: Key[] = []
```
RobinMalfait, line 8: We could make `topologicalSort` a Generator which has a nice API, but that's way slower compared to this array-based implementation:

Image

RobinMalfait: Tested this with a very simple graph:
```ts
let graph = new Map<string, Set<string>>([
  ['A', new Set(['B', 'C'])],
  ['B', new Set(['D'])],
  ['C', new Set(['D'])],
  ['D', new Set(['E'])],
  ['E', new Set()],
])
```

But I don't think the real graphs will be much more complex anyway...

RobinMalfait:
Note: we can likely use the topologicalSort utility when handling `@apply`, but didn't want to touch that as part of this PR because it might have to change a little bit to get the graph in a shape we can work with.

Might do a follow PR for this.

packages/tailwindcss/src/utils/topological-sort.ts
thecrypticace: Yeah I'd keep the array in this case. The perf tradeoff makes sense when the memory hit is possibly unbounded but you're likely to hit other perf and memory problems with large CSS files first.

packages/tailwindcss/src/variants.ts
thecrypticace: This is fine imo. The problem here is the nested style rules are the things that disqualify it because otherwise we'll have to implement flattening ourselves.

packages/tailwindcss/src/index.ts
thecrypticace: i think this is fine. Keeping the last defined one would probably be more CSS-y *but* it would make overriding builtin variants different from custom ones and I'd prefer that they act the same.

RobinMalfait: Yep agree

packages/tailwindcss/src/variants.ts
RobinMalfait: Exactly, could be a follow-up PR, but this PR on its own is already an improvement.

packages/tailwindcss/src/variants.ts
```diff
@@ -1198,3 +1203,30 @@ export function substituteAtSlot(ast: AstNode[], nodes: AstNode[]) {
...
+      throw new Error(`Cannot use \`@variant\` with unknown variant: ${variant}`)
+    }
+
+    let result = applyVariant(node, variantAst, designSystem.variants)
+    if (result === null) {
+      throw new Error(`Cannot use \`@variant\` with variant: ${variant}`)
+    }
+
+    // Update the variant at-rule node, to be the `&` rule node
+    replaceWith(node)
```
RobinMalfait, line 1228: If we never mutate the selector of the `&` node, then we could skip some layers by using:

```suggestion
    replaceWith(node.nodes)
```

But the final result should be the same. This is also the same behavior we had before so I think we can keep it like this.

packages/tailwindcss/src/compat/plugin-api.ts
```diff
@@ -154,7 +154,7 @@ export function buildPluginApi({
 
       // CSS-in-JS object
       else if (typeof variant === 'object') {
-        designSystem.variants.fromAst(name, objectToAst(variant))
+        designSystem.variants.fromAst(name, objectToAst(variant), designSystem)
```
thecrypticace, line 157: i kinda wish we didn't have to pass the design system into something hanging off the design system :D

thecrypticace (approved):

packages/tailwindcss/src/compat/plugin-api.ts
RobinMalfait: Yeah, one way of solving this is that we put a `variantFromAst` onto the `DesignSystem` itself 🤔 

or, we can pass in a callback if we want to do something with the `ast` (body in this case) so the logic lives in the callsite.

I think for now, this is fine because it's all private anyway.

rozsazoltan:
Cool.

PR #17389, opened 2025-03-26
Add `text-shadow-*` utilities

philipp-spiess:
This PR adds new `text-shadow-*` utilities and default values courtesy of @danhollick's.

Usage is similar to the normal `shadow-*` utilities, for example:

```html
<h1 class="text-center text-7xl tracking-tight text-white text-shadow-xl">
  Some fancy <br />
  headline
</h1>
```

Since this PR also adds first-class support for the `--text-shadow` theme namespace, it also means it resolves #17047.

## Test plan

- Copied @danhollick's [demo playground](https://play.tailwindcss.com/lbA3Y6VY8u) to the `vite` setup and ran a production build to ensure it looks correct: https://tailwind-text-shadows-preview-2kdnjb32b.vercel.app/

PR #18901, opened 2025-09-08
Add `@container-size` utility

RobinMalfait:
This PR adds a new `@container-size` utility instead of `@container-[size]`.

The main reason we didn't do this before is because we only have container width related container queries, and not block based ones so we never needed `size` and `inline-size` was enough.

However, `@container-size` is still useful if you are using container query related units such as `cqb` which are using the block size of the container not the inline size.

I also added a little helper such that `@container-size` is only available in `insiders` and `4.3.0` (and later) so `4.1.x` releases won't have this utility yet. This will require some CHANGELOG changes such that we don't include this when releasing the next minor release.


packages/tailwindcss/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,42 @@
...
+
+let majorMinorPatchRegex = /^(\d+)\.(\d+)\.(\d+)(?:-(\w+))?/
+let parsedVersion = new DefaultMap((version: string) => {
+  let result = majorMinorPatchRegex.exec(version)
+  if (!result) throw new Error(`Invalid version: ${version}`)
+  let [, major, minor, patch, prerelease = null] = result
+  return { major: Number(major), minor: Number(minor), patch: Number(patch), prerelease }
+})
+
+export function greaterThanOrEqual(a: string, b: string) {
```
RobinMalfait: Didn't want to pull in a full semver package just to test the major, minor patch and a pre-release channel.

packages/tailwindcss/src/utilities.ts
```diff
@@ -5712,6 +5713,12 @@ export function createUtilities(theme: Theme) {
       value = candidate.value.value
     } else if (candidate.value.kind === 'named' && candidate.value.value === 'normal') {
       value = 'normal'
+    } else if (
+      featureFlagVersion('4.2.0', '0.0.0-insiders') &&
```
RobinMalfait: The main reason I added this is such that we can merge these kinds of PRs instead of keeping them open until we are ready to release a `4.2.x` with new features.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
```
RobinMalfait: Added this here, but open to suggestions on how we would tackle this because this should only be released once we release `4.2.x`, not in a `4.1.x` patch release.

packages/tailwindcss/src/utilities.ts
RobinMalfait: This also ensures that if we want to make changes to the API, that we can do that until we release a next minor version.

packages/tailwindcss/src/utils/feature-flags.ts
philipp-spiess: Not sure why we need to ship runtime checks for this, the feature flagging setup we used for preparing the 4.1 branch was also enabling it in the insiders build wasn't it? We'd document the feature as not released yet until we eventually cut the 4.2 release. This way we avoid any runtime checks and can even compile it out for patch releases

philipp-spiess: So just relying on this env: https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/tsup.config.ts#L16

philipp-spiess: https://github.com/tailwindlabs/tailwindcss/pull/17389/files#diff-a1365c16b6dd1facff1314fc1536d533e46a7c7eabf2d4df9a50963f3a777086

RobinMalfait: Did it ship for insiders as well? 🤔 

If that's the case, then yeah that's even better. I think we got rid of all that code though, right?

This is semi-runtime, it does a runtime check, but the `version` of package.json is inlined at build-time already so there is no IO involved anymore.

Will update this with the build-time feature flags instead 👍

philipp-spiess: Yeah defaults to insiders unless running a stable release job: https://github.com/search?q=repo%3Atailwindlabs%2Ftailwindcss%20FEATURES_ENV&type=code

CHANGELOG.md
philipp-spiess: I think what we did in the past was to prefix this with `_Experimental_` so we'd not forget: https://github.com/tailwindlabs/tailwindcss/pull/17389/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

philipp-spiess (approved):

benface:
Is this ever going to get released? 😇 Also, maybe `@container-scroll-state` should be added too?

PR #18907, opened 2025-09-09
Do not generate `grid-column` when configuring `grid-column-start` or `grid-column-end`; Similar for `grid-row`

RobinMalfait:
This PR fixes an issue where configuring a custom `--grid-column-start` or `--grid-column-end` also generated a `grid-column` utility due to the overlapping namespace.

```css
@theme {
  --grid-column-start-custom: custom-start;
  --grid-column-end-custom: custom-end;
}
```

Would then generate:
```css
.col-end-custom {
  grid-column: var(--grid-column-end-custom);
}
.col-start-custom {
  grid-column: var(--grid-column-start-custom);
}
.col-start-custom {
  grid-column-start: var(--grid-column-start-custom);
}
.col-end-custom {
  grid-column-end: var(--grid-column-end-custom);
}
```

Instead of the expected:
```css
.col-start-custom {
  grid-column-start: var(--grid-column-start-custom);
}
.col-end-custom {
  grid-column-end: var(--grid-column-end-custom);
}
```

This fix is also applied to `grid-row` when configuring `--grid-row-start` and `--grid-row-end`

Fixes: #18906

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Handle `@variant` inside `@custom-variant` ([#18885](https://github.com/tailwindlabs/tailwindcss/pull/18885))
 - Merge suggestions when using `@utility` ([#18900](https://github.com/tailwindlabs/tailwindcss/pull/18900))
 - Ensure that file system watchers created when using the CLI are always cleaned up ([#18905](https://github.com/tailwindlabs/tailwindcss/pull/18905))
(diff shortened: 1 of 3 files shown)


philipp-spiess (commented):
I think the same issue exists for `--grid-row`?

From my WIP PR:

Image

RobinMalfait:
@philipp-spiess ah yep; I'm sure there are more cases but I still want to fix this in a more generic way somehow.

For now, updated the PR to handle both `--grid-column` and `--grid-row`

philipp-spiess (approved):

PR #18917, opened 2025-09-11
Add support for `--max-breakpoint-*` theme variables

madasebrof:
Summary

This PR adds support for `--max-breakpoint-*` theme variables, enabling desktop-first breakpoint workflows similar to Webflow. This complements the existing mobile-first `--breakpoint-*` system by providing max-width media
queries.

Motivation

Currently, Tailwind v4 only supports mobile-first breakpoints via --breakpoint-* variables, which generate min-width media queries. Many designers and teams prefer desktop-first workflows where styles are defined for larger
screens and then overridden for smaller devices.

This is particularly useful for:
- Teams migrating from Webflow or other desktop-first CSS frameworks
- Projects that start with desktop designs and adapt down to mobile
- Situations where max-width queries provide cleaner, more intuitive CSS

Changes

Implementation

- Added --max-breakpoint-* variant support in src/variants.ts
- Creates static variants for each --max-breakpoint-* theme variable
- Generates max-width media queries using the same format as existing max-* variants
- Uses descending sort order (largest breakpoints first) for proper CSS cascade

Usage

Define desktop-first breakpoints in your theme:

```css
@theme {
  --max-breakpoint-phone: 501px;
  --max-breakpoint-tablet: 1001px;
  --max-breakpoint-desktop: 1441px;
}
```

Use them with intuitive, semantic class names:

```html
<!-- Desktop-first: starts as flex-row, becomes flex-col on phones -->
<div class="flex flex-row phone:flex-col">
  <div>Content</div>
  <div>Sidebar</div>
</div>

<!-- Mixed approach: combine with existing min-width breakpoints -->
<div class="md:grid md:grid-cols-2 tablet:block phone:text-center">
  <!-- ... -->  
</div>
```

Generated CSS:

```css
@media not all and (min-width: 1441px) {
  .desktop\:hidden { display: none; }
}

@media not all and (min-width: 1001px) {
  .tablet\:block { display: block; }
}

@media not all and (min-width: 501px) {
  .phone\:flex-col { flex-direction: column; }
}
```

Key Features

- Semantic naming: Allow the use of meaningful breakpoint names like phone, tablet, desktop instead of size-based names
- Desktop-first workflow: Define base styles for large screens, override for smaller devices
- Seamless integration: Works alongside existing `--breakpoint-*` variables
- Consistent behavior: Uses the same media query format and parsing logic as existing `@max-width` variants
- Proper ordering: Automatically sorts breakpoints in descending order for optimal CSS cascade

Testing

Added comprehensive test coverage including:
- Basic functionality with custom breakpoint names
- Breakpoint ordering (descending for max-width queries)
- Mixed usage with existing min-width breakpoints
- Edge cases and proper CSS generation

All existing tests continue to pass, ensuring no breaking changes.

Backward Compatibility

This is a purely additive change that doesn't affect existing functionality:
- Existing `--breakpoint-*` variables continue to work exactly as before
- No changes to existing API or behavior
- Can be used alongside existing breakpoint system

Example Use Cases

Webflow-style migration:
```css
@theme {
  --max-breakpoint-tablet: 991px;
  --max-breakpoint-mobile-landscape: 767px;
  --max-breakpoint-mobile-portrait: 479px;
}
```

Component-specific responsive design:

```html
<nav class="desktop:flex desktop:justify-between tablet:block mobile:text-center">
<!-- Navigation that stacks on smaller screens -->
</nav>

Mixed responsive strategies:
<div class="lg:grid lg:grid-cols-3 tablet:grid tablet:grid-cols-2 mobile:block">
<!-- Hybrid approach using both min and max width breakpoints -->
</div>
```

This feature provides developers with more flexibility in their responsive design workflows while maintaining Tailwind's utility-first philosophy and consistent API patterns.

### DIFF
--- a/packages/tailwindcss/src/variants.test.ts
+++ b/packages/tailwindcss/src/variants.test.ts
@@ -877,6 +877,109 @@ test('custom breakpoint', async () => {
   `)
 })
 
+test('max-breakpoint (desktop-first)', async () => {
+  expect(
+    await compileCss(
+      css`
+        @theme {
+          --max-breakpoint-phone: 501px;
+          --max-breakpoint-tablet: 1001px;
+        }
+        @tailwind utilities;
+      `,
+      ['phone:flex-col', 'tablet:flex-wrap'],
+    ),
+  ).toMatchInlineSnapshot(`
+    "@media not all and (min-width: 1001px) {
+      .tablet\\:flex-wrap {
+        flex-wrap: wrap;
+      }
+    }
+
+    @media not all and (min-width: 501px) {
+      .phone\\:flex-col {
+        flex-direction: column;
+      }
+    }"
+  `)
+})
+
+test('max-breakpoint ordering', async () => {
+  expect(
+    await compileCss(
+      css`
+        @theme {
+          /* Test ordering - larger values should come first in DESC order */
+          --max-breakpoint-mobile: 600px;
+          --max-breakpoint-tablet: 1200px;
+          --max-breakpoint-desktop: 1800px;
+        }
+        @tailwind utilities;
+      `,
+      ['mobile:flex', 'tablet:flex', 'desktop:flex'],
+    ),
+  ).toMatchInlineSnapshot(`
+    "@media not all and (min-width: 1800px) {
+      .desktop\\:flex {
+        display: flex;
+      }
+    }
+
+    @media not all and (min-width: 1200px) {
+      .tablet\\:flex {
+        display: flex;
+      }
+    }
+
+    @media not all and (min-width: 600px) {
+      .mobile\\:flex {
+        display: flex;
+      }
+    }"
+  `)
+})
+
+test('mixed breakpoints and max-breakpoints', async () => {
+  expect(
+    await compileCss(
+      css`
+        @theme {
+          --breakpoint-sm: 640px;
+          --breakpoint-md: 768px;
+          --max-breakpoint-phone: 501px;
+          --max-breakpoint-tablet: 1001px;
+        }
+        @tailwind utilities;
+      `,
+      ['sm:flex', 'phone:flex-col', 'md:grid', 'tablet:hidden'],
+    ),
+  ).toMatchInlineSnapshot(`
+    "@media (min-width: 640px) {
+      .sm\\:flex {
+        display: flex;
+      }
+    }
+
+    @media (min-width: 768px) {
+      .md\\:grid {
+        display: grid;
+      }
+    }
+
+    @media not all and (min-width: 1001px) {
+      .tablet\\:hidden {
+        display: none;
+      }
+    }
+
+    @media not all and (min-width: 501px) {
+      .phone\\:flex-col {
+        flex-direction: column;
+      }
+    }"
+  `)
+})
+
 test('max-*', async () => {
   expect(
     await compileCss(
--- a/packages/tailwindcss/src/variants.ts
+++ b/packages/tailwindcss/src/variants.ts
@@ -1019,6 +1019,55 @@ export function createVariants(theme: Theme): Variants {
       )
     }
 
+    // Max breakpoints (desktop-first)
+    {
+      let maxBreakpoints = theme.namespace('--max-breakpoint')
+      let resolvedMaxBreakpoints = new DefaultMap((variant: Variant) => {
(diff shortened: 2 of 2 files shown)


rozsazoltan:
Related: https://github.com/tailwindlabs/tailwindcss/discussions/18916

RobinMalfait:
Hey!

Thanks for the PR, but going to say no to this one for now. With Tailwind CSS v4 we aim to simplify things, and we explicitly didn't port the min/max/raw media queries configuration.

In v4, you would configure your breakpoints like normal, but if you want them to act as max-width queries to make them desktop first instead of mobile first then you would prefix the variant with `max-*`.

This way you can have both a `min-sm` and a `max-sm` and make everything explicit where you use it.

In your case, you can then define your breakpoint values as:
```diff
  @theme {
-   --max-breakpoint-phone: 501px;
-   --max-breakpoint-tablet: 1001px;
-   --max-breakpoint-desktop: 1441px;
+   --breakpoint-phone: 500px;
+   --breakpoint-tablet: 1000px;
+   --breakpoint-desktop: 1440px;
  }
```
Another small benefit is that you can define the real breakpoints, instead of `breakpoint + 1px`.

Next you would use it as `max-phone:…` or `desktop:` if you want to target 1440px and up.

If more people want this functionality then we can rethink this, but you can already achieve it by being explicit where you use the variant instead. This way we keep things simple and reduce the amount of ways you can do a particular thing.

Thanks again for the PR!

madasebrof:
Okay--no worries. Thanks for the feedback.

I might publish this to NPM for myself, as this functionality is critical to an app I am building (a website builder), and I find the mobile-first way of thinking about building websites fundamentally confusing and unintuitive. But then again, maybe `max-phone` is okay... kinda weird, but I do hate to fork...

I honestly feel like this the one and only "design choice" of tailwind I disagree with, and one decision that Webflow got right. I just feel like it's very unintuitive to think about responsive web design where you add thing with the word `sm` to mean `this applies to anything that is LARGER than sm`. 

Like think if we sized clothes that way. If a `small` meant, `this shirt fits anyone that is BIGGER than a small person`. It would be super bizarre. In clothing, `small` means, this will fit anyone that is `smaller` than this given size, and `medium` mean it will fit anyone that is a `medium` size or less. If you are driving under a tunnel, you know it has a maximum height and as long as you are that height or less, you are okay.

Thus, `phone` intuitively means, this class only applies to things that are phone-sized or smaller, and `tablet` means, this is for things that are tablet sized and smaller. Thus, the intuitive cascade of normal (e.g. no modifier) -> tablet -> phone. If  I didn't know tailwind at all, and I see this `class="text-2xl tablet:text-xl phone:text-md"` I would get the idea. Whereas the inverse `class="text-md sm:text-xl md:text-2xl"`, I would have to invert my natural thinking.

Also, this fits in with how you actually need to design websites.

In website design, mobile design is generally a subtractive process. E.g. you start with a "full site", then you trim back, reduce font sizes, reduce padding/margin & reorganize as you get to smaller and smaller screens. If you work the other way around, e.g. start with mobile screens, it is much more difficult: you are forced to think about the desktop design at the same time as you design the mobile stuff, as you'll need the structure for later, e.g. a hero section.

Make sense???

Anyway, just my two cents. Thanks again for tailwind. It's amazing.

☮️

rozsazoltan:
I think I would stick with `@custom-variant`. You used it incorrectly in the discussion, but I attached working examples.
* https://github.com/tailwindlabs/tailwindcss/discussions/18916#discussioncomment-14369845
* https://github.com/tailwindlabs/tailwindcss/discussions/18916#discussioncomment-14369902

madasebrof:
I feel very foolish. I just realized I had a `<script src="https://cdn.tailwindcss.com"></script>` lurking in my index.html!!! (was loading TW3 as well, which was injecting css after TW4) Literally a day to fix.

So, no need for this pull request, I can just use a custom variant (which is what I started off doing yesterday, but couldn't get it to work!)

```css
@import "tailwindcss";

@source "./**/*.html";
@source "./src/**/*.{js,ts,jsx,tsx}";

@custom-variant tablet (@media (max-width: 1001px));
@custom-variant phone (@media (max-width: 501px));
```

Well, at least I now really get TW4!

Sorry for the fire drill. Thanks again.

PR #18918, opened 2025-09-11
Show Lightning CSS warnings when optimizing/minifying in production

### LINKED ISSUE #15872: v4: @tailwindcss/vite: Invalid css in @theme section causes the styles to be silently broken in production build only

"@tailwindcss/vite": "^4.0.0"

"vite": "^6.0.5"

v22.13.0

Chrome

macOS

**Reproduction URL**

https://github.com/tkafka/tailwind-v4-css-parsing-bug/tree/main

**Describe your issue**

The above demo demonstrates, how a single typo in `@theme` section in `style.css` causes the page to have broken css in production.
The dev build is fine, and there is no error message in production.
This can let to a single typo silently breaking the production.

## Reproduction steps:

1. Install dependencies:

```bash
pnpm install
```

2. Dev build looks fine:

```bash
pnpm dev
```

Image

3. Production build has broken styling, no error is produced:

```bash
pnpm build && pnpm preview
```

Image

This is caused by a typo in `style.css` in `--shadow-inset-bad` - it has invalid `rgb(var(from --color-secondary-500) r g b / 0.08)` instead of `rgb(from var(--color-secondary-500) r g b / 0.08)` (`from` is in the wrong place).

Commenting out the `--shadow-inset-bad` style causes the production build to have a proper styling.

This seems like an error caused by different behavior of css parsing in development vs. in production. I believe that either the parser should work in a same way in development and production mode, or it should report found errors in css.

RobinMalfait:
This PR improves the DX by showing all the Lightning CSS warnings when using a "production" build (or using `--optimize` or `--minify` flags when using the CLI).

Right now Tailwind CSS itself doesn't care about the exact syntax you are using in the CSS as long as it looks valid. We do this because otherwise we would have to parse a lot more CSS syntax and validate it even though it would be valid CSS in 99.99% of the cases.

Even worse, if you want to use newer CSS syntax that Tailwind CSS doesn't validate yet, then you would get warnings for valid CSS.

Another reason why we don't do this is because the browser already does a great job at ignoring invalid CSS.

So the linked issue #15872 would still silently fail in development mode. In this case, everything would work, except the shadow with the invalid syntax.

But in production mode, you would now get a proper warning from Lightning CSS, because they try to optimize the CSS and remove invalid CSS.

One potential issue here is that we run Lightning CSS on the generated CSS, not on the input CSS. So the current output shows the warnings in the output CSS not the input CSS. Any thoughts if we would just skip the line numbers?

## Test plan

1. Everything works as before
2. In production mode, you would get warnings printed to the terminal. This is done in `@tailwindcss/node` so the CLI/Vite/PostCSS plugins would all get the same behavior.

Screenshots:

If you have a single issue:
image

If you have multiple issues:
image

Fixes: #15872

### DIFF
--- a/packages/@tailwindcss-node/src/optimize.ts
+++ b/packages/@tailwindcss-node/src/optimize.ts
@@ -1,6 +1,7 @@
 import remapping from '@jridgewell/remapping'
 import { Features, transform } from 'lightningcss'
 import MagicString from 'magic-string'
+import pc from 'picocolors'
 
 export interface OptimizeOptions {
   /**
@@ -59,6 +60,49 @@ export function optimize(
   // nesting is applied. This creates a more optimized output.
   let result = optimize(Buffer.from(input), map)
   map = result.map?.toString()
+
+  // Because of `errorRecovery: true`, there could be warnings, so let's let the
+  // user know about them.
+  if (process.env.NODE_ENV !== 'test' && result.warnings.length > 0) {
+    let lines = input.split('\n')
+
+    let output = [
+      `Found ${result.warnings.length} ${result.warnings.length === 1 ? 'warning' : 'warnings'} while optimizing generated CSS:`,
+    ]
+
+    for (let [idx, warning] of result.warnings.entries()) {
+      output.push('')
+      if (result.warnings.length > 1) {
+        output.push(`Issue #${idx + 1}:`)
+      }
+
+      let context = 2
+
+      let start = Math.max(0, warning.loc.line - context - 1)
+      let end = Math.min(lines.length, warning.loc.line + context)
+      let gutterWidth = (warning.loc.line + context).toString().length
+
+      let snippet = lines.slice(start, end).map((line, idx) => {
+        if (start + idx + 1 === warning.loc.line) {
+          return `${pc.dim(`${(start + idx + 1).toString().padStart(gutterWidth, ' ')} \u2502`)} ${line}`
+        } else {
+          return pc.dim(`${(start + idx + 1).toString().padStart(gutterWidth, ' ')} \u2502 ${line}`)
+        }
+      })
+
+      snippet.splice(
+        warning.loc.line - start,
+        0,
+        `${' '.repeat(gutterWidth)} ${pc.dim('\u2506')}${' '.repeat(warning.loc.column - 1)} ${pc.yellow(`${pc.dim('^--')} ${warning.message}`)}`,
+        `${' '.repeat(gutterWidth)} ${pc.dim('\u2506')}`,
+      )
+
+      output.push(...snippet)
+    }
+    output.push('')
+
+    console.warn(output.join('\n'))
+  }
 
   result = optimize(result.code, map)
   map = result.map?.toString()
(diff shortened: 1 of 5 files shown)


RobinMalfait:
Update: I dropped the line numbers because I think those are more confusing because they don't actually map to the correct spot in the input. So I think just _some_ indication that something went wrong is already a big improvement.

image

packages/@tailwindcss-node/src/optimize.ts
```diff
@@ -1,6 +1,7 @@
 import remapping from '@jridgewell/remapping'
 import { Features, transform } from 'lightningcss'
 import MagicString from 'magic-string'
+import pc from 'picocolors'
```
philipp-spiess: Can we inline this dependency? https://unpkg.com/picocolors@1.1.1/picocolors.js 

Probably don't even need all of the options here

philipp-spiess (changes requested):
Let's not add a new dependency, especially when it's only like 30 LOC

philipp-spiess:
Also I think we probably want a way to silence these warnings, WDYT? Have you tried running e.g. tailwindcss.com CSS through it? Will it cause any unexpected warnings?

packages/@tailwindcss-node/src/optimize.ts
RobinMalfait: Inlined it 👍 

Only reason I added the dependency is because we were already using it in other packages (`@tailwindcss/cli` and `@tailwindcss/upgrade`). But since we don't use it in the `@tailwindcss/postcss` or `@tailwindcss/vite`, you're right let's just inline them. If they were used there already it's not a big deal.

RobinMalfait:
> Also I think we probably want a way to silence these warnings, WDYT? Have you tried running e.g. tailwindcss.com CSS through it? Will it cause any unexpected warnings?

You sure? Running it on Tailwind Plus or tailwindcss.com doesn't result in any warnings. The warnings also only show up when using Lightning CSS which is typically only when you explicitly have `--minify` or `--optimize` in the CLI, or if you use `NODE_ENV='production'` in PostCSS / Vite land so it won't spam your development workflow.

philipp-spiess:
Yeah that sounds good then 👍

pnpm-lock.yaml
```diff
@@ -6,39 +6,9 @@ settings:
 
 catalogs:
   default:
-    '@types/node':
```
RobinMalfait: Cleaning this up as well because nothing should change here now that I removed the dependency again.

philipp-spiess (approved):

lvzhenbo:
@RobinMalfait 
The victim of this PR has appeared. When I used both the @apply directive and Vue's deep syntax simultaneously, the warning appeared. I need a way to turn off the warning.
image

RobinMalfait:
@lvzhenbo will be fixed by https://github.com/tailwindlabs/tailwindcss/pull/19094

sgarfinkel:
@RobinMalfait Can you add a way to opt out of the warnings? I'd like to turn them off--they are busy and are being caused by transitives that I can't easily patch or fix. Since these log straight to console there's no particularly easy way to filter them.

PR #18922, opened 2025-09-12
Do not migrate `variant = 'outline'` during upgrades

RobinMalfait:
This PR improves the upgrade tool for shadcn/ui projects where the `variant = "outline"` is incorrectly migrated to `variant = "outline-solid"`.

This PR also handles a few more cases:

```ts
// As default argument
function Button({ variant = "outline", ...props }: ButtonProps) { }

// With different kinds of quotes (single, double, backticks)
function Button({ variant = 'outline', ...props }: ButtonProps) { }

// Regardless of whitespace
function Button({ variant="outline", ...props }: ButtonProps) { }

// In JSX
<Button variant="outline" />

// With different quotes and using JavaScript expressions
<Button variant={'outline'} />

// As an object property
buttonVariants({ variant: "outline" })
```


philipp-spiess (approved):
FINE

PR #18938, opened 2025-09-15
Fix native ESM config loading in v3

### LINKED ISSUE #14152: Upgrade dependency postcss-load-config to support postcss configs that use ESM with TypeScript

3.4.3

vite 5.2.11 (@ electron-forge 7.4.0 x electron 30.0.2)
just using the vite layer to debug (`npx vite`)

22.1.0

N/A

win11

**Reproduction URL**

https://github.com/aentwist/tailwind-upgrade-postcss-repro

**Describe your issue**

My project cannot handle postcss config that uses ESM and TypeScript. Apparently there was a fix for this in v5 https://github.com/postcss/postcss-load-config/issues/239.

```
$ npm why postcss-load-config
postcss-load-config@4.0.2 dev
node_modules/postcss-load-config
  postcss-load-config@"^4.0.1" from tailwindcss@3.4.3
  node_modules/tailwindcss
    dev tailwindcss@"^3.4.3" from the root project

https://github.com/tailwindlabs/tailwindcss/blob/f65023efb97832660dc17cf954504f9f156047ba/package.json#L84

Actually though, [they are now on v6](https://www.npmjs.com/package/postcss-load-config?activeTab=versions).

The breaking changes are very light - a [node upgrade](https://github.com/postcss/postcss-load-config/blob/824a7de5f625310505504b2d79526ee8d608194f/CHANGELOG.md#breaking-changes-1) that drops support for versions of node that have been [unmaintained for a year minimum now](https://nodejs.org/en/about/previous-releases#release-schedule) [v5], and [requiring yaml users to install their own yaml dependency](https://github.com/postcss/postcss-load-config/blob/824a7de5f625310505504b2d79526ee8d608194f/CHANGELOG.md#breaking-changes) [v6].

I see that unfortunately tailwind still supports old versions of node (14) https://github.com/tailwindlabs/tailwindcss/blob/f65023efb97832660dc17cf954504f9f156047ba/package.json#L116, so I guess this *is* a breaking change. In that case, may as well upgrade all the way to whatever latest is when the opportunity comes.

---

For now, I will use commonjs and TypeScript, by adding a .cts extension to the file. This seems to work fine, but I know that for some reason in these configurations the postcss config file produces no warning if not present, however if present produces a seemingly related warning when running the app (`vite`) and loading the page.

```
(node:19428) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
```

I was hoping that just supporting ESM and TypeScript would make it go away. Sadly for completeness, I cannot figure how to pass that node option through vite to get the full trace.

My postcss config is the recommended starter,

```ts
export default {
  plugins: {
    tailwindcss: {},
    autoprefixer: {},
  },
};

```

I've tried various combinations of file extension and export syntax with `"type": "module"` and cannot make this warning disappear. For TS, I'm using `"module": "nodenext"`.

### LINKED ISSUE #14423: import.meta.resolve doesn't work in the tailwind.config file

v3.4.10

vite 5.3.4, vue 3.4.31

v20.12.0

macOS

[Reproduction URL](https://github.com/mkeyy0/tailwind-import-meta-issue)

**Describe your issue**

Tailwind doesn't support ESM natively and uses some tools to achieve that. And for some properties of `import.meta` like `url` works nice, but in my case, I need to use `import.meta.resolve` to resolve the path to my monorepo UI library in `node_modules`.
I know that I can use `require.resolve` as a workaround, but I'm not sure that is a good way to mix esm and cjs globals. So for me, it would be nice to add support of `import.meta.resolve` to the `tailwind.config` file

thecrypticace:
Unfortunately for backwards compatibility purposes (with `loadConfig` at least) we can't switch things to use `import(…)` because there's baked in knowledge that the config is loaded synchronously for v3.

This PR does two things:
- Defers to `require(…)` which allows newer versions that support `require(esm)` to work natively. This works around the need to switch to `import(…)` for those versions.
- Allows newer versions of `postcss-load-config` enabling better ESM+TypeScript support for PostCSS configs in the CLI.

We support v4, v5, and v6 of `postcss-load-config` simultaneously so any of those versions should work. I've verified that newer node versions seem to install v6 while earlier ones like Node v14 install v4 of `postcss-load-config`. So this should be a backwards compatible change.

- [x] needs tests for `import.meta.resolve(…)`
- [x] needs tests for ESM postcss configs

Fixes #14152
Fixes #14423

### DIFF
--- a/src/cli/build/plugin.js
+++ b/src/cli/build/plugin.js
@@ -43,10 +43,13 @@ async function loadPostCssPlugins(customPostCssPath) {
           config.plugins = []
         }
 
+        // We have to await these because in v5 and v6 of postcss-load-config
+        // these functions return promises while they don't in v4. Awaiting a
+        // non-promise is basically a no-op so this is safe to do.
         return {
           file,
-          plugins: loadPlugins(config, file),
-          options: loadOptions(config, file),
+          plugins: await loadPlugins(config, file),
+          options: await loadOptions(config, file),
         }
       })()
     : await postcssrc()
--- a/src/lib/load-config.ts
+++ b/src/lib/load-config.ts
@@ -35,23 +35,8 @@ export function loadConfig(path: string): Config {
   let config = (function () {
     if (!path) return {}
 
-    // Always use jiti for now. There is a a bug that occurs in Node v22.12+
-    // where imported files return invalid results
-    return lazyJiti()(path)
-
-    // Always use jiti for ESM or TS files
-    if (
-      path &&
-      (path.endsWith('.mjs') ||
-        path.endsWith('.ts') ||
-        path.endsWith('.cts') ||
-        path.endsWith('.mts'))
-    ) {
-      return lazyJiti()(path)
-    }
-
     try {
-      return path ? require(path) : {}
+      return require(path)
     } catch {
       return lazyJiti()(path)
     }
--- a/.github/workflows/ci-stable.yml
+++ b/.github/workflows/ci-stable.yml
@@ -27,7 +27,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 18]
+        node-version: [14, 18, 20, 22, 24]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 18]
+        node-version: [14, 18, 20, 22, 24]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -31,7 +31,7 @@ jobs:
           - vite
           - webpack-4
           - webpack-5
-        node-version: [18]
+        node-version: [18, 20]
       fail-fast: false
 
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve support for raw `supports-[…]` queries in arbitrary values ([#13605](https://github.com/tailwindlabs/tailwindcss/pull/13605))
 - Fix `require.cache` error when loaded through a TypeScript file in Node 22.18+ ([#18665](https://github.com/tailwindlabs/tailwindcss/pull/18665))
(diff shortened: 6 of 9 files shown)


src/lib/load-config.ts
```diff
@@ -35,18 +35,10 @@ export function loadConfig(path: string): Config {
...
-      path &&
-      (path.endsWith('.mjs') ||
-        path.endsWith('.ts') ||
-        path.endsWith('.cts') ||
-        path.endsWith('.mts'))
-    ) {
+    // Always use jiti for TS files
+    // Newer versions of Node support `require(esm)` so we'll try that first
+    // in that case.
+    if (path && (path.endsWith('.ts') || path.endsWith('.cts') || path.endsWith('.mts'))) {
```
thecrypticace: Since Node now has TypeScript support (for type stripping at least) should we drop this check too?

thecrypticace: @RobinMalfait @philipp-spiess any thoughts on this? I kinda think we should drop this check entirely and only fallback to Jiti if loading fails.

src/cli/build/plugin.js
```diff
@@ -43,10 +43,13 @@ async function loadPostCssPlugins(customPostCssPath) {
...
 
+        // We have to await these because in v5 and v6 of postcss-load-config
+        // these functions return promises while they don't in v4. Awaiting a
+        // non-promise is basically a no-op so this is safe to do.
         return {
           file,
-          plugins: loadPlugins(config, file),
-          options: loadOptions(config, file),
+          plugins: await loadPlugins(config, file),
+          options: await loadOptions(config, file),
```
RobinMalfait, line 52: Thank god this function was already async!

src/lib/load-config.ts
RobinMalfait: I think we should drop this entire block 🤔 

I feel like the block below should be sufficient:
```ts
    try {
      return path ? require(path) : {}
    } catch {
      return lazyJiti()(path)
    }
```
Essentially always rely on `require`, and if that fails for _some_ reason, use `lazyJiti`

But I vaguely remember there was a reason why we didn't do that. Was it related to ESM where modules now would have a `.default` key or something 🤔

RobinMalfait (approved):
Looks good, but I think we can drop that extension check.

src/lib/load-config.ts
thecrypticace: I think it was just a case of "Node probably isn't gonna support these so we can just always use Jiti to save a bit of time". But that's not the case anymore so…

thecrypticace: Updated 👍

PR #18940, opened 2025-09-15
Add plugin option documentation to the postcss plugin readme

### LINKED ISSUE #2061: The parameters for the @tailwindcss/postcss plugin are missing
The documentation does not mention the parameters for the `@tailwindcss/postcss` plugin. At least, when searching for "post css" in the search bar, the page only offers the installation guide.

```ts
export type PluginOptions = {
  // The base directory to scan for class candidates.
  base?: string

  // Optimize and minify the output CSS.
  optimize?: boolean | { minify?: boolean }
}
```
[Permalink: `@tailwindcss/postcss/src/idnex.ts#L44-L50`](https://github.com/tailwindlabs/tailwindcss/blob/837e240b4d8474fae786cda8eaf9ae3f76a52a75/packages/@tailwindcss-postcss/src/index.ts#L44-L50)

thecrypticace:
Closes https://github.com/tailwindlabs/tailwindcss.com/issues/2061

In lieu of finding a place to put this on the docs we can at least document this in the readme for the plugin.

### DIFF
--- a/packages/@tailwindcss-postcss/README.md
+++ b/packages/@tailwindcss-postcss/README.md
@@ -34,3 +34,97 @@ For help, discussion about best practices, or feature ideas:
 ## Contributing
 
 If you're interested in contributing to Tailwind CSS, please read our [contributing docs](https://github.com/tailwindcss/tailwindcss/blob/next/.github/CONTRIBUTING.md) **before submitting a pull request**.
+
+---
+
+## `@tailwindcss/postcss` plugin API
+
+### Changing where the plugin searches for source files
+
+You can use the `base` option (defaults to the current working directory) to change the directory in which the plugin searches for source files:
+
+```js
+import tailwindcss from "@tailwindcss/postcss"
+
+export default {
+ plugins: [
+  tailwindcss({
+    base: path.resolve(__dirname, "./path)
+  })
+ ]
+}
+```
+
+### Enabling or disabling Lightning CSS
+
+By default, this plugin detects whether or not the CSS is being built for production by checking the `NODE_ENV` environment variable. When building for production Lightning CSS will be enabled otherwise it is disabled.
+
+If you want to always enable or disable Lightning CSS the `optimize` option may be used:
+
+```js
+import tailwindcss from "@tailwindcss/postcss"
+
+export default {
+ plugins: [
+  tailwindcss({
+    // Disables Lightning CSS
+    optimize: false,
+
+    // Disables Lightning CSS
+    optimize: true,
+  })
+ ]
+}
+```
+
+You may also pass options to `optimize` to enable Lighting CSS but prevent minification:
+
+```js
+import tailwindcss from "@tailwindcss/postcss"
+
+export default {
+ plugins: [
+  tailwindcss({
+    // Enables Lightning CSS but disables minification
+    optimize: { minify: false },
+  })
+ ]
+}
+```
+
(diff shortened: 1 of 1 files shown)


packages/@tailwindcss-postcss/README.md
```diff
@@ -34,3 +34,97 @@ For help, discussion about best practices, or feature ideas:
...
+```js
+import tailwindcss from "@tailwindcss/postcss"
+
+export default {
+ plugins: [
+  tailwindcss({
+    // Disables Lightning CSS
+    optimize: false,
+
+    // Disables Lightning CSS
```
philipp-spiess:
```suggestion
    // Enables Lightning CSS
```

but also it seems a bit wasted to include both variations since it's a boolean it's kinda inferred that true does the opposite

packages/@tailwindcss-postcss/README.md
```diff
@@ -34,3 +34,97 @@ For help, discussion about best practices, or feature ideas:
...
+    transformAssetUrls: false,
+
+    // Enable `url(…)` rewriting (the default)
+    transformAssetUrls: true,
+  })
+ ]
+}
+```
+
+You may also pass options to `optimize` to enable Lighting CSS but prevent minification:
```
philipp-spiess, line 113: feels like this should be under the `Enabling or disabling Lightning CSS` heading since it's the same field

packages/@tailwindcss-postcss/README.md
```diff
@@ -34,3 +34,85 @@ For help, discussion about best practices, or feature ideas:
...
+ plugins: [
+  tailwindcss({
+    // Disables Lightning CSS
+    optimize: false,
+
+    // Enables Lightning CSS
+    optimize: true,
+
+    // Enables Lightning CSS but disables minification
+    optimize: { minify: false },
```
philipp-spiess, line 85: This still feels a bit off to me. It's trying to be a valid code example with all the chrome around it but then defining the same key three times. Is this a pattern we've used before? 

If not, my proposal is that we split this up into two full examples and instead of having one for `true` vs. `false` we just document one or the other with the same title "If you want to always enable or disable Lightning CSS the `optimize` option may be used:"

WDYT?

thecrypticace: I guess I don't understand what you mean by this then?
> feels like this should be under the Enabling or disabling Lightning CSS heading since it's the same field

packages/@tailwindcss-postcss/README.md
```diff
@@ -34,3 +34,85 @@ For help, discussion about best practices, or feature ideas:
...
+  })
+ ]
+}
+```
```
philipp-spiess:
```suggestion
If you want to always enable or disable Lightning CSS the `optimize` option may be used:

```js
import tailwindcss from "@tailwindcss/postcss"

export default {
 plugins: [
  tailwindcss({
    // Enable or disable Lightning CSS
    optimize: false,
  })
 ]
}
```

It's also possible to keep Lightning CSS enabled but disable minification:

```js
import tailwindcss from "@tailwindcss/postcss"

export default {
 plugins: [
  tailwindcss({
    optimize: { minify: false },
  })
 ]
}
```

---- 


@thecrypticace That's kinda what I mean. Quite some reputation, I give you that though!

philipp-spiess (approved):

PR #18948, opened 2025-09-16
Use `default` export condition for `@tailwindcss/vite`

bvandrc:
## Summary

In `@tailwindcss/vite` 's `package.json`, change the `exports` key from `include` to `default` since there is no `require` case.

Ran into an issue using the `tsx` package to run a script that has a sub-dependency that imports from `@tailwindcss/vite`, where `tsx` converts things to cjs to run, and since there is no `require` case for this package, it can't find the file. Changing to `default` covers the cases for both `import` and `require`.

## Test plan

No testing needed. Functionality is the same.


thecrypticace:
Wouldn't this be a tsx bug? I did some testing and it does fail to load but if you're in a `.mts` file it works fine.

thecrypticace:
Ah I guess `default` should be preferred now that `require(esm)` is a thing since `import` and `require` are mutually exclusive.

We could also use `module-sync` but my guess is `default` should be the preferred option now.

thecrypticace (approved):

PR #18960, opened 2025-09-17
Bump Prettier

RobinMalfait:
This PR bumps prettier and solves one of our `- *` formatting issues. Not all, but a few!

### DIFF
--- a/packages/tailwindcss/src/intellisense.test.ts
+++ b/packages/tailwindcss/src/intellisense.test.ts
@@ -496,7 +496,7 @@ test('Custom functional @utility', async () => {
 
     @utility example-* {
       font-size: --value(--text);
-      line-height: --value(--text- * --line-height);
+      line-height: --value(--text-* --line-height);
       line-height: --modifier(--leading, 'normal');
     }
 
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/intellisense.test.ts
```diff
@@ -513,7 +513,7 @@ test('Custom functional @utility', async () => {
     }
 
     @utility -negative-* {
-      margin: --value(--tab-size- *);
+      margin: --value(--tab-size-*);
```
RobinMalfait, line 516: This is fixed which is nice

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -27897,7 +27897,7 @@ describe('custom utilities', () => {
     test('resolving any arbitrary values', async () => {
       let input = css`
         @utility tab-* {
-          tab-size: --value([ *]);
+          tab-size: --value([*]);
```
RobinMalfait, line 27900: This is also fixed, which is also nice

packages/tailwindcss/src/intellisense.test.ts
```diff
@@ -496,7 +496,7 @@ test('Custom functional @utility', async () => {
 
     @utility example-* {
       font-size: --value(--text);
-      line-height: --value(--text- * --line-height);
+      line-height: --value(--text-* --line-height);
```
RobinMalfait, line 499: This is better, but not really worse compared to before

PR #19028, opened 2025-09-29
Show version mismatch when running upgrade tool

RobinMalfait:
This PR fixes an issue where sometimes people try to run the upgrade tool, reset the changes and then try again.

If this happens, then the `package.json` and/or your lock file will point to the old Tailwind CSS v3 version, but the actual installed version will be v4.

This will also cause the upgrade tool to now upgrade from v4 to v4, which is not what most people want if they were trying to upgrade from v3 to v4. This in turn will cause some issues because now we won't try to migrate the config file, or v3-specific classes that also exist in v4 but are only safe to upgrade from v3 to v4.

This PR uses `npm ls tailwindcss` to determine the actual installed version. This command already errors if there is a mismatch between the installed version and the version in `package.json` or the lock file. This also happens to work in pnpm and bun projects (added integration tests for these). 

If for whatever reason we can't determine the expected version, we fall back to the old behavior of just upgrading. In this scenario, the changes introduced in https://github.com/tailwindlabs/tailwindcss/pull/19026 will at least give you a hint of what version was actually installed.

### Test plan

1. Tested it in a v3 project where I performed the following steps:
   1. Run the upgrade tool in full (`npx tailwindcss-upgrade`)
   2. Reset the changes (`git reset --hard && git clean -df`)
   1. Run the upgrade tool again

   This resulted in the following output: image

2. Added some integration tests to make sure this also works in pnpm, bun and normal npm projects.

[ci-all]

### DIFF
--- a/packages/@tailwindcss-upgrade/src/index.ts
+++ b/packages/@tailwindcss-upgrade/src/index.ts
@@ -4,6 +4,7 @@ import { Scanner } from '@tailwindcss/oxide'
 import { globby } from 'globby'
 import fs from 'node:fs/promises'
 import path from 'node:path'
+import pc from 'picocolors'
 import postcss from 'postcss'
 import { migrateJsConfig } from './codemods/config/migrate-js-config'
 import { migratePostCSSConfig } from './codemods/config/migrate-postcss'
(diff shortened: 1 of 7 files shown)


packages/@tailwindcss-upgrade/src/index.ts
```diff
@@ -62,6 +63,27 @@ async function run() {
...
+        `${pc.red('-')} ${`${pc.dim('"tailwindcss":')} ${`${pc.dim('"')}${pc.blue(version.expectedTailwindVersion(base))}${pc.dim('"')}`}`} (expected version in package.json / lockfile)`,
+        `${pc.green('+')} ${`${pc.dim('"tailwindcss":')} ${`${pc.dim('"')}${pc.blue(version.installedTailwindVersion(base))}${pc.dim('"')}`}`} (installed version in \`node_modules\`)`,
+        pc.dim('```'),
+        '',
+        `Make sure to run ${highlight(`${pkgManager} install`)}, and try again.`,
+      ].join('\n'),
+      {
+        prefix: '↳ ',
+      },
+    )
```
RobinMalfait, line 83: Open to suggestions here!

thecrypticace: I *think* i would say "Tailwind CSS version mismatch" but other than that this seems fine to me.

thecrypticace (approved):

packages/@tailwindcss-upgrade/src/index.ts
thecrypticace: aside: I don't think it *needs* to change since `tailwindcss` is also prevent in the diff so if you don't wanna change it that's fine imo. 🤷‍♂️

RobinMalfait: I think this is a good change if we want to show more issues. But I used the '↳' as well because it's sort of a continuation of the previous Tailwind CSS version

PR #19030, opened 2025-09-29
v3: Remove irrelevant utility rules when matching important classes

### LINKED ISSUE #18678: Error: "Parsing css source code failed" in Tailwind V3

TailwindCSS: v3.4.17

Nextjs: 15.4.5
React: 19.1.0

NodeJS: 22

Browser: Chrome

OS: macOS

**Reproduction URL**

https://github.com/NishargShah/tailwindcss-v3-issue

**Describe your issue**

Hi Tailwind team,

I’m working on a client project where I had to change **padding values for large screens (min-width: 1728px)** using a custom plugin in `tailwind.config.ts`.

Everything was working fine until I used the `--turbopack` flag in the dev script.

After many hours of debugging, I found a specific issue that causes the development server to break. I also created a **minimal repo to reproduce the bug**.

---

### 🧩 What I Did

1. In my `tailwind.config.ts`, I added a plugin to support custom padding values.

2. In `page.tsx` at 7th line I used the class `!pt-1.25`.

    Please take a note, it only reproduce when important (!) is there, without important it will work as aspected.

3. Then I ran this command to generate CSS:

   ```bash
   npx tailwindcss -i ./src/styles/globals.css -o ./src/styles/output.css
   ```

4. When I start the Next.js dev server with `--turbopack`:

   ```bash
   next dev --turbopack
   ```

   I get a **Next.js error screen**, and the server fails to load the app.

5. But if I **remove `!pt-1.25`**, everything works perfectly.

6. Also, if I run the dev server **without `--turbopack`**, it works fine.

---

#### Error in CSS File:

Browser error screen

#### Error from dev server:

Next.js error screen

---

### 🔍 What I Tried

* I thought it might be a **PostCSS issue**, but not sure.
* After testing multiple cases, the **main issue is with custom value `1.25` and `--turbopack` in dev mode**.
* This issue does **not** happen in production builds, only during dev with `--turbopack`.

---

### ✅ Temporary Workaround

If I remove the `--turbopack` flag from the script, like this:

```bash
next dev
```

…it works fine. But I prefer to use `--turbopack` for better speed.

---

### 📦 Possible Cause?

It might be that:

* `1.25` is not being processed correctly in some part of the pipeline (maybe PostCSS or internal Tailwind logic).
* Or maybe `--turbopack` in Next.js is strict and crashes when there CSS syntax error.

---

Please check if:

* This is a Tailwind bug or something from Next.js Turbo integration.
* Any workaround/fix is possible for supporting custom values like `1.25`.

Thanks a lot for your great work and support 🙌

thecrypticace:
Fixes #18678

When given a simple plugin like this:
```js
export default {
  content: [{ raw: '!a' },],
  plugins: [
    ({ addBase }) => addBase({
      '@media (min-width: 1728px)': {
        '.a': { 'padding-top': '1rem !important' },
        '.b': { 'padding-right': '1rem !important' },
      }
    }),
  ],
}
```

If this plugin saw the utility `!a` it would correctly modify both rules to eliminate any irrelevant classes. Unfortunately, in the case of the 2nd rule this meant it's left with an empty selector which is invalid.

We now detect this case and remove the rule if that happens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `require.cache` error when loaded through a TypeScript file in Node 22.18+ ([#18665](https://github.com/tailwindlabs/tailwindcss/pull/18665))
 - Support `import.meta.resolve(…)` in configs for new enough Node.js versions ([#18938](https://github.com/tailwindlabs/tailwindcss/pull/18938))
 - Allow using newer versions of `postcss-load-config` for better ESM and TypeScript PostCSS config support with the CLI ([#18938](https://github.com/tailwindlabs/tailwindcss/pull/18938))
+- Remove irrelevant utility rules when matching important classes ([#19030](https://github.com/tailwindlabs/tailwindcss/pull/19030))
 
 ## [3.4.17] - 2024-12-17
 
--- a/src/lib/generateRules.js
+++ b/src/lib/generateRules.js
@@ -143,7 +143,14 @@ function applyImportant(matches, classCandidate) {
         className === classCandidate ? `!${className}` : className
       )
 
-      r.selector = ast.toString()
+      let newSelector = ast.toString()
+
+      if (newSelector.trim() === '') {
+        r.remove()
+        return
+      }
+
+      r.selector = newSelector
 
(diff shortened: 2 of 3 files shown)


RobinMalfait (approved):
Nice, yep this would be bad:

Image

PR #19031, opened 2025-09-29
Ensure first class inside className in React is migrated

govizlora:
### 1. Handling React className migration

The PR fixes an issue when migrating React components to tailwind v4 with the migration tool, that the first class after `className="` is ignored.

For example, when migrating
```JSX
<div className="shadow"></div>
```
`shadow` will not be migrated to `shadow-sm` .

This is because in `is-safe-migration.ts`, it tests the line before candidate with regex `/(?<!:?class)=['"]$/`. This basically skips the migration for anything like `foo="shadow"`, with only exception for Vue (eg. `class="shadow"`).

The PR changes the regex from

```regex
/(?<!:?class)=['"]$/
````
to
```regex
/(?<!:?class|className)=['"]$/
```
which essentially adds a new exception specifically for React's `className="shadow"` case.

### 2. Removing redundant rules

Besides, I found that several other rules in `CONDITIONAL_TEMPLATE_SYNTAX` being redundant since they are already covered by the rule above, so I removed them. If we prefer the previous explicit approach, I can revert it.

## Test plan

Tests added for both the Vue and React classes to prevent false negative cases.


thecrypticace (approved):

thecrypticace:
Thanks!

PR #19038, opened 2025-09-30
Prepare v3.4.18 release

thecrypticace:


thecrypticace:
We're still running some CI stuff on Node 14. I think we can probably drop those. What do you think @RobinMalfait? Like I don't see any reason do run Node CI on anything < Node 20 *really*.

RobinMalfait:
That seems fine, I think we can run CI on the latest LTS and newer.

Only thing we have to make sure is that if we build/publish with an older node version that the generated files don't change when changing it to a newer version. If the artifacts did change, then maybe we bump to the Node 18 or LTS?

RobinMalfait (approved):

thecrypticace:
Okay, I'll set that up after we do the release.

PR #19054, opened 2025-10-04
Add interactive diagram for tailwindcss 

Abdullah85MBA:
## Summary

I was trying to understand how tailwindcss works under the hood so I created a diagram for it using [code-canvas.com](http://code-canvas.com). I think it will be pretty helpful for contributors to quickly onboard to the inner-works of tailwindcss.

Here is the generated diagram for tailwindcss grouped by the application's use-cases and their runtime data flows: [Open interactive diagram ->](https://www.code-canvas.com/?session=unauthenticatedGithub&repo=tailwindcss&owner=tailwindlabs&branch=main&OnboardingTutorial=true)

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@
 
 For full documentation, visit [tailwindcss.com](https://tailwindcss.com).
 
+### How does `tailwindcss` work?
+
+Check out this interactive walkthrough of the `tailwindcss` codebase on CodeCanvas [here](https://www.code-canvas.com/?session=unauthenticatedGithub&repo=tailwindcss&owner=tailwindlabs&branch=main&OnboardingTutorial=true).
+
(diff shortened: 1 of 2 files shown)


eveeifyeve:
why not https://mermaid.js.org ?

RobinMalfait:
Hey! I appreciate the PR but I don't think we need something like this. The service you linked to took *minutes* to generate something and I wonder if this actually useful to anyone trying to contribute, not just cool looking. This is what it generated after a few minutes: 
image

It also included words like `JIT`, which is a term we don't use in this codebase anymore. So I think it's going to be more confusing and I think it's also going to be much more clear if you look at the actual code in your editor and jump to definitions and follow function calls instead.

If we ever need some visual on how everything works, I think we'd just generate an SVG or use mermaid directly in the README or CONTRIBUTING markdown files.

Going to close this PR for now.

PR #19059, opened 2025-10-06
Introduce `canonicalizeCandidates` on the internal Design System

RobinMalfait:
This PR introduces a new `canonicalizeCandidates` function on the internal Design System.

The big motivation to moving this to the core `tailwindcss` package is that we can use this in various places:

- The Raycast extension
- The VS Code extension / language server
- 3rd party tools that use the Tailwind CSS design system APIs

> This PR looks very big, but **I think it's best to go over the changes commit by commit**. Basically all of these steps already existed in the upgrade tool, but are now moved to our core `tailwindcss` package.

Here is a list of all the changes:
- Added a new `canonicalizeCandidates` function to the design system
- Moved various migration steps to the core package. I inlined them in the same file and because of that I noticed a specific pattern (more on this later).
- Moved `printCandidate` tests to the `tailwindcss` package
- Setup tests for `canonicalizeCandidates` based on the existing tests in the upgrade tool.

I noticed that all the migrations followed a specific pattern:

1. Parse the raw candidate into a `Candidate[]` AST
2. In a loop, try to migrate the `Candidate` to a new `Candidate` (this often handled both the `Candidate` and its `Variant[]`)
3. If something changed, print the new `Candidate` back to a string, and pass it to the next migration step.

While this makes sense in isolation, we are doing a lot of repeated work by parsing, modifying, and printing the candidate multiple times. This let me to introduce the `big refactor` commit. This changes the steps to:

1. Up front, parse the raw candidate into a `Candidate[]` _once_.
2. Strip the variants and the important marker from the candidate. This means that each migration step only has to deal with the base `utility` and not care about the variants or the important marker. We can re-attach these afterwards.
3. Instead of a `rawCandidate: string`, each migration step receives an actual `Candidate` object (or a `Variant` object).
4. I also split up the migration steps for the `Candidate` and the `Variant[]`.

All of this means that there is a lot less work that needs to be done. We can also cache results between migrations. So `[@media_print]:flex` and `[@media_print]:block` will result in `print:flex` and `print:block` respectively, but the `[@media_print]` part is only migrated once across both candidates.

One migration step relied on the `postcss-selector-parser` package to parse selectors and attribute selectors. I didn't want to introduce a package just for this, so instead used our own `SelectorParser` in the migration and wrote a small `AttributeSelectorParser` that can parse the attribute selector into a little data structure we can work with instead.

If we want, we can split this PR up into smaller pieces, but since the biggest chunk is moving existing code around, I think it's fairly doable to review as long as you go commit by commit.

---

With this new API, we can turn:
```
[
  'bg-red-500',
  'hover:bg-red-500',
  '[@media_print]:bg-red-500',
  'hover:[@media_print]:bg-red-500',
  'bg-red-500/100',
  'hover:bg-red-500/100',
  '[@media_print]:bg-red-500/100',
  'hover:[@media_print]:bg-red-500/100',
  'bg-[var(--color-red-500)]',
  'hover:bg-[var(--color-red-500)]',
  '[@media_print]:bg-[var(--color-red-500)]',
  'hover:[@media_print]:bg-[var(--color-red-500)]',
  'bg-[var(--color-red-500)]/100',
  'hover:bg-[var(--color-red-500)]/100',
  '[@media_print]:bg-[var(--color-red-500)]/100',
  'hover:[@media_print]:bg-[var(--color-red-500)]/100',
  'bg-(--color-red-500)',
  'hover:bg-(--color-red-500)',
  '[@media_print]:bg-(--color-red-500)',
  'hover:[@media_print]:bg-(--color-red-500)',
  'bg-(--color-red-500)/100',
  'hover:bg-(--color-red-500)/100',
  '[@media_print]:bg-(--color-red-500)/100',
  'hover:[@media_print]:bg-(--color-red-500)/100',
  'bg-[color:var(--color-red-500)]',
  'hover:bg-[color:var(--color-red-500)]',
  '[@media_print]:bg-[color:var(--color-red-500)]',
  'hover:[@media_print]:bg-[color:var(--color-red-500)]',
  'bg-[color:var(--color-red-500)]/100',
  'hover:bg-[color:var(--color-red-500)]/100',
  '[@media_print]:bg-[color:var(--color-red-500)]/100',
  'hover:[@media_print]:bg-[color:var(--color-red-500)]/100',
  'bg-(color:--color-red-500)',
  'hover:bg-(color:--color-red-500)',
  '[@media_print]:bg-(color:--color-red-500)',
  'hover:[@media_print]:bg-(color:--color-red-500)',
  'bg-(color:--color-red-500)/100',
  'hover:bg-(color:--color-red-500)/100',
  '[@media_print]:bg-(color:--color-red-500)/100',
  'hover:[@media_print]:bg-(color:--color-red-500)/100',
  '[background-color:var(--color-red-500)]',
  'hover:[background-color:var(--color-red-500)]',
  '[@media_print]:[background-color:var(--color-red-500)]',
  'hover:[@media_print]:[background-color:var(--color-red-500)]',
  '[background-color:var(--color-red-500)]/100',
  'hover:[background-color:var(--color-red-500)]/100',
  '[@media_print]:[background-color:var(--color-red-500)]/100',
  'hover:[@media_print]:[background-color:var(--color-red-500)]/100'
]
```

Into their canonicalized form:
```
[
  'bg-red-500',
  'hover:bg-red-500',
  'print:bg-red-500',
  'hover:print:bg-red-500'
]
```

The list is also unique, so we won't end up with `bg-red-500 bg-red-500` twice.

While the canonicalization itself is fairly fast, we still pay a **~1s** startup cost for some migrations (once, and cached for the entire lifetime of the design system). I would like to keep improving the performance and the kinds of migrations we do, but I think this is a good start.

The cost we pay is for:

1. Generating a full list of all possible utilities based on the `getClassList` suggestions API.
2. Generating a full list of all possible variants.

The canonicalization step for this list takes **~2.9ms** on my machine.

Just for fun, if you use the `getClassList` API for intellisense that generates all the suggestions and their modifiers, you get a list of **263788** classes. If you canonicalize all of these, it takes **~500ms** in total. So roughly **~1.9μs** per candidate.

This new API doesn't result in a performance difference for normal Tailwind CSS builds.

The other potential concern is file size of the package. The generated `tailwindcss.tgz` file changed like this:
```diff
- 684652 bytes (684.65 kB)
+ 749169 bytes (749.17 kB)
```
So the package increased by ~65 kB which I don't think is the end of the world, but it is important for the `@tailwindcss/browser` build which we don't want to grow unnecessarily. For this reason we remove some of the code for the design system conditionally such that you don't pay this cost in an environment where you will never need this API.

The `@tailwindcss/browser` build looks like this:
```shell
`dist/index.global.js 255.14 KB` (main)
`dist/index.global.js 272.61 KB` (before this change)
`dist/index.global.js 252.83 KB` (after this change, even smaller than on `main`)
```


packages/tailwindcss/src/utils/dimensions.ts
```diff
@@ -1,8 +1,10 @@
 import { DefaultMap } from '../../../tailwindcss/src/utils/default-map'
 
+const DIMENSION_REGEX = /^(?<value>-?(?:\d*\.)?\d+)(?<unit>[a-z]+|%)$/i
+
 // Parse a dimension such as `64rem` into `[64, 'rem']`.
 export const dimensions = new DefaultMap((input) => {
-  let match = /^(?<value>-?(?:\d*\.)?\d+)(?<unit>[a-z]+|%)$/i.exec(input)
+  let match = DIMENSION_REGEX.exec(input)
```
RobinMalfait, line 7: The only reason I touched this is because I noticed it before I wanted to move it from the upgrade package to core.

packages/@tailwindcss-upgrade/src/codemods/template/migrate-deprecated-utilities.ts
```diff
@@ -6,11 +6,11 @@ import { computeUtilitySignature } from './signatures'
 
 const DEPRECATION_MAP = new Map([['order-none', 'order-0']])
 
-export async function migrateDeprecatedUtilities(
+export function migrateDeprecatedUtilities(
   designSystem: DesignSystem,
   _userConfig: Config | null,
   rawCandidate: string,
-): Promise<string> {
+): string {
```
RobinMalfait: I was annotating migrations to know whether it was a v3 → v4 upgrade, or if it was a v4 optimization. Noticed that this was using async which is not necessary in this case.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -65,7 +65,7 @@ const enum Convert {
 }
 
 function themeToVar(designSystem: DesignSystem, rawCandidate: string): string {
-  let convert = createConverter(designSystem)
+  let convert = converterCache.get(designSystem)
```
RobinMalfait, line 180: If you look at this commit, use the `?w=1` GitHub flag. The only real changes here are:

1. Re-indent because of the wrapping `DefaultMap`
2. The removal of the `prettyPrint` flag, that was only used in the CSS part (which still exists in the CSS version where whitespace is fine.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -238,11 +238,11 @@ const converterCache = new DefaultMap((ds: DesignSystem) => {
       return [result, modifier]
     }
 
-    function pathToVariableName(path: string) {
+    function pathToVariableName(path: string, shouldPrefix = true) {
```
RobinMalfait, line 348: This was actually a bug we shipped and only noticed it because the canonicalization tests run all examples in multiple catagories:
1. The candidate on its own
2. The candidate with a variant in front
3. The candidate with an important `!` at the end
4. The candidate with a `tw:` prefix in front

Then I noticed that this broke a test. Without this, the following candidate would be using `--tw-` as well:
```diff
- tw:bg-[--theme(--color-red-500/50%)]/50
+ tw:bg-[--theme(--tw-color-red-500/50%)]/50
```
The problem is that the `--theme(…)` does take the value _without_ the `--tw-` flag: https://play.tailwindcss.com/GMFKMFGLFz

You could argue that this is a `--theme(…)` bug and that we should handle it there. But this is the behavior of `--theme(…)` today, so didn't want to break it.

packages/@tailwindcss-upgrade/src/codemods/template/signatures.ts
```diff
@@ -1,13 +1,13 @@
-import { substituteAtApply } from '../../../../tailwindcss/src/apply'
```
RobinMalfait, line 1: It was kind of a smell that all these imports in the upgrade package were using `tailwindcss` internals 😅

packages/tailwindcss/src/canonicalize-candidates.test.ts
```diff
@@ -97,7 +97,7 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s',
 
     /// theme(…) to `var(…)`
     // Keep candidates that don't contain `theme(…)` or `theme(…, …)`
-    ['[color:red]', '[color:red]'],
+    ['[color:red]', 'text-[red]'],
```
RobinMalfait, line 100: So you will notice that tests changed in this commit. This is because we were testing each migration in isolation before. But now we're testing the entire `canonicalizeCandidates` function at once.

It just happens to be that this is an even _more_ canonicalized compared to before.

philipp-spiess:
I feel like this would better be present in another package or a separate and completely isolated file, not the core one. Specifically since core is part of the browser build which is pretty sensitive to code size and this is moving a bunch of code into it now. Have you thought about this?

This PR also adds `cloneAstNode` which should probably be in a separate PR with its own description.

packages/tailwindcss/src/selector-parser.test.ts
```diff
@@ -131,6 +131,34 @@ describe('parse', () => {
...
+
+  it('parses nesting selector before attribute selector', () => {
+    expect(parse('&[data-foo]')).toEqual([
+      { kind: 'selector', value: '&' },
+      { kind: 'selector', value: '[data-foo]' },
+    ])
+  })
+
+  it('parses nesting selector after an attribute selector', () => {
+    expect(parse('[data-foo]&')).toEqual([
```
RobinMalfait, line 143: While converting from the `postcss-selector-parser` setup to our own `SelectorParser`, I noticed that this case wasn't covered properly and broke some tests.

philipp-spiess:
I see that you mentioned file size, how the hell does core have more than 3MB 🤔 That seems very unexpected.

I think something with your measurement is wrong: https://bundlephobia.com/package/@tailwindcss/browser@4.1.14

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -1110,6 +1112,516 @@ function* tryValueReplacements(
...
+
+enum AttributePart {
+  Start,
+  Attribute,
+  Value,
+  Modifier,
+  End,
+}
+
+function parseAttributeSelector(value: string) {
```
RobinMalfait: Don't read too much into this function. This was a temporary replacement for the attribute selector parser. I added a proper commit with an implementation + tests + benchmark. Just needed this to make the tests pass.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -25,65 +25,146 @@ import * as ValueParser from './value-parser'
 
 export function canonicalizeCandidates(ds: DesignSystem, candidates: string[]): string[] {
   let result = new Set<string>()
+  let cache = canonicalizeCandidateCache.get(ds)
```
RobinMalfait, line 29: This commit is a big boy, but it changes from:
1. Parse string
2. Loop over AST
3. Print AST to string
4. Repeat

To 

1. Parse string
2. Canonicalize variants and the base utility separately
3. Pass the Candidate and Variant AST objects to each canonicalization step
4. At the end, stringify again

This change looks big, but we:
- Replace `rawCandidate: string` with `candidate: Candidate` in the function signature
- Don't loop over the `parseCandidate()` result anymore, because we already have the object
- Some `continue`'s are replaces with `return candidate` because we are not in a loop anymore

This also split up the canonicalization steps for Variants and Candidates, originally they were handled in the same migration steps.

RobinMalfait: Might be useful to look at this commit with `?w=1` as well due to the re-indents.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -25,65 +25,146 @@ import * as ValueParser from './value-parser'
...
-
-      return designSystem.printCandidate({
-        ...candidate,
-        root: `bg-linear-to-${direction}`,
-      })
+    if (!DIRECTIONS.includes(direction)) {
+      return candidate
     }
+
+    candidate.root = `bg-linear-to-${direction}`
```
RobinMalfait, line 164: We already cloned the candidate when it came out of the design system, so we can mutate it freely without causing issues.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -425,18 +501,7 @@ function* walkVariants(candidate: Candidate) {
     }
   }
 
-  for (let variant of candidate.variants) {
-    yield* inner(variant, null)
-  }
-}
-
-function baseCandidate<T extends Candidate>(candidate: T) {
```
RobinMalfait: Before running the canonicalization steps, we already drop the `important` flag and `variants` so each step is already dealing with the base candidate (utility).

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -843,199 +842,167 @@ function bareValueUtilities(designSystem: DesignSystem, rawCandidate: string): s
...
-    // The below logic makes use of mutation. Since candidates in the
-    // DesignSystem are cached, we can't mutate them directly.
-    let candidate = structuredClone(readonlyCandidate) as Writable<typeof readonlyCandidate>
-
-    // Create a basic stripped candidate without variants or important flag. We
-    // will re-add those later but they are irrelevant for what we are trying to
-    // do here (and will increase cache hits because we only have to deal with
-    // the base utility, nothing more).
-    let targetCandidate = baseCandidate(candidate)
-    let targetCandidateString = printUnprefixedCandidate(designSystem, targetCandidate)
```
RobinMalfait: Similar here, we already are dealing with the base candidate so we can just drop all of this entirely.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -94,62 +175,57 @@ const enum Convert {
   MigrateThemeOnly = 1 << 1,
 }
 
-function themeToVar(designSystem: DesignSystem, rawCandidate: string): string {
+function themeToVarUtility(designSystem: DesignSystem, candidate: Candidate): Candidate {
   let convert = converterCache.get(designSystem)
 
-  for (let candidate of parseCandidate(designSystem, rawCandidate)) {
-    let clone = structuredClone(candidate) as Writable<typeof candidate>
-    let changed = false
```
RobinMalfait: We were tracking whether something changed or not, and if it did we would re-print the Candidate AST, otherwise we would just re-use the incoming `rawCandidate`.

Now, we are directly mutating the incoming `candidate: Candidate` which we will return as well, so this code can be dropped.

packages/tailwindcss/src/attribute-selector-parser.bench.ts
```diff
@@ -0,0 +1,35 @@
...
+  '[data-foo]',
+  '[data-state=expanded]',
+  '[data-state = expanded ]',
+  '[data-state*="expanded"]',
+  '[data-state*="expanded"i]',
+  '[data-state*=expanded i]',
+]
+
+const ATTRIBUTE_REGEX =
+  /\[\s*(?<attribute>[a-zA-Z_-][a-zA-Z0-9_-]*)\s*((?<operator>[*|~^$]?=)\s*(?<quote>['"])?\s*(?<value>.*?)\4\s*(?<sensitivity>[is])?\s*)?\]/
```
RobinMalfait, line 15: This one isn't even 100% correct, but it's slower already so didn't want to make it more correct 😅

thecrypticace:
Yeah core isn't 3.5MB right now. It's about 680KB (which is still a lot but that's because CJS + ESM duplication).

The build process doesn't clear the `dist` folder so it probably got a bunch of chunks from previous builds bundled into the .tgz folder on Robin's local machine.

RobinMalfait:
> Yeah core isn't 3.5MB right now. It's about 680KB (which is still a lot but that's because CJS + ESM duplication).
> 
> The build process doesn't clear the `dist` folder so it probably got a bunch of chunks from previous builds bundled into the .tgz folder on Robin's local machine.

Yep, that was the reason why! I'm not _that_ stupid and I did clear them manually before. But I'm stupid enough not to realise that Turborepo just put them all back when running a build as if nothing ever happened. 😅

packages/tailwindcss/src/compat/selector-parser.ts
```diff
@@ -369,7 +371,7 @@ export function parse(input: string) {
             ast.push(node)
           }
         }
-        buffer = String.fromCharCode(currentChar)
```
thecrypticace, line 372: I was thinking we did this in purpose for perf reasons but it might not actually make a difference compared to everything else.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -0,0 +1,1510 @@
...
+  // Try a few options to find a suitable replacement utility
+  for (let replacementCandidate of tryReplacements(targetSignature, candidate)) {
+    let replacementString = designSystem.printCandidate(replacementCandidate)
+    let replacementSignature = signatures.get(replacementString)
+    if (replacementSignature !== targetSignature) {
+      continue
+    }
+
+    // Update the candidate with the new value
+    replaceObject(candidate, replacementCandidate)
```
thecrypticace: I think all the uses of `replaceObject(candidate, replacementCandidate); return replacementCandidate` can be just `return replacementCandidate` right?

It'd be cool if we could get rid of `replaceObject` entirely for variants too but that's probably not realistic.

packages/tailwindcss/src/attribute-selector-parser.ts
thecrypticace, line 1: Does it make sense to merge this with the main selector parser or nah?

thecrypticace (approved):
Overall things look good 💯 

Massive PR but will be awesome to have this stuff in core. If we can speed up the initialization step enough I'll be able to make this a "default on" diagnostic for intellisense which would be pretty cool.

packages/tailwindcss/src/attribute-selector-parser.ts
RobinMalfait: I didn't because we don't _need_ it anywhere else. But we can make it part of the SelectorParser and introduce a new `attribute` kind 🤔 

I think the `postcss-selector-parser` actually has a `selector` kind that is just a group, each node is `class`, `id`, `attribute` etc.

The only other spot in core where we use the selector parser is inside of the compat plugin API

packages/tailwindcss/src/canonicalize-candidates.ts
RobinMalfait: You know... very likely. Let me test that out!

packages/tailwindcss/src/attribute-selector-parser.ts
thecrypticace: I'm kinda indifferent — it just feels odd to have a "selector parser" and a separate "attribute selector parser". Reads like the selector parser doesn't handle attributes at all. (it does just doesn't split out their "parts").

Maybe @philipp-spiess can weigh in on what he thinks.

packages/@tailwindcss-browser/tsup.config.ts
```diff
@@ -13,4 +13,20 @@ export default defineConfig({
...
     'process.env.FEATURES_ENV': '"stable"',
   },
+  esbuildPlugins: [
+    {
+      name: 'patch-intellisense-apis',
+      setup(build) {
+        build.onLoad({ filter: /intellisense.ts$/ }, () => {
+          return {
+            contents: `
+              export function getClassList() { return [] }
```
itsdouges, line 23: Does this mean these APIs aren't available when running the browser version?

packages/tailwindcss/src/attribute-selector-parser.bench.ts
itsdouges: Are you balancing correctness, bundle size and performance? There could always be work done on a proper parser here but of course comes with more code as a cost

packages/@tailwindcss-browser/tsup.config.ts
thecrypticace: These APIs are internal and were already inaccessible in the `@tailwindcss/browser` package. They were on an object (i.e. not tree-shakable) which meant they still took up some space. This replaces those APIs so they take up much less space since they can't be used anyway.

basically this is just a code-size reduction and doesn't eliminate any actual functionality.

itsdouges: Ah ok

thecrypticace: You can still import the `tailwindcss` package directly — even in the browser — it'll work just fine and will have these APIs.

This is done to help eliminate some bundle size concerns with the CDN version (e.g. the one that scans class="…" attributes in HTML and compiles on the fly)

itsdouges: Ah that must be it great, thanks mate

packages/tailwindcss/src/attribute-selector-parser.bench.ts
RobinMalfait: The hand-rolled parser is bigger in file size than the regex, but not by a lot compared to the other things we ship in Tailwind's core package.  Performance and correctness is much more important in these cases. Maintainability is also a big one, and the regex is not super maintainable, especially because it is not 100% correct yet.

packages/tailwindcss/src/attribute-selector-parser.ts
philipp-spiess: So, right now the selector parser just skips over attribute selectors until the closing `]`, right? Seems like in the future we might merge those two but right now if there's no overlap I don't think we have to bother. In fact, using the attribute selector parser from within the selector parser in the attribute branch should be fairly trivial now. I'd say we leave it like this for now but maybe add a comment here explaining why we don't do more work yet and where that work can be found: https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/src/compat/selector-parser.ts#L378-L406

packages/@tailwindcss-browser/tsup.config.ts
philipp-spiess: Yeah! The browser package is here only for the code that scans the mounted DOM class lists to figure out what utilities to compile.

philipp-spiess (approved):

PR #19095, opened 2025-10-09
Canonicalization constant folding and handling zeros

RobinMalfait:
The main goal of this PR was to support canonicalization of zero like values. We essentially want to canonicalize `-mt-0` as `mt-0`, but also `mt-[0px]`, `mt-[0rem]`, and other length-like units to just `mt-0`.

To do this, we had to handle 2 things:

1. We introduced some more constant folding, including making `0px` and `0rem` fold to `0`. We only do this for length units. We also normalize `-0`, `+0`, `-0.0` and so on to `0`.
2. While pre-computing utilities in our lookup table, we make sure that we prefer `mt-0` over `-mt-0` if both result in the same signature.

Moved some of the constant folding logic into its own function and added a bunch of separate tests for it.

## Test plan

Added more unit tests where we normalize different zero-like values to `0`.

Running the canonicalization logic:
```js
designSystem.canonicalizeCandidates([
  '-m-0',
  '-m-[-0px]',
  '-m-[-0rem]',
  '-m-[0px]',
  '-m-[0rem]',
  'm-0',
  'm-[-0px]',
  'm-[-0rem]',
  'm-[0px]',
  'm-[0rem]',
  'm-[calc(var(--spacing)*0)]',
  'm-[--spacing(0)]',
  'm-[--spacing(0.0)]',
  'm-[+0]',
  'm-[-0]',
  '-m-[-0]',
  '-m-[+0]',
]) // → ['m-0']
```

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
@@ -61,7 +61,7 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s', 
     // handle the `0px * -1` case which translates to `0px` not `-0px`.
     //
     // This translation is actually fine, because now, we will prefer the
-    // non-negative version first so we can replace `-mt-[0px]` with `mt-[0px]`.
+    // non-negative version first so we can replace `-mt-[0px]` with `mt-0`.
     ['mt-[0px]', 'mt-0'],
     ['-mt-[0px]', 'mt-0'],
 
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
@@ -62,8 +62,8 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s', 
     //
     // This translation is actually fine, because now, we will prefer the
     // non-negative version first so we can replace `-mt-[0px]` with `mt-[0px]`.
-    ['mt-[0px]', 'mt-[0px]'],
-    ['-mt-[0px]', 'mt-[0px]'],
+    ['mt-[0px]', 'mt-0'],
+    ['-mt-[0px]', 'mt-0'],
 
     // Shorthand CSS Variables should be converted to the new syntax, even if
     // the fallback contains functions. The fallback should also be migrated to
--- a/packages/tailwindcss/src/constant-fold-declaration.test.ts
+++ b/packages/tailwindcss/src/constant-fold-declaration.test.ts
@@ -23,6 +23,12 @@ it.each([
   ['calc(2rem - 0.5rem)', '1.5rem'],
   ['calc(3rem * 6)', '18rem'],
   ['calc(5rem / 2)', '2.5rem'],
+
+  // Nested partial evaluation
+  ['calc(calc(1 + 2) + 2rem)', 'calc(3 + 2rem)'],
+
+  // Evaluation only handles two operands right now, this can change in the future
+  ['calc(1 + 2 + 3)', 'calc(1 + 2 + 3)'],
 ])('should constant fold `%s` into `%s`', (input, expected) => {
   expect(constantFoldDeclaration(input)).toBe(expected)
 })
--- a/packages/tailwindcss/src/constant-fold-declaration.test.ts
+++ b/packages/tailwindcss/src/constant-fold-declaration.test.ts
@@ -1,0 +1,88 @@
+import { expect, it } from 'vitest'
+import { constantFoldDeclaration } from './constant-fold-declaration'
+
+it.each([
+  // Simple expression
+  ['calc(1 + 1)', '2'],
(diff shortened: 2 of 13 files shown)


packages/@tailwindcss-upgrade/src/codemods/template/migrate.test.ts
```diff
@@ -62,8 +62,8 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s',
     //
     // This translation is actually fine, because now, we will prefer the
     // non-negative version first so we can replace `-mt-[0px]` with `mt-[0px]`.
```
RobinMalfait:
```suggestion
    // non-negative version first so we can replace `-mt-[0px]` with `mt-0`.
```

packages/tailwindcss/src/constant-fold-declaration.ts
```diff
@@ -0,0 +1,126 @@
...
+      // ]
+      if (valueNode.nodes.length !== 5) return
+
+      let lhs = dimensions.get(valueNode.nodes[0].value)
+      let operator = valueNode.nodes[2].value
+      let rhs = dimensions.get(valueNode.nodes[4].value)
+
+      // Nullify entire expression when multiplying by `0`, e.g.: `calc(0 * 100vw)` -> `0`
+      //
+      // TODO: Ensure it's safe to do so based on the data types?
```
RobinMalfait, line 61: I think this is fairly safe, and feels like something we only have to add support for when running into actual issues 🤔

packages/tailwindcss/src/value-parser.ts
```diff
@@ -172,6 +237,33 @@ export function parse(input: string) {
         break
       }
 
+      // Typically for math operators they have to have spaces around them. But
+      // therea re situations in `theme(colors.red.500/10)` where we use `/`
+      // without spaces. Let's make srue this is a separate word as well.
```
RobinMalfait:
```suggestion
      // Typically for math operators, they have to have spaces around them. But
      // there are situations in `theme(colors.red.500/10)` where we use `/`
      // without spaces. Let's make sure this is a separate word as well.
```

packages/tailwindcss/src/constant-fold-declaration.ts
thecrypticace: Let me see if I can break this

RobinMalfait: Probaaaably, but once you do, the questions is: how common / real is the actual issue.

packages/tailwindcss/src/constant-fold-declaration.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  ['calc(1 + (1 + 1))', '3'],
+  ['calc(3 - (1 + 2))', '0'],
+  ['calc(2 * (1 + 3))', '8'],
+  ['calc(8 / (2 + 2))', '2'],
+
+  // With units
+  ['calc(1rem * 2)', '2rem'],
+  ['calc(2rem - 0.5rem)', '1.5rem'],
+  ['calc(3rem * 6)', '18rem'],
+  ['calc(5rem / 2)', '2.5rem'],
```
thecrypticace, line 25: Some additional tests / comments:

```suggestion
  ['calc(5rem / 2)', '2.5rem'],

  // Nested partial evaluation
  ['calc(calc(1 + 2) + 2rem)', 'calc(3 + 2rem)'],

  // Evaluation only handles two operands right now, this can change in the future
  ['calc(1 + 2 + 3)', 'calc(1 + 2 + 3)'],
```

packages/tailwindcss/src/constant-fold-declaration.ts
thecrypticace: I think normalizing things with units to without could cause problems (like rem, em, etc… should probably normalize to 0px instead) but  then how would you know `margin-left: 0` normalizes to `0px` without encoding data type info and that'd be annoying.

All that to say… I *think* this is fine for now.  We can improve this later if need be.

packages/tailwindcss/src/constant-fold-declaration.test.ts
RobinMalfait: I thought about doing `1 + 2 + 3` but didn't want to parse and think about operator precedence for this ... I debated of using the parentheses hack though 😅 

```ts
let input = '1 + 2 + 3 * 4'
function operatorPrecedenceHack(input: string): string {
  return `((${input
    .replace(/\s\(/g, '(((')
    .replace(/\)/g, ')))')
    .replace(/\s*([-+])\s*/g, ')) $1 ((')
    .replace(/\s*([*/])\s*/g, ') $1 (')}))`
}

console.log(operatorPrecedenceHack(input)) // ((1)) + ((2)) + ((3) * (4))
```

thecrypticace: hahaha I think it'd be good just to document it in tests that it *currently* doesn't work is all.

thecrypticace (approved):

PR #19097, opened 2025-10-09
Fix resolving colors via `theme(…)` in compat mode with nested objects

RobinMalfait:
This PR fixes an issue when loading (nested) colors from a config file and later referencing it via the `theme(…)` function in CSS.

Given a config like this:

```js
module.exports = {
  theme: {
    colors: {
      foo: 'var(--foo-foo)',
      'foo-bar': 'var(--foo-foo-bar)',
    },
  },
}
```

We internally map this into the design system. The issue here is that the `foo` and `foo-bar` are overlapping and it behaves more like this:

```js
{
  foo: {
    DEFAULT: 'var(--foo-foo)',
    bar: 'var(--foo-foo-bar)'
  },
}
```

So while we can easily resolve `colors.foo-bar`, the `colors.foo` would result in the object with a `DEFAULT` key. This PR solves that by using the `DEFAULT` key if we end up with an object that has it.

If you end up resolving an object (`theme(colors)`) then the behavior is unchanged.

## Test plan

1. Added a test based on the config in the issue (which failed before this fix).
2. Also simplified the test case after identifying the problem (with the `DEFAULT` key).

Fixes: #19091

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix Safari devtools rendering issue due to `color-mix` fallback ([#19069](https://github.com/tailwindlabs/tailwindcss/pull/19069))
 - Suppress Lightning CSS warnings about `:deep`, `:slotted` and `:global` ([#19094](https://github.com/tailwindlabs/tailwindcss/pull/19094))
+- Fix resolving colors via `theme(…)` in compat mode with nested objects ([#19097](https://github.com/tailwindlabs/tailwindcss/pull/19097))
 
 ## [4.1.14] - 2025-10-01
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix Safari devtools rendering issue due to `color-mix` fallback ([#19069](https://github.com/tailwindlabs/tailwindcss/pull/19069))
 - Suppress Lightning CSS warnings about `:deep`, `:slotted` and `:global` ([#19094](https://github.com/tailwindlabs/tailwindcss/pull/19094))
+- Fix resolving colors via `theme(…)` in compat mode with nested objects ([#19097](https://github.com/tailwindlabs/tailwindcss/pull/19097))
```
RobinMalfait: Not super sure what to call this. Because the issue is related to nested objects and the `DEFAULT` key, but that's not what you see / use as the user (at least nog in this case). It just happens to be that your input is mapped to such an object.

Maybe something with "overlapping namespace key" could work, but that doesn't make it more clear...

thecrypticace (approved):

CHANGELOG.md
thecrypticace: "Fix resolving theme keys when starting with the name of another theme key in JS configs and plugins"?

RobinMalfait: A mouthful, love it. Going to be lazy, nothing to see here...

```suggestion
- Fix resolving theme keys when starting with the name of another theme key in JS configs and plugins ([#19097](https://github.com/tailwindlabs/tailwindcss/pull/19097))
```

PR #19100, opened 2025-10-09
Move modifier of `not-*`, `has-*`, and `in-*` variant to sub variant

### LINKED ISSUE #15772: The `not-` variant does not work as expected with named groups (`group/{name}`)

v4.0.0

Tailwind Play (https://play.tailwindcss.com/)

-

Chrome 131

Windows 11

**Reproduction URL**

https://play.tailwindcss.com/DKPgTfOLF0

**Describe your issue**

```
<!-- Named group (does NOT work) -->  
<div class="group/test bg-red-50 p-8">  
  <div class="not-group/test-hover:text-red-500">Text (should be red when group is NOT hovered)</div>  
</div>  

<!-- Unnamed group (works correctly) -->  
<div class="group bg-green-50 p-8">  
  <div class="not-group-hover:text-green-500">Text (green when group is NOT hovered)</div>  
</div>  
```

Current Behavior:

* The `not-group/test-hover:text-red-500` class does not apply styles when the named group (`group/test`) is not hovered.
* The `not-group-hover:text-green-500` class works as intended for the unnamed group.

Expected Behavior:

Both named and unnamed groups should support the `not-` variant (e.g., `not-group/{name}-hover` and `not-group-hover`) to apply styles when the parent group is not in the specified state.

RobinMalfait:
This PR fixes an issue where a compound variant with a modifier such as `not-group-hover/name:flex` would not generate anything because the `/name` modifier belongs to the `not` variant, and not the compounded `group-hover` variant.

This PR is a **workaround** (and definitely not perfect) by special casing the `not`, `has`, and `in` variants such that their modifiers are moved internally to the sub variant as-if the `/name` existed on `group-hover`.

We don't do it for other compound variants such as `group` and `peer` because then `group-peer-focus/name:underline` would result in a breaking change:

```diff
- .group-peer-focus\\/name\\:flex:is(:where(.group\\/name):is(:where(.peer):focus ~ *) *)
+ .group-peer-focus\/name\:flex:is(:where(.group):is(:where(.peer\/name):focus ~ *) *) 
```

In case the diff is not clear, the name has moved:
image

This is also a limited workaround, because if you need multiple modifiers it won't work. I would've loved to special case this _inside_ the `not`, `has`, and `in` code that handles these variants, but we handle the variants in a depth-first way, so by the time you are handling the `not` variant, the sub variant was already handled...

In a perfect world, you can use something like `not-group/name-hover` but then it becomes unambiguous because is `name` the name, is `name-hover`?

## Test plan

Added a new test that wouldn't generate anything before this fix.

Fixes: #15772

### DIFF
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -806,22 +806,15 @@ export function parseVariant(variant: string, designSystem: DesignSystem): Varia
         case 'compound': {
           if (value === null) return null
 
-          let subVariant = designSystem.parseVariant(value)
-          if (subVariant === null) return null
-
           // Forward the modifier of the `not` and `has` variants to its
           // subVariant. This allows for `not-group-hover/name:flex` to work.
-          if (modifier && (root === 'not' || root === 'has') && 'modifier' in subVariant) {
-            // Forward modifiers to the subVariant
-            let parsedModifier = parseModifier(modifier)
-            if (parsedModifier !== null) {
-              // Can't mutate directly because the `subVariant` is a shared
-              // object and then `group-hover` would be equivalent to
-              // `group-hover/name` which would be incorrect.
-              subVariant = { ...subVariant, modifier: parsedModifier }
-              modifier = null
-            }
+          if (modifier && (root === 'not' || root === 'has')) {
+            value = `${value}/${modifier}`
+            modifier = null
           }
+
+          let subVariant = designSystem.parseVariant(value)
+          if (subVariant === null) return null
 
           // These two variants must be compatible when compounded
           if (!designSystem.variants.compoundsWith(root, subVariant)) return null
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -809,6 +809,20 @@ export function parseVariant(variant: string, designSystem: DesignSystem): Varia
           let subVariant = designSystem.parseVariant(value)
           if (subVariant === null) return null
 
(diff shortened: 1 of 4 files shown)


packages/tailwindcss/src/candidate.ts
```diff
@@ -806,6 +806,13 @@ export function parseVariant(variant: string, designSystem: DesignSystem): Varia
         case 'compound': {
           if (value === null) return null
 
+          // Forward the modifier of the `not` and `has` variants to its
+          // subVariant. This allows for `not-group-hover/name:flex` to work.
+          if (modifier && (root === 'not' || root === 'has')) {
+            value = `${value}/${modifier}`
+            modifier = null
+          }
```
RobinMalfait: Ideally this was located in the handling of `not` and `has` variants, but since it's depth first the sub variant would've already been handled.

I'm kind of okay with special casing `not` and `has` because right now you can't create compound variants outside of core anyway so we do have control over all of it.

If you were to override the `not` or `has` variant via `@variant not (…)` then it wouldn't be a `compound` variant anymore so we wouldn't hit this branch.

thecrypticace: Ah yeah that's a good point. If only core-defined stuff will ever hit this branch I'm fine with the hardcoded workaround for now.

Should we special case `in` as well — just for consistency / completeness sake?

RobinMalfait: I forgot about in. Oops

thecrypticace (approved):

PR #19131, opened 2025-10-15
Add `optimize` option to `@tailwindcss/vite` plugin

benface:
Adds an `optimize` option to the Vite plugin that matches the API and behavior of the PostCSS plugin. By default, it checks `NODE_ENV` to determine whether to optimize (enabling Lightning CSS in production).

Supports three formats:
- `optimize: false` - disable optimization
- `optimize: true` - enable optimization with minification
- `optimize: { minify: false }` - enable optimization without minification

🤖 Generated with [Claude Code](https://claude.ai/code)


thecrypticace (approved):

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -65,8 +73,22 @@ export default function tailwindcss(): Plugin[] {
 
       async configResolved(_config) {
         config = _config
-        minify = config.build.cssMinify !== false
         isSSR = config.build.ssr !== false && config.build.ssr !== undefined
+
+        // By default we optimize CSS during the build phase but if the user
+        // provides explicit options we'll use to those instead
```
RobinMalfait:
```suggestion
        // By default we optimize CSS during the build phase but if the user
        // provides explicit options we'll use those instead
```

RobinMalfait (approved):

PR #19148, opened 2025-10-18
Incorporate changes from latest `ignore` crate

thecrypticace:
This updates our internal fork of the `ignore` crate to incorporate changes from the latest release [v0.4.24](https://github.com/BurntSushi/ripgrep/compare/ignore-0.4.23...ignore-0.4.24).

Aside: We should look into opening issues about the changes we had to make to see if they could be addressed in the crate itself so we can get rid of our fork.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,9 +83,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.13"
+version = "0.5.15"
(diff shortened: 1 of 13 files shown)


RobinMalfait (approved):
Nice! I tested it on some projects and everything still works as expected.

RobinMalfait (commented):
We could add a changelog entry in case things break then we know why, but not sure if it's necesary.

PR #19149, opened 2025-10-19
Remove Oxide postinstall script

philipp-spiess:
This PR effectively reverts #17929.

The bug in npm that required it was fixed a couple of months ago and with recent changes to pnpm that requires manually approving all postinstall scripts, this is creating some unnecessary noise.


RobinMalfait (approved):

RobinMalfait (commented):
Just need to run pnpm install to update the lockfile

RobinMalfait (approved):

agusterodin:
Does this mean I can remove "@tailwindcss/oxide" from "onlyBuiltDependencies" in pnpm?

PR #19169, opened 2025-10-20
Stop suggesting legacy utilities

thecrypticace:
It was already supposed to work this way — and it appeared via the tests that it did — but the tests weren't loading the design system normally so when we split the legacy utilities into a separate file it stopped testing this properly.

The setup here is a bit iffy but the gist is:
- We allow static utilities to be suggested by default
- A static utility can *opt out* of suggestions by explicitly registering themselves to return none.


RobinMalfait (approved):
Yeah makes sense, didn't realise that we weren't loading the design system for intellisense tests properly 😬

thecrypticace:
Yeah… me either 🤦‍♂️

PR #19171, opened 2025-10-21
Improve memory usage during canonicalization

RobinMalfait:
This PR drastically improves the memory usage when performing canonicalization if you swap out the underlying DesignSystem often. This will be most noticeable in Intellisense.

The big issue we had is that we used module scoped Map objects where we cache data based on the DesignSystem. If you then create new design systems (often), then the cache would just keep growing and growing.

This PR solves that by essentially storing all the caches on the Design System itself. This way, when you throw away a Design System, all the caches go with it.

Another approach would've been to use a WeakMap, but then we would have to make sure that no strong references to the DesignSystem exist anywhere else, otherwise we would still have the same memory issues.

Note: make sure to go commit by commit and use `?w=1` to ignore whitespace changes.

## Test plan

1. All existing tests pass

Not super sure how to test this as part of the test suite without making it slow But I also don't think that's super necessary either. Here is an experiment I did where I introduce 5 design systems:
image

On the current `main` branch, this looks like:
image

In this PR, the memory usage looks like:
image

The memory usage is stable, but to actually prove that we can still track multiple design systems, let's track them all in a `Set` so garbage collection cannot get rid of the unused design system objects:
image

Now we're sort of back to the current situation on `main`:
image


thecrypticace (approved):
Tested this in IntelliSense too and the memory usage there is now stable 🔥

PR #19172, opened 2025-10-21
Discard candidates with an empty data type

### LINKED ISSUE #1479: Invalid edit range on completion resolve response

Using Zed.

N/A

"tailwindcss": "3.4.18",
"@tailwindcss/language-server": "0.14.28"

npm

macOS

**Tailwind CSS Stylesheet (v4) or config file (v3)**

None

**VS Code settings**

Using Zed

**Reproduction URL**

Can be reproducible in new Vite react typescript app. More: https://vite.dev/guide/. 

**Describe your issue**

Use classname such as `"*:mx-[^]"` where ^ is your cursor. Type "100000" quickly. **You get edit ranges where start is greater than end.** 

Zed Issue: https://github.com/zed-industries/zed/issues/36223

You can also run Zed locally and add dbg! log here:

https://github.com/zed-industries/zed/blob/a56122e1448ce6022a8a1be08167f94c6498fc66/crates/project/src/lsp_command.rs#L2209

to see this.

thecrypticace:
Fixes https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1479

Maybe should close https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1480 — perhaps we can find a workaround there for older versions?

We're building up a class name in code to validate if something is a valid variant: `{variant}:[color:red]`

if `{variant}` got replaced with `bg-[` then we'd produce `bg-[:[color:red]` and this parsed as a valid candidate:
```
bg-[:[color:red]
^^                root: `bg`
   ^              data type: `` (empty string) — this should be invalid
     ^^^^^^^^^^   value: `[color:red`
```

The value isn't valid _but_ the syntax for arbitrary values is pretty lax in core. Oxide already won't pick something like this up though so no problem there. Only a problem for something like IntelliSense or clients using the compile() API directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Experimental_: Add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
 
+### Fixed
+
+- Discard candidates with an empty data type ([#19172](https://github.com/tailwindlabs/tailwindcss/pull/19172))
+
 ## [4.1.15] - 2025-10-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
I think we can add a changelog entry for this one.

PR #19213, opened 2025-10-27
feat: add around shadow which will cover all sides in a component

mehediScriptDev:
Hello Tailwind team,

I've noticed there isn't an "around-shadow" utility in Tailwind. The existing shadow classes are great, as they only apply shadows to the left, right, and bottom. It feels like the light is coming from the top.

But sometimes we need a shadow for cards or other elements that must cover all sides equally, and there's no built-in class for this.

So, I am proposing a new set of utilities: shadow-around-xs, shadow-around-sm, shadow-around-md, etc.

I am proposing these as an addition, without touching or changing any of the existing shadow properties.

### DIFF
--- a/packages/tailwindcss/theme.css
+++ b/packages/tailwindcss/theme.css
@@ -363,6 +363,13 @@
   --shadow-xl: 0 20px 25px -5px rgb(0 0 0 / 0.1), 0 8px 10px -6px rgb(0 0 0 / 0.1);
   --shadow-2xl: 0 25px 50px -12px rgb(0 0 0 / 0.25);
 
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! I appreciate the PR, but if we want to add shadows like this, we have to do some research to make sure the default shadows make sense and match the current style of the default theme we ship. We would also have to think about the API (naming) and have to create documentation with good examples for this as well.

Because of that, I'm going to say no to this PR for now and we can revisit this if there is more demand for native shadows like this.

I would recommend to just add these CSS variables to your own projects where you want to use them.

Thanks again!

PR #19245, opened 2025-10-31
Support setting a default letter spacing for a custom font family

adamwathan:
This PR makes it possible to set a default letter spacing a given font family, just like how you can set the default `font-feature-settings` and `font-variation-settings`.

The use case is for things like heading fonts, where maybe you want the letter spacing to always be tighter and you don't want to add `tracking-tight` every time you use it.

The letter spacing can still be overridden with explicit tracking utilities.

```css
@theme {
  --font-display: Mona Sans;
  --font-display--letter-spacing: -0.025em;
}
```

### DIFF
--- a/packages/tailwindcss/src/compat/config.test.ts
+++ b/packages/tailwindcss/src/compat/config.test.ts
@@ -738,6 +738,83 @@ describe('default font family compatibility', () => {
     `)
   })
 
+  test('overriding `fontFamily.sans[1].letterSpacing` sets letter-spacing', async ({ expect }) => {
+    let input = css`
+      @theme default {
+        --default-font-family: var(--font-family-sans);
+        --default-font-feature-settings: var(--font-family-sans--font-feature-settings);
+        --default-font-variation-settings: var(--font-family-sans--font-variation-settings);
+      }
+      @config "./config.js";
+      @tailwind utilities;
+    `
+
+    let compiler = await compile(input, {
+      loadModule: async () => ({
+        module: {
+          theme: {
+            fontFamily: {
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/compat/config.test.ts
```diff
@@ -989,6 +1066,83 @@ describe('default font family compatibility', () => {
...
+      ".font-mono {
+        font-family: Potato Mono;
+        font-feature-settings: "cv06";
+        font-variation-settings: "XHGT" 0.7;
+        letter-spacing: -0.02em;
+      }
+      "
+    `)
+  })
+
```
thecrypticace, line 1145: I kinda feel like this test is redundant (if `sans` works then `mono` should work too) — but it matches some of the existing stuff so probably fine.

packages/tailwindcss/src/compat/config.test.ts
```diff
@@ -738,6 +738,83 @@ describe('default font family compatibility', () => {
     `)
   })
 
+  test('overriding `fontFamily.sans[1].letterSpacing` sets letter-spacing', async ({ expect }) => {
+    let input = css`
+      @theme default {
+        --default-font-family: var(--font-family-sans);
```
thecrypticace, line 744: Not for this PR but our default theme uses --theme() instead of var(…). We should probably mirror the actual usage in these tests.

thecrypticace (approved):
LGTM

thecrypticace:
We've decided we're not gonna do this one right now. If we want this feature in the future we'll re-open this PR / make a new one.

PR #19254, opened 2025-11-03
Handle utilities with `1` keys in a JS config

thecrypticace:
Fixes #19250

When converting a JS config to a series of theme keys we special case a key of `1`, treat it as if it were tuple access, and turn that into nested theme keys.

This isn't 100% correct though because a user can use `1` in an object instead of an array — in which case we should generate normal theme keys with a `1` in the variable name instead of one with a nested theme key separator (`--`).

Because the conversion works based on key paths we need to treat a path like `colors.a.1.b` as both `--color-a-1-b` *and* `--color-a--b`. This ensures a utility like `text-a-1-b` works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


philipp-spiess (approved):
I think it makes sense. It's gross, sure, but at least it's contained in the compat layer.

thecrypticace:
I'm gonna try to come up with a better design for this because I'm not a big fan of this implementation. I'd like to use actual type information when generating the theme keys from the config.

PR #19263, opened 2025-11-04
Substitute `@variant` inside utilities

thecrypticace:
Fixes #19258


RobinMalfait (approved):
A bit odd to include new features in old APIs. But given other things already worked I don't think it hurts to include these 🤔

PR #19269, opened 2025-11-05
Escape variant modifiers when used in selectors and at-rules

### LINKED ISSUE #19268: `+` character is not escaped in modifier name

4.1.16

Vite 7.2.0

24.10.0

Chrome, Safari, Firefox

macOS

**Reproduction URL**

https://play.tailwindcss.com/hmryLWAdTn

**Describe your issue**

For Reasons™ (the internal plugin sort of reasons), I would love to use `+` in modifier names (e.g. `group/foo+bar`), and it seems to be supported (🤞) in the sense that Tailwind will recognize a class that includes one and generate CSS for it... the only issue is that it doesn't escape the `+` properly in the generated CSS:

Image

It should be `.group\/foo\+bar`.

Thank you very much!

thecrypticace:
Fixes #19268

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Substitute `@variant` inside legacy JS APIs ([#19263](https://github.com/tailwindlabs/tailwindcss/pull/19263))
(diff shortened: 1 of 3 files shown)


thecrypticace:
I should've checked Oxide but I believe we'd already decided that modifiers can't contain "special" chars as a restriction:

Screenshot 2025-11-05 at 16 28 45

(notice how the first one doesn't have *any* kind of decoration — that means Oxide won't even see it)

So even *if* we escape these it won't get picked up anyway.

PR #19270, opened 2025-11-06
fix: format

WuMingDao:
## Summary
Use format command


RobinMalfait:
Hey! Thanks for the PR, however we like to keep the code from this crate in its original state as much as possible to make future updates from the real crate an easier experience without introducing unnecessary merge conflicts.

PR #19298, opened 2025-11-12
Add `corner-shape` utility classes (`corner-*`)

baptisthecht:
This PR adds utilities to control the shape of rounded corners of an element. These map to the new [`corner-shape`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/corner-shape) property in CSS.

You can use the `corner-*` utilities to control the shape of all corners simultaneously or use utilities like `corner-t-*`, `corner-bl-*`, etc… to control individual corners.

| Utility | Affected corners |
| --- | --- |
| `corner-*` | all |
| `corner-t-*` | top left, top right |
| `corner-b-*` | bottom left, bottom right |
| `corner-l-*` | top left, bottom left |
| `corner-r-*` | top right, bottom right |
| `corner-tl-*` | top left |
| `corner-tr-*` | top right |
| `corner-bl-*` | bottom left |
| `corner-br-*` | bottom right |
| `corner-s-*` | (logical) start start, end start |
| `corner-e-*` | (logical) start end, end end |
| `corner-ss-*` | (logical) start start |
| `corner-se-*` | (logical) start end |
| `corner-ee-*` | (logical) end end |
| `corner-es-*` | (logical) end start |

Each utility has a handful of pre-defined values that map to their counterparts in CSS. Shown here are the versions for shaping all corners but they're present individual corner utilities as well:

| Utility | Value |
| --- | --- |
| `corner-round` | `corner-shape: round` |
| `corner-scoop` | `corner-shape: scoop` |
| `corner-bevel` | `corner-shape: bevel` |
| `corner-notch` | `corner-shape: notch` |
| `corner-square` | `corner-shape: square` |
| `corner-squircle` | `corner-shape: squircle` |

Additionally, arbitrary values are supported so you can use custom `superellipse(…)` values e.g. `corner-[superellipse(1.333)]`.

For an idea of what each of these do you can use Chrome to look at [this Play](https://play.tailwindcss.com/gv7MV1rIPx) (or any other supported browser). Otherwise there's a screenshot below:

corner shape examples


thecrypticace (changes requested):
- I don't _think_ we need to add anything to the compatibility config since these are new utilities. Seems fine if these are only customizable via CSS.
- Please add utility tests to the utilities.test.ts file

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -11124,9 +11184,9 @@ const prefixes = [
   'border-l',
 ]
 
-for (let prefix of prefixes) {
+for (const prefix of prefixes) {
```
RobinMalfait: Change all of these back to `let` please.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add utilities for the experimental `corner-shape` property (e.g. `corner-scoop`, `corner-tl-bevel`)
```
RobinMalfait: Let's mark this as experimental as well.

packages/tailwindcss/src/utilities.ts
```diff
@@ -2199,6 +2199,37 @@ export function createUtilities(theme: Theme) {
     }
   }
 
+  {
```
RobinMalfait: Let's add this behind a feature flag.
See: 

https://github.com/tailwindlabs/tailwindcss/blob/21f2bf99b9764929b5268ce2174840086ef191cc/packages/tailwindcss/src/feature-flags.ts#L1-L1

and wrap this in an if-statement:

```suggestion
  if (enableCornerShapeUtilities) {
```

philipp-spiess:
One issue with this approach is that it will look quite different now on browsers that support `corner-shape` vs those that don't since the perceived radius is much smaller with e.g. a `corner-shape: squircle`:

Check out this for example: https://play.tailwindcss.com/j0QURKzlvU

CleanShot 2025-11-17 at 10 34 35@2x

I would argue that `rounded-md` is a much better fallback for `rounded-xl corner-squircle`. Especially now that only one browser supports it yet!

thecrypticace:
> I would argue that rounded-md is a much better fallback for rounded-xl corner-squircle …

@philipp-spiess I've been thinking about this too and have been trying to decide if we should try to use `@supports` in this + some kind of calculation via `calc(…)` or if we should delegate that to the user.

Aside: The current implementation in Chrome renders with fairly poor aliasing for small values and I wonder if any tweaks to that rendering would change how different sizes compare optically (even if just by a small amount).

philipp-spiess:
@thecrypticace Yeah I think you're right, getting this right is non-trivial and should be up to the user. We do have the `supports-corner-shape` variant to help with this already.

packages/tailwindcss/src/utilities.ts
```diff
@@ -2199,6 +2199,37 @@ export function createUtilities(theme: Theme) {
...
+      ['corner-bl', ['corner-bottom-left-shape']],
+    ] as const) {
+      functionalUtility(root, {
+        themeKeys: ['--corner-shape'],
+        handle: (value) => properties.map((property) => decl(property, value)),
+        staticValues: Object.fromEntries(
+          shapes.map((shape) => [shape, properties.map((property) => decl(property, shape))]),
+        ),
+      })
+    }
```
thecrypticace, line 2230: @RobinMalfait @philipp-spiess What are your thoughts on modifying this to accept `corner-[<number>]` that expands to `corner-shape: superellipse(<number>)`. Or alternatively a second `corner-superellipse-[<number>]` utility?

We *could* also punt that decision down the road since we won't ship these in a stable release until v4.2 but just something to think about. I think a nice way to write custom super ellipse factors would be nice.

RobinMalfait: While I like the succinct `corner-[<number>]` version, I'm afraid if other corner-shape functions will exist in the future, that this will become ambiguous. Writing `corner-superellipse-[<number>]` is much safer imo, sadly also much longer.

We could also just use `corner-[<number>]`, and _if_ in the future more functions like `superellipse(…)` (I had to type this 3 times, so this is not great...) exist, then we can introduce the more specific versions then. Maybe that's the better play here. Thoughts?

thecrypticace: hmm yeah — I say lets push that decision to later

brandonmcconnell: I was wrestling with the same question when building my related plugin [@toolwind/corner-shape](https://github.com/toolwind/corner-shape) recently.

In the interim, I've added a `corner-superellipse/[…]` utility, but having worked with it for a few days now, my thoughts have changed a bit.

- prefer a value over a modifier: `corner-superellipse-[…]` > `corner-superellipse/[…]`
- prefer full term: I've been doing a lot of research about concepts tangential to superellipses that might conflict in the future if I were to use `corner-[…]` to represent superellipses.
  
  So far, I've found none, so it feels safe, but maybe there are still non-mathematically prominent ideas they'd load `corner-shape` with that could be confuse this use case.
  
  For that reason, I would probably avoid `corner-[…]` unless CSSWG settles on `corner-shape: NUMBER` as a shorthand for superellipses.
  
  In the meantime, what would be shorter and still seemingly safe would be `corner-super-[…]`, as I can't find any other mathematically concepts for other shapes or algos like superellipse that start with `super`. You'll find some results about `supercircle` but it's essentially a concept built upon the superellipse formula.

All this said, while I considered `corner-super-[…]`, I think saving the few characters by trimming off `ellipse` could actually be more confusing for users and agents.

My vote would be for `corner-superellipse-[…]` which would accept any of these use cases:

- `corner-superellipse-[NUMBER]`
- `corner-superellipse-[fn(…)]` (calc/var/etc.)
- `corner-superellipse-(--val)` shorthand variable syntax

This option is not only future-proof, but it's clearer to users and agents, and exposing `superellipse` as option like this actually saves the user quite a few characters/tokens, with all the affordances of a dedicated value.

Compare:
```
corner-[superellipse(var(--val))]
corner-superellipse-(--val)
```

PR #19299, opened 2025-11-12
Fix @scope variant output and add test for scoped variant

SHARMA1525:
This pull request fixes the output structure for custom variants that use conditional at-rules such as @scope.
Previously, the generated CSS placed the selector before the at-rule, which resulted in invalid nesting. After this change, the compiler correctly emits the conditional at-rule as the outer wrapper, ensuring valid and expected CSS output.
Changes Made:
file variants.scope.test.ts:

Added a new test case verifying that custom variants using @scope produce correctly structured CSS.

Ensures the at-rule wraps the selector and that theme variables are applied properly.
file ast.ts :
 Added a transform in the [optimizeAst]

function that detects when a style rule contains only a single conditional at-rule child (@scope, @media, @supports, @container) and restructures the AST to emit the at-rule as the outer wrapper with the rule inside.

Test Plan:

Verified that the new test case for @scope passes successfully.

Confirmed that all existing test suites continue to pass without regressions.

Validated the generated CSS output structure to ensure correctness and consistency with other conditional at-rules.

Result:

Correctly wraps conditional at-rules like @scope around variant selectors.

Maintains backward compatibility with existing variant behavior.

All tests pass successfully across supported environments.


thecrypticace (changes requested):
This implementation may end up modifying user written CSS incorrectly — so definitely don't think we should be doing this in `optimizeAst`.

A "better" way to do this would be to handle this in variants directly — not sure if we need a marker node other than `@slot` to notate that things can be wrapped _on the outside_ though as our entire variant implementation was built on nesting *inside* the class. 🤔

SHARMA1525:
> This implementation may end up modifying user written CSS incorrectly — so definitely don't think we should be doing this in `optimizeAst`.
> 
> A "better" way to do this would be to handle this in variants directly — not sure if we need a marker node other than `@slot` to notate that things can be wrapped _on the outside_ though as our entire variant implementation was built on nesting _inside_ the class. 🤔

hey @thecrypticace , Thanks for the feedback! I’ve updated the implementation so conditional at-rules like @scope are now handled directly in the variants layer instead of optimizeAst.
Changes were made in compile.ts, variants.ts, and plugin-api.ts to correctly wrap rules without affecting user-authored CSS.
All tests, including the new variants.scope.test.ts, are passing . Kindly Review!

PR #19308, opened 2025-11-14
perf(vite): add plugin hook filter

TheAlexLichter:
## Summary

This PR adds plugin [hook filters](https://vite.dev/guide/api-plugin#hook-filters) to the Vite plugin. They are backwards-compatible, having no impact on older Vite versions as the check for `isPotentialCssRootFile` is still included.

It can be dropped once you don't need to support older Vite versions.

## See also

https://github.com/e18e/ecosystem-issues/issues/171

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -97,23 +97,30 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
       name: '@tailwindcss/vite:generate:serve',
       apply: 'serve',
       enforce: 'pre',
+      transform: {
+        filter: {
+          id: {
+            exclude: [/\/\.vite\//, SPECIAL_QUERY_RE, COMMON_JS_PROXY_RE],
+            include: [/css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
+          },
+        },
+        async handler(src, id) {
+          if (!isPotentialCssRootFile(id)) return
 
-      async transform(src, id, options) {
-        if (!isPotentialCssRootFile(id)) return
+          using I = new Instrumentation()
+          DEBUG && I.start('[@tailwindcss/vite] Generate CSS (serve)')
 
-        using I = new Instrumentation()
-        DEBUG && I.start('[@tailwindcss/vite] Generate CSS (serve)')
+          let root = roots.get(id)
 
-        let root = roots.get(id)
+          let result = await root.generate(src, (file) => this.addWatchFile(file), I)
+          if (!result) {
+            roots.delete(id)
+            return src
+          }
 
-        let result = await root.generate(src, (file) => this.addWatchFile(file), I)
-        if (!result) {
-          roots.delete(id)
-          return src
-        }
-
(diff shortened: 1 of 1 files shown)


packages/@tailwindcss-vite/src/index.ts
```diff
@@ -97,23 +97,30 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
       name: '@tailwindcss/vite:generate:serve',
       apply: 'serve',
       enforce: 'pre',
+      transform: {
+        filter: {
+          id: {
+            exclude: [/\/\.vite\//, SPECIAL_QUERY_RE, COMMON_JS_PROXY_RE],
+            include: [/css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
+          },
+        },
```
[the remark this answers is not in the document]
TheAlexLichter: @coderabbitai The syntax is backwards-compatible as the original check is still included in the `handler` function.

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -97,23 +97,30 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
       name: '@tailwindcss/vite:generate:serve',
       apply: 'serve',
       enforce: 'pre',
+      transform: {
+        filter: {
+          id: {
+            exclude: [/\/\.vite\//, SPECIAL_QUERY_RE, COMMON_JS_PROXY_RE],
+            include: [/css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
```
thecrypticace: I think it should be this instead?

```suggestion
            include: [/\.css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
```

We use `path.extname(…)` which grabs the extension after the last `.` in the ID.

thecrypticace:
Ah, good call! Should probably add a test for `rolldown-vite` to our integrations too (I'll do that later in a separate PR).

Left one comment for you

packages/@tailwindcss-vite/src/index.ts
TheAlexLichter: Good catch ☺️

packages/@tailwindcss-vite/src/index.ts
```diff
@@ -123,31 +130,39 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
       apply: 'build',
       enforce: 'pre',
 
-      async transform(src, id) {
-        if (!isPotentialCssRootFile(id)) return
+      transform: {
+        filter: {
+          id: {
+            exclude: [/\/\.vite\//, SPECIAL_QUERY_RE, COMMON_JS_PROXY_RE],
+            include: [/css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
```
TheAlexLichter:
```suggestion
            include: [/\.css(?:\?.*)?$/, /&lang\.css/, INLINE_STYLE_ID_RE],
```

thecrypticace (approved):

thecrypticace:
Probably will end up doing a release next week at some point if I can get a couple of bug fixes into core.

Will be testable via our insiders build in like 20m or so:
```
```

PR #19329, opened 2025-11-18
Drop `sucrase` dependency from v3

thecrypticace:
Fixes #19327

It depends on a vulnerable version of `glob` and we used it mainly for faster config loading in v3. We have to delegate to babel when `import.meta` is used anyway so it seems fine to just always use babel instead.


thecrypticace:
We don't actually need to do this now since both `glob` and `sucrase` have updated stuff and the NPM advisory has been updated to reflect that the new version of `glob` is not vulnerable.

PR #19335, opened 2025-11-19
v3: Don’t break `sibling-*()` functions when used inside `calc(…)`

thecrypticace:
The implementation of v3's math operator normalization uses a safe-list of function names. Need to add `sibling-index()` and `sibling-count()` to this list otherwise when used inside math functions like `calc()` they'll get spaces around the `-`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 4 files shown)


RobinMalfait (approved):
Going to approve it, but I wonder how hard it would be to backport the v4 implementation instead which should handle this (and future functions) 🤔

thecrypticace:
@RobinMalfait I added a thing to handle custom functions as well but yeah maybe could backport from v4. Lemme look into that to see what it'd be like. 🤔

thecrypticace:
@RobinMalfait ported that from v4 👍

RobinMalfait (approved):

PR #18115, opened 2025-05-22
Fix lowercase+uppercase keys (camelCase/PascalCase) unexpectedly treated to kebab-case when using JavaScript config file

woohm402:
## Summary

Fixes #18114.

It Seems in `apply-config-to-theme.ts#keyPathToCssProperty` function the theme key is transformed as:
- `.` is replaced with `-`
- `/([a-z])([A-Z])/` is replaced with `$1-$2.toLowerCase()`

However in `theme.ts#resolve`, when failed to find exact key, it only replaces
- `.` is replaced with `-`

It seems the second transformation step is missing in theme.ts#resolve, causing inconsistencies when using camelCase or PascalCase theme keys.

## Test plan

All unit tests pass, including the new test case covering this behavior.

PR #19337, opened 2025-11-19
Don’t unconditionally convert config keys to `kebab-case`

### LINKED ISSUE #18114: Utility class name transforms lowercase+uppercase keys (camelCase/PascalCase) to kebab-case unexpectedly when using JavaScript config file

v4.1.7

`vite@6.2.0` & `@tailwindcss/vite@4.1.7`

v20.18.1

Chrome v136.0.7103.114

macOS

**Reproduction URL**

In my real project I use vite, but for minimal reproduction, I used `@tailwindcss/cli` here.

https://github.com/woohm402/bug-reproduction/tree/9479cdf

Reproduction guide is written in readme.

**Describe your issue**

When using a JavaScript-based config (tailwind.config.js) and defining theme keys that include a lowercase letter followed by an uppercase letter (e.g., lightGreen or LightGreen), the expected utility class (e.g., bg-lightGreen) is not generated. Instead, Tailwind generates bg-light-green, converting the key to kebab-case.

```js
export default {
  content: ["./index.html"],
  theme: {
    backgroundColor: {
      lightGreen: "#94d3a2",
    },
  },
};
```

- Expected: bg-lightGreen
- Actual: bg-light-green

| generated output.css | devtools element tab |
| --- | --- |
| Image | Image |

thecrypticace:
Fixes #18114
Closes #18115

This PR changes JS config handling such that we always preserve casing for theme keys (when possible).

Now there are *two* exceptions to this rule:

1. Top-level keys in the theme *do* get converted to kebab-case. As CSS variables are not generated from these this shouldn't be a big issue. 

All of our internal plugins look for kebab-case keys. So, for example, take the path `backgroundColor.red.500`. This must translate to `--background-color-red-500`. But if you had something like `backgroundColor.lightBlue` it would be perfectly fine for that to translate to `--background-color-lightBlue` internally (and thus the utility be written as `bg-lightBlue`.

2. Tuple object keys are converted to kebab-case as well.

These keys are converted to "nested" key syntax internally and typtically represent CSS property names.

For example:
```js
export default {
  theme: {
    fontSize: {
      xs: ["1.5rem", { lineHeight: "1.3" }]
    },
    
    fontFamily: {
      sans: ["Potato Mono", { fontVariationSettings: '"XHGT" 0.7' }]
    }
  }
}
```

The `lineHeight` key here must be converted to `line-height` because it represents a CSS property name. The theme key that represents this value is `--text-xs--line-height`. The same situation applies for the `fontVariationSettings` where the theme key is `--font-sans--font-variation-settings`.


RobinMalfait (approved):
It makes sense that if your _own_ names for values were described in camelCase that it should stay as camelCase instead of being converted to kebab-case.

This could be a breaking change technically, but then you already had a breaking change in the first place (because your camelCase name was converted to a kebab-case name).

If it causes too much trouble for people, we could enable both (and probably show a warning that you have a casing mismatch and that you should either convert the source or the usages).

thecrypticace:
Yeah, I think *if* this causes issues then we can try to do something similar to #18115 *in addition to this PR* to address that.

PR #19348, opened 2025-11-20
Improve compatibility with special default values in JS configs

### LINKED ISSUE #19345: Upgrade tool doesn't migrate `ringColor.DEFAULT` correctly
It should be migrated to `--default-ring-color` instead of `--ring-color`

thecrypticace:
Fixes #19345

In v3 the `ringColor.DEFAULT` option was used as the default color for ring utilities (when it was defined). This currently gets translated as `--ring-color` but that doesn't work this way in v4. Instead it should translate to `--default-ring-color` and *not* `--ring-color`.

I've also tweaked the upgrade tool to handle this properly as well.

### DIFF
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -2347,8 +2347,7 @@ test(
         --shadow-*: initial;
         --shadow: 0 1px 3px 0 rgb(0 0 0 / 0.1), 0 1px 2px -1px rgb(0 0 0 / 0.1);
 
-        --ring-width-*: initial;
-        --ring-width: 4px;
+        --default-ring-width: 4px;
 
         --blur: var(--custom-default-blur);
 
@@ -2381,7 +2380,7 @@ test(
         <div class="blur blur-xs"></div>
         <div class="backdrop-blur backdrop-blur-xs"></div>
         <div class="rounded rounded-sm"></div>
-        <div class="ring"></div>
+        <div class="ring-3"></div>
       </div>
       "
     `)
--- a/integrations/upgrade/index.test.ts
+++ b/integrations/upgrade/index.test.ts
@@ -2380,7 +2380,7 @@ test(
         <div class="blur blur-xs"></div>
         <div class="backdrop-blur backdrop-blur-xs"></div>
         <div class="rounded rounded-sm"></div>
-        <div class="ring-3"></div>
+        <div class="ring"></div>
       </div>
       "
     `)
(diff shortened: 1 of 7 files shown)


packages/tailwindcss/src/compat/apply-config-to-theme.ts
```diff
@@ -50,6 +50,18 @@ export function applyConfigToTheme(
...
+    // as to match the behavior of v3
+    if (path[0] === 'ringColor' && path[1] === 'DEFAULT') {
+      designSystem.theme.add(
+        '--default-ring-color',
+        '' + value,
+        ThemeOptions.INLINE | ThemeOptions.REFERENCE | ThemeOptions.DEFAULT,
+      )
+
+      continue
+    }
```
philipp-spiess: I'm a bit confused why this is handled in isolation here since we had a lot of other `DEFAULT` special keys that _are_ supported in v4 too via the `--default-*` pattern, e.g.: `--default-ring-width`, `--default-outline-width`, `--default-transition-duration`. So I'd expect there to a code path a bit more generic to handle all of these variants. 

From my quick look it seems that we didn't handle this in `apply-config-to-theme` at all right now so not sure if this is an oversight or if there's an alternative explanation?

thecrypticace: Yeah this is a good point. We *do* handle a few default things with regards to fonts but nothing else. Lemme do a bit of poking around to see if we need to handle a few more things.

integrations/upgrade/index.test.ts
```diff
@@ -2381,7 +2380,7 @@ test(
         <div class="blur blur-xs"></div>
         <div class="backdrop-blur backdrop-blur-xs"></div>
         <div class="rounded rounded-sm"></div>
-        <div class="ring"></div>
+        <div class="ring-3"></div>
```
RobinMalfait: This doesn't seem right I think. This test is making sure that if default values are different or implementations are different for a utility that we don't migrate because you would end up with a different result.

- `ring` in v3 mapped to `3px`, but in the config we set it to `4px` via `DEFAULT: 4px` which means that `ring` maps to `4px`.
- `ring` in v4 maps to `1px`, but if we use that `--default-ring-width` of `4px` then I think `ring` also means `4px`.

Now it looks like the `ring` was converted to `ring-3` which means that we go from `4px` (`ring` with default of `4px`) to `3px` (because of `ring-3`) which I believe is incorrect.

RobinMalfait (approved):

integrations/upgrade/index.test.ts
RobinMalfait: For future reference: We missed a configuration change where we had to update a variable namespace. See last commit: https://github.com/tailwindlabs/tailwindcss/pull/19348/commits/867206496a541ab812ee2b78ac1343dec50bae44

PR #19361, opened 2025-11-22
fix(ring-utility): remove extra comma after --tw-ring-inset (fixes #18923)"

saranganet:
This PR fixes issue #18923 by removing the extra comma after `--tw-ring-inset` in the ring utility.

### Changes Made
- Removed the trailing comma after `--tw-ring-inset` in the ring utility definition

### Testing
- [x] Tested the changes locally
- [x] Verified that the ring utility works as expected

Fixes #18923


thecrypticace:
The empty fallback is required for correctness. This change will stop several utilities from working whenever there's not an inset ring. See [this Play](https://play.tailwindcss.com/wf5iyWq1pp?file=css) for an example.

There's UI tests for this as well if you run:
```
pnpm run test:ui
```

PR #19364, opened 2025-11-23
Fix: Remove iOS Safari dialog max-width/max-height constraints in Preflight

swarnim02:
Fixes #19363 - Added dialog reset to Preflight CSS to remove iOS Safari's default max-width/max-height constraints on dialog elements.


swarnim02:
@thecrypticace please review this pr once
Whats the current status

thecrypticace:
So, this style is at least consistent across all browsers. Safari, Chrome, and FF on all platforms apply this UA style. It's *not* just a Safari + iOS thing. This style actually comes from the HTML spec:
https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3

I don't know why they chose this so I'm not sure if it's reasonable for us to reset it in preflight or not. Personally it feels like a maybe but will need to look into it further.

However, given that the styling behavior is at least consistent across all browsers, this would technically be a breaking change to add this in a patch release — even if it does only affect `<dialog>`s that were opened with `showModal()`.

I'm going to close this PR but we may revisit in the future if we decide that the benefit outweighs the potential impact.

PR #19368, opened 2025-11-24
Preserve aspect ratio of `svg` elements when changing their width

benface:
## Summary

Not sure if there's any reason why the "Constrain images and videos to the parent width and preserve their intrinsic aspect ratio" rule of Preflight doesn't apply to `svg` elements, as they behave pretty much the same as `img`s, in that they also accept `width` and `height` attributes and have an intrinsic aspect ratio. Adding `max-width: 100%` to an `svg` with these attributes ensures that it doesn't overflow its container, and adding `height: auto` ensures that changing its width (either due to the `max-width: 100%`, or to a `w-*` class) also changes the height proportionally, instead of it being locked to the value of the `height` attribute.

## Test plan

"Works ~~on my machine~~ in my projects"

### DIFF
--- a/packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
+++ b/packages/@tailwindcss-postcss/src/__snapshots__/index.test.ts.snap
@@ -128,7 +128,7 @@ exports[`\`@import 'tailwindcss'\` is replaced with the generated CSS 1`] = `
     display: block;
   }
 
-  img, video {
+  img, svg, video {
     max-width: 100%;
     height: auto;
   }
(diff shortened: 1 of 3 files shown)


thecrypticace:
For my reference later see also: https://github.com/tailwindlabs/tailwindcss/issues/506

thecrypticace:
So, it's possible that `<svg>` elements are *less* affected by the issues pointed out in #506 — especially given that inline SVGs often use styles + classes. However, I think *not* introducing this in a patch release is probably best just out of caution. Maybe we can revisit this stuff if/when we want to do a v5?

Gonna close this one for now but is something to keep in mind for the future.

PR #19369, opened 2025-11-24
Add `text-align` and `text-transform` to inherited properties of `button`, `input`, etc.

benface:
## Summary

This could be considered a bug fix, because `text-align` and `text-transform` usually inherit like those other "font styles" that are reset by Preflight on some elements (`button`, `input`, etc.) due to being set to a fixed value by user agent stylesheets. It seems like an oversight.

## Test plan

Pray this is not a breaking change?

### DIFF
--- a/packages/tailwindcss/preflight.css
+++ b/packages/tailwindcss/preflight.css
@@ -246,6 +246,8 @@ textarea,
   font-variation-settings: inherit; /* 1 */
   letter-spacing: inherit; /* 1 */
   color: inherit; /* 1 */
+  text-align: inherit; /* 1 */
+  text-transform: inherit; /* 1 */
   border-radius: 0; /* 2 */
   background-color: transparent; /* 3 */
   opacity: 1; /* 4 */


thecrypticace:
This is definitely breaking changes (almost *any* change to Preflight is).

For example, in this UI the buttons will now be aligned incorrectly *and* be uppercase: https://play.tailwindcss.com/5L25lT02oc

This is obviously a contrived example and maybe more real world impact would be smaller but it's definitely still a breaking change.

I don't think we'd want to do either of these things (personally I'd definitely not want `text-align` to be in there) but even if we did we'd have to come up with some kind of solution to opt in to these changes (otherwise we'd need to tag Tailwind CSS v5).

benface:
Hmm, right, I guess people actually rely on `<button>` having centered text by default. 🤔 I find it very strange personally, and inconsistent with the philosophy of Tailwind/Preflight, but fair enough! Thank you for considering it.

thecrypticace:
Gonna close this one since it's a breaking change — the `text-align` one especially feels like it could pop up in a bunch of places were we to ship this.

Issue #19343, opened 2025-11-20
`@tailwindcss/upgrade` doesn't upgrade `extends.content` correctly


simon-abbott:
**What version of Tailwind CSS are you using?**

v3.4.18

**What build tool (or framework if it abstracts the build tool) are you using?**

Not relevant

**What version of Node.js are you using?**

For example: v24.11.1

**What browser are you using?**

Not relevant

**What operating system are you using?**

macOS

**Reproduction URL**

https://play.tailwindcss.com/SZJTT4w0iF

**Describe your issue**

I am trying to upgrade from v3 to v4 using the auto-upgrade tool. I have a custom `content` value in my config. When I run the upgrade tool (`npx @tailwindcss/upgrade`) it migrates that setting over to be `--content-slash`, but then my linter errors because `after:content-slash` isn't a valid class.

ath1614:
@thecrypticace I would like to take this issue up. i have a solid solution and would like to raise a pr!

ath1614:
@thecrypticace Please review the change i made! Thanks

thecrypticace:
This will be fixed in the next release. We still recommend using arbitrary values for the `content-*` utility but `--content-*` theme keys will now work as expected.

PR #19377, opened 2025-11-26
Fix content-slash migration in @tailwindcss/upgrade tool

ath1614:
## Summary

Fixes issue #19343 where `@tailwindcss/upgrade` incorrectly converts `content-slash` to `--content-slash`, causing linter errors because `after:content-slash` isn't a valid class in Tailwind CSS v4.

The problem occurs when upgrading from v3 to v4 - the upgrade tool was missing the migration mapping for `content-slash` utilities. In v4, the correct syntax for using a slash character in the content utility is `content-['/']`.

**Changes made:**
- Added `'content-slash': "content-['/']"` mapping to `LEGACY_CLASS_MAP` in `migrate-simple-legacy-classes.ts`
- Added comprehensive test cases covering both basic usage and variant combinations
- Fixed TypeScript compatibility issue in tests with appropriate type casting

This ensures that utilities like `after:content-slash` are correctly migrated to `after:content-['/']` during the upgrade process.

## Test plan

**Verified the fix with:**

1. **Unit tests pass:**
   ```bash
   npm test -- packages/@tailwindcss-upgrade/src/codemods/template/migrate-simple-legacy-classes.test.ts
   ```
   All 14 tests pass, including the new test cases for `content-slash` migration.

2. **Test cases added:**
   - `content-slash` → `content-['/']` (basic migration)
   - `after:content-slash` → `after:content-['/']` (with pseudo-element variant)

3. **Migration verification:**
   The upgrade tool now correctly converts:
   - `content-slash` → `content-['/']`
   - `after:content-slash` → `after:content-['/']`
   - `before:content-slash` → `before:content-['/']`
   - Any variant combination with `content-slash`

**No breaking changes** - all existing functionality remains intact and all existing tests continue to pass.

### DIFF
--- a/packages/@tailwindcss-upgrade/src/codemods/template/migrate-simple-legacy-classes.test.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/migrate-simple-legacy-classes.test.ts
@@ -12,10 +12,12 @@ test.each([
   ['flex-shrink-0', 'shrink-0'],
   ['decoration-clone', 'box-decoration-clone'],
   ['decoration-slice', 'box-decoration-slice'],
+  ['content-slash', 'content-[\'/\']'],
 
   ['max-lg:hover:decoration-slice', 'max-lg:hover:box-decoration-slice'],
   ['max-lg:hover:decoration-slice!', 'max-lg:hover:box-decoration-slice!'],
   ['max-lg:hover:!decoration-slice', 'max-lg:hover:box-decoration-slice!'],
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hardcoding `content-slash` to migrate this way is incorrect. This was a custom theme value in the project from the original issue. The utility doesn't exist normally.

I have a solution here — just need to spend some time today to implement it.

Thanks anyway

ath1614:
@thecrypticace  Thanks

PR #19380, opened 2025-11-26
fix: add empty transform to vite scan plugin for Vite 7.0+ compatibility

ath1614:
## Summary

Fixes issue #19376 where `@tailwindcss/vite` plugin causes "Cannot read properties of undefined (reading 'call')" error in Vite 7.0.0, especially during massive code changes, formatting, or branch switches.

The problem occurs because Vite 7.0.0's plugin container system expects all plugins to have a `transform` property, but the first plugin in the Tailwind CSS plugin array (`@tailwindcss/vite:scan`) only had `configureServer` and `configResolved` hooks without a `transform` property.

**Root cause:**
- Vite 7.0.0's `pluginContainer.ts` tries to call `plugin.transform.call()` on all plugins
- The scan plugin was missing the `transform` property, causing `plugin.transform` to be `undefined`
- This resulted in the error when Vite tried to access `plugin.transform.call()`

**Changes made:**
- Added empty `transform()` function to the `@tailwindcss/vite:scan` plugin that returns `null`
- This ensures all plugins in the array have the expected `transform` property for Vite 7.0.0 compatibility

## Test plan

**Verified the fix with:**

1. **Build verification:**
   ```bash
   pnpm build
   ```
   The plugin builds successfully without TypeScript errors.

2. **Plugin structure verification:**
   - All three plugins now have the `transform` property defined
   - The scan plugin's `transform` returns `null`, telling Vite to skip transformation
   - Existing functionality remains unchanged

3. **Compatibility verification:**
   - ✅ Works with Vite 7.0.0's plugin container system
   - ✅ Backward compatible with older Vite versions
   - ✅ No performance impact (empty function returns immediately)
   - ✅ Prevents the "Cannot read properties of undefined" error

**No breaking changes** - all existing functionality remains intact. The fix is minimal and only adds the missing property to prevent the undefined error.

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -114,6 +114,11 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
           minify = opts.optimize.minify !== false
         }
       },
+
+      // Add empty transform to prevent Vite 7.0+ from trying to call undefined transform
+      transform() {
+        return null
+      },
     },
 
     {


ath1614:
@thecrypticace Can you please review this and tell me if there is any scope of improvement?

thecrypticace:
The transform hook is optional in Vite. If it's throwing this is likely a Vite bug.

thecrypticace:
See my comment here: https://github.com/tailwindlabs/tailwindcss/issues/19376#issuecomment-3580965156

I'm going to work on triaging this and filing a bug with Vite. Seems like something is causing a stale / incorrect cache and that is ultimately the bug here from what I can tell.

ath1614:
@thecrypticace  Thanks Man!

ath1614:
I thought i found the solution!

PR #19381, opened 2025-11-26
Handle backwards compatibility for `content` theme from JS configs

### LINKED ISSUE #19343: `@tailwindcss/upgrade` doesn't upgrade `extends.content` correctly

thecrypticace:
Fixes #19343

This PR makes it so the `content-*` utilities read from the `--content-*` theme namespace. This change is **purely for backwards compatibility with Tailwind CSS v3**. It is recommended you use arbitrary values with the `content-*` utility instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Skip over arbitrary property utilities with a top-level `!` in the value ([#19243](https://github.com/tailwindlabs/tailwindcss/pull/19243))
 - Support environment API in `@tailwindcss/vite` ([#18970](https://github.com/tailwindlabs/tailwindcss/pull/18970))
 - Preserve case of theme keys from JS configs and plugins ([#19337](https://github.com/tailwindlabs/tailwindcss/pull/19337))
+- Improve backwards compatibility for `content` theme key from JS configs ([#19381](https://github.com/tailwindlabs/tailwindcss/pull/19381))
 - Upgrade: Handle `future` and `experimental` config keys ([#19344](https://github.com/tailwindlabs/tailwindcss/pull/19344))
 
 ### Added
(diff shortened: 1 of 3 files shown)


packages/tailwindcss/src/compat/content-compat.ts
```diff
@@ -0,0 +1,26 @@
...
+export function registerContentCompat(userConfig: ResolvedConfig, designSystem: DesignSystem) {
+  let content = userConfig.theme.content || {}
+
+  if (typeof content !== 'object' || content === null) return
+
+  for (let [key, value] of Object.entries(content)) {
+    if (typeof value !== 'string') continue
+
+    designSystem.utilities.static(`content-${key}`, () => [
+      decl('--tw-content', value),
```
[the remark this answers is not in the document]
thecrypticace: Yes but also no. The value could also be a CSS variable so this is fine.

integrations/upgrade/js-config.test.ts
```diff
@@ -1840,6 +1840,92 @@ describe('border compatibility', () => {
...
+        }
+        @utility content-slashslash {
+          --tw-content: '//';
+          content: var(--tw-content);
+        }
+        @property --tw-content {
+          syntax: '*';
+          inherits: false;
+          initial-value: '';
+        }
```
RobinMalfait: It does expose a little bit of internal information, but I don't hate this.
As mentioned before, reading from `--content` could also work. Not sure what's best 🤔

thecrypticace: Yeah, those are our two options. The one where we read from `--content` is a *much* smaller change. So it depends on which one we want to do.

thecrypticace:
I've updated this to read from `--content` instead.

RobinMalfait (approved):

PR #19385, opened 2025-11-28
[v3 Changelog] Fix date

hebbet:
## Summary
version 3.4.17 was released in 2025 and not 2024

## Test plan

Not needed


thecrypticace:
Ah whoops. I thought this was editing an entry on the main branch so I pushed a fix into the v3 branch just now b/c I went to check it (so don't need the PR). Thanks for pointing this out though!

PR #19410, opened 2025-12-04
Safe UTF-8 Extraction and Robust <style> Block Detection in Upgrade Codemods

Saniya777-byte:
This PR improves the safety and robustness of Tailwind's candidate extraction and codemod migration logic in two critical areas:

1. **Rust UTF-8 Extraction**
   - Replaces `unsafe { String::from_utf8_unchecked(...) }` with safe `String::from_utf8(...)` in candidate extraction.
   - Skips any invalid UTF-8 candidates rather than panicking, providing a safer and more resilient extraction process.
2. **TypeScript Style Block Handling**
   - Switches from manual string-index logic to parsing using `node-html-parser` for detecting `<style>` blocks in codemod logic.
   - Ensures accurate detection of all `<style>` tags (including those with attributes and complex nesting), so migrations never accidentally affect style blocks.
   - Adds `node-html-parser` as a dependency to the `@tailwindcss-upgrade` package.
3. **pnpm Workspace Guidance**
   - Includes documentation for installing dependencies within the correct workspace package, avoiding root install issues.

These changes modernize the affected modules, prevent potential runtime errors from malformed files, and support robust migrations across real-world HTML/template structures.

---

## Test plan

**1. UTF-8 Candidate Extraction**
- Added test cases using invalid and valid UTF-8 byte candidates
- Ran `cargo test` to ensure the scanner skips bad candidates without panicking

**2. Codemod `<style>` Block Handling**
- Added migration examples with plain, nested, and attribute-rich `<style>` tags for multiple template syntaxes
- Confirmed classes inside `<style>` blocks are not migrated
- Ran:
  ```sh
  cd packages/@tailwindcss-upgrade
  pnpm install
  pnpm test

### DIFF
--- a/crates/oxide/src/scanner/mod.rs
+++ b/crates/oxide/src/scanner/mod.rs
@@ -553,11 +553,17 @@ where
             a
         })
         .into_iter()
-        .map(|s| unsafe { String::from_utf8_unchecked(s.to_vec()) })
+        .filter_map(|s| match String::from_utf8(s.to_vec()) {
+            Ok(s) => Some(s),
+            Err(_e) => {
+                // Optionally log or handle invalid UTF-8 here
+                // eprintln!("Skipped invalid UTF-8 candidate, error: {:?}", _e);
+                None
+            }
+        })
         .collect();
 
-    // SAFETY: Unstable sort is faster and in this scenario it's also safe because we are
-    //         guaranteed to have unique candidates.
+    // Unstable sort is performant & remains correct with unique candidates
     result.par_sort_unstable();
 
     result
--- a/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
+++ b/packages/@tailwindcss-upgrade/src/codemods/template/is-safe-migration.ts
@@ -1,4 +1,5 @@
 import { parseCandidate } from '../../../../tailwindcss/src/candidate'
+import { parse as parseHtml } from 'node-html-parser'
 import type { DesignSystem } from '../../../../tailwindcss/src/design-system'
 import { DefaultMap } from '../../../../tailwindcss/src/utils/default-map'
 import * as version from '../../utils/version'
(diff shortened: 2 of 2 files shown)


crates/oxide/src/scanner/mod.rs
```diff
@@ -553,11 +553,17 @@ where
...
+                // Optionally log or handle invalid UTF-8 here
+                // eprintln!("Skipped invalid UTF-8 candidate, error: {:?}", _e);
+                None
+            }
+        })
         .collect();
 
-    // SAFETY: Unstable sort is faster and in this scenario it's also safe because we are
-    //         guaranteed to have unique candidates.
+    // Unstable sort is performant & remains correct with unique candidates
```
thecrypticace, line 566: While not a traditional safety message this is 100% intentional because behavior would be incorrect in the face of non-unique candidates + an unstable sort.

crates/oxide/src/scanner/mod.rs
```diff
@@ -553,11 +553,17 @@ where
...
         .into_iter()
-        .map(|s| unsafe { String::from_utf8_unchecked(s.to_vec()) })
+        .filter_map(|s| match String::from_utf8(s.to_vec()) {
+            Ok(s) => Some(s),
+            Err(_e) => {
+                // Optionally log or handle invalid UTF-8 here
+                // eprintln!("Skipped invalid UTF-8 candidate, error: {:?}", _e);
+                None
+            }
+        })
```
thecrypticace, line 563: This is an intentional performance optimization. `String::from_utf8` does a validation pass over the string which should be uncnessary here. The extractor's implementation *should* guarantee that the resulting candidates are already UTF-8.

If you have a test case for this that crashes when scanning files please open a separate issue and I will take a look.

Saniya777-byte:
Thanks for the clarification!

That makes sense—I wasn't aware the extractor always guarantees valid UTF-8 candidates. I'll leave the existing performance-oriented implementation as is. If I ever encounter a crash or edge case where invalid candidates slip through, I'll open a separate issue with a reproducible example.

PR #19427, opened 2025-12-09
v4: Allow `@apply` to be used with CSS mixins

thecrypticace:
The CSS Custom Functions and Mixins spec [is using `@apply` with dashed idents](https://drafts.csswg.org/css-mixins-1/#apply-rule) for native mixin support. We shouldn't attempt to treat these as utilities and try to compile them since they'll fail.

There one intentional limitation with regards to mixins and our use of `@apply`: We do not allow users to mix utilities and mixins in the same at-rule. In other words, all of the following `@apply` rules are invalid:

```css
.foo {
  /* Invalid because the rules contain both mixins and utilities */
  @apply --my-mixin underline;
  @apply --my-mixin() underline;
  @apply underline --my-mixin;
  @apply underline --my-mixin();
}
```

Aside: Lightning CSS does not yet support this syntax so the results of a production build won't produce the correct code but we'll at least handle these correctly in Tailwind CSS itself.

Fixes #19422


RobinMalfait:
Opened an issue on Lightning CSS' repo: https://github.com/parcel-bundler/lightningcss/issues/1232
Also opened a corresponding PR to try and keep unknown at-rules printed in the same location: https://github.com/parcel-bundler/lightningcss/pull/1233

RobinMalfait (approved):
To properly support it we need to wait for Lightning CSS, but if you are using the CDN, or not using Lightning CSS then this should start working.

PR #19429, opened 2025-12-09
Fix provenance setup for v3

thecrypticace:


package.json
```diff
@@ -5,7 +5,10 @@
   "license": "MIT",
   "main": "lib/index.js",
   "types": "types/index.d.ts",
-  "repository": "https://github.com/tailwindlabs/tailwindcss.git",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tailwindlabs/tailwindcss.git#v3"
```
RobinMalfait, line 10: I think this is correct, but it's weird that GitHub doesn't go to the `v3` branch when you open this link 😅

RobinMalfait (approved):

package.json
thecrypticace: `npm repo` does tho

Issue #19445, opened 2025-12-15
`color-mix()` should not be transpiled with custom properties


hirasso:
**What version of Tailwind CSS are you using?**

v4.1.18

**What build tool (or framework if it abstracts the build tool) are you using?**

vite, reproducible in playground

**What version of Node.js are you using?**

n/a

**What browser are you using?**

n/a

**What operating system are you using?**

n/a

**Reproduction URL**

https://play.tailwindcss.com/vfK8S6Q1Ry?file=css (look at the class `.via-ease-in-sine` in the "Generated CSS" panel)

**Describe your issue**

I've got a custom directive using `color-mix` with two custom properties, as [described here](https://github.com/tailwindlabs/tailwindcss/discussions/19441).

## What I expected

I'd expect no transpilation to happen if I use custom properties in color-mix:

```css
.via-ease-in-sine {
  --tw-gradient-via-stops: var(--tw-gradient-position),
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 0%) 0%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 0.8%) 8.1%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 2.9%) 15.5%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 6.2%) 22.5%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 10.1%) 29%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 14.8%) 35.3%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 20.1%) 41.2%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 26.2%) 47.1%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 32.8%) 52.9%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 40%) 58.8%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 47.8%) 64.7%,
    color-mix(in oklab, var(--tw-gradient-from), var(--tw-gradient-to) 56.6%) 71%,
[...]

PR #19456, opened 2025-12-17
feat: only polyfill internal, intentional usages of `color-mix`

hirasso:
Previous discussion: #19445 

> [!NOTE] 
> This is a proof of concept PR. I don't expect it to be merged, but thought it would be easier to reason about the sense and nonsense of my idea with an actual implementation at hand

## Summary

Prefix internal usage of `color-mix` for adding alpha to colors with `tw-`, so that optimzations (replacing the rule with the first color value, adding a polyfill) isn't applied to user-land CSS.

## Alternatives considered

Use a custom comment instead, e.g. `/* tw:polyfill-color-mix */ color-mix()`. Went for the "hacky" approach with the non-native `tw-color-mix()` function for now, as this approach is much more straight-forward in the CSS parsing and AST functions.

## Test plan & other Notes

- Added a (first-failing) test "does not apply optimizations to native color-mix()"
- Updated all tests that were failing
- Apply the alpha in the `:placeholder` rule in preflight manually, to keep the css valid case that file would be loaded without directly from somewhere

## Breaking Change

What speaks against this: It would be a breaking change. People might be relying on tailwindcss polyfilling their custom `color-mix` CSS.

Feel free to close this. But at least now we have something to look at in the future, should it come up somewhere else again 🙂


hirasso:
Yeah ok, I'll close this and spare everyone the trouble I went through :)

PR #19480, opened 2025-12-23
Allow `@variant` to be used inside `addBase`

thecrypticace:
This seems like an unintentional omission


RobinMalfait (approved):
I think this makes sense, we already handle `@apply` and we also handle `@variant` in the other JS based APIs.

Thanks!

PR #19499, opened 2025-12-29
chore: remove duplicate esbuild devDependencies

43081j:
pnpm has got itself in a state it can't seem to get out of.

basically, it has 3 versions of esbuild install (0.23.x, 0.27.x, 0.25.x)
thanks to the fact we have `tsx` installed.

we don't depend on `tsx` anywhere, nor do any dependencies in the entire
tree. pnpm is holding onto it for no reason.

it is an _optional_ peer of vite and postcss-load-config (via tsup), one
which we don't depend on anywhere and so it shouldn't be installed.

to fix this, i did a bunch of pnpm-fu to remove and re-add the target packages, and it now seems to have cleaned up the duplicates. 

**At some point, we should probably just remove the lock file and re-generate it**. I think there'll be many other cases like this.

## Test plan

Existing tests should cover this.


43081j:
FYI i caught it up as the problem still exists in main.

there are multiple copies of esbuild, and tsx is installed but unused.

RobinMalfait:
Hey! Thanks for the PR, but removing `pnpm-lock.yaml` and re-generating it results in the same lockfile.

When running `pnpm why esbuild tsx`, you get:
image

`tsx` exists because it's used by `tsup`, multiple versions exist of `esbuild` because of the `esbuild` used by `tsx` is different then the one used by `vitest` for example.

Fwiw, they are all dev-dependencies, so you are not paying for these dependencies when you install Tailwind CSS.

Going to keep this as-is.

43081j:
indeed this is purely a DX thing for tailwind maintainers.

what you said isn't quite right though. tsx is an _optional_ peer of all of these packages.

vite:
https://github.com/vitejs/vite/blob/c917f1ef9d9c6ef131af96d89089d8ec680b18f2/packages/vite/package.json#L186-L188

postcss-load-config:
https://github.com/postcss/postcss-load-config/blob/33179e97575403a8040a3f53a6b7c32f2551b753/package.json#L71-L73

etc.

tailwind itself does not use tsx - so there's no reason it is installed.

it should only have installed itself if we pulled it in somewhere in a non-optional way (which would then satisfy the _optional_ peer of both of the above).

also:

```sh
git clean -xdf
rm pnpm-lock.yaml
pnpm i
pnpm why -r tsx
# NO OUTPUT
```

removing the lockfile and reinstalling does remove tsx - because it is an optional peer, one we don't use.

RobinMalfait:
Yeah that makes sense. I don't believe this is a DX issue, but fewer dependencies is always better.

I opened a new PR: #20102, and made sure to add you as a co-author. Thanks!

PR #19506, opened 2025-12-31
aligning item in display block

nitin01924:
here, the display is set to block and doing the "vertical-align: middle;", while it is not in block.
there should be the "align-items: center; " and take use of flexbox.

### DIFF
--- a/integrations/cli/index.test.ts
+++ b/integrations/cli/index.test.ts
@@ -2405,7 +2405,7 @@ test(
           list-style: none;
         }
         img, svg, video, canvas, audio, iframe, embed, object {
-          display: block;
+          display: flex;
           align-items: center;
         }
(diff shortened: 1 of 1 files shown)


wongjn:
Hi, thank you for your contribution. As per [this maintainer comment](https://github.com/tailwindlabs/tailwindcss/issues/9046#issuecomment-1207578129) regarding `vertical-align`:

> Basically the reason this exists is because the `display` of these replaced elements can still be changed to inline by users and the `vertical-align` is there to target those scenarios. You can find more details in the link in the comment above the code: https://github.com/jensimmons/cssremedy/issues/14#issuecomment-634934210

Furthermore, this could be a breaking change, so I'm going to close this, but thank you again for you contribution.

PR #19515, opened 2026-01-03
Add radius utilities and deprecate rounded

adamwathan:
_This is just a test of using Claude Code in the Cloud, not really planning to merge this but very curious to see how it did!_

_This was my prompt:_

> Create a new set of utilities called `radius-*` that duplicate the existing `rounded-*` utilities but in that new namespace, and treat the old utilities as legacy/deprecated. I'm pretty sure there's somewhere in the codebase where legacy utilities are handles separately from regular ones so they aren't suggested in IntelliSense and stuff.
> 
> Make the new utilities also support the spacing scale, so `radius-4` would be `border-radius: 1rem`. Make sure to support all four corners, just left, just top, just bottom right, etc. like we do with other things.
> 
> Most of this work should be just in the `tailwindcss` package (this is a monorepo so there are multiple packages). Make sure to add tests, try to follow existing testing conventions as much as possible.

---

Introduce new `radius-*` utilities that support both the `--radius` theme and the spacing scale (e.g., `radius-4` = `calc(--spacing * 4)`).

- Add `radius-*` utilities with all corner variants (radius-t, radius-tl, etc.)
- Support spacing scale values (radius-4, radius-0.5, etc.)
- Support px values (radius-px)
- Move `rounded-*` utilities to legacy-utilities.ts
- Mark `rounded-*` utilities as legacy (hidden from IntelliSense suggestions)
- Add comprehensive tests for both new and legacy utilities


dandaniel5:
? in my case today
`absolute top-0 right-0 z-50  bg-white/20 backdrop-blur-sm hover:bg-white/40 cursor-pointer active:scale-95 transition-all  flex flex-col items-center gap-4 p-4 rounded-l-full rounded-tr-none`

neither of this two works
`rounded-l-full rounded-tr-none etc.`

works only
`rounded-full etc.`

tailwindcss 4.2.0, tailwind-merge 3.5.0, next 16.1.6

dandaniel5:
also absolute -top-8 not working proppa
however right-4  is work alright

PR #19534, opened 2026-01-08
Add funding info to package.json

jeffreyvr:
Just noticed funding property in `package.json` files had not been set, perhaps by choice, but just in case.

## Summary

Adding funding information to package.json files, such that users of Tailwind CSS might become aware of sponsor opportunities.

Screenshot 2026-01-08 at 17 28 50

## Test plan

NA. Will give funding info upon package installation and running `npm fund`.


mariocandela:
Well done!

PR #19539, opened 2026-01-09
display: none !important; #19538

Lucifer007-max:
Bug  #19538 

This change keeps existing behavior intact while reducing selector impact by moving the :where() wrapper to the full selector.

It does not remove !important or change semantics, but improves composability for component libraries and advanced overrides

### DIFF
--- a/packages/tailwindcss/preflight.css
+++ b/packages/tailwindcss/preflight.css
@@ -388,6 +388,6 @@ input:where([type='button'], [type='reset'], [type='submit']),
   Make elements with the HTML hidden attribute stay hidden by default.
 */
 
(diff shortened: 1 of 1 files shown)


adamwathan:
> Hey! We do this to make sure the `hidden` attribute on an element always takes precedence over the CSS, and the precedent we took it from originally was Bootstrap which also does it with `!important`. It is a breaking change from v3 but it's a breaking change to switch it back as well 😕 I think this is just one of those things where some people like it, some people don't but we can't change it at this point without tagging a v5 even if we wanted to, so going to leave it as is for now. Thanks regardless!

PR #19587, opened 2026-01-21
Fix utility generaFix utility generation for escaped theme keystion for escaped theme keys

KashviYadav09:
### Summary

This PR fixes an issue where utilities are not generated when theme keys contain escaped special characters (e.g. `brand\:primary`).

Tailwind allows escaped characters in theme configuration keys, but named utility values were not being normalized before theme lookup, which caused matching to fail.

### What’s changed

- Added normalization for escaped theme keys
- Ensured named utility values are unescaped before resolving against the theme
- Added tests to cover escaped theme keys in configuration and utility generation

### Example

```js
// tailwind.config.js
theme: {
  colors: {
    'brand\\:primary': '#ff0000',
  },
}

### DIFF
--- a/packages/tailwindcss/src/candidate.ts
+++ b/packages/tailwindcss/src/candidate.ts
@@ -5,12 +5,16 @@ import { isValidArbitrary } from './utils/is-valid-arbitrary'
 import { segment } from './utils/segment'
 import * as ValueParser from './value-parser'
 import { walk, WalkAction } from './walk'
+import { normalizeEscapedKey } from './utils/escape'
+
 
 const COLON = 0x3a
 const DASH = 0x2d
 const LOWER_A = 0x61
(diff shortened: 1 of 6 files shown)


RobinMalfait:
Hey! Thanks for the PR, but going to say no to this one for now unless more people run into this issue.

While we can merge this and fix it, our Oxide scanner (the part that extracts classes from template files) won't pick up values such as `bg-brand\:primary` so this PR on its own won't fix the issue you are seeing. We can allow it in Oxide as well, but it would make things more complex.

This also introduces some new Tailwind CSS syntax that is unfamiliar. I would instead recommend to use `bg-brand-primary`. It's the same amount of characters but without the added complexity and it works as-is today.

If more people run into this then I'm more than happy to re evaluate, but for now I'm going to close it.

PR #19581, opened 2026-01-20
Upgrade lightningcss to 1.31.0

t0byman:
## Summary

Upgrade lightningcss to [1.31.0](https://github.com/parcel-bundler/lightningcss/releases/tag/v1.31.0), mainly to [add support for scroll-state container queries](https://github.com/parcel-bundler/lightningcss/issues/887)

## Test plan

I ran `pnpm test` and got 25 failed tests, that's 3 more than there were with lightningcss 1.30.2 Not sure how to proceed.

PR #19608, opened 2026-01-28
Bump dependencies

RobinMalfait:
This PR bumps a bunch of dependencies. This also moves a few dependencies that we use in multiple packages to the pnpm catalog.

Closes: #19603, #19604, #19576, #19575, #19573, #19565, #19547, #19546, #19545, #19609, #19581, #19620, #19619

- #19603
- #19604
- #19576
- #19575
- #19573
- #19565
- #19547
- #19546
- #19545
- #19609
- #19581
- https://github.com/tailwindlabs/tailwindcss/pull/19620 
- #19620
- #19619

## Test Plan

All tests in CI should still pass. [ci-all]


packages/tailwindcss/src/index.test.ts
```diff
@@ -1561,7 +1561,8 @@ describe('Parsing theme values from CSS', () => {
       ),
     ).toMatchInlineSnapshot(`
       ":root, :host {
-        --animate-very-long-animation-name: very-long-animation-name var(--very-long-animation-name-configuration, 2.5s ease-in-out 0s infinite normal none running);
+        --animate-very-long-animation-name: very-long-animation-name
+                    var(--very-long-animation-name-configuration, 2.5s ease-in-out 0s infinite normal none running);
```
RobinMalfait, line 1565: The weird formatting here is because it's formatted this way in the input of the test above. The latest version of Lightning CSS doesn't normalize whitespace in custom properties anymore, which maps to the spec https://drafts.csswg.org/css-variables/#serializing-custom-props

> Specified values of [custom properties](https://drafts.csswg.org/css-variables/#custom-property) must be serialized exactly as specified by the author (with one exception, below). Simplifications that might occur in other properties, such as dropping comments, normalizing whitespace, reserializing numeric tokens from their value, etc., must not occur.

> The one exception to serialization being precisely as specified in the source file is that an empty custom property (such as --foo:;) must serialize with a single space as its value.

packages/tailwindcss/src/index.test.ts
```diff
@@ -5650,7 +5651,7 @@ describe('`color-mix(…)` polyfill', () => {
 
       @supports (color: color-mix(in lab, red, red)) {
         .text-red-500\\/50 {
-          color: color-mix(in oklab, var(--color-red-500) 50%, transparent);
+          color: color-mix(in oklab,var(--color-red-500)50%,transparent);
```
RobinMalfait, line 5654: The input doesn't contain spaces, so this one doesn't contain spaces either. While this looks a little bit odd, it's correct. This is also the output we would produce in Tailwind Play or the CDN because we don't use Lightning CSS there.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -5764,27 +5764,27 @@ test('rotate-x', async () => {
 
     .-rotate-x-\\(--var\\) {
       --tw-rotate-x: rotateX(calc(var(--var) * -1));
-      transform: var(--tw-rotate-x, ) var(--tw-rotate-y, ) var(--tw-rotate-z, ) var(--tw-skew-x, ) var(--tw-skew-y, );
+      transform: var(--tw-rotate-x,  ) var(--tw-rotate-y,  ) var(--tw-rotate-z,  ) var(--tw-skew-x,  ) var(--tw-skew-y,  );
```
RobinMalfait, line 5767: These are a little bit silly but not the end of the world. This is because we run Lightning CSS twice to get better output. But they have a bug where they always introduce 1 additional whitespace character in these fallbacks.

I fixed this upstream: https://github.com/parcel-bundler/lightningcss/pull/1142

So we likely will have a similar PR in the near future where we fix these again.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -28599,7 +28600,7 @@ describe('custom utilities', () => {
         .text-sm {
           font-size: var(--text-sm, .8755rem);
           line-height: var(--text-sm--line-height, 1.255rem);
-          text-rendering: optimizeLegibility;
+          text-rendering: optimizelegibility;
```
RobinMalfait, line 28603: This is a Lightning CSS change, see: https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.foo%20%7B%5Cn%20%20text-rendering%3A%20optimizeLegibility%3B%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%221.31.1%22%7D

Note: make sure you switch the version in the dropdown to 1.31.1 to see the change. If you're looking at this in the future, you can toggle between 1.30.2 and 1.31.0 to see the difference.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -29582,7 +29583,7 @@ describe('custom utilities', () => {
           }
 
           .example-\\[7\\/9\\] {
-            --value: 7 / 9;
+            --value: 7/9;
```
RobinMalfait, line 29586: This is because we take the input as-is and don't normalize whitespace here.

packages/tailwindcss/src/utilities.ts
```diff
@@ -421,7 +421,7 @@ export function createUtilities(theme: Theme) {
           if (value === null && desc.supportsFractions && candidate.value.fraction) {
             let [lhs, rhs] = segment(candidate.value.fraction, '/')
             if (!isPositiveInteger(lhs) || !isPositiveInteger(rhs)) return
-            value = `calc(${candidate.value.fraction} * 100%)`
+            value = `calc(${lhs} / ${rhs} * 100%)`
```
RobinMalfait, line 424: In `calc` expressions whitespace can be important for certain operators. The main reason this changed is because it will produce nicer output that's always correct.

packages/tailwindcss/tests/ui.spec.ts
```diff
@@ -1911,7 +1911,7 @@ test('outline style is optional', async ({ page }) => {
     html`<div id="x" class="outline-2 outline-white"></div>`,
   )
 
-  expect(await getPropertyValue('#x', 'outline')).toEqual('rgb(255, 255, 255) solid 2px')
+  expect(await getPropertyValue('#x', 'outline')).toEqual('2px solid rgb(255, 255, 255)')
```
RobinMalfait, line 1914: Normalized this because the order of arguments is different between WebKit and Chromium for some reason...

playgrounds/v3/package.json
```diff
@@ -9,18 +9,18 @@
...
-    "react-dom": "^19.2.3",
+    "next": "^16.1.6",
+    "react": "^19.2.4",
+    "react-dom": "^19.2.4",
     "tailwindcss": "^3"
   },
   "devDependencies": {
-    "@types/node": "^20.14.8",
-    "@types/react": "^19.2.7",
+    "@types/node": "^25.0.10",
```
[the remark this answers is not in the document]
RobinMalfait: Not relevant anymore. This is using `catalog:` now which points to the shared `^20.19.0` version.

packages/tailwindcss/src/utilities.test.ts
```diff
@@ -26645,12 +26646,12 @@ test('inset-shadow', async () => {
...
       --tw-inset-shadow-alpha: 25%;
-      --tw-inset-shadow: inset 12px 12px var(--tw-inset-shadow-color, #08c), inset 12px 12px var(--tw-inset-shadow-color, var(--value, #08c));
+      --tw-inset-shadow: inset 12px 12px var(--tw-inset-shadow-color, #08c),inset  12px 12px var(--tw-inset-shadow-color, var(--value, #08c));
     }
 
     @supports (color: lab(from red l a b)) {
       .inset-shadow-\\[12px_12px_\\#0088cc\\,12px_12px_var\\(--value\\,\\#0088cc\\)\\]\\/25 {
-        --tw-inset-shadow: inset 12px 12px var(--tw-inset-shadow-color, oklab(59.9824% -.067 -.124 / .25)), inset 12px 12px var(--tw-inset-shadow-color, oklab(from var(--value, #08c) l a b / 25%));
+        --tw-inset-shadow: inset 12px 12px var(--tw-inset-shadow-color, oklab(59.9824% -.067 -.124 / .25)),inset  12px 12px var(--tw-inset-shadow-color, oklab(from var(--value, #08c) l a b / 25%));
       }
```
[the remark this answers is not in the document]
RobinMalfait: This has been cleaned up.

PR #16274, opened 2025-02-05
Take `NODE_PATH` into account when resolving modules

philipp-spiess:
Fixes #15847

This PR changes the node specific resolver config to takes the eventual `NODE_PATH` env into account.

## Test plan

See #15847

Screenshot 2025-02-05 at 12 58 20

PR #19617, opened 2026-01-29
support NODE_PATH in standalone build

SteffenDE:
References https://github.com/tailwindlabs/tailwindcss/pull/19391.
References https://github.com/tailwindlabs/tailwindcss/pull/16274.

Right now, when using the standalone build of the TailwindCSS CLI, you cannot use a custom `NODE_PATH`, but you can when using it via Node.js directly.

A custom NODE_PATH allows you to resolve imports from multiple locations. For example, in [Phoenix LiveView](https://github.com/phoenixframework/phoenix_live_view/), we have a feature where you can write scripts in templates that we extract at compile time to a custom folder and users can import those in their application bundle by saying

```javascript
import { hooks as colocatedHooks } from "phoenix-colocated/my_app"
```

where the "phoenix-colocated" folder lives in a different location than the usual `node_modules` folder. This works fine with the default esbuild setup, as it respects `NODE_PATH`, so we can pass it a custom location.

We want to also support colocating CSS in templates soon, but the same approach doesn't work with the standalone Tailwind CLI we ship with default Phoenix projects. It works when running Tailwind through Node.js, but we don't want to tell users they need to install it, just to use the feature.

This patch changes the lookup logic for the standalone CLI to also account for `NODE_PATH`. Note that you can pass multiple paths, that are split according the the OS PATH separator.


SteffenDE:
Any chance to get this merged? :)

RobinMalfait (approved):

RobinMalfait:
Thanks! Will be available in the next release, and once merged it will be available in the `insiders` version soon.

SteffenDE:
Nice, thank you!

PR #19626, opened 2026-01-30
feat: handle backslash in `@utility` name

### LINKED ISSUE #19607: @utility doesn't allow for escaped utility names

v4.1.18

**Reproduction URL**

https://play.tailwindcss.com/LrHsRLSUuF?file=css

**Describe your issue**

The `@utility` function allows you to create utilities with special characters in them like a forward slash and it will automatically escape them (`@utility my/utility { ... }` -> `.my\/utility { ... }`)

However, if you pass it an already escaped utility name, it will throw an error: 

```
`@utility my\/utility` defines an invalid utility name. Utilities should be alphanumeric and start with a lowercase letter.
``` 

This is a problem as some linters like biome will correctly fail to parse the unescaped forward slash version, as it's not valid css: https://biomejs.dev/playground/?code=QAB1AHQAaQBsAGkAdAB5ACAAbQB5AC8AdQB0AGkAbABpAHQAeQAgAHsACgAgACAAYwBvAGwAbwByADoAIAByAGUAZAA7AAoAfQA%3D&language=css

Best solution would probably be to relax the naming rules around utilities and ensure if the special characters are already escaped, they don't get escaped again

rozsazoltan:
Resolves #19607

## Summary

I believe there is no obstacle to simply ignoring backslashes in the name. This way, various validators - which are not aware of Tailwind CSS's specific syntax (which allows the `/` character to be used directly in utility names) - can still be bypassed using backslashes. For example, instead of `@utility push-1/2`, one could use `@utility push-1\/2`, while the end result would be identical.

## Test plan

I took a previous utility test as a baseline and extended it with backslashes, and I expect the same result in the output as in the original test case:

* https://github.com/rozsazoltan/tailwindcss/blob/main/packages/tailwindcss/src/utilities.test.ts#L28553-L28567 (the original test case I started from)
* https://github.com/rozsazoltan/tailwindcss/blob/feat/handle-blackslash-in-utility-name/packages/tailwindcss/src/index.test.ts#L4640-L4654 (the current PR's test)


RobinMalfait (approved):
Thanks!

PR #19669, opened 2026-02-12
fix: support decimal values in aspect ratio utilities

shtse8:
## Problem

The `aspect-ratio` utility only accepts integer values for bare fractions (e.g. `aspect-4/3`), but the CSS `aspect-ratio` property supports any positive number. Using a decimal like `aspect-8.5/11` silently produces no output.

Reported in #19663.

## Solution

Added an `isPositiveNumber` helper (alongside the existing `isPositiveInteger`) that accepts finite non-negative numbers including decimals. Updated the `aspect` utility's `handleBareValue` to use `isPositiveNumber` instead of `isPositiveInteger`.

### Before
```html
<div class="aspect-8.5/11">  <!-- ❌ No CSS generated -->
```

### After
```html
<div class="aspect-8.5/11">  <!-- ✅ aspect-ratio: 8.5 / 11 -->
```

## Changes

- `packages/tailwindcss/src/utils/infer-data-type.ts` — Added `isPositiveNumber()` utility function
- `packages/tailwindcss/src/utilities.ts` — Changed aspect ratio validation from `isPositiveInteger` to `isPositiveNumber`
- `packages/tailwindcss/src/utilities.test.ts` — Added test case for `aspect-8.5/11`

Fixes #19663

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -3156,15 +3156,20 @@ test('aspect-ratio', async () => {
         }
         @tailwind utilities;
       `,
-      ['aspect-video', 'aspect-[10/9]', 'aspect-4/3'],
+      ['aspect-video', 'aspect-[10/9]', 'aspect-4/3', 'aspect-8.5/11'],
     ),
   ).toMatchInlineSnapshot(`
-    ":root, :host {
+    "/*! tailwindcss v4.1.18 | MIT License | https://tailwindcss.com */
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey! Thanks for the PR. But implemented it a bit different in another PR (https://github.com/tailwindlabs/tailwindcss/pull/19688) to not allow all positive numbers, but only multiples of `.25`. This makes it consistent with other utilities that accept bare values (without the square brackets).

PR #19670, opened 2026-02-13
fix: restore full page reload for watched external files on Vite 7.1+

Arxsher:
# PR: Fix @source file changes not triggering full page reload on Vite 7.1+

## Description
This PR addresses issue #19637 where template files (PHP, HTML, Blade, etc.) watched via the `@source` directive fail to trigger a full page reload when using Vite 7.1 or newer.

## Root Cause
Vite 7.1 introduced the Environment API, which supersedes the legacy WebSocket API for HMR. Specifically:
- `server.ws.send` is deprecated/ignored for certain external file updates in favor of `server.hot.send`.
- The `@tailwindcss/vite` plugin currently collects `ViteDevServer` instances but lacks a `handleHotUpdate` hook to explicitly trigger reloads for non-module files added via `addWatchFile`.

## Changes
- Implemented a `handleHotUpdate` hook in the `@tailwindcss/vite` plugin.
- The hook identifies changes to files that are not part of the standard Vite module graph (e.g., `.php`, `.html`) but are watched by Tailwind.
- Triggers a `full-reload` using the new `server.hot.send` API if available (Vite 7.1+), with a fallback to `server.ws.send` for backward compatibility.

## Verification
- Reproduced the issue in a standalone Vite 7.1.0 project using a mock plugin with the legacy API.
- Confirmed that the browser fails to reload upon editing a watched `.php` file.
- Verified that migrating to `server.hot.send` restores the expected reload behavior.

[ci-all]


Arxsher:
Thanks for the cleanup @RobinMalfait, Appreciate the help getting this ready.

RobinMalfait:
@Arxsher thank you for the PR! Just doing a few more tests here and then I think we can get it merged!

RobinMalfait:
Alright, did more testing and added integration tests. I also reworked the entire `handleHotUpdate` logic because there were some issues with the implementation. Wrote more about it in the commit itself: https://github.com/tailwindlabs/tailwindcss/pull/19670/changes/6f749ababb419a23213b576e6caf26b00ba1bead

But longs story short:

1. Update to the `hotUpdate` API since `handleHotUpdate` will be deprecated in a future version
2. Because of the `addWatchFile` call, the `.php` files _will_ be in the moduleGraph. But they are just marked as an `type: 'asset'`
3. We also have to make sure that we are updating based on the current environment
4. We have to make sure that the file is being watched by Tailwind CSS. We check the actual scanned files instead of looking at the watched files in general.

Not sure what your test setup looked like, but in my testing we never did a full reload because the very first check to know whether a file was in the moduleGraph bailed out early.

RobinMalfait (commented):
Thanks!

Arxsher:
Thanks for the detailed breakdown and the refactor, @RobinMalfait! The standalone reproduction I used was a bit more isolated, which explains why the module graph behavior looked different there. Really glad to see the robust solution with the new `hotUpdate` API and the comprehensive tests.

Arxsher:
Thanks for the breakdown and the refactor, @RobinMalfait! My standalone setup was likely too simplified, so I missed that those files were still hitting the module graph. The switch to the new `hotUpdate` API and the extra tests look solid.

PR #19679, opened 2026-02-15
Fallback to `config.createResolver` for `client` and `ssr` environments in `@tailwindcss/vite`

4ndv:
## Summary

Sometimes even if Vite Envrionment API is available, some plugins are still override `config.createResolver` function to inject own aliases

Since technically `config.createResolver` was only [properly deprecated](https://github.com/vitejs/vite/pull/20031) in Vite 7.0.0, it's still a valid(-ish) to do so, even if it wasn't ever officially supported

Vite already handles this in its internal css resolvers, but not exposes the code to do so as part of public API, so I've copied and adapted it

Fixes #19677

## Test plan

Tested by copying built package into my repro from the issue, also ran vite integration tests


RobinMalfait (approved):
Thanks! Added an integration test to make sure this solves the issue you are seeing.

PR #19680, opened 2026-02-15
Allow decimal values in aspect ratio utilities

veeceey:
Fixes #19663

The `aspect-*` utility currently only accepts integer ratios (e.g. `aspect-4/3`) because `handleBareValue` validates both sides of the fraction with `isPositiveInteger`. This means perfectly valid ratios like `aspect-8.5/11` are silently dropped from the stylesheet.

CSS `aspect-ratio` supports any positive number ratio, so there's no reason to restrict it to integers. This PR adds an `isPositiveNumber` check and uses it for the aspect utility, so `aspect-8.5/11` now correctly generates:

```css
.aspect-8\.5\/11 {
  aspect-ratio: 8.5 / 11;
}
```

Tested by adding `aspect-8.5/11` to the existing aspect-ratio test case and confirming the output. All 382 utility tests pass, no regressions.

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -3156,7 +3156,7 @@ test('aspect-ratio', async () => {
         }
         @tailwind utilities;
       `,
-      ['aspect-video', 'aspect-[10/9]', 'aspect-4/3'],
+      ['aspect-video', 'aspect-[10/9]', 'aspect-4/3', 'aspect-8.5/11'],
     ),
   ).toMatchInlineSnapshot(`
     ":root, :host {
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey! Thanks for the PR. But implemented it a bit different in another PR (https://github.com/tailwindlabs/tailwindcss/pull/19688) to not allow all positive numbers, but only multiples of `.25`. This makes it consistent with other utilities that accept bare values (without the square brackets).

Also, it looks like somebody already implemented a PR for the issue, so in the future make sure to look for existing PRs otherwise you might be working on something that was already solved and I don't want you to waste precious time 👍

veeceey:
Thanks for the feedback @RobinMalfait! The approach of restricting to multiples of `.25` for consistency with other bare-value utilities makes a lot of sense. I should have checked for existing PRs before starting — will be more diligent about that going forward. Appreciate you taking the time to explain the reasoning!

PR #19687, opened 2026-02-17
Add '.jj' to ignored content directories

bb:
Ignore .jj directory, just like .git, .hg, .svn.

https://www.jj-vcs.dev/latest/


RobinMalfait (approved):
Thanks!

PR #19688, opened 2026-02-17
Allow multiples of `.25` in `aspect-*` fractions

RobinMalfait:
This PR reduces the restrictions of the `aspect-*` utilities when dealing with fractional values.

Up until now, the numbers had to be positive integers, so `aspect-1/2` was valid, but `aspect-8.5/11` was not.

This PR allows for any multiple of `.25` as a valid value, so `aspect-8.5/11` is now valid, but `aspect-8.3/11` is not, this will still require `aspect-[8.3/11]` arbitrary value syntax to be valid.

This behavior of allowing multiples of `.25` is consistent with other utilities that handle bare values such as `w-2.5`.

## Test plan

1. Existing tests pass
2. Added a test for `aspect-8.5/11`

Fixes: #19663
Closes: #19680, #19669


shtse8:
Hey @RobinMalfait, thanks for picking this up! Would you mind adding a `Co-authored-by` credit since I raised the issue and submitted the original PR (#19669)? Would appreciate it 🙏

```
Co-authored-by: shtse8 <shtse8@gmail.com>
```

PR #19691, opened 2026-02-17
Fix @utility to accept already-escaped utility names

### LINKED ISSUE #19607: @utility doesn't allow for escaped utility names

shtse8:
This PR fixes a bug where using CSS escape sequences in `@utility` names would throw a validation error, even though escaping is required by strict CSS linters like Biome.

## Problem

Writing `@utility my/utility` is technically invalid CSS (an unescaped `/` in an identifier). Strict linters like Biome will flag this with a parse error. The correct CSS syntax is `@utility my\/utility` (with the slash escaped).

However, Tailwind currently throws:

> \`@utility my\/utility\` defines an invalid utility name. Utilities should be alphanumeric and start with a lowercase letter.

This is because the CSS parser preserves raw escape sequences in `node.params`, and the validator didn't account for CSS escape sequences.

## Fix

Unescape the utility name before validation and registration in `createCssUtility`, using the existing `unescape` utility from `utils/escape.ts`. This means:

- `@utility my\/utility` (escaped) is accepted and treated as equivalent to `@utility my/utility` (unescaped)
- The utility is registered under the unescaped name (`my/utility`) so that candidates in HTML continue to match correctly
- No double-escaping: the generated CSS selector is `.my\/button` in both cases

## Test plan

1. Existing tests pass
2. Added a test verifying `@utility my\/utility` generates `.my\/utility { … }`
3. Added a test verifying the escaped and unescaped forms produce identical output

Fixes #19607

### DIFF
--- a/packages/tailwindcss/src/index.test.ts
+++ b/packages/tailwindcss/src/index.test.ts
@@ -4787,6 +4787,59 @@ describe('@utility', () => {
       `[Error: \`@utility ui/button/sm\` defines an invalid utility name. Utilities should be alphanumeric and start with a lowercase letter.]`,
     )
   })
+
+  // https://github.com/tailwindlabs/tailwindcss/issues/19607
+  test('@utility supports escaped utility names', async () => {
+    // Linters like biome require the `/` to be escaped as `\/` in CSS.
+    // Tailwind should accept the already-escaped form `my\/utility` and treat
+    // it as equivalent to the unescaped `my/utility`.
+    await expect(
+      compileCss(
+        css`
+          @utility my\/utility {
+            display: inline-flex;
+            background: blue;
+          }
+          @tailwind utilities;
+        `,
+        ['my/utility'],
+      ),
+    ).resolves.toMatchInlineSnapshot(
+      `
+      ".my\\/utility {
+        background: #00f;
+        display: inline-flex;
+      }"
+    `,
+    )
+  })
+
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! I appreciate the PR but please only open PRs for issues that don't have an open PR already. 

Going to close this in favor of the PR that solves this issue already: https://github.com/tailwindlabs/tailwindcss/pull/19626

PR #19700, opened 2026-02-19
Correct CHANGELOG to mention inset not inline

ginnyTheCat:
I noticed this when reading the changelog to migrate a project to the new version. I'm not sure whether retroactively changing the changelog is considered acceptable behaviour in this project or whether appending a notice in the next changelog would be better.

Funnily this was actually guessed in #19613, though hard to notice among the other many false guesses.


RobinMalfait (approved):
Thanks!

PR #19723, opened 2026-02-25
Fix webpack loader cache key for resource queries

colinaaa:
## Summary

`@tailwindcss/webpack` currently uses `this.resourcePath` as the cache key, which ignores the resource query. When the same CSS file is imported multiple times with different `resourceQuery` values, all of those imports share a single `CacheEntry`. That means the utilities discovered for one entry can leak into the CSS output for another entry.

This PR changes the cache key to use `this.resource` (path + query) instead, while still using `this.resourcePath` for all filesystem work.

## Test plan

- `pnpm test:integrations -- webpack/loader.test.ts`  
  - Confirms all existing webpack loader integration tests pass.  
  - Confirms the new `@tailwindcss/webpack loader isolates cache by resource including query` test passes, verifying that two entries importing the same CSS file with different queries produce isolated outputs (`dist/a.css` only contains `only-a` / `--color-red-500`, and `dist/b.css` only contains `only-b` / `--color-blue-500`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Experimental_: Add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
I've updated the integration test to make it pass in CI. The `\n` were causing a bit of issues because we were embedding it in JS to write to the file, so I think we should've used `\\n` but I got rid of it anyway.

Thanks!

PR #19725, opened 2026-02-25
Guard object lookups against inherited prototype properties

### LINKED ISSUE #19721: Tailwind v4 crashes when scanned source contains plain string `row-constructor` (`V.map is not a function`)

- `tailwindcss@4.2.1`
- `@tailwindcss/postcss@4.2.1`

(Also seen in `4.1.x` in my project)

### What build tool (or framework) are you using?

PostCSS (minimal reproduction), also reproducible in Vite + `@tailwindcss/postcss`

`v25.4.0`

N/A (build-time crash)

macOS (Apple Silicon)

---

## Describe your issue

Tailwind v4 crashes during CSS build if a scanned source file contains the plain string:

`row-constructor`

This is **not** an actual class used in HTML/JSX, just a string in source code (e.g. a test title / log message).

The crash is:

`V.map is not a function`

It appears Tailwind parses `row-constructor` as the `row-*` utility with candidate value `constructor`, then hits a prototype property (`constructor`) instead of an own static value entry and crashes.

### Actual behavior

Build fails with an error like:

```txt
tailwindcss: .../app.css:1:1: V.map is not a function
TypeError: V.map is not a function
Expected behavior
Tailwind should ignore unsupported/invalid candidate values safely and not crash.

Minimal reproduction
1) Create a minimal project
mkdir tw-row-constructor-repro
cd tw-row-constructor-repro
npm init -y
npm i -D tailwindcss@4.2.1 @tailwindcss/postcss@4.2.1 postcss
mkdir src
2) Add a scanned source file containing the string row-constructor
test.ts

const label = 'row-constructor'
console.log(label)
3) Add Tailwind CSS entry with explicit source scanning (v4 style)
app.css

@source "./src";
4) Run PostCSS with @tailwindcss/postcss
node -e "import postcss from 'postcss'; import tw from '@tailwindcss/postcss'; import fs from 'fs'; postcss([tw()]).process(fs.readFileSync('app.css','utf8'),{from:process.cwd()+'/app.css'}).then(()=>console.log('PASS')).catch(e=>{console.error(e); process.exit(1);});"
5) Result
Build crashes with:

TypeError: V.map is not a function
...
tailwindcss: .../app.css:1:1: V.map is not a function
Why I think this is a Tailwind bug (not app code)
The trigger is a plain string literal in scanned source.
Removing or changing row-constructor avoids the crash.
Restricting Tailwind scan sources to exclude that file works as a workaround.
This reproduces in a minimal project (no framework-specific code required).
Workarounds (for others)
Avoid scanning non-frontend/test source files
Restrict v4 scan sources with @source
Rename the string (e.g. row constructor) to avoid row-constructor
Suspected cause (optional hint)
This looks like a prototype lookup issue when resolving a static value for the row utility:

candidate value becomes constructor
lookup returns Object.prototype.constructor (a function)
Tailwind then calls .map(...) on it
If helpful, I can provide a patched local debug trace that showed the candidate value was constructor.

adamwathan:
When user-controlled candidate values like "constructor" are used as
keys to look up values in plain objects (staticValues, plugin values,
modifiers, config), they can match inherited Object.prototype properties
instead of returning undefined. This caused crashes like "V.map is not
a function" when scanning source files containing strings like
"row-constructor".

Use Object.hasOwn() checks before all user-keyed object lookups in:
- utilities.ts (staticValues lookup)
- plugin-api.ts (values, modifiers, and variant values lookups)
- plugin-functions.ts (get() config traversal function)

Fixes #19721

https://claude.ai/code/session_011CYSGw3DLh2Z8xnuyoaCgC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Experimental_: Add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
 
+### Fixed
+
+- Guard object lookups against inherited prototype properties ([#19725](https://github.com/tailwindlabs/tailwindcss/pull/19725))
+
 ## [4.2.1] - 2026-02-23
 
 ### Fixed
--- a/packages/tailwindcss/src/compat/plugin-api.test.ts
+++ b/packages/tailwindcss/src/compat/plugin-api.test.ts
@@ -4592,4 +4592,48 @@ describe('config()', () => {
 
     expect(fn).toHaveBeenCalledWith('defaultvalue')
   })
+
+  // https://github.com/tailwindlabs/tailwindcss/issues/19721
+  test('matchUtilities does not match Object.prototype properties as values', async ({
+    expect,
+  }) => {
+    let input = css`
+      @tailwind utilities;
+      @plugin "my-plugin";
+    `
+
+    let compiler = await compile(input, {
+      loadModule: async (id, base) => {
+        return {
+          path: '',
+          base,
+          module: plugin(function ({ matchUtilities }) {
+            matchUtilities(
+              {
+                test: (value) => ({ '--test': value }),
+              },
+              {
+                values: {
+                  foo: 'bar',
+                },
+              },
+            )
+          }),
+        }
+      },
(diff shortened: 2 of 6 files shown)


RobinMalfait (approved):
Made a few small changes, but looks good otherwise.

PR #19759, opened 2026-03-06
fix(node): handle character class globs in source() root check

siewcapital:
## Summary

Fixes a false-positive validation error when using `source(...)` with glob character classes such as `[a-z]`.

`@tailwindcss/node` validates that the static prefix of the `source(...)` path exists before scanning. The previous implementation only treated `*` and `{` as glob markers, so patterns like `./src/[a-z]/**/*.html` were interpreted as literal directories and failed with:

> The `source(...)` does not exist or is not a directory.

This change updates the static-prefix detection to also stop at `[` and `?`, which are also valid glob metacharacters.

## Changes

- In `ensureSourceDetectionRootExists`, expand glob-symbol detection from `/[*{]/` to `/[*{[?]/`.
- Add `compile.test.ts` in `@tailwindcss/node` to cover `source("./src/[a-z]/**/*.html")` and verify compilation succeeds.

## Test plan

- `pnpm vitest packages/@tailwindcss-node/src/compile.test.ts`
- `pnpm vitest packages/@tailwindcss-node/src/*.test.ts`

### DIFF
--- a/packages/@tailwindcss-node/src/compile.test.ts
+++ b/packages/@tailwindcss-node/src/compile.test.ts
@@ -1,0 +1,27 @@
+import fs from 'node:fs/promises'
+import path from 'node:path'
+import { expect, test } from 'vitest'
+import { compile } from './compile'
+
+test('source() accepts glob patterns with character classes', async () => {
+  let base = await fs.mkdtemp(path.join(process.cwd(), '.tmp-tailwind-source-glob-'))
+
+  try {
+    await fs.mkdir(path.join(base, 'src'), { recursive: true })
+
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! I appreciate the PR but we want to keep things simple and only support `*`, `**` and brace expansion. Handing ranges with `[…]` syntax would make things more complex and would be a breaking change, especially for projects using Next.js where that syntax exists in the file name e.g. https://nextjs.org/docs/app/api-reference/file-conventions/dynamic-routes#convention

For that reason, I'm going to say no to this PR right now. Thanks again!

PR #19777, opened 2026-03-10
Allow multiple @utility definitions with same name but different value types

mvanhorn:
## Summary

Fixes #16948

When defining multiple CSS `@utility foo-*` with different value types (e.g., one for colors, one for numbers), only the first handler was tried. If it returned `null` (value didn't match), the compile loop stopped, preventing subsequent handlers from being attempted.

```css
@utility foo-* {
  color: --value(--color-*);
}
@utility foo-* {
  font-size: --spacing(--value(number));
}
```

Previously, `foo-red-500` worked but `foo-123` did not (or vice versa depending on definition order).

The fix distinguishes between CSS `@utility` handlers and JS plugin `matchUtilities` handlers:
- **CSS `@utility`** (no typed options): `null` means "try the next handler" - allows multiple definitions with different value types to coexist
- **JS `matchUtilities`** (with explicit types): `null` means "the value was invalid for this type, stop" - preserves existing behavior where typed utilities prevent invalid values from falling through

## Test plan

- Added test: two `@utility foo-*` definitions with different value types - verifies both `foo-red-500` (color) and `foo-123` (number) produce correct CSS
- All 4621 existing tests pass (including the `matchUtilities` type-safety tests)
- `pnpm build && pnpm test` passes

This contribution was developed with AI assistance (Claude Code).


RobinMalfait (approved):
Thanks! Did additional testing related to legacy `matchUtilities` behavior and I think the current implementation is fine. 

We could also allow `matchUtilities` that have `any` as a type to continue, but that could result in more (unwanted) changes in current projects. We could also _not_ care about the types, and just ignore the results of the current plugin when we return `null` (just like we do in `@utility`) but that would also probably cause unwanted changes in current projects.

PR #19796, opened 2026-03-14
Add `--stream` flag to `canonicalize` subcommand

aptinio:
## Summary

- Adds `--stream` flag to `tailwindcss canonicalize` that reads candidate groups from stdin line by line and writes canonicalized results to stdout
- Keeps the design system loaded across requests, making it suitable as a long-running sidecar process
- Empty lines pass through, keeping request/response pairs aligned

## Motivation

Non-JS tools (formatters, editor plugins, etc.) currently have no lightweight way to canonicalize Tailwind classes. The existing batch mode works for one-off use, but tools that need to canonicalize repeatedly pay the cost of loading the design system each time.

With `--stream`, a tool can start `tailwindcss canonicalize --stream` once and send candidate groups over stdin as needed:

```sh
$ echo -e "py-3 p-1 px-3\nmt-2 mr-2 mb-2 ml-2" | tailwindcss canonicalize --stream
p-3
m-2
```

Related discussion: https://github.com/tailwindlabs/tailwindcss/discussions/19736

### DIFF
--- a/packages/@tailwindcss-cli/src/commands/canonicalize/canonicalize.test.ts
+++ b/packages/@tailwindcss-cli/src/commands/canonicalize/canonicalize.test.ts
@@ -120,7 +120,7 @@ describe('runCommandLine', { timeout: 30_000 }, () => {
     let input = Readable.from('py-3 p-1 px-3\nmt-2 mr-2 mb-2 ml-2\n')
     let { stream: output, collect: collectOutput } = createOutput()
 
-    await streamStdin({ css, cwd: path.dirname(css), input, output })
+    await streamStdin({ css, cwd: path.dirname(css), format: 'text', input, output })
 
     expect(collectOutput()).toBe('p-3\nm-2\n')
   })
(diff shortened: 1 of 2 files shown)


RobinMalfait (approved):
Hey! I appreciate the PR. The `canonicalize` command was still in a proof of concept phase but was already merged so we can test things out.

That said, I do like the `--stream` option so that we don't have to spawn the command every time, just like you mentioned. One thing you can do is collect all groups first, and then call it once. But I can see that that might not be as ergonomic.

I'm also conflicted on the `--stream` name, we might change this before we properly release this. I like `--stdio` as an option as well, but we already communicate via stdin/stdout even without this long running process so let's keep `--stream` for now.

Thanks!

aptinio:
Thanks for the review and the merge!

On batching — in my case (`Phoenix.LiveView.HTMLFormatter`), the formatter hook processes each class attribute value one at a time, so collecting groups up front isn't an option. Streaming fits that model well.

One thing I'd love to see before this is properly released: renaming the subcommand to something like `sort-classes`, with canonicalization as a flag (e.g. `--canonicalize`). That'd likely better match what most consumers are using it for — class sorting — which is what the Prettier plugin does too. Happy to put up a PR for that if you're interested.

RobinMalfait:
The initial version used to be called `candidates` which could be invoked with some flags: `candidates --sort --canonicalize --collapse`, but we changed it to just `canonicalize` because that's what we're trying to do, generate the most canonical form of a set of classes which means that it will be sorted, that each utility is written in its canonical form and that multiple utilities would be collapsed into single classes when possible. While `sort-classes` could work, it does feel incomplete.

Maybe we rename it in the future, or introduce separate commands, but for now we want to keep it as canonicalize.

aptinio:
That makes sense — canonicalize covers more than just ordering. Thanks for the context on the earlier `candidates` iteration.

PR #19797, opened 2026-03-14
fix: remove unused braces dependencies

SidWuChou:
Removes unused braces and @types/braces from @tailwindcss/upgrade package.json as reported in issue #19794


RobinMalfait (approved):
Thanks! We use the `braces` package in one of the other packages, but good catch here!

PR #19799, opened 2026-03-14
docs: add noopener to README external link

pavan-sh:
## Summary

Adds rel="noopener noreferrer" to the external Tailwind CSS homepage link in README.md that uses target="_blank". This is a small security hardening to prevent window.opener-based tabnabbing and to avoid leaking referrer information when opening the link in a new tab.

## Test plan

Verified the README change renders correctly on GitHub (link still opens in a new tab, now with rel="noopener noreferrer" present).


RobinMalfait:
Thanks for the PR, but this is unnecessary because this literally links to the documentation page not to a random URL.

PR #19803, opened 2026-03-15
fix(vite): resolve tsconfig paths in CSS and JS resolvers

nightah:
Change `aliasOnly` from `true` to `false` when calling Vite's resolver so that the full resolution pipeline runs, including the oxc resolver responsible for tsconfig path resolution.

When `aliasOnly` was `true`, only the @rollup/plugin-alias plugin ran, which meant `resolve.tsconfigPaths: true` had no effect on CSS `@import` or JS `@plugin` resolution in `@tailwindcss/vite`.

Closes #19802.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix crash in canonicalization step when handling utilities with empty property maps ([#19727](https://github.com/tailwindlabs/tailwindcss/pull/19727))
 - Skip full reload for server only modules scanned by client CSS when using `@tailwindcss/vite` ([#19745](https://github.com/tailwindlabs/tailwindcss/pull/19745))
 - Add support for Vite 8 in `@tailwindcss/vite` ([#19790](https://github.com/tailwindlabs/tailwindcss/pull/19790))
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey! Looks like CI is failing. Do you mind making sure that things work as expected? Thanks!

nightah:
Thanks for the ping @RobinMalfait. I believe this is now fixed.

When `aliasOnly` was changed to `false`, the full Vite resolution pipeline could resolve bare specifiers like `tailwindcss` to JS entry points (e.g. `lib.mjs`) instead of CSS files. The enhanced-resolve fallback, which respects `mainFields: ['style']`, never ran.

That said, I'm unsure whether the remaining test failures are related to my change or an environmental issue.
Following the contribution guidelines after getting to `pnpm build` and `pnpm test:integrations` against a4be9838 for a baseline, the following tests failed for me:

- oxide/wasm.test.ts: wasm32-wasi loading (timeout)
- postcss/source.test.ts: (these look like they didn't even fail in CI)
- vite/nuxt.test.ts: build (timeout)
- vite/qwik.test.ts: dev mode (pnpm install warning causing failure)

RobinMalfait (approved):
Running all the integration tests locally can result in issues when ports are already occupied. In CI, we split them so that doesn't happen. What I typically do locally is only run a single integration test at a time (typically the one I'm working on).

Since they pass in CI, I wouldn't worry about them locally 👍 

Thanks for the contribution!

PR #19804, opened 2026-03-15
Cleanup inconsistencies related to (regex) escapes

RobinMalfait:
This PR does some generic cleanup to the codebase.

I'm playing with oxfmt and oxlint and noticed some unnecessary escapes. Might add these dependencies to the project later (and rolldown for building). But baby steps for now.

## Test plan

1. All tests should still pass

### DIFF
--- a/packages/tailwindcss/src/utils/infer-data-type.ts
+++ b/packages/tailwindcss/src/utils/infer-data-type.ts
@@ -184,7 +184,7 @@ function isPercentage(value: string): boolean {
 
 /* -------------------------------------------------------------------------- */
 
-const IS_FRACTION = new RegExp(`^${HAS_NUMBER.source}\s*/\s*${HAS_NUMBER.source}$`)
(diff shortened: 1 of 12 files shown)


packages/tailwindcss/src/utils/infer-data-type.ts
```diff
@@ -184,7 +184,7 @@ function isPercentage(value: string): boolean {
 
 /* -------------------------------------------------------------------------- */
 
-const IS_FRACTION = new RegExp(`^${HAS_NUMBER.source}\s*/\s*${HAS_NUMBER.source}$`)
+const IS_FRACTION = new RegExp(`^${HAS_NUMBER.source}\\s*/\\s*${HAS_NUMBER.source}$`)
```
RobinMalfait, line 187: This was technically a bug, because it would never match `1 / 2` (with spaces). However, in all of our cases, we don't have spaces anyways and `1/2` did match this regex.

The only reason it worked, is because we had the literal `s` with `*`, meaning that `s` has to occur 0 or more times. Which means that `100s/s200` technically matched...

PR #19824, opened 2026-03-18
Replace custom brace expansion with brace-expansion library for faster pattern matching

marco-carvalho:
## Summary

Replace custom brace expansion with brace-expansion library for faster pattern matching

**Why:**
- **2x faster** on the project's own benchmark pattern (1.49M expansions: ~72ms → ~24ms).
- **Less code to maintain** — 90 lines of custom parsing reduced to a 10-line wrapper with a single guard for step=0 (which causes an infinite loop in the library).
- **Built-in DoS protection** — the library caps expansions at 100,000 by default, which the custom implementation did not.
- **Zero-padding support** — `{001..9}` now correctly produces `001, 002, …, 009` instead of stripping leading zeros.

**Behavioral changes:**

| Behavior | Before | After |
| - | - | - |
| Zero-padding (`{00..05}`) | Stripped → `0, 1, …` | Preserved → `00, 01, …` |
| Decimal ranges (`{1.1..2.2}`) | `1.1..2.2` | `{1.1..2.2}` (braces preserved as literal) |
| Unbalanced braces | Throws `Error` | Graceful degradation (unmatched braces kept as literal text) |
| Step=0 (`{0..5..0}`) | Throws `Error` | Still throws (wrapper guard) |

## Test plan

- Ran the existing brace-expansion unit tests with updated expectations:
  ```
  cd packages/tailwindcss && npx vitest run src/utils/brace-expansion.test.ts
  # 22 tests passed
  ```
- Ran the full tailwindcss test suite to verify no regressions:
  ```
  cd packages/tailwindcss && npx vitest run
  # 4,117 tests passed | 1 skipped
  ```
- Benchmarked before/after on the project's own large expansion pattern (1.49M items):
  ```
  Before (custom):            mean 72.40ms, 13.8 ops/s
  After  (brace-expansion):   mean 24.27ms, 41.2 ops/s
  Speedup:                    2.1x
  ```

### DIFF
--- a/packages/tailwindcss/package.json
+++ b/packages/tailwindcss/package.json
@@ -130,6 +130,7 @@
     "@jridgewell/remapping": "^2.3.5",
     "@tailwindcss/oxide": "workspace:^",
     "@types/node": "catalog:",
+    "brace-expansion": "^5.0.4",
     "dedent": "1.7.1",
     "lightningcss": "catalog:",
     "magic-string": "^0.30.21",
--- a/packages/tailwindcss/src/utils/brace-expansion.test.ts
+++ b/packages/tailwindcss/src/utils/brace-expansion.test.ts
@@ -19,9 +19,12 @@ describe('expand(…)', () => {
     ['a/{10..0..5}/b', ['a/10/b', 'a/5/b', 'a/0/b']],
     ['a/{10..0..-5}/b', ['a/0/b', 'a/5/b', 'a/10/b']],
 
-    // Numeric range with padding (we do not support padding)
-    ['a/{00..05}/b', ['a/0/b', 'a/1/b', 'a/2/b', 'a/3/b', 'a/4/b', 'a/5/b']],
-    ['a{001..9}b', ['a1b', 'a2b', 'a3b', 'a4b', 'a5b', 'a6b', 'a7b', 'a8b', 'a9b']],
+    // Numeric range with zero-padding
+    ['a/{00..05}/b', ['a/00/b', 'a/01/b', 'a/02/b', 'a/03/b', 'a/04/b', 'a/05/b']],
+    [
+      'a{001..9}b',
+      ['a001b', 'a002b', 'a003b', 'a004b', 'a005b', 'a006b', 'a007b', 'a008b', 'a009b'],
+    ],
 
     // Numeric range with step
     ['a/{0..5..2}/b', ['a/0/b', 'a/2/b', 'a/4/b']],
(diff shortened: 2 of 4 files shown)


RobinMalfait:
Hey! What kind of issues were you running into that triggered this PR?

Brace expansion is only used for our `@source inline('…')` feature (https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-variants) and we explicitly don't support certain features because they don't make sense in a Tailwind CSS class context.

For example:
- Zero-padding support: we don't support utilities like `w-0001`, so it doesn't make sense for us to support this
- Decimal ranges: similar to decimal ranges, we support multiples of .25, but `w-1.2` is not supported, so this also doesn't really make sense here.
- Unbalanced braces: this throws on purpose to inform the user that something is wrong
- Built-in DoS protection: Tailwind CSS is not executed at runtime, and the `@source inline(…)` feature doesn't handle arbitrary inputs except from the hardcoded list in the CSS. Worst case if you manage to get it to expand forever, your memory goes up and/or your dev process crashes. A production website will not be affected. 

So I'm really curious about the issues you were running into.

marco-carvalho:
hi @RobinMalfait there wasn't a specific issue, just a codebase exploration for maintenance wins, but your points are well taken, so happy to close this.

RobinMalfait:
No worries, was just interested in what caused this. You did peak my interest with the `2x faster` part, so wanted to see where we can make things faster even if this is not a hot-path code. But first, I wanted to gather some data:

On a small pattern like: `{text,bg}-{red,green,blue}-{50,{100..900..100},950}`, there will be 66 generated classes. In this scenario, the `braces-expansion` package is ~4.6 times slower than the current implementation:
image

On a large pattern like: `{{xs,sm,md,lg}:,}{border-{x,y,t,r,b,l,s,e},bg,text,cursor,accent}-{{red,orange,amber,yellow,lime,green,emerald,teal,cyan,sky,blue,indigo,violet,purple,fuchsia,pink,rose,slate,gray,zinc,neutral,stone}-{50,{100..900..100},950},black,white}{,/{0..100}}`, there will be 1 493 280 generated classes.

It looked like in this scenario the `braces-expansion` packages is almost ~2x faster, but that's only because it does indeed cap out at 100k generated classes:
image

If we increase that limit to a much bigger number, the `braces-expansion` package is ~10x slower compared to the current implementation:
image

Here is the exact code I used for the benchmark:
```js
import { expand as braceExpansionExpand } from 'brace-expansion'
import { expand as bracesExpand } from 'braces'
import { expand } from './brace-expansion'

import { bench, boxplot, do_not_optimize, run, summary } from 'mitata'

const PATTERN =
  '{{xs,sm,md,lg}:,}{border-{x,y,t,r,b,l,s,e},bg,text,cursor,accent}-{{red,orange,amber,yellow,lime,green,emerald,teal,cyan,sky,blue,indigo,violet,purple,fuchsia,pink,rose,slate,gray,zinc,neutral,stone}-{50,{100..900..100},950},black,white}{,/{0..100}}'
// const PATTERN = '{text,bg}-{red,green,blue}-{50,{100..900..100},950}'
const braceExpansionExpandOptions = {
  max: 1_500_000,
}

// Verify that the results are the same (at least length wise)
let a = bracesExpand(PATTERN)
let b = braceExpansionExpand(PATTERN, braceExpansionExpandOptions)
let c = expand(PATTERN)

console.log('Results', a.length, b.length, c.length)

summary(() => {
  boxplot(() => {
    bench('npm: `braces`', () => {
      do_not_optimize(bracesExpand(PATTERN))
    })

    bench('npm: `braces-expansion`', () => {
      do_not_optimize(braceExpansionExpand(PATTERN, braceExpansionExpandOptions))
    })

    bench('local: ./brace-expansion', () => {
      do_not_optimize(expand(PATTERN))
    })
  })
})

await run()
```

PR #19833, opened 2026-03-21
Fix: Set neutral, zinc, and stone color palettes to use 'none' for Hue in OKLCH

nmurrell07:
## Summary

This PR fixes the Abney Effect and hue-shifting issue when mixing neutral colors with other colors using `color-mix()` in OKLCH space.

## Problem

When mixing a themed green color (e.g., `green-700`) with a neutral color (e.g., `neutral-700`) using the CSS `color-mix(in oklch, ...)` function, the resulting color unexpectedly shifts toward yellow instead of becoming a desaturated dark green.

This happens because Tailwind's default neutral, zinc, and stone palettes define their OKLCH hue as `0` (which represents the Red/Magenta axis in OKLCH) instead of `none`. When performing a linear interpolation in a color-mix:
- `green-700` (Hue ≈ 150)
- `neutral-700` (Hue = 0)
- Result: The browser calculates the average hue: $(150 + 0) / 2 = 75$, which is exactly the Yellow hue range.

## Solution

Neutral colors in OKLCH should have their Hue component set to `none` (powerless). This ensures that when mixed with other colors, the hue of the non-neutral color is preserved, resulting in a natural desaturation rather than a hue shift to yellow.

## Changes

- Updated `neutral` palette: Changed all colors from `oklch(X% 0 0)` to `oklch(X% none none)`
- Updated `zinc` palette: Changed all colors from `oklch(X% 0 0)` to `oklch(X% none none)`
- Updated `stone` palette: Changed all colors from `oklch(X% 0 0)` to `oklch(X% none none)`

## Test Plan

- [ ] Verify that `color-mix(in oklch, green-700, neutral-700)` now produces a desaturated dark green instead of a yellowish color
- [ ] Verify that existing color-mix operations with neutral colors work as expected
- [ ] Run existing tests to ensure no regressions

Fixes #19830


RobinMalfait:
Thanks for the PR! But unfortunately, this PR doesn't fix the issue you linked to. It's also doing something slightly different than your other PR that's related to this https://github.com/tailwindlabs/tailwindcss/pull/19836

PR #19836, opened 2026-03-21
Fix neutral colors to use powerless hue (none) instead of 0

nmurrell07:
## Summary

This PR fixes an issue where neutral colors were using `0` as the hue value in their OKLCH representation. According to the CSS spec, powerless hues should be represented as `none` instead of `0` to indicate that the color has no chroma.

## Issue

This change ensures that neutral colors work correctly with `color-mix` and other color manipulation functions that expect powerless hues.

## Changes

- Changed all neutral color values from `oklch(... 0 0)` to `oklch(... 0 none)`
- This affects neutral colors 50-950

## Test Plan

- [ ] Run existing tests to ensure no regressions
- [ ] Verify that color-mix works correctly with neutral colors

## References

- Related to issue #19830 (neutral colors using 0 instead of none for powerless hue)


themavik (commented):
Moving the `neutral` ramp to `oklch(... none)` fixes powerless-hue neutrals. nit: `stone` still carries a numeric hue in this file—flag only if you want every gray ramp consistent.

HardikDewra (commented):
The fix is correct in intent - neutral colors with zero chroma should use `none` for the hue component per the CSS Color Level 4 spec. When chroma is 0, hue is mathematically undefined (powerless), and using `0` causes `color-mix()` to interpolate toward red/magenta instead of preserving the other color's hue.

However, I'd flag one thing: this PR only changes the hue from `0` to `none` while keeping chroma as `0`. PR #19833 by the same author takes it further and sets both chroma AND hue to `none` for neutral colors. There's a meaningful difference:

- `oklch(98.5% 0 none)` - chroma is explicitly 0, hue is powerless
- `oklch(98.5% none none)` - both chroma and hue are powerless

Per the CSS spec, when chroma is exactly `0`, the hue is automatically considered powerless during interpolation regardless of whether you write `0` or `none`. But writing `none` for chroma makes the intent explicit and is arguably more spec-aligned.

The two PRs seem to overlap in scope. Might be worth consolidating into one - #19833 appears to be the more thorough fix since it also covers zinc and stone palettes, and marks chroma as `none` too.

RobinMalfait:
Thanks for the PR! But unfortunately, this PR doesn't fix the issue you linked to. It's also doing something slightly different than your other PR that's related to this https://github.com/tailwindlabs/tailwindcss/pull/19833

PR #19837, opened 2026-03-22
fix(canonicalize): collapse arbitrary values into shorthand utilities

aptinio:
The guard on `dynamicUtilities` restricted root-swapping to named values, so arbitrary values like `px-[1.2rem] py-[1.2rem]` were never collapsed into `p-[1.2rem]`.

This is what caused #19835 — in `--stream` mode, the collapse only happened if an earlier line caused the shorthand to be registered in `STATIC_UTILITIES_KEY` as a side effect, making the output non-deterministic. The underlying issue is that arbitrary value collapse wasn't supported at all.

The fix relaxes the guard from `parsedCandidate.value?.kind !== 'named'` to `parsedCandidate.value === null`. `cloneCandidate` and `printCandidate` already handle arbitrary values, so the root-swapping machinery works without other changes.

The iteration in `dynamicUtilities` is over `designSystem.utilities.keys('functional')` — a fixed set of roots, not input-proportional — so the performance cost of including arbitrary values should be negligible.

Fixes #19835.


themavik (commented):
ping

themavik (commented):
`collapseCandidates` now continues only when `parsedCandidate.value === null`, so arbitrary functional utilities participate in collapse—matches the `canonicalizeCandidates` regression for `px-[1.2rem]` / `py-[1.2rem]` merging to `p-[1.2rem]`.

RobinMalfait (approved):
Beautiful, thanks!

aptinio:
Thanks for improving and merging! :pray:

PR #19846, opened 2026-03-24
Make upgrade tooling more stable

### LINKED ISSUE #18972: v4 upgrader removing the content of all files in my project

```
⚡️? 78% ➜ npx @tailwindcss/upgrade --version
≈ tailwindcss v4.1.13
```

Vite: 6.3.6

For example: v21.7.3

n/a

macOS Tahoe 26.0

**Reproduction URL**

This is not a rendering issue, it's a problem with the `@tailwindcss/upgrade` package, so I'm not sure how to provide a "reproduction" but happy to do so if someone can explain.

**Describe your issue**

I realize this is going to sound crazy, but I swear it's happening and I can't figure out why.  When I run `npx @tailwindcss/upgrade" to upgrade my app from v3 to v4, it deletes every file in my project tree.  Not just CSS files, but literally every file.

Specifically, it happens during the "Migrating templates..." step.  Originally, it looked like it was hung, but when I killed the process, I checked my git tree and thousands of files had been deleted.  Thankfully, I had run a commit right before I ran the updater so I could revert without issue, but I suspect this is not expected behaviour.

In this example, I show that there are no changes to be made, I run the upgrader, and then it's showing all of these files having been deleted (it would delete more if I didn't `CTRL-C` out of the upgrader).

```sh
2:36:03 PM in ipms on  filament-v4-upgrade via 🥟 v1.2.22 via  v8.4.12
⚡️? 78% ➜ git status
On branch filament-v4-upgrade
nothing to commit, working tree clean

2:36:05 PM in ipms on  filament-v4-upgrade via 🥟 v1.2.22 via  v8.4.12
⚡️? 78% ➜ npx @tailwindcss/upgrade
≈ tailwindcss v4.1.13

│ Searching for CSS files in the current directory and its subdirectories…

│ Migrating stylesheets…

│ ↳ Migrated stylesheet: `./resources/css/app.css`

│ Updating dependencies…

│ ↳ Updated package: `tailwindcss`

│ ↳ Updated package: `@tailwindcss/postcss`

│ ↳ Updated package: `@tailwindcss/vite`

│ ↳ Updated package: `prettier-plugin-tailwindcss`

│ Migrating templates…

^C

2:36:17 PM in ipms on  filament-v4-upgrade [!] via 🥟 v1.2.22 via  v8.4.12 took 8s
⚡️? 78% ➜ git status
On branch filament-v4-upgrade
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .editorconfig
        modified:   README.md
        modified:   _ide_helper_models.php
        modified:   app/Casts/Markdown.php
        modified:   app/Console/Commands/AddDuesToAllCommand.php
        modified:   app/Console/Commands/AddInstructionalVideosCommand.php
        modified:   app/Console/Commands/ImportContest2024Command.php
        modified:   app/Console/Commands/ImportMembersCommand.php
        modified:   app/Console/Commands/MigrateDbEngineCommand.php
        modified:   app/Enums/ClassifiedAdContactType.php
        modified:   app/Enums/ClassifiedAdType.php
        modified:   app/Enums/Config/DuesFrequency.php
        modified:   app/Enums/ContestEntryAgeClass.php
        modified:   app/Enums/ContestEntryRanking.php
        modified:   app/Enums/ExternalMembershipOrganization.php
        modified:   app/Enums/InstructionalVideoPanel.php

no changes added to commit (use "git add" and/or "git commit -a")

2:36:19 PM in ipms on  filament-v4-upgrade [!] via 🥟 v1.2.22 via  v8.4.12
⚡️? 78% ➜
```

I'm not really sure what else to do.  I tried downloading and compiling the package myself to try and understand why it's even thinking it should touch any non-CSS file but I'm getting this error when I try to `pnpm run build` the main tailwindcss project:

```
error[E0463]: can't find crate for `std`
  |
  = note: the `wasm32-wasip1-threads` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-wasip1-threads`
```

even though I've already confirmed that target is installed:

```
⚡️? 78% ➜ rustup target add wasm32-wasip1-threads
info: component 'rust-std' for target 'wasm32-wasip1-threads' is up to date
```

Any help would be greatly appreciated. Thanks!

RobinMalfait:
This PR is an attempt to make the upgrade tooling more stable.

### TL;DR

1. When migrating from Tailwind CSS v3 → Tailwind CSS v4, only migrate files listed in the `config.content` instead of relying on v4's auto content detection feature
2. Skip writing files that have not been changed
3. Write changed files in a safe way: first write to a temporary file, then rename the file atomically
4. Never migrate files that are git ignored, even if they are listed in the `config.content` file
5. Always ignore `.env` and `.env.*` files when scanning for files. Most people will have this in their `.gitignore` file, but if not, then this is a fallback mechanism.

---

Looking at the #18972 issue, it looks like some people are running into weird situations where some of the contents is just gone. 

I have never been able to reproduce this on my own devices and in my own projects unfortunately. But there is definitely _something_ happening that's not right that people are running into.

Therefore, this PR is an attempt to fix what I think _might_ be wrong, but I'm not 100% sure if these fixes are enough, or if something else is still happening here.

This builds on top of the #19779 PR which has some small fixes, but is incomplete to make this work.

### What's happening

Looking at some of the comments, it looks like a few things are happening such as:

1. The upgrade tool is emptying out my files — it looks like these are only happening if you ctrl+c while the process is taking a while. It could be that a lot of files are being checked and therefore the tooling looks like its stuck.
6. The upgrade tool is looking at files it shouldn't look at — in Tailwind CSS v4 we have this concept of the auto-content detection. This means that we will look at any plain text file that is not git ignored.

#### Emptying out files

The files being emptied looks like it's because how `fs.writeFile` behaves by default. It opens the file handle with the `w` flag, which will first truncate the file before writing the new contents. This is not a single atomic operation, so a killed process in the middle will cause invalid state.

When we migrate your template files, everything is happening in promises to migrate things at the same time. When a lot of files are being scanned, truncating might have happened already before we write the new content. Since we migrate a bunch of files in parallel, a ctrl+c could cause data loss in multiple files.

To mitigate this, I switched to an alternative way of writing files. 

1. First, we do some quick checks where if the contents didn't change we just bail out immediately. Files that don't include Tailwind CSS classes won't change, and therefore we don't need to override these files with the same contents.
2. When the migrated contents is empty, we bail out as well. I'm 100% sure that this is not the spot where the "emptying out" happens, I still believe it happens in the `writeFile` itself, but added it just in case.
7. Next, I introduced a safe write, where we first write to a temporary file in the same folder. We could write it to `/tmp`, but then we can't guarantee that we are on the same file system.
   
   If we ctrl+c at this stage, then the worst case scenario is that you have additional temporary files in your project, but your original files are still there.
   
   Once that file was written, we will use the atomic `fs.rename`. This should be atomic as long as we are on the same file system, so either the rename didn't happen yet, or it completed.

I added an integration test for this, but I had to change the `writeFile` implementation slightly. In the test, we will truncate the file first, after that we will write the new contents. This is so that we have enough time to kill the current process and allows us to verify that we didn't clear out the file. Again, this is a hacky way of testing this, just because I can't reproduce this issue myself, let alone reproduce it reliable in a CI environment.

Note: we are also using `realpath` to make sure that we are updating the real file. Otherwise, if we were dealing with a symlinked file, we would override the symlink with a "hard" copy instead.

#### Touching files that should not be touched

During the migration, we rely on the Tailwind CSS v4 auto detection logic which means that it will scan any plain text file that is not git ignored. Therefore changes to php files could happen because in theory they could contain Tailwind CSS classes.

To solve this, when migrating from Tailwind CSS v3 to Tailwind CSS v4, we will _only_ take the sources into account that were listed in the `config.content` array. Since this was a requirement in Tailwind CSS v3, it should be safe to rely on this array.

Additionally, this will make sure that we are dealing with way fewer files to migrate as well.

On top of that, files that match the patterns in the content array that are git ignored will also be skipped. This is to prevent that we mutate files in `node_modules` for example.

In one of the comments I read that `.env` files were emptied out. In most cases people will have these files gitignored but I explicitly added `.env` and `.env.*` as files to never ever touch by default when scanning.

Last but not least, this also updates the output a little bit of the upgrade tool in case we skip content files (because of git ignore) and if we changed a file.

image

- "Git ignored folder, skipping: `./node_modules`": this is because the content array looks like this while the `node_modules` are being ignored: 
image

- "Migrated `./resources/views/vendor/filament-panels/components/logo.blade.php`": this is because **I** made a change to showcase this feature.

Fixes: #18972
Closes: #19779

### Test plan

1. Existing tests still pass
1. Added a dedicated integration test to ensure that we only take `config.content` into account when migrating from Tailwind CSS v3 to Tailwind CSS v4 projects.
1. Added a dedicated integration test to make sure that files listed in `config.content` that are also git ignored, will still be skipped.
1. Added a dedicated integration test to ensure that when `writeFile` is cancelled mid-write that our old files are still present.
1. Added a dedicated integration test to ensure that we ignore `.env` and `.env.*` files even if you didn't git ignore them.

[ci-all] To verify on Windows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Canonicalization: collapse `overscroll-{x,y}-*` into `overscroll-*` ([#19842](https://github.com/tailwindlabs/tailwindcss/pull/19842))
 - Read from `--placeholder-color` instead of `--background-color` for `placeholder-*` utilities ([#19843](https://github.com/tailwindlabs/tailwindcss/pull/19843))
 - Upgrade: Ensure files are not emptied out when killing the upgrade process while it's running ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
+- Upgrade: Use `config.content` when migrating from Tailwind CSS v3 to Tailwind CSS v4 ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 
 ## [4.2.2] - 2026-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Read from `--placeholder-color` instead of `--background-color` for `placeholder-*` utilities ([#19843](https://github.com/tailwindlabs/tailwindcss/pull/19843))
 - Upgrade: Ensure files are not emptied out when killing the upgrade process while it's running ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 - Upgrade: Use `config.content` when migrating from Tailwind CSS v3 to Tailwind CSS v4 ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
+- Upgrade: Never migrate files that are ignored by git ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 
 ## [4.2.2] - 2026-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Canonicalization: collapse `overflow-{x,y}-*` into `overflow-*` ([#19842](https://github.com/tailwindlabs/tailwindcss/pull/19842))
 - Canonicalization: collapse `overscroll-{x,y}-*` into `overscroll-*` ([#19842](https://github.com/tailwindlabs/tailwindcss/pull/19842))
 - Read from `--placeholder-color` instead of `--background-color` for `placeholder-*` utilities ([#19843](https://github.com/tailwindlabs/tailwindcss/pull/19843))
+- Upgrade: Ensure files are not emptied out when killing the upgrade process while it's running ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 
 ## [4.2.2] - 2026-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade: Ensure files are not emptied out when killing the upgrade process while it's running ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 - Upgrade: Use `config.content` when migrating from Tailwind CSS v3 to Tailwind CSS v4 ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 - Upgrade: Never migrate files that are ignored by git ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
+- Add `.env` and `.env.*` to default ignored content files ([#19846](https://github.com/tailwindlabs/tailwindcss/pull/19846))
 
 ## [4.2.2] - 2026-03-18
 
--- a/crates/oxide/src/scanner/fixtures/ignored-files.txt
+++ b/crates/oxide/src/scanner/fixtures/ignored-files.txt
@@ -2,3 +2,5 @@ package-lock.json
 pnpm-lock.yaml
 bun.lockb
 .gitignore
+.env
+.env.*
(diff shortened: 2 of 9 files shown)


RobinMalfait:
Update: I _was_ finally able to reproduce it when running with `node` instead of `bun` and while re-trying a few dozen times just to get the timing right. 

While watching whenever the git status changed to files I knew should not be touched, I tried to quickly ctrl+c and eventually got it. 

I also added logging to ensure that the file in question didn't actually change (so no migration) but it was still written to disk, which was cancelled and ended up emptying the file.

image


Good news is, I can't reproduce this now anymore, at all, with the changes implemented in this PR.

PR #19858, opened 2026-03-26
Canonicalize negative arbitrary values

RobinMalfait:
This PR adds a few more canonicalizations for some cases I noticed on our templates.

When dealing with arbitrary values, and the utility is a "negative" utility, then we will try to put the `-` inside of the arbitrary value:

```diff
- -left-[9rem]
+ left-[-9rem]
```

The idea is that the arbitrary value is already an escape hatch for when a value is not available by default. The `-` in front uses an implicit `calc(<expression> * -1)` which might be confusion if you have an value like this already.

This also can allow for some further optimizations. For example
```diff
- -mt-[492px]
  ↓↓↓↓↓↓↓                           Into a simpler arbitrary value
+ mt-[-492px]
  ↓↓↓↓↓↓↓                           Into a bare value
+ -mt-123
```

This PR also improve the constant folding of calc expressions a bit more such that nested calc expressions with 2 constants and an unknown can be folded. Bit of a mouthful, but it allows us to handle this:
```diff
- mt-[calc(-1*calc(-1*var(--foo)))]
  ↓↓↓↓↓↓↓                           The -1 * -1 becomes a no-op
+ mt-[var(--foo)]
  ↓↓↓↓↓↓↓                           Into the shorthand for CSS variables
+ mt-(--foo)
```

Now that we can handle moving the `-` into the arbitrary value, there are also cases where we can get the `-` _out_ of the arbitrary value:
```diff
- mt-[calc(-1*var(--foo))]
  ↓↓↓↓↓↓↓                           Simplify calc, move `-` to the front
+ -mt-[var(--foo)]
  ↓↓↓↓↓↓↓                           Into the shorthand for CSS variables
+ -mt-(--foo)
```

Another missing piece that this PR adds is the concept of canonicalizing or normalizing calc expressions. This is a separate step used when calculating the signature for each utility. This allows us to normalize `calc(-1*var(--foo))` and `calc(var(--foo)*-1)`. Without this they would not be considered the same, but now it will.

It's only used when comparing values, it won't unify the actual arbitrary values with this logic (at least for now).

With the additional constant folding logic and the canonicalization when comparing signatures it unlocks the necessary power to perform the above transformations.

## Test plan

1. Existing tests still pass
2. Added additional tests for the constant folding logic
3. Added tests for the canonicalization of calc expressions
4. Added new tests where we move the `-` inside the value, or move the `-` outside of the arbitrary value.

### DIFF
--- a/packages/tailwindcss/src/canonicalize-calc-expressions.test.ts
+++ b/packages/tailwindcss/src/canonicalize-calc-expressions.test.ts
@@ -1,0 +1,24 @@
+import { expect, it } from 'vitest'
+import { canonicalizeCalcExpressions } from './canonicalize-calc-expressions'
+
+it.each([
+  ['calc(-1 * var(--foo))', 'calc(var(--foo) * -1)'],
+  ['calc(1rem + var(--foo))', 'calc(var(--foo) + 1rem)'],
+  ['calc(2rem * calc(3px * var(--foo)))', 'calc(calc(var(--foo) * 3px) * 2rem)'],
+  ['calc(var(--b) + var(--a))', 'calc(var(--a) + var(--b))'],
+  ['calc(3px * 2rem)', 'calc(2rem * 3px)'],
+  ['calc(5px * 3px)', 'calc(3px * 5px)'],
+  ['calc(1 * 1rem)', 'calc(1rem * 1)'],
+  ['calc(10 + 2)', 'calc(2 + 10)'],
+])('`%s` → `%s` (%#)', (input, expected) => {
+  expect(canonicalizeCalcExpressions(input)).toBe(expected)
+})
+
+it.each([
+  ['calc(1rem - var(--foo))'],
+  ['calc(1rem / 2)'],
+  ['calc(var(--a) + 1rem)'],
+  ['calc(2rem * 3px)'],
+])('should keep `%s` (%#)', (input) => {
+  expect(canonicalizeCalcExpressions(input)).toBe(input)
+})
--- a/packages/tailwindcss/src/canonicalize-calc-expressions.ts
+++ b/packages/tailwindcss/src/canonicalize-calc-expressions.ts
@@ -77,6 +77,6 @@ function shouldSwap(lhs: ValueParser.ValueAstNode, rhs: ValueParser.ValueAstNode
     }
   }
 
-  // Both nodes are not non-dimensions, sort them as strings
+  // Both nodes are unknown values (not dimensions), sort them as strings
   return ValueParser.toCss([lhs]).localeCompare(ValueParser.toCss([rhs])) > 0
 }
--- a/packages/tailwindcss/src/canonicalize-calc-expressions.ts
+++ b/packages/tailwindcss/src/canonicalize-calc-expressions.ts
@@ -1,0 +1,82 @@
+import { dimensions } from './utils/dimensions'
+import * as ValueParser from './value-parser'
+import { walk, WalkAction } from './walk'
+
+// Assumption: We already assume that we receive somewhat valid `calc()`
+// expressions. So we will see `calc(1 + 1)` and not `calc(1+1)`
+export function canonicalizeCalcExpressions(input: string): string {
+  let [canonicalized, valueAst] = canonicalizeCalcExpressionsAst(ValueParser.parse(input))
+
+  return canonicalized ? ValueParser.toCss(valueAst) : input
+}
+
+export function canonicalizeCalcExpressionsAst(
+  ast: ValueParser.ValueAstNode[],
+): [canonicalized: boolean, ast: ValueParser.ValueAstNode[]] {
+  let canonicalized = false
+
+  walk(ast, {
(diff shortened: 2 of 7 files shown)


packages/tailwindcss/src/canonicalize-calc-expressions.ts
```diff
@@ -0,0 +1,82 @@
...
+    let [rhsValue, rhsUnit] = rhsDimension
+
+    // Within dimensions, keep unit-bearing values ahead of unitless numbers so
+    // `1rem` sorts before `1`.
+    if (lhsUnit === null && rhsUnit !== null) return true
+    if (lhsUnit !== null && rhsUnit === null) return false
+
+    // Then sort dimensions numerically, and finally by unit for ties.
+    if (lhsValue !== rhsValue) {
+      return lhsValue - rhsValue > 0
```
RobinMalfait, line 72: A `localeCompare` for `2` and `10` would result in `10 2` otherwise, so we compare the actual values here.

packages/tailwindcss/src/canonicalize-calc-expressions.test.ts
```diff
@@ -0,0 +1,24 @@
+import { expect, it } from 'vitest'
+import { canonicalizeCalcExpressions } from './canonicalize-calc-expressions'
+
+it.each([
+  ['calc(-1 * var(--foo))', 'calc(var(--foo) * -1)'],
```
RobinMalfait, line 5: I don't think the order matters to much but I did:

1. Move constants to the end
2. When comparing values without unit, sort by the value
3. When it's not a dimension (function, var, ...) then sort alphabetically (`localeCompare`)

packages/tailwindcss/src/canonicalize-candidates.test.ts
```diff
@@ -446,6 +446,28 @@ describe.each([['default'], ['with-variant'], ['important'], ['prefix']])('%s',
...
+      ['-mt-[-492px]', 'mt-123'],
+      ['-mt-[calc(-1*-492px)]', '-mt-123'], // Constant folding with calc expressions
+      ['-mt-(--my-var)', '-mt-(--my-var)'], // Keep as-is
+      ['-mt-[var(--my-var)]', '-mt-(--my-var)'], // Keep as-is, but convert to shorthand
+      ['mt-[calc(var(--my-var)*-1)]', '-mt-(--my-var)'], // Move `-` out
+      ['mt-[calc(-1*var(--my-var))]', '-mt-(--my-var)'], // Move `-` out
+      ['-mt-[calc(var(--my-var)*-1)]', 'mt-(--my-var)'], // Move `-` out
+      ['-mt-[calc(-1*var(--my-var))]', 'mt-(--my-var)'], // Move `-` out
+      ['mt-[calc(-1*calc(-1*var(--my-var)))]', 'mt-(--my-var)'], // Move `-` out
+      ['-mt-[calc(-1*calc(-1*var(--my-var)))]', '-mt-(--my-var)'], // Move `-` out
```
RobinMalfait, line 470: These are obviously a bit silly. But needed to make sure that we always have correct utilities. While it's silly, it's also very satisfying to see the simplification here.

packages/tailwindcss/src/canonicalize-candidates.ts
```diff
@@ -139,6 +145,25 @@ function createSignatureOptionsCache(): DesignSystem['storage'][typeof SIGNATURE
   })
 }
 
+const COMPARE_CANDIDATES_KEY = Symbol()
+function createSignatureComparison(designSystem: DesignSystem) {
+  return new DefaultMap((options: SignatureOptions) => {
+    let signatures = designSystem.storage[UTILITY_SIGNATURE_KEY].get(options)
+
+    return function hasSameSignature(a: Candidate | string, b: Candidate | string): boolean {
```
RobinMalfait, line 153: We use the same thing in a few places, so wanted to abstract it.

I think a future (but very low priority) is to compute the signature of the incoming candidate ahead of time. But since we're dealing with cached values it should be fine as-is.

packages/tailwindcss/src/constant-fold-declaration.ts
```diff
@@ -70,7 +210,9 @@ export function constantFoldDeclaration(input: string, rem: number | null = null
               (lhs[1] !== null && rhs[1] === null) // Unit * Unitless, e.g.: `1rem * 2`
             ) {
               folded = true
-              return WalkAction.ReplaceSkip(ValueParser.word(`${lhs[0] * rhs[0]}${lhs[1] ?? ''}`))
+              return WalkAction.ReplaceSkip(
+                ValueParser.word(`${lhs[0] * rhs[0]}${lhs[1] ?? rhs[1] ?? ''}`),
+              )
```
RobinMalfait, line 215: Added a test for this as well, but `calc(2rem * 4)` resulted in `8rem`, but `calc(4 * 2rem)` resulted in `8` because we always took the unit from the `lhs`.

PR #19859, opened 2026-03-26
fix error serving with older vite versions when config.css was not defined

mruthh:
## Summary
When attempting to serve a Vue app using Vite 4.3.5 with the tailwind/vite plugin, I received the following error:

```
[vite] Internal server error: Cannot read properties of undefined (reading 'devSourcemap')
  Plugin: @tailwindcss/vite:generate:serve
```

The issue was a missing `?.` check. My `vite.config` file did not contain a `css` property and that crashed the plugin.

## Test plan
- In `pnpm-workspace.yaml`, change the vite version to `^4.3.5`.
- Run `pnpm install`
- In `playgrounds/vite`, run `pnpm build && pnpm vite`
- The example app should serve normally. Note that without the change from the PR, it fails to serve.

### DIFF
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -122,7 +122,7 @@ export default function tailwindcss(opts: PluginOptions = {}): Plugin[] {
       config!.root,
       // Currently, Vite only supports CSS source maps in development and they
       // are off by default. Check to see if we need them or not.
-      config?.css.devSourcemap ?? false,
+      config?.css?.devSourcemap ?? false,
       customCssResolver,
       customJsResolver,
     )


RobinMalfait:
Hey! I appreciate the PR, but I would highly recommend to upgrade to a newer Vite version.

While we can merge this as is, if you are on the older Vite version the build won't even work due to type checking reasons.

The `@tailwindcss/vite` package supports Vite 5.2 and up, and note that even Vite 5 is officially unsupported by the Vite team now: https://vite.dev/releases#supported-versions 

Additionally, if you ignore all the types, run the Vite playground, it will work ~ish. If you make a change, no new CSS will be generated, there is no HMR and a hard reload won't work either. In this case you will have to restart the server for the changes to take effect.

So while this is a small fix, it's not complete and doesn't add support for Vite 4. Since Vite 4 and 5 are not supported anymore, I don't think we should spend time supporting it either.

I appreciate the PR, but going to close it because of the above mentioned reasons. Thanks!

PR #19862, opened 2026-03-26
Use `json` preprocessor for new-line delimited JSON files

thecrypticace:
## Summary

This specializes the `.jsonl` and `.ndjson` file extensions so they're preprocessed like JSON instead of by the standard scanner. This prevents them from creating thousands of sub machines and reduces scanning time (see #17125 where this was done for `.json` files).

It seems reasonable to handle new-line delimited JSON files as well otherwise scanning these files can take quite a long time.

It's quite unlikely that these will contain classes so, alternatively, these *could* go in the binary extensions list so they get ignored entirely.

## Test plan

I ran manual tests inside the `oxide` crate against some large-ish JSONL files (5MB–15MB). These changes bring down scanning time from 2s–3s on my M3 Max (via `cargo test --release …`) to less than 20ms. 

I also ran tests through a full CLI build pipeline on a low-spec linux box. This change brought scanning time down from ~90s to ~300ms for a single ~15MB file.


RobinMalfait (approved):
Love it, thanks!

PR #19869, opened 2026-03-29
Remove :-moz-focusring rule from preflight

jmlee9707:
## Summary

  Firefox has natively supported :focus-visible since version 85, making the :-moz-focusring { outline: auto; }
   rule in preflight redundant. Additionally, this rule was overriding Firefox's own iframe:focus-visible {
  outline-style: none; }, causing unwanted outlines on focused iframes.

  This PR removes the rule entirely rather than excluding specific elements, since the underlying vendor prefix
   is no longer needed.

  Fixes #19795

## Test plan

 - Verified in Firefox that focusable elements (buttons, inputs, links) still display proper focus-visible
  outlines without the :-moz-focusring rule
  - Verified that iframes no longer show unwanted outlines on focus in Firefox
  - Snapshot and source map tests updated to reflect the removal


jmlee9707:
Closing this PR as I wasn't aware that #19866 was already open addressing the same issue. Thanks!

PR #19876, opened 2026-03-30
Append harmless comment to README

### LINKED ISSUE #16824: [v4] Utility modifiers only work with arguments
With Tailwind 4 (4.0.9 currently); modifiers only seem to work with utilities that do have an argument, this works:

```css
@utility text-* {
  --alpha: calc(--modifier(integer) * 1%);
  color: --alpha(--value(--*, [color]) / var(--alpha, 100%));
}
```

With `text-default`, `text-default/50`, `text-blue`, `text-blue/50`, etc.

This doesn't work:

```css
@utility text {
  --alpha: calc(--modifier(integer) * 1%);
  color: --alpha(red / var(--alpha, 100%));
}
```

With just `text`, `text/50`, etc. as die modifier isn't parsed.

Demo: https://play.tailwindcss.com/ECphsE1vzA?file=css

---

Related discussion: https://github.com/tailwindlabs/tailwindcss/discussions/15775

JBR-0100:
Closes #16824

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,4 @@ For help, discussion about best practices, or feature ideas:
 ## Contributing
 
 If you're interested in contributing to Tailwind CSS, please read our [contributing docs](https://github.com/tailwindlabs/tailwindcss/blob/main/.github/CONTRIBUTING.md) **before submitting a pull request**.
+<!-- Auto-generated contribution -->


RobinMalfait:
Please stop.

PR #19879, opened 2026-03-30
fix(oxide): simplify @source pattern handling and ensure gitignore negation

JBR-0100:
This PR simplifies the pattern handling in the Oxide scanner. Previously, complex logic for guessing patterns could fail to correctly whitelist files in git-ignored directories. By consistently prefixing all @source patterns with '!' in the internal Gitignore builder, we ensure they are always discovered even when buried in ignored paths.


RobinMalfait:
Hey! I appreciate the PR, but would love to see an actual issue it solves (by opening an issue and/or adding tests first). Also please don't add unnecessary things to PRs such as the auto-generated contribution comment.

Going to close this for now, but if you still want to PR something like this, let's open an issue first so we have a record of something that's actually wrong today.

Thanks!

PR #19891, opened 2026-04-02
Bump pnpm, and ensure packages are 7 days old

RobinMalfait:
This PR bumps pnpm to 10.33, we required at least version 10.16 for the `minimumReleaseAge` setting to take effect.

We also now add the `minimumReleaseAge` to be 7 days, newer packages are not allowed. 

[ci-all]


lajczi (commented):
lgtm

PR #19907, opened 2026-04-05
fix(node): use Module.registerHooks instead of deprecated Module.register for Node v26+

cyphercodes:
## Description

Node.js v26 deprecated `Module.register()` in favor of `Module.registerHooks()`, causing a deprecation warning when using Tailwind CSS with Node v26:

```
[DEP0205] DeprecationWarning: `module.register()` is deprecated. Use `module.registerHooks()` instead.
```

This change uses the new `Module.registerHooks()` API when available (Node v26+), falling back to the old `Module.register()` API for older Node versions to maintain backward compatibility.

## Changes

- Check for `Module.registerHooks` first (new API in Node v26+)
- Fall back to `Module.register` for older Node versions
- Added comment explaining the version requirements

## Related Issue

Fixes #19893

### DIFF
--- a/packages/@tailwindcss-node/src/index.ts
+++ b/packages/@tailwindcss-node/src/index.ts
@@ -14,8 +14,14 @@ if (!process.versions.bun) {
   let localRequire = Module.createRequire(import.meta.url)
 
   // `Module#register` was added in Node v18.19.0 and v20.6.0
+  // `Module#registerHooks` was added in Node v26.0.0 and is the preferred API
   //
   // Not calling it means that while ESM dependencies don't get reloaded, the
   // actual included files will because they cache bust directly via `?id=…`
-  Module.register?.(pathToFileURL(localRequire.resolve('@tailwindcss/node/esm-cache-loader')))
+  let loaderUrl = pathToFileURL(localRequire.resolve('@tailwindcss/node/esm-cache-loader'))
+  if (Module.registerHooks) {
+    Module.registerHooks({ resolve: loaderUrl })
+  } else {
+    Module.register?.(loaderUrl)
+  }
 }


afurm:
Good migration to registerHooks. One thing to consider - if both registerHooks fails and register falls back, should we warn? Currently silent failure could be hard to debug. Also, any reason not to use feature detection for Node 26+ and skip register entirely?

packages/@tailwindcss-node/src/index.ts
```diff
@@ -14,8 +14,14 @@ if (!process.versions.bun) {
   let localRequire = Module.createRequire(import.meta.url)
 
   // `Module#register` was added in Node v18.19.0 and v20.6.0
+  // `Module#registerHooks` was added in Node v26.0.0 and is the preferred API
```
WarningImHack3r, line 17:
```suggestion
  // `Module#registerHooks` was added in Node v23.5.0 and v22.15.0 and is the preferred API since v25.9.0,
  // runtime-deprecating `Module#register` since v26
```
Source: https://nodejs.org/docs/latest/api/module.html#moduleregisterhooksoptions

packages/@tailwindcss-node/src/index.ts
```diff
@@ -14,8 +14,14 @@ if (!process.versions.bun) {
...
   //
   // Not calling it means that while ESM dependencies don't get reloaded, the
   // actual included files will because they cache bust directly via `?id=…`
-  Module.register?.(pathToFileURL(localRequire.resolve('@tailwindcss/node/esm-cache-loader')))
+  let loaderUrl = pathToFileURL(localRequire.resolve('@tailwindcss/node/esm-cache-loader'))
+  if (Module.registerHooks) {
+    Module.registerHooks({ resolve: loaderUrl })
```
WarningImHack3r, line 23: This is incorrect usage. As per [the docs](https://nodejs.org/docs/latest/api/module.html#customization-hooks), the exported `resolve` function from `esm-cache.loader.mts` can be used as-is here, no need for a loader URL anymore.

WarningImHack3r (commented):
This AI-made PR doesn't compile. I made a PR that should work much better and be much more comprehensive at #20028

PR #19918, opened 2026-04-08
[@tailwindcss/upgrade] Don’t migrate inline style properties

dbpolito:
Prevent the upgrade tool from rewriting CSS properties inside inline `style` attributes.

This fixes cases like `style="flex-grow: 1"` being changed to `style="grow: 1"` and adds regression tests.


afurm:
The `INLINE_STYLE_ATTRIBUTE` regex only matches against `currentLineBeforeCandidate`, but what about multi-line style attributes where a CSS property wraps to the next line? E.g. `style="flex-grow:\n1"` — the candidate `grow` would appear on a line that doesn't have the `style=` prefix.

RobinMalfait:
@afurm is that an actual issue you have in your projects?

RobinMalfait (approved):
Thanks! I also added some additional tests to make sure that multi-line style attributes don't get migrated either.

PR #19928, opened 2026-04-13
Fix negation of theme() in arbitrary values

snaptopixel:
## Problem

Using a negative modifier with `theme()` inside an arbitrary value produces no CSS output.

- ✅ `ms-[theme(margin.1)]` — works
- ✅ `-ms-[10px]` — works
- ❌ `-ms-[theme(margin.1)]` — no output

## Cause

`negateValue()` in `src/util/negateValue.js` doesn't recognize `theme()` as a negatable function. It falls through all conditions and returns `undefined`, causing the utility to be silently dropped.

## Fix

Added `'theme'` to the `numericFunctions` list so `theme()` values get wrapped in `calc(... * -1)`, which the PostCSS `theme()` resolver then processes downstream.

## Tests

- **Unit test** (`negateValue.test.js`): `negateValue('theme(margin.1)')` returns `calc(theme(margin.1) * -1)`
- **Integration test** (`arbitrary-values.test.js`): `-mt-[theme(spacing.1)]` produces `margin-top: -0.25rem`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure classes are extracted when a variant precedes an arbitrary variant containing a quoted attribute selector (e.g. `focus-visible:[&:not([aria-selected="true"])]:bg-red-500`) ([#20374](https://github.com/tailwindlabs/tailwindcss/pull/20374))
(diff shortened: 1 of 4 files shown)


RobinMalfait:
Edit: This is for Tailwind CSS v3.

~~Hey! Thanks for the PR, but where do you see this issue? This should work as-is today already:~~
~~image~~
~~https://play.tailwindcss.com/bGXXIU9Wai~~

RobinMalfait (approved):
Thanks!

Will be available in the next v3 release.

snaptopixel:
Thanks for merging @RobinMalfait!

PR #19942, opened 2026-04-20
Add Vite support for configurable Lightning CSS options and polyfills

lubomirblazekcz:
## Summary

This updates the `@tailwindcss/vite` plugin so the `optimize` option can pass through additional Lightning CSS settings instead of only supporting `minify`. In addition to `minify`, Vite users can now configure `drafts`, `nonStandard`, `include`, `exclude`, and `targets`.

It also adds a `polyfills` option to the Vite plugin and threads that through to the Tailwind compiler, matching the behavior that already exists in the PostCSS and Webpack integrations. CSS Modules continue to disable the `@property` fallback polyfill so they do not emit global non-pure selectors.

While making this change, the Vite integration now only overrides `minify` when it is explicitly provided in `optimize`, which preserves Vite's existing `build.cssMinify` behavior when users only want to customize other Lightning CSS options.

## Test plan

- `pnpm vitest --root=./integrations run vite/index.test.ts -t "optimize option: disabled|optimize option: enabled with minify disabled|optimize option: advanced Lightning CSS settings|optimize option: advanced Lightning CSS exclude|polyfills option: disabled"`

Screenshots not applicable; this change is covered by the Vite integration tests above.

### DIFF
--- a/integrations/vite/index.test.ts
+++ b/integrations/vite/index.test.ts
@@ -1217,6 +1217,199 @@ test(
 )
 
 test(
+  'optimize option: advanced Lightning CSS settings',
+  {
+    fs: {
+      'package.json': txt`
+        {
+          "type": "module",
+          "dependencies": {
+            "@tailwindcss/vite": "workspace:^",
+            "tailwindcss": "workspace:^"
+          },
+          "devDependencies": {
+            "lightningcss": "^1",
+            "vite": "^7"
+          }
+        }
+      `,
+      'vite.config.ts': ts`
+        import { Features } from 'lightningcss'
+        import tailwindcss from '@tailwindcss/vite'
+        import { defineConfig } from 'vite'
+
+        export default defineConfig({
+          build: { cssMinify: false },
+          plugins: [
+            tailwindcss({
+              optimize: {
+                include: Features.Nesting,
+                targets: { chrome: 999 << 16 },
+                drafts: { customMedia: false },
+                nonStandard: { deepSelectorCombinator: true },
+              },
+            }),
+          ],
+        })
+      `,
+      'index.html': html`
+        <head>
+          <link rel="stylesheet" href="./src/index.css" />
+        </head>
+        <body>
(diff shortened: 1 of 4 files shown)


lubomirblazekcz:
I'm considering removing the additional Lightning CSS options as setting from this PR. Setting `optimize: false` and leaving the rest to vite lightningcss perhaps makes more sense. I've added that primarily because of when someone has different targets in vite configuration.

But I think ability to configuring polyfills would be neat. Thoughts?

RobinMalfait:
Hey! I appreciate the PR but going to pass on this for now. Right now Lightning CSS is almost an implementation detail, and exposing some of its internal features makes it harder for us to switch tools later. 

If you already use Vite, which you very likely are since you use `@tailwindcss/vite`, then they use Lightning CSS as well and you should be able to configure Lightning CSS that way (which you also hinted at in your comment).

The Polyfills feature is also used for internals right now, and allows us to hide polyfills for intellisense purposes (otherwise with the polyfills there is more noise). Since Lightning CSS runs later anyways, you will get these polyfills based on your configuration anyway so I don't think there is a good use case for exposing this information right now.

Are there any issues you were running into that required you to expose these polyfills? Feel free to open an issue if that's the case so we can see if this is the best solution for the problem you're experiencing.

Would love to keep the conversation going, but going to close this in the meantime since this PR is mainly focused around the Lightning CSS options which I don't think we should do right now. 

Thanks!

lubomirblazekcz:
@RobinMalfait 

Yeah, I agree on the Lightning CSS options. I came to the same conclusion in my previous comment. As for the polyfills, that was the main thing I wanted to change. My reasoning is that even if I decide to turn off optimize and provide my own options via Vite, the polyfills are still included at compile time. I might want to target newer browsers where they aren’t needed, especially `color-mix()`, which adds a lot of additional code.

Should I create an issue for that? Is that something that makes sense to have configurable via Vite? I’m open to creating a PR just for that too :)

RobinMalfait:
If you can make an issue for that, that would be amazing, thanks!

PR #19949, opened 2026-04-21
fix(vite): avoid resolving JS plugins to browser CSS entries

### LINKED ISSUE #19950: Unknown file extension ".css" for ...
Describe: 
Looks like same issue with #19946 , but occur in v4.2.4, more info in [daisyUI](https://github.com/saadeghi/daisyui/issues/4505#issuecomment-4290067403)

Reproduction:
https://stackblitz.com/~/github.com/13RTK/reproduct

Source code:
https://github.com/13RTK/reproduct

For example: v4.2.4

Vite 8.0.9

v24.15.0

All browsers

For example: macOS, Windows

**Reproduction URL**

A Tailwind Play link or public GitHub repo that includes a minimal reproduction of the bug. **Please do not link to your actual project**, what we need instead is a _minimal_ reproduction in a fresh project without any unnecessary code. This means it doesn't matter if your real project is private/confidential, since we want a link to a separate, isolated reproduction anyways.

A reproduction is **required** when filing an issue — any issue opened without a reproduction will be closed and you'll be asked to create a new issue that includes a reproduction. We're a small team and we can't keep up with the volume of issues we receive if we need to reproduce each issue from scratch ourselves.

**Describe your issue**

Describe the problem you're seeing, any important steps to reproduce and what behavior you expect instead.

ArcherGu:
Edit: some edits by @RobinMalfait 

---

## Summary

Fix a regression in `@tailwindcss/vite` introduced by `#19803` where JS plugin resolution could incorrectly resolve a package to its `browser` CSS entry.

In cases like `daisyui`, Vite can resolve `@plugin "daisyui"` to `daisyui.css` instead of the package's JS entry, which causes Tailwind to try to load a CSS file as a JS plugin and fail with:

```txt
Unknown file extension ".css"
```

This change keeps the `aliasOnly: false` behavior from `#19803` so tsconfig path resolution still works, but adds a JS-entry guard to `customJsResolver` in `@tailwindcss/vite`. If Vite resolves a plugin request to a non-JS file like `.css`, the custom resolver now returns `undefined` so Tailwind's internal fallback resolver can resolve the package as a JS plugin entry instead.

I also added integration coverage for a package whose `main`/`module` points to JS while `browser` points to CSS, and verified that `@plugin "pkg"` still resolves to the JS entry in both build and dev mode.

## Test plan

Added new integration tests in `integrations/vite/resolvers.test.ts` covering a package with:

- `main` / `module` -> JS
- `browser` -> CSS
- `@plugin "pkg"` -> should resolve to JS, not CSS

Verified with:

```sh
pnpm test:integrations vite/resolvers.test.ts -t "browser points to CSS"
pnpm test:integrations vite/resolvers.test.ts -t "resolves tsconfig paths"
```

These verify that:

- `@plugin` no longer resolves to a CSS browser entry
- the original tsconfig paths fix from `#19803` still works in both build and dev mode

---

Maintainer edits:

Instead of hardcoding file extensions, first try to resolve aliases and then fallback to the default resolving system we had before. We still check for a `.css` extension, even in the JS resolver because some dependencies (like `daisyUI`) put the CSS file there instead of in an `exports.style`. If we detect that, we still fallback to the default resolving logic.

This should be compatible with the original issue we were trying to fix where we wanted to make Vite aliases work.

Fixes: #19950

[ci-all]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - _Experimental_: add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
 
+### Fixed
+
+- Ensure `@plugin` resolves package JavaScript entries instead of browser CSS entries when using `@tailwindcss/vite` ([#19949](https://github.com/tailwindlabs/tailwindcss/pull/19949))
+
 ## [4.2.4] - 2026-04-21
 
 ### Fixed
--- a/integrations/vite/resolvers.test.ts
+++ b/integrations/vite/resolvers.test.ts
@@ -452,6 +452,67 @@ test(
   },
 )
 
+test(
+  'resolves package plugins in dev mode when package exports CSS files',
+  {
+    fs: {
+      'package.json': json`
+        {
+          "type": "module",
+          "dependencies": {
+            "@tailwindcss/vite": "workspace:^",
+            "daisyui": "^5",
+            "tailwindcss": "workspace:^"
+          },
+          "devDependencies": {
+            "vite": "^8"
+          }
+        }
+      `,
+      'vite.config.ts': ts`
+        import tailwindcss from '@tailwindcss/vite'
+        import { defineConfig } from 'vite'
+
+        export default defineConfig({
+          build: { cssMinify: false },
+          plugins: [tailwindcss()],
+        })
+      `,
+      'index.html': html`
+        <html>
+          <head>
+            <link rel="stylesheet" href="./src/index.css" />
+          </head>
+          <body>
+            <button class="btn btn-primary">Hello, world!</button>
+          </body>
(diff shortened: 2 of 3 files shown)


RobinMalfait:
Additional tests from the reproduction provided in the linked issue is working now:
image
image

Additionally, verified with the original issue we tried to solve in 4.2.4 to make sure that still works as well:
image

The integration tests are also green where we test regressions for this, but needed to make sure these reproductions also still passed.

RobinMalfait (approved):
Thanks! Made some changes by removing the explicit extension list, and by trying to resolve the aliases first. Then fallback to the default resolving logic we used to have. If we resolve to a `.css` file in the JS resolvers we bail as well.

JakubMindNexus:
Would be great to have it released - it's a small blocker for my library

kenn:
same here, been waiting for this fix to be released. I hope 4.2.5 lands soon.

PR #19961, opened 2026-04-23
feat: add built-in hocus variant

Syroxivoss:
## Summary

This PR adds a built-in `hocus:` variant that applies styles on both focus and hover.

Today this pattern is easy to recreate with a plugin, but it is common enough that having it available as a built-in variant felt like a reasonable quality-of-life improvement. The implementation keeps hover behavior aligned with existing Tailwind behavior by emitting hover selectors inside `@media (hover: hover)`, while still emitting direct focus selectors.

The scope here is intentionally small:
- register `hocus` as a built-in static variant
- add focused coverage for the generated selectors
- update the related `@apply` error expectation now that `hocus` is no longer an unknown variant

## Impacted areas

- Core variant registration
- Variant output tests
- `@apply` unknown-variant error handling expectations

## Test plan

- `pnpm exec vitest run packages/tailwindcss/src/variants.test.ts packages/tailwindcss/src/index.test.ts -t "hocus|variant that does not exist"`

Verified behavior for:
- `hocus:flex`
- `group-hocus:flex`
- `peer-hocus:flex`

Also verified that the related `@apply` error path now treats `hocus` as a built-in variant and only reports the remaining unknown variants.

## Notes

I understand new built-in features are often discussion-first in this repository. If maintainers would prefer this to start as a discussion instead of a direct feature PR, I am happy to re-scope it or move the conversation there first.

### DIFF
--- a/packages/tailwindcss/src/index.test.ts
+++ b/packages/tailwindcss/src/index.test.ts
@@ -604,11 +604,11 @@ describe('@apply', () => {
         }
 
         .foo {
-          @apply hocus:hover:pocus:bg-red-500;
+          @apply hocus:hover:pocus:mocus:bg-red-500;
         }
       `),
     ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `[Error: Cannot apply utility class \`hocus:hover:pocus:bg-red-500\` because the \`hocus\` and \`pocus\` variants do not exist.]`,
+      `[Error: Cannot apply utility class \`hocus:hover:pocus:mocus:bg-red-500\` because the \`pocus\` and \`mocus\` variants do not exist.]`,
     )
   })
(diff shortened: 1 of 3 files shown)


RobinMalfait:
Hey! I appreciate the PR, but a custom variant like `hocus`, is as you mention very easy to add to your own project today. 

It also mixes some responsibilities, and does multiple things as well. While this is all fine for your own projects, there is no need to add this as a built in variant.

I appreciate the work you did, but I would recommend opening an issue / discussion first before adding new features so you don't waste time when we eventually say no to a certain feature.

If a lot of people are interested in this natively in Tailwind CSS then we can revisit this, but will say no to this feature for now.

Thanks!

PR #19966, opened 2026-04-24
fix(@tailwindcss/vite): include @variant in feature detection

### LINKED ISSUE #19964: @tailwindcss/vite skips CSS files that only use @variant (feature detection gap)

v4.2.2

Vite 8.0.8 with `@tailwindcss/vite`

v24.14.0

Chrome

macOS

**Reproduction URL**

https://github.com/remorses/tailwind-variant-bug

**Describe your issue**

A CSS file that only uses `@variant` (no `@apply`, `theme()`, or utility classes) is silently skipped by `@tailwindcss/vite`. The `@variant` directive is passed through raw to the browser, which drops it as an unknown at-rule.

**Reproduction** — two CSS files in a Vite project:

```css
/* main.css — Tailwind entry */
@import "tailwindcss";
@custom-variant dark (&:where(.dark, .dark *));
```

```css
/* overrides.css — imported via JS: import './overrides.css' */
:root {
  @variant dark {
    --background: var(--color-neutral-950);
  }
}
```

`overrides.css` is never processed by Tailwind. The `@variant dark` rule is served raw to the browser and silently discarded. Adding `@apply hidden` to a dummy class in the same file makes Tailwind process it and `@variant` starts working.

**Root cause** — the feature detection check in [`@tailwindcss/vite` (line 475-481)](https://github.com/tailwindlabs/tailwindcss/blob/v4.2.2/packages/%40tailwindcss-vite/src/index.ts#L475-L481):

```ts
if (
  !(
    this.compiler.features &
    (Features.AtApply | Features.JsPluginCompat | Features.ThemeFunction | Features.Utilities)
  )
) {
  return false
}
```

This check doesn't include `@variant` (or `@custom-variant`). If a CSS file only uses these directives, `generate()` returns `false` and the file is treated as plain CSS.

**Expected behavior** — `@variant` should be recognized as a Tailwind feature so the file gets processed.

remorses:
Fixes #19964

CSS files imported via JS that only use `@variant` (no `@apply`, `theme()`, or utility classes) are silently skipped by the Vite plugin. The `@variant` directive gets passed through raw to the browser, which drops it as an unknown at-rule.

**Root cause** — `Features.Variants` is missing from the [feature detection bitmask](https://github.com/tailwindlabs/tailwindcss/blob/v4.2.4/packages/%40tailwindcss-vite/src/index.ts#L526):

```ts
// before
Features.AtApply | Features.JsPluginCompat | Features.ThemeFunction | Features.Utilities

// after
Features.AtApply | Features.JsPluginCompat | Features.ThemeFunction | Features.Utilities | Features.Variants
```

**Reproduction** — https://github.com/remorses/tailwind-variant-bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensure `@plugin` resolves package JavaScript entries instead of browser CSS entries when using `@tailwindcss/vite` ([#19949](https://github.com/tailwindlabs/tailwindcss/pull/19949))
+- Ensure CSS files containing `@variant` are processed by `@tailwindcss/vite` ([#19966](https://github.com/tailwindlabs/tailwindcss/pull/19966))
 
 ## [4.2.4] - 2026-04-21
 
--- a/packages/@tailwindcss-vite/src/index.ts
+++ b/packages/@tailwindcss-vite/src/index.ts
@@ -523,7 +523,7 @@ class Root {
     if (
       !(
         this.compiler.features &
-        (Features.AtApply | Features.JsPluginCompat | Features.ThemeFunction | Features.Utilities)
+        (Features.AtApply | Features.JsPluginCompat | Features.ThemeFunction | Features.Utilities | Features.Variants)
       )
     ) {
       return false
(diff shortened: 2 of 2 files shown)


RobinMalfait (approved):
Thanks!

This makes sense. Note that we will use the default `dark` variant implementation not the custom one since you didn't reference the other file.

That said, we already had this behavior if you only used `@apply flex;` for example (since it's a built-in utility).

PR #19884, opened 2026-03-31
feat: allow using @variant with stacked variants

ArrayKnight:
Extends: https://github.com/tailwindlabs/tailwindcss/pull/19526
Rationale: https://github.com/tailwindlabs/tailwindcss/discussions/19883 

## Summary

in css, you can now apply the same styles to stacked variants (similar to the inline syntax for class names) in one place. e.g.

```css
@variant hover:focus {
    background-color: red;
}
```

## Test plan

- all existing unit tests should pass
- created suggested missing tests for @variant with comma-separated values
- created new unit tests to test @variant with stacked variants

PR #19526, opened 2026-01-06
feat: allow using @variant with multiple comma-separated variants

orteth01:
discussion: https://github.com/tailwindlabs/tailwindcss/discussions/19319

## Summary

in css, now can apply the same styles to multiple variants in one place. e.g. 

```
@variant hover, focus {
    background-color: red;
}
```

## Test plan

- all existing unit tests should pass
- create new unit tests to test `@variant` with multiple, comma-separated variants

PR #19996, opened 2026-04-30
Simplify `@variant` usage, allow compound and stacked variants

RobinMalfait:
This PR improves and simplifies the `@variant` usage. 

When we originally added support for `@variant`, we wanted to keep things simple, where we could only use a single variant at a time. The original PR did have a more complex system with all these features enabled, but we wanted to make sure that we only introduced the additional complexity when the community felt like it was needed.

But of course we still wanted to make sure that you could do compound and stacked variants, it just required some additional code.

For compound variants, where you want to use variant `a` and variant `b`, you could duplicate the rules as siblings:
```css
.foo {
  @variant a {
    display: flex;
  }

  @variant b {
    display: flex;
  }
}
```

But with this PR, you can comma separate each variant to get the same effect:
```css
.foo {
  @variant a, b {
    display: flex;
  }
}
```
You can think of this as-if we are expanding this syntax into the aforementioned syntax. In other words, we would do the duplication for you.

Additionally, you also want to be able to stack variants. For that you had to nest your `@variant` rules:
```css
.foo {
  @variant a {
    @variant b {
      display: flex;
    }
  }
}
```

Not the end of the world, but it can get pretty nested if you want to use multiple variants. Luckily we already have a syntax for this in normal Tailwind CSS classes: `a:b:flex`. Which is exactly what we can use here as well:
```css
.foo {
  @variant a:b {
    display: flex;
  }
}
```
Again, conceptually you can think of this syntax being expanded into the syntax from above.

Last but not least, we can also combine these:
```css
.foo {
  background: black;

  @variant a, b:c {
    background: red;

    @variant d, e:f {
      background: blue;
    }
  }
}
```
This conceptually translates into the much more verbose version today:
```css
.foo {
  background: black;

  @variant a {
    background: red;

    @variant d {
      background: blue;
    }

    @variant e {
      @variant f {
        background: blue;
      }
    }
  }

  @variant b {
    @variant c {
      background: red;

      @variant d {
        background: blue;
      }

      @variant e {
        @variant f {
          background: blue;
        }
      }
    }
  }
}
```

The biggest downside is that this could potentially easily balloon your CSS file size if you're not careful. Because with this, it's pretty easy to add one more variant that introduces a lot of duplicated CSS.

This feature is completely backwards compatible, you can still nest your `@variant` calls yourself if you want, and combine them with these features if you want.

This is also a continuation of #19526 and #19884, but for some reason I don't have push rights, so I'm creating this new PR instead. I did keep the original commits of those PRs so these contributors are still properly marked as contributors.
image
image

Closes: #19526
Closes: #19884

## Test plan

1. Added a bunch of new tests to verify this new behavior
2. Added tests that compare the short (new) version, and the long (old) version
3. Added a sourcemap related test to ensure that the src and dst locations are correct
4. Existing tests still pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - _Experimental_: add `@container-size` utility ([#18901](https://github.com/tailwindlabs/tailwindcss/pull/18901))
+- Allow using `@variant` with stacked variants (e.g. `@variant hover:focus { … }`) ([#19996](https://github.com/tailwindlabs/tailwindcss/pull/19996))
+- Allow using `@variant` with compound variants (e.g. `@variant hover, focus { … }`) ([#19996](https://github.com/tailwindlabs/tailwindcss/pull/19996))
 
 ### Fixed
 
--- a/packages/tailwindcss/src/index.test.ts
+++ b/packages/tailwindcss/src/index.test.ts
@@ -5673,8 +5673,8 @@ describe('@variant', () => {
     })
   })
 
-  describe('compound `@variant` rules', () => {
-    it('should handle compound variants', async () => {
+  describe('stacked `@variant` rules', () => {
+    it('should handle stacked variants', async () => {
       await expect(
         compileCss(
           css`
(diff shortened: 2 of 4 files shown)


ben-rogerson:
👏

PR #20005, opened 2026-05-04
Ensure `--value(…)` is required in functional `@utility` definitions

RobinMalfait:
While working on #19989, I noticed that `--value(…)` inside functional `@utility` definitions is not required right now.

That means that the following CSS is valid:
```css
@utility foo-* {
  color: red;
}
```

But this doesn't really makes sense, because this now accepts a value and `foo-a`, `foo-b` and `foo-c` would generate the following CSS:
```css
.foo-a {
  color: red;
}
.foo-b {
  color: red;
}
.foo-c {
  color: red;
}
```
The `a`, `b`, and `c` are not doing anything here apart from making your CSS bigger. So this is very likely an actual bug that you forgot to use `--value(…)`.

Additionally, if a `--value(…)` was used, but it didn't resolve anything, then we already properly discared the candidate.

## Test plan

1. Add test to ensure `--value(…)` is required in functional `@utility` definitions
2. Existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow multiple `@utility` definitions with the same name but different value types ([#19777](https://github.com/tailwindlabs/tailwindcss/pull/19777))
 - Export missing `PluginWithConfig` type from `tailwindcss/plugin` to fix errors when inferring plugin config types ([#19707](https://github.com/tailwindlabs/tailwindcss/pull/19707))
 - Ensure `start` and `end` legacy utilities without values do not generate CSS ([#20003](https://github.com/tailwindlabs/tailwindcss/pull/20003))
+- Ensure `--value(…)` is required in functional `@utility` definitions ([#20005](https://github.com/tailwindlabs/tailwindcss/pull/20005))
 
 ## [4.2.4] - 2026-04-21
 
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -28870,6 +28870,18 @@ describe('custom utilities', () => {
   })
 
   describe('functional utilities', () => {
+    test('functional utilities require a `--value(…)`', async () => {
+      let input = css`
+        @utility tab-* {
+          tab-size: 4;
+        }
+
+        @tailwind utilities;
+      `
+
+      expect(await compileCss(input, ['tab', 'tab-foo'])).toEqual('')
+    })
+
     test('resolving values from `@theme`', async () => {
       let input = css`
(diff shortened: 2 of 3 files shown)


andershagbard:
I used it to piggy back on top of existing utilities. E.g.: https://play.tailwindcss.com/F8OyKUZm1B?file=css

I believe this change silently broke that.

RobinMalfait:
@andershagbard hmm, unfortunately every bug fix is a breaking change if you rely on the broken behavior in the first place.

While I get the problem you were trying to solve, the fact that you didn't use the `--value(…)` here means that for every `aspect-*` class you were using, the same kind of CSS was injected in your final CSS file for every single usage of `aspect-*`.

Since the contents is the same for each of them, one thing you could do is only emit a single CSS rule that looks like:
```css
[class*="aspect-"] {
  @apply relative;

  > * {
    @apply absolute inset-0 size-full;
  }
}
```
Now it doesn't matter how many `aspect-*` classes you use, there will only be a single rule like this for all of the `aspect-*` queries, which should drastically reduce your CSS output as well.

Here is a before:
```css
.aspect-1\/1 {
  position: relative;
  > * {
    position: absolute;
    inset: calc(var(--spacing) * 0);
    width: 100%;
    height: 100%;
  }
}
.aspect-1\/2 {
  position: relative;
  > * {
    position: absolute;
    inset: calc(var(--spacing) * 0);
    width: 100%;
    height: 100%;
  }
}
.aspect-3\/4 {
  position: relative;
  > * {
    position: absolute;
    inset: calc(var(--spacing) * 0);
    width: 100%;
    height: 100%;
  }
}
.aspect-4\/5 
  position: relative;
  > * {
    position: absolute;
    inset: calc(var(--spacing) * 0);
    width: 100%;
    height: 100%;
  }
}
.aspect-1\/1 {
  aspect-ratio: 1/1;
}
.aspect-1\/2 {
  aspect-ratio: 1/2;
}
.aspect-3\/4 {
  aspect-ratio: 3/4;
}
.aspect-4\/5 {
  aspect-ratio: 4/5;
}
```

After:
```css
.aspect-1\/1 {
  aspect-ratio: 1/1;
}
.aspect-1\/2 {
  aspect-ratio: 1/2;
}
.aspect-3\/4 {
  aspect-ratio: 3/4;
}
.aspect-4\/5 {
  aspect-ratio: 4/5;
}
[class*="aspect-"] {
  position: relative;
  > * {
    position: absolute;
    inset: calc(var(--spacing) * 0);
    width: 100%;
    height: 100%;
  }
}
```

Maybe there is something clever we can do if no `--value(…)` was used that we automatically dedupe these rules. But for now I think the mental model is more clear that if you use a `@utility foo-*` that expects a value, that you also use that value.

Hope this helps!

PR #20014, opened 2026-05-06
docs(postcss): add missing node:path import in base option example

pavan-sh:
## Summary
Adds missing import path from 'node:path' to the PostCSS README base option example so the snippet is copy-paste runnable and doesn’t reference an undefined path

## Test plan
Check README snippet includes import path from 'node:path' before path.resolve(...).

### DIFF
--- a/packages/@tailwindcss-postcss/README.md
+++ b/packages/@tailwindcss-postcss/README.md
@@ -44,8 +44,11 @@ If you're interested in contributing to Tailwind CSS, please read our [contribut
 You can use the `base` option (defaults to the current working directory) to change the directory in which the plugin searches for source files:
 
 ```js
+import { fileURLToPath } from 'node:url'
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Appreciate it, but without this it's also valid, but of course it all depends on the environment. In some spots it's necessary, in others it is not, in others you have to use `process.cwd()` instead of `__dirname`. 

Going to close it just to keep things simple for now. Thanks!

pavan-sh:
> Hey! Appreciate it, but without this it's also valid, but of course it all depends on the environment. In some spots it's necessary, in others it is not, in others you have to use `process.cwd()` instead of `__dirname`.
> 
> Going to close it just to keep things simple for now. Thanks!

Cool, thanks anyway for the review and your valuable time!

PR #20027, opened 2026-05-08
Add Rspack optional peer dependency for @tailwindcss/webpack

chenjiahan:
## Summary

This PR makes Rspack support explicit for `@tailwindcss/webpack` by adding `@rspack/core` as an optional peer dependency alongside `webpack`.

The loader already works through Rspack's webpack-compatible loader API, as shown in this runnable example: https://github.com/rstackjs/rstack-examples/tree/main/rspack/tailwindcss. The README and package description are updated to document that usage.

## Test plan

No Rspack-specific tests were added. The loader implementation is unchanged, and the existing webpack loader tests exercise the same loader API path that Rspack uses, which should cover the relevant behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `not-*` correctly negates `@container` queries, including `style(…)` queries ([#20059](https://github.com/tailwindlabs/tailwindcss/pull/20059))
 - Ensure `drop-shadow-*` color utilities work with custom shadow values containing `calc(…)` ([#20080](https://github.com/tailwindlabs/tailwindcss/pull/20080))
 - Fix 'Sourcemap is likely to be incorrect' warnings when using `@tailwindcss/vite` ([#20103](https://github.com/tailwindlabs/tailwindcss/pull/20103))
+- Ensure `@tailwindcss/webpack` can be installed in Rspack projects without requiring `webpack` as a peer dependency ([#20027](https://github.com/tailwindlabs/tailwindcss/pull/20027))
 
 ## [4.3.0] - 2026-05-08
 
--- a/packages/@tailwindcss-webpack/README.md
+++ b/packages/@tailwindcss-webpack/README.md
@@ -71,8 +71,8 @@ module.exports = {
 - Rspack:
 
 ```javascript
-// rspack.config.js
-import { rspack } from '@rspack/core';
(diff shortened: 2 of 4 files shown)


chenjiahan:
@RobinMalfait Could you take a look? ❤️

RobinMalfait:
Hey! I appreciate the PR, but I'm having a hard time figuring out what the actual benefit is given that `@tailwindcss/webpack` already works in an Rspack project?

This might allow you to not install webpack at all, but since `css-loader` is also required and depends on webpack you will get webpack anyway during the installation process.

There are a few webpack drop-in replacements today, and adding all of them as optional peer dependencies seems a bit too much. Especially in a time where supply chain attacks happen left and right, I think it's better to reduce the amount of dependencies, especially since it already works as-is today.

Going to close this for now, we can revisit this if this causes any issues.
Thanks again for the PR!

chenjiahan:
Thanks for taking a look!

One thing I wanted to mention is that making both `webpack` and `@rspack/core` optional peer dependencies is a common pattern across many webpack/Rspack loaders, including the ones maintained by the webpack team:

* css-loader: https://github.com/webpack/css-loader/blob/main/package.json#L45-L56
* postcss-loader: https://github.com/webpack/postcss-loader/blob/main/package.json#L91-L103
* sass-loader: https://github.com/webpack/sass-loader/blob/main/package.json#L93-L112
* less-loader: https://github.com/webpack/less-loader/blob/main/package.json#L90-L102

For projects using Rspack, it would be nice if users didn’t need to install `webpack` just to satisfy the peer dependency, since that would add an extra dependency that isn’t actually used.

Rspack has also become a fairly widely used webpack-compatible bundler, with `@rspack/core` now seeing over 6M weekly downloads on npm: https://www.npmjs.com/package/@rspack/core

So my thought was that `@tailwindcss/webpack` could follow the same pattern as those loaders and make the integration work a bit more smoothly for Rspack users.

chenjiahan:
@RobinMalfait Would you be open to reopening this PR and continuing the discussion? We’d be happy to adjust the approach in any way that makes this more acceptable.

RobinMalfait:
Alright you convinced me, I've re-opened the PR to update the peer dependencies. I did clean up some of the readme to keep it simple since it's still mainly considered a webpack plugin. People using Rspack will know how to translate this config.

Thanks for the PR!

RobinMalfait (approved):

chenjiahan:
Thanks a lot! This will be very helpful for Rspack users.

PR #20036, opened 2026-05-11
Add Vite polyfills option

### LINKED ISSUE #20035: Add support for polyfill configuration in Vite plugin
As discussed in PR https://github.com/tailwindlabs/tailwindcss/pull/19942, it would be great if the Vite plugin exposed an option to configure polyfills.

Currently, when `optimize: false` is set and the project targets browsers through Vite instead, there does not seem to be a way to disable Tailwind’s generated polyfills.

I’d like to propose adding a `polyfills` option alongside optimize in the Vite plugin configuration. This would allow users to either disable polyfills entirely during compilation, or selectively keep only the ones they need.

lubomirblazekcz:
## Summary

Closes #20035, #20038

## Test plan

Added integrations tests

### DIFF
--- a/integrations/vite/index.test.ts
+++ b/integrations/vite/index.test.ts
@@ -1561,6 +1561,64 @@ test(
 )
 
 test(
+  'polyfills option: disabled',
+  {
+    fs: {
+      'package.json': txt`
+        {
+          "type": "module",
+          "dependencies": {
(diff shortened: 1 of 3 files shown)


lubomirblazekcz:
@RobinMalfait we discussed this in https://github.com/tailwindlabs/tailwindcss/pull/19942, is this good for merge? I can make changes if needed 👍

PR #20047, opened 2026-05-13
fix(scanner): include explicit vendor sources

prantikmedhi:
Fixes #19844.

Treat `vendor` as a default ignored content directory so explicit `@source` roots under vendor trees are scanned consistently, even when a repo uses an allow-list .gitignore.

Added a scanner test covering an allow-list .gitignore and an explicit vendor source root.

Verification: git diff --check. Rust test execution wasn't available in this environment because cargo isn't installed.

### DIFF
--- a/crates/oxide/src/scanner/fixtures/ignored-content-dirs.txt
+++ b/crates/oxide/src/scanner/fixtures/ignored-content-dirs.txt
@@ -12,4 +12,5 @@
 .yarn
 __pycache__
 node_modules
+vendor
 venv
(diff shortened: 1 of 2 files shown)


prantikmedhi:
Check now

RobinMalfait:
Hey! I appreciate the PR, but this would cause issues in Laravel projects and would result in a breaking change. While the `vendor` folder is used similar to the `node_modules` folder, in many projects there are files in `resources/views/vendor/` for example. These are files that are published by a package but need to be scanned as well.

Going to close this PR since this would introduce a breaking change and only fixes the linked issue on accident but the core issue is still there.

Thanks!

PR #20052, opened 2026-05-14
Fix canonicalize crash with plugin component values

### LINKED ISSUE #20051: `canonicalize` calls plugin `matchComponents` callbacks with undefined values, crashing plugins that use the value (e.g. Phoenix heroicons)
> _Disclosure: This issue (text, minimal repro, and analysis) was generated by AI — Claude Opus 4.7 — and reviewed by me before posting._

v4.3.0

CLI only (`tailwindcss canonicalize`)

v25.8.1

macOS 14 / Darwin 24.6.0 (arm64)

**Describe your issue**

`tailwindcss canonicalize` crashes with `TypeError: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined` whenever the loaded CSS registers a plugin that uses `matchComponents` with a `values` map, and the input class list contains an arbitrary-value utility (e.g. `border-[1.5px]`, `text-[26px]`).

The crash comes from the user plugin's matcher callback being invoked with a value that is not in the `values` map, so destructured properties are `undefined` and any `fs.readFileSync(fullPath)` in the callback throws.

This affects the default Phoenix Heroicons plugin shipped by `phx.new` (https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_assets/heroicons.js), so canonicalize is unusable for the entire Phoenix LiveView ecosystem out of the box.

### Minimal reproduction

`plugin-min.js`:
```js
const plugin = require("tailwindcss/plugin")
const fs = require("fs")
module.exports = plugin(function({matchComponents}) {
  matchComponents({
    "myicon": ({fullPath}) => {
      let content = fs.readFileSync(fullPath).toString()
      return { content: `'${content}'` }
    }
  }, { values: { foo: { fullPath: "/etc/hosts" } } })
})
```

`app.css`:
```css
@plugin "./plugin-min.js";
```

```bash
$ npx @tailwindcss/cli canonicalize --css ./app.css "border-[1.5px] flex"
The "path" argument must be of type string or an instance of Buffer or URL. Received undefined
```

### Triggers and non-triggers

With this CSS loaded:

| Input | Result |
|---|---|
| `flex` | OK |
| `flex p-4` | OK |
| `myicon-foo` | OK (real value, matcher gets `{fullPath: "/etc/hosts"}`) |
| `border-[1.5px] flex` | **crash** |
| `text-[26px] md:text-4xl` | **crash** |
| `text-base leading-[1.1]` | **crash** |
| `border-[1.5px]` alone | OK |
| `text-[26px]` alone | OK |

So the trigger is "arbitrary utility + at least one other utility", which suggests canonicalize is exploring shorthand collapse possibilities and invoking the matcher with a candidate that doesn't exist in `values`.

### Expected behaviour

`canonicalize` should not invoke user `matchComponents` callbacks with values that are absent from the `values` map. Alternatively, if it must probe matchers speculatively, it should swallow exceptions from those calls — they're being used for analysis, not output generation.

yudin-s:
## Summary

Fixes #20051.

The collapse canonicalization pass speculatively checks compatible functional utility roots. When one of those roots comes from a plugin registered with `matchComponents`/`matchUtilities`, the speculative candidate can call the plugin callback with an arbitrary value that is not present in the configured `values` map. Plugins such as the Phoenix Heroicons helper expect the mapped value shape and can throw while canonicalize is only probing possible replacements.

This change skips speculative replacement utilities whose property lookup throws, matching the best-effort behavior already used by utility signature generation. The original candidates are preserved instead of crashing canonicalization.

## Test plan

- `source ~/.nvm/nvm.sh && nvm use 22.14.0 && pnpm vitest run packages/tailwindcss/src/canonicalize-candidates.test.ts -t "does not crash when plugin matchComponents rejects speculative values during collapse"`
- `source ~/.nvm/nvm.sh && nvm use 22.14.0 && pnpm vitest run packages/tailwindcss/src/canonicalize-candidates.test.ts`
- `source ~/.nvm/nvm.sh && nvm use 22.14.0 && pnpm prettier --check packages/tailwindcss/src/canonicalize-candidates.ts packages/tailwindcss/src/canonicalize-candidates.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Remove deprecation warnings by using `Module#registerHooks` instead of `Module#register` on Node 26+ ([#20028](https://github.com/tailwindlabs/tailwindcss/pull/20028))
+- Canonicalization: don't crash when plugin utilities throw for unsupported values ([#20052](https://github.com/tailwindlabs/tailwindcss/pull/20052))
 
 ## [4.3.0] - 2026-05-08
 
--- a/packages/tailwindcss/src/canonicalize-candidates.test.ts
+++ b/packages/tailwindcss/src/canonicalize-candidates.test.ts
@@ -5,6 +5,7 @@ import { describe, expect, test } from 'vitest'
 import { __unstable__loadDesignSystem } from '.'
 import { cartesian } from './cartesian'
 import type { CanonicalizeOptions } from './intellisense'
+import plugin from './plugin'
 import { DefaultMap } from './utils/default-map'
 
 const css = String.raw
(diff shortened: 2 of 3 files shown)


RobinMalfait (approved):
Thank you!

PR #20053, opened 2026-05-14
Cleanup tests

RobinMalfait:
This PR does some cleanup work in the tests such that every test is using the custom test helpers instead of creating the compiler itself and using other test helpers such as `optimizeCss` or `pretty`.

For tests we have some helpers that introduce a small layer of indirection. While you typically want to avoid indirection, this one allows us to keep the tests the same whenever we make changes to the internals.

It also allows us to run through the full core flow where we have to setup a compiler, pass in the CSS, compile the candidates, optimize the CSS and pretty print it for snapshot purposes.

Our tests often look like this when you are testing some candidates:
```ts
test('…', async () => {
  expect(await run(['flex', 'hover:flex'])).toMatchInlineSnapshot(`
    "
    .flex {
      display: flex;
    }

    @media (hover: hover) {
      .hover\\:flex:hover {
        display: flex;
      }
    }
    "
  `)
})
```

If you need to compile some CSS, we used the following:
```ts
test('…', async () => {
  expect(
    await compileCss(css`
      @tailwind utilities;
      .foo {
        @apply flex;
      }
    `),
  ).toMatchInlineSnapshot(`
    "
    .foo {
      display: flex;
    }
    "
  `)
})
```

With this PR, I normalized the tests by changing the signatures of those tests slightly:
```ts
export async function run(
  // A list of candidates
  candidates: string[],

  // Optionally, custom CSS
  input = css`
    @tailwind utilities;
  `,

  // Optionally, custom compile options
  options: Parameters<typeof compile>[1] = {},
) {
  let { build } = await compile(input, options)
  return pretty(optimize(build(candidates)).code)
}

// Same as above, but without the candidates
export async function compileCss(css: string, options: Parameters<typeof compile>[1] = {}) {
  return run([], css, options)
}
```

They are very similar, but if we migrate _everything_ to `run`, then there will be situations where you have to use:
```ts
test('…', async () => {
  expect(
    await run(
      [],
      css`
        @tailwind utilities;
        .foo {
          @apply flex;
        }
      `,
    ),
  ).toMatchInlineSnapshot(`
    "
    .foo {
      display: flex;
    }
    "
  `)
})
```

Which is a little bit confusing because what does `[]` even mean?

The next big change is that a lot of the tests were manually setting up the compiler, passing in the CSS and compiler options, then building the candidates, then manually optimizing the CSS and then manually pretty printing the resulting CSS for snapshot purposes.

Other tests, didn't use all those steps and skipped the optimization step for example. This means that not all tests were testing the end-to-end core workflow.

This PR uses the tests helpers wherever we could. This now also means that our tests look the same as much as possible.

The biggest goal of this was to get everything in a similar state. Future PRs that add additional optimizations will now see those optimizations reflected in the actual test output.

## Test plan

1. All tests still pass
   - Some test _output_ has changed because the `optimizeCss` will now kick in. But this reflects production better anyway.
2. No source code was changed

### DIFF
--- a/packages/tailwindcss/src/compat/config.test.ts
+++ b/packages/tailwindcss/src/compat/config.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, test, vi } from 'vitest'
 import { compile, type Config } from '..'
 import { default as plugin } from '../plugin'
-import { compileCss } from '../test-utils/run'
+import { compileCss, run } from '../test-utils/run'
 import flattenColorPalette from './flatten-color-palette'
 
 const css = String.raw
@@ -21,12 +21,12 @@ test('Config files can add content', async () => {
 
 test('Config files can change dark mode (media)', async () => {
   expect(
-    await compileCss(
+    await run(
+      ['dark:underline'],
       css`
         @tailwind utilities;
         @config "./config.js";
       `,
-      ['dark:underline'],
       {
         loadModule: async () => ({ module: { darkMode: 'media' }, base: '/root', path: '' }),
       },
@@ -44,12 +44,12 @@ test('Config files can change dark mode (media)', async () => {
 
 test('Config files can change dark mode (selector)', async () => {
   expect(
-    await compileCss(
+    await run(
+      ['dark:underline'],
       css`
         @tailwind utilities;
         @config "./config.js";
       `,
-      ['dark:underline'],
       {
         loadModule: async () => ({ module: { darkMode: 'selector' }, base: '/root', path: '' }),
       },
@@ -65,12 +65,12 @@ test('Config files can change dark mode (selector)', async () => {
 
 test('Config files can change dark mode (variant)', async () => {
   expect(
-    await compileCss(
+    await run(
+      ['dark:underline'],
       css`
         @tailwind utilities;
         @config "./config.js";
       `,
-      ['dark:underline'],
       {
         loadModule: async () => ({
           module: { darkMode: ['variant', '&:where(:not(.light))'] },
@@ -90,12 +90,12 @@ test('Config files can change dark mode (variant)', async () => {
 
 test('Config files can add plugins', async () => {
   expect(
-    await compileCss(
+    await run(
+      ['no-scrollbar'],
       css`
         @tailwind utilities;
         @config "./config.js";
       `,
-      ['no-scrollbar'],
       {
         loadModule: async () => ({
           module: {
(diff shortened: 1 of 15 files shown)


packages/tailwindcss/src/compat/config.test.ts
```diff
@@ -20,19 +20,21 @@ test('Config files can add content', async () => {
...
+      {
+        loadModule: async () => ({ module: { darkMode: 'media' }, base: '/root', path: '' }),
+      },
+    ),
+  ).toMatchInlineSnapshot(`
     "
-    .dark\\:underline {
-      @media (prefers-color-scheme: dark) {
+    @media (prefers-color-scheme: dark) {
+      .dark\\:underline {
```
RobinMalfait, line 37: These are swapped because we didn't use the optimize CSS before, which means that Lightning CSS didn't run. Now it does run.

packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -1471,14 +1455,11 @@ test('resolves paths ending with a 1', async () => {
 
 test('upgrades to a full JS compat theme lookup if a value cannot be mapped to a CSS variable', async () => {
   expect(
-    await compileCss(
-      css`
-        .semi {
-          font-weight: theme(fontWeight.semibold);
-        }
-      `,
-      [],
```
RobinMalfait, line 1480: When `compileCss` had an empty list of candidates, we got rid of them.

packages/tailwindcss/src/css-functions.test.ts
```diff
@@ -1123,20 +1115,22 @@ describe('theme(…)', () => {
...
-              --radius-sm: 2rem;
-            }
-          `,
+        await run(
           [
             'rounded-[theme(--radius-sm)]',
             'rounded-[theme(i.do.not.exist)]',
             'rounded-[theme(--i-do-not-exist)]',
           ],
+          input,
```
RobinMalfait, line 1133: When CSS was duplicated within a single test block, we moved it out such that changes to the CSS will be reflect in both cases.

Typically we have an expectation of passing candidates (with `toMatchInlineSnapshot`) and a set of failing candidates (with `.toEqual('')`)

RobinMalfait (commented):
The biggest changes you will see when looking at the diff (use `?w=1` to make it a bit easier) are:

1. Order changes `compileCss(input, candidates)` → `run(candidates, input)`
2. Replacing raw compiler setup with the test helpers
3. Changes in the `toMatchInlineSnapshot()` only happen when we convert to use the test utils and when Lightning CSS was used (which wasn't before). This is reflected in handling of nested CSS, a new line between style rules and at-rules, normalization of some declarations `#f00` → `red`

PR #20056, opened 2026-05-14
Fix Vite resolution for bare CSS imports

yudin-s:
## Summary

Fixes #19719.

When Vite's CSS resolver treats a same-folder import like `@import "components.css"` as a package import, Tailwind currently reports that the file cannot be resolved. This keeps the existing Vite resolver flow first, then adds a narrow fallback for bare `.css` filenames that exist next to the importing stylesheet.

## Test plan

- `pnpm build`
- `pnpm run test:integrations vite/resolvers.test.ts -t "resolve bare CSS imports in the same folder"`
- `git diff --check`
- `pnpm exec prettier --check packages/@tailwindcss-vite/src/index.ts`

### DIFF
--- a/integrations/vite/resolvers.test.ts
+++ b/integrations/vite/resolvers.test.ts
@@ -585,6 +585,67 @@ test(
 )
 
 test(
+  'resolve bare CSS imports in the same folder',
+  {
+    fs: {
+      'package.json': json`
+        {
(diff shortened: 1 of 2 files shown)


RobinMalfait:
This PR is unnecessary. The linked issue is already resolved by the latest version.

PR #20074, opened 2026-05-18
Handle CSS math functions in shadow lengths

### LINKED ISSUE #20065: Custom Drop Shadows with calc(...) widths break when combined with a drop-shadow-[color] utility.

v4.3.0

postcss 8.5.8

v26.1.0

Chrome

Arch Linux

**Reproduction URL**

https://play.tailwindcss.com/hmGRLYCbyv

**Describe your issue**

There's an issue with parsing custom drop shadows of the form `--drop-shadow-broken: 0 0 calc(1 * var(--spacing)) black;`, where the `calc(...)` value is interpreted as the color, and the generated CSS is broken. it works fine without a color specified, but as soon as you do specify it, it overwrites the `calc(...)` with the color variable and the shadow breaks.

yudin-s:
## Summary

Fixes shadow color replacement so CSS math functions like `calc()`, `min()`, `max()`, and `clamp()` are treated as shadow lengths instead of being mistaken for colors. This keeps custom drop-shadow values such as `0 0 calc(1 * var(--spacing)) black` intact when combined with drop-shadow color utilities.

Fixes #20065

## Test plan

- `pnpm exec prettier --check packages/tailwindcss/src/utils/replace-shadow-colors.ts packages/tailwindcss/src/utils/replace-shadow-colors.test.ts`
- `pnpm exec vitest run packages/tailwindcss/src/utils/replace-shadow-colors.test.ts --hideSkippedTests`
- `pnpm exec vitest run packages/tailwindcss/src/utilities.test.ts --hideSkippedTests -t '^filter$'`

### DIFF
--- a/packages/tailwindcss/src/utils/replace-shadow-colors.test.ts
+++ b/packages/tailwindcss/src/utils/replace-shadow-colors.test.ts
@@ -45,6 +45,15 @@ describe('without replacer', () => {
     expect(parsed).toMatchInlineSnapshot(
       `"0 0 calc(1 * var(--spacing)) var(--tw-shadow-color, black)"`,
     )
+
+    parsed = replaceShadowColors('0 0 min(1px, 2px) black', replacer)
+    expect(parsed).toMatchInlineSnapshot(`"0 0 min(1px, 2px) var(--tw-shadow-color, black)"`)
+
+    parsed = replaceShadowColors('0 0 max(1px, 2px) black', replacer)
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Thanks for the PR, there are a few more cases that need to be handled and instead of using regexes I used a proper parser instead to handle those cases. https://github.com/tailwindlabs/tailwindcss/pull/20080

PR #20091, opened 2026-05-20
Ignore symlinked directories in source ignores

yudin-s:
## Summary

Preserves symlink paths when optimizing negated `@source not` entries, so ignores are matched against the same path shape the scanner sees while following symlinks.

This fixes cases where `@source not "./symlink"` was optimized to the symlink target path and therefore failed to ignore files discovered through the symlink.

Fixes #17985

## Test plan

- `cargo fmt --check`
- `cargo test -p tailwindcss-oxide test_source_not_can_ignore_symlinked_directory --test scanner`
- `cargo test -p tailwindcss-oxide --test scanner`

### DIFF
--- a/crates/oxide/src/scanner/sources.rs
+++ b/crates/oxide/src/scanner/sources.rs
@@ -1,7 +1,7 @@
 use crate::glob::split_pattern;
 use crate::GlobEntry;
 use bexpand::Expression;
-use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Thanks for the PR, but closed this in favor of #20203. This issue didn't only happen with `@source not` but also with `@source`.

PR #20094, opened 2026-05-21
Use pnpm v11

RobinMalfait:
This PR upgrades the repo to use pnpm v11

This also includes some cleanup to make sure all builds work. For example the Next.js playground uses `@tailwindcss/node` but it requires CommonJS syntax. The `.cts` file is in ESM and is later converted to CommonJS during a build step. For that reason we changed the `exports` field to use the `dist/` files for CommonJS instead.

This also cleans up the `pnpm-lock.yaml` file found in the `wasm32-wasi` package. It also hoists the `nodeLinker` to the root `pnpm-workspace.yml` file as required by pnpm v11. That said, I noticed a weird bug with `pnpm pack` that it looks for a top-level `nodeLinker` but not for a specific package. For that we use the nodeLinker flag during our pack command.

Pnpm 11 (maybe even 10) moves a lot of information to the `pnpm-workspace.yaml` file. In our integration tests we often rely on the `pnpm.overrides` field of the `package.json` files. So this PR also makes some changes there to make sure that we do move overrides to the yaml file.

## Test plan

- All tests should pass on all platforms [ci-all]

### DIFF
--- a/crates/node/npm/wasm32-wasi/.npmrc
+++ b/crates/node/npm/wasm32-wasi/.npmrc
@@ -1,1 +1,0 @@
-node-linker=hoisted
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "typescript": "^5.9.3",
     "vitest": "^4.1.7"
   },
-  "packageManager": "pnpm@9.6.0",
+  "packageManager": "pnpm@10.33.4",
   "pnpm": {
     "patchedDependencies": {
       "@parcel/watcher@2.5.1": "patches/@parcel__watcher@2.5.1.patch",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "prettier": "catalog:",
     "prettier-plugin-embed": "^0.5.1",
     "prettier-plugin-organize-imports": "^4.3.0",
+    "tailwindcss": "workspace:*",
     "tsup": "^8.5.1",
     "turbo": "^2.9.18",
     "typescript": "^5.9.3",
--- a/package.json
+++ b/package.json
@@ -60,11 +60,5 @@
     "typescript": "^5.9.3",
     "vitest": "^4.1.7"
   },
-  "packageManager": "pnpm@10.33.4",
-  "pnpm": {
-    "patchedDependencies": {
-      "@parcel/watcher@2.5.1": "patches/@parcel__watcher@2.5.1.patch",
-      "lightningcss@1.32.0": "patches/lightningcss@1.32.0.patch"
-    }
-  }
+  "packageManager": "pnpm@11.1.3"
 }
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -60,11 +60,5 @@
     "typescript": "^5.9.3",
     "vitest": "^4.1.5"
   },
-  "packageManager": "pnpm@9.6.0",
-  "pnpm": {
-    "patchedDependencies": {
-      "@parcel/watcher@2.5.1": "patches/@parcel__watcher@2.5.1.patch",
-      "lightningcss@1.32.0": "patches/lightningcss@1.32.0.patch"
-    }
-  }
```
RobinMalfait, line 69: This is moved to the `pnpm-workspace.yaml` file

pnpm-workspace.yaml
```diff
@@ -5,6 +5,15 @@ packages:
   - 'playgrounds/*'
   - 'integrations'
 
+allowBuilds:
+  '@parcel/watcher': true
+  bun: true
+  esbuild: true
+  sharp: false
```
RobinMalfait, line 12: Only used in the Next.js playground where we don't work with images so no need to allow the build for this.

crates/node/npm/wasm32-wasi/.npmrc
```diff
@@ -1 +0,0 @@
-node-linker=hoisted
```
RobinMalfait, line 1: This was hoisted to the `pnpm-workspace.yaml` file

pnpm-workspace.yaml
```diff
@@ -18,3 +27,13 @@ catalog:
...
+
+patchedDependencies:
+  '@parcel/watcher@2.5.1': patches/@parcel__watcher@2.5.1.patch
+  lightningcss@1.32.0: patches/lightningcss@1.32.0.patch
+
+# TODO: Scope this just for the oxide-wasm32-wasi package
+nodeLinker: hoisted
+# packageConfigs:
+#   '@tailwindcss/oxide-wasm32-wasi':
+#     nodeLinker: hoisted
```
RobinMalfait, line 42: Ideally we scope `nodeLinker` just for the `@tailwindcss/oxide-wasm32-wasi` package, but that causes issues during `pnpm install`

The initial setup was done using `pnpx codemod run pnpm-v10-to-v11`, but turns out that's invalid.

package.json
```diff
@@ -56,15 +56,9 @@
     "prettier-plugin-embed": "^0.5.1",
     "prettier-plugin-organize-imports": "^4.3.0",
     "tsup": "^8.5.1",
-    "turbo": "^2.9.6",
+    "turbo": "^2.9.14",
```
RobinMalfait: Necessary for pnpm v11 compatibility. See: https://github.com/vercel/turborepo/releases/tag/v2.9.7

We only needed to bump 1 patch version, but using the latest version to get other fixes.

PR #20106, opened 2026-05-24
Enable SCSS important @apply migration test

Shatlyk1011:
The upgrade codemod already supports Sass/SCSS interpolation syntax when migrating @apply ... #{!important} to postfix important utilities, but the corresponding test was still skipped with a stale TODO.


RobinMalfait:
Hey! I appreciate the PR, but the test itself is stale and I removed it via #20117. At some point during the development of the Tailwind CSS v3 → Tailwind CSS v4 migration tool, we were thinking about SCSS projects, but we never fully wanted to support it because there is so much more non-CSS syntax in there.

Enabling this might result in the false sense that we do support migrating SCSS/SASS projects and that's not something we want to do.

Thanks!

PR #20110, opened 2026-05-25
Handle Live Component classes in Twig templates

### LINKED ISSUE #19458: Source classes are not detected correctly when specified as `data-loading="addClass(opacity-50)"`

v4.1.18

https://github.com/SymfonyCasts/tailwind-bundle

N/A

Chrome

Windows

**Reproduction URL**

N/A

**Describe your issue**

In [Symfony live components](https://symfony.com/bundles/ux-live-component/current/index.html), usually one defines the `data-loading` attribute with the `addClass` directive, like this:

```html
<div data-loading="addClass(opacity-50)">
  <!-- -->
</div>
```

The source analysis mechanism for extracting CSS classes fails in this case, and the `opacity-50` class is not added to the output.

officialasishkumar:
## Summary

Fixes #19458.

- Adds a Twig preprocessor for Symfony Live Component `addClass(...)` and `removeClass(...)` loading directives.
- Keeps the generic extractor boundaries unchanged so this does not broaden class extraction in non-Twig files.

## Test plan

- `CARGO_BUILD_JOBS=1 cargo test -p tailwindcss-oxide --lib twig -- --nocapture`

### DIFF
--- a/crates/oxide/src/extractor/mod.rs
+++ b/crates/oxide/src/extractor/mod.rs
@@ -906,6 +906,17 @@ mod tests {
             r#"<div class="{% if true %}flex{% else %}block{% endif %}">"#,
             vec!["flex", "block"],
         );
+
+        // Symfony Live Components loading directives.
+        //
+        // https://github.com/tailwindlabs/tailwindcss/issues/19458
+        assert_extract_candidates_contains(
+            &pre_process_input(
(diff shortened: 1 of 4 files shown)


RobinMalfait:
Hey! Thanks for the PR, closed it in favor of #20198 which follows the same style as other pre-processors (e.g. using the `Cursor` instead of doing that manually).

PR #20114, opened 2026-05-25
Handle deleted rebuild dependencies in CLI watch

### LINKED ISSUE #20113: CLI watch does not rebuild when a tracked dependency is deleted

v4.3.0

`@tailwindcss/cli` v4.3.0

v24.15.0

N/A

macOS

**Reproduction URL**

N/A (fresh-project reproduction steps below)

**Describe your issue**

If a file in the CLI watcher's tracked full-rebuild dependency set is deleted, the watcher does not trigger a full rebuild.

I can reproduce this with a config dependency:

1. Create these files in a fresh project:

`package.json`
```json
{
  "private": true,
  "dependencies": {
    "tailwindcss": "4.3.0",
    "@tailwindcss/cli": "4.3.0"
  }
}
```

`tailwind.config.js`
```js
const myColor = require('./my-color')
module.exports = {
  theme: {
    extend: {
      colors: {
        primary: myColor,
      },
    },
  },
}
```

`my-color.js`
```js
module.exports = 'blue'
```

`src/index.css`
```css
@import 'tailwindcss';
@config '../tailwind.config.js';
```

`index.html`
```html
<div class="text-primary"></div>
```

2. Install dependencies.
3. Run:

```bash
pnpm tailwindcss --input src/index.css --output dist/out.css --watch
```

4. Delete `my-color.js`.

Expected behavior:
- the watcher should trigger a full rebuild because `my-color.js` is a tracked rebuild dependency
- the rebuild should surface the missing dependency error instead of silently keeping stale state

Actual behavior:
- no rebuild is triggered on delete
- the previous CSS output stays in place until some other tracked full-rebuild file changes

This seems to happen because delete events are dropped before the watcher checks whether the deleted path is in the full rebuild dependency set.

fallintoplace:
Fixes #20113.

## What changed

- forward delete events from the CLI watcher to the rebuild decision logic instead of dropping them up front
- keep ignoring deleted content files for incremental rebuilds, but still force a full rebuild when the deleted path is a tracked dependency
- add a CLI integration regression test that deletes a config dependency during watch mode and verifies the watcher surfaces the error and recovers on the next rebuild

## Why

The watcher already tracks config/plugin/import dependencies as `fullRebuildPaths`, but delete events were being filtered out before that check happened. That meant deleting a tracked dependency could leave the watcher serving stale output until some other full-rebuild file changed.

## Impact

Deleting a tracked dependency in `--watch` mode now behaves like other full-rebuild dependency changes: the watcher rebuilds immediately instead of silently keeping stale state.

## Validation

- `pnpm exec vitest --root=./integrations run cli/config.test.ts`
- `pnpm exec prettier --check packages/@tailwindcss-cli/src/commands/build/index.ts integrations/cli/config.test.ts`
- `pnpm --dir packages/@tailwindcss-cli lint` currently reports unrelated repo-wide TypeScript errors outside this change

### DIFF
--- a/integrations/cli/config.test.ts
+++ b/integrations/cli/config.test.ts
@@ -299,6 +299,8 @@ test(
     await process.onStderr((m) => m.includes('ENOENT') && m.includes('my-color.js'))
 
     await fs.write('my-color.js', js`module.exports = 'red'`)
+    // Touch the input CSS to force a new full rebuild after the failed reload.
+    // At this point the missing dependency is no longer tracked yet.
     await fs.write(
       'src/index.css',
       css`
--- a/integrations/cli/config.test.ts
+++ b/integrations/cli/config.test.ts
@@ -249,6 +249,73 @@ test(
 )
 
 test(
+  'Config dependency deletion triggers a rebuild error in watch mode',
+  {
+    fs: {
+      'package.json': json`
+        {
+          "dependencies": {
+            "tailwindcss": "workspace:^",
+            "@tailwindcss/cli": "workspace:^"
+          }
+        }
+      `,
+      'index.html': html`
+        <div class="text-primary"></div>
+      `,
+      'tailwind.config.js': js`
+        const myColor = require('./my-color')
+        module.exports = {
+          theme: {
+            extend: {
+              colors: {
+                primary: myColor,
+              },
+            },
+          },
+        }
+      `,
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Closing this in favor of #20137 where a different approach was used. Thanks!

PR #20126, opened 2026-05-29
Use a finite radius for rounded-full

### LINKED ISSUE #20125: Completely out of bounds use of border-radius causing rendering issues on some engines.

Pale Moon

Windows

**Reproduction URL**

https://tailwindcss.com/

**Describe your issue**

Investigating some strange rendering issues noted on sites that use tailwind css, I found out that there are a few common classes in tailwind that use completely out of bounds values for border radius. This causes issues on some layout and rendering engines (notably Goanna, but possibly others) resulting in radial lines going out from the corners instead of rounding element corners. While the symptom is engine-specific, the root cause and potential fallout is not.

Specifically, the following seems to be common in all tailwind css deployments:
```CSS
	.rounded-full {
		border-radius:3.40282e38px
	}
```
Specifying a 10^38 sized radius is absolutely insane, and leads to unspecified behavior. Ideally, CSS pixel values should always remain within `int` number space (<2G). Where did this specific value even come from? I suggest never exceeding 10,000px as a maximum radius meaning "fully circular".

yudin-s:
## Summary

Fixes #20125.

`rounded-full` currently emits `calc(infinity * 1px)`. When production optimization runs through Lightning CSS, that value is serialized as an extremely large finite pixel value, which can cause rendering issues in some engines.

This switches the built-in fallback for `rounded-full` to `9999px`, while still allowing users to override it with `--radius-full` in their theme.

## Test plan

- `pnpm exec vitest run packages/tailwindcss/src/utilities.test.ts -t rounded`
- `pnpm exec vitest run packages/tailwindcss/src/utilities.test.ts`
- `pnpm exec prettier --check packages/tailwindcss/src/utilities.ts packages/tailwindcss/src/utilities.test.ts`
- `git diff --check`
- `pnpm build`

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -11989,7 +11989,7 @@ test('rounded', async () => {
     }
 
     .rounded-full {
-      border-radius: 3.40282e38px;
+      border-radius: 9999px;
     }
 
     .rounded-none {
--- a/packages/tailwindcss/src/utilities.ts
+++ b/packages/tailwindcss/src/utilities.ts
@@ -2340,7 +2340,7 @@ export function createUtilities(theme: Theme) {
         handle: (value) => properties.map((property) => decl(property, value)),
         staticValues: {
           none: properties.map((property) => decl(property, '0')),
-          full: properties.map((property) => decl(property, 'calc(infinity * 1px)')),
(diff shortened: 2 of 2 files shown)


RobinMalfait:
Hey! Going to keep `calc(infinity * 1px)` for now since this is the right intent compared to guessing a big number.

PR #20131, opened 2026-05-29
fix(v4): output static components inside @layer components (issue #15045)

saitejabandaru-in:
### Description

This pull request resolves issue #15045 by correctly placing styles registered via the JavaScript plugin compatibility API `addComponents` inside the static `@layer components` cascade layer rather than `@layer utilities` in Tailwind CSS v4.

### Background

In Tailwind CSS v3, plugins using `addComponents` (such as `daisyui` or `@tailwindcss/forms`) registered custom components whose styles were output inside `@tailwind components` (predefined `@layer components`). This allowed developers to cleanly override component-level styles with utility classes (which are ordered later in `@layer utilities`) in their HTML.

In v4, the compatibility layer defined `addComponents` as a simple alias for `addUtilities`:
```typescript
    addComponents(components, options) {
      this.addUtilities(components, options)
    },
```
This causes component styles to be generated inside the dynamic `@layer utilities` layer. Because they share the same cascade layer as standard utility classes, they are ordered based on property counts and property-overlap heuristics rather than cascade layering. This breaks specificity, making it extremely difficult or impossible to override component classes (like `.btn`, `.card`) with utility classes (like `bg-red-100`) in markup without resorting to `!important` everywhere.

### Changes

1. **Static Component Compilation**: Updated the `addComponents` implementation in `plugin-api.ts` to parse component styles using `objectToAst` and wrap them inside an explicit `@layer components` at-rule, pushing it directly to the root `ast` array (similar to how `addBase` works for global baseline styles):
   ```typescript
   addComponents(components, options) {
     if (referenceMode) return
     let componentNodes = objectToAst(components)
     
     // Prefix all class selectors with the configured theme prefix
     if (designSystem.theme.prefix) {
       walk(componentNodes, (node) => {
         if (node.kind === 'rule') {
           let selectorAst = SelectorParser.parse(node.selector)
           walk(selectorAst, (node) => {
             if (node.kind === 'selector' && node.value[0] === '.') {
               node.value = `.${designSystem.theme.prefix}\\\\:${node.value.slice(1)}`
             }
           })
           node.selector = SelectorParser.toCss(selectorAst)
         }
       })
     }

     featuresRef.current |= substituteFunctions(componentNodes, designSystem)
     let rule = atRule('@layer', 'components', componentNodes)
     walk([rule], (node) => {
       node.src = src
     })
     ast.push(rule)
   }
   ```
2. **Prefix Support**: Properly applied configured theme prefixing to class selectors inside component rules, ensuring backward compatibility with prefix-based configurations.
3. **Tests**: Updated `plugin-api.test.ts` to verify that `addComponents` outputs compiled rules cleanly wrapped inside `@layer components` in the generated CSS, rather than placing them straight into the flat utilities listing.

This is a critical backward compatibility fix that enables a seamless transition to Tailwind CSS v4 for standard component libraries (such as daisyUI, flowbite, tailwindcss-forms, etc.) that rely on cascade layering specificity.

### DIFF
--- a/packages/tailwindcss/src/compat/plugin-api.test.ts
+++ b/packages/tailwindcss/src/compat/plugin-api.test.ts
@@ -4146,7 +4146,7 @@ describe('matchUtilities()', () => {
 })
 
 describe('addComponents()', () => {
-  test('is an alias for addUtilities', async () => {
+  test('adds styles to @layer components', async () => {
     expect(
       await run(
         ['btn', 'btn-blue', 'btn-red'],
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Appreciate the PR, but we're not sure yet if we even want this. While this is a breaking change, it was an intentional breaking change to simplify the system. 

The PR currently is incomplete, but since we're not sure if we even want this, I'm going to close this for now. Thanks!

PR #20196, opened 2026-06-03
Simplify CSS when using utilities that use a `*-0` or `*-1` value

akx:
## Why?

While inspecting a Tailwind-powered site, I noticed selectors like

```css
.inset-0{inset:calc(var(--spacing) * 0)}
.inset-x-0{inset-inline:calc(var(--spacing) * 0)}
.top-0{top:calc(var(--spacing) * 0)}
```

which seem a bit silly, not to mention more complex for the end-user device parsing the CSS.

## Summary

This PR adjusts helpers to not generate `calc(... * 0)` expressions.

https://github.com/tailwindlabs/tailwindcss/pull/19095 does a similar thing on CSS AST, but that doesn't run during the build proper.

## Test plan

Ran `pnpm build && pnpm test && pnpm test:integration`. (Some unrelated tests failed on my machine, hopefully less in CI.)


RobinMalfait:
Thanks! I had a similar change in a local branch the other day when working on that PR you linked! Technically using just `0` isn't always safe because then we often lose the data type. It can also make some calculations that were invalid, valid all of a sudden (e.g.: `0deg * 0rem` is invalid, but a result of `0` would be valid).

That said, we can definitely make some assumptions here. In order for Tailwind CSS to work with the `--spacing` variable, we need to assume that:

1. The `--spacing` value is set (it can be overridden at runtime, but it must still be set)
2. The `--spacing` value must be set to a `<length>` type such as `0.25rem`.

These conditions must be true in order for the framework to function properly.

I'm going to make some (small) changes here:

1. Instead of hardcoding this logic in most places, we can use the `--spacing(…)` helper in the utilities instead. This has the benefit that we only have to tackle this in a single spot, and that CSS-based plugins can make use of this optimization automatically as well.
2. There's a spot where you omitted a CSS variable when the `0` value is used. I'm going to put this back for 2 reasons:
   1. This changes the order in the CSS because fewer properties were used
   2. In the HTML, some other classes could be using this information. Deleting it would make accessing this variables' value invalid.
3. There is another optimization we can do here with the `1` value. `calc(var(--spacing) * 1)` is also a little bit silly, we can replace this with just `var(--spacing)`.

akx:
Good stuff, thanks for taking a look @RobinMalfait!

I was also thinking about the `* 1` case, but decided to leave it out of this PR to test the waters, as it were :)

RobinMalfait (approved):

PR #20201, opened 2026-06-05
Add riscv64 (linux-gnu) prebuilt binary

gounthar:
Adds `riscv64gc-unknown-linux-gnu` to the oxide release matrix so a prebuilt `@tailwindcss/oxide-linux-riscv64-gnu` package gets published. There is no riscv64 binary today, so on riscv64 Linux boards `@tailwindcss/oxide` can't load its native module, and anything pulling Tailwind v4 (the CLI, the Vite plugin, PostCSS) is stuck there.

### Changes
- `.github/workflows/release.yml`: a new matrix entry mirroring the aarch64 cross entry. napi-cross (`@napi-rs/cross-toolchain`) doesn't ship a riscv64 toolchain, so this one uses the system `gcc-riscv64-linux-gnu` plus the matching linker env instead of `--use-napi-cross`, and strips with `riscv64-linux-gnu-strip`.
- `crates/node/package.json`: register `riscv64gc` in `napi.targets` and add the optionalDependency.
- `crates/node/npm/linux-riscv64-gnu`: the new platform package, mirroring `linux-arm64-gnu`.
- `pnpm-lock.yaml`: the new workspace package.

### Verification
- Cross-compile on ubuntu-latest with the recipe in this PR produces `tailwindcss-oxide.linux-riscv64-gnu.node` (ELF 64-bit RISC-V).
- A native build on a real riscv64 board produces the same artifact, loads it (the `Scanner` export is there), and runs.

One thing worth flagging: napi-cross supports arm/arm64/x64/s390x/ppc64le but not riscv64 yet, which is why this entry uses the system cross toolchain rather than `--use-napi-cross` like your other Linux targets. If you'd rather it followed the napi-cross path once that gains riscv64, or want it done another way, happy to adjust.


sharaddh (commented):
fine

RobinMalfait:
Hey, thanks for the PR!

Instead of adding and maintaining prebuilt native binaries for every platform, we've made the `wasm32-wasi` build of `@tailwindcss/oxide` an automatic fallback in #20383. When no native binding matches your OS/CPU, your package manager now installs the wasm build alongside the other optional dependencies, and the loader picks it up automatically.

This is not yet released, but is available as part of the `insiders` version. You can already try it by using:

CLI:
```sh
```
PostCSS:
```sh
npm install tailwindcss@insiders @tailwindcss/postcss@insiders
```
Vite:
```sh
```
webpack:
```sh
npm install tailwindcss@insiders @tailwindcss/webpack@insiders
```

Could you give that a try on your platform and let us know how it goes? If it works well for you, we'd like to close this PR in favor of the wasm fallback since it covers every platform at once. Thanks again!

gounthar:
Tried it on a Banana Pi BPI-F3 (SpaceMiT K1, riscv64, Ubuntu 24.04, Node 22).

The oxide half works. `npm install tailwindcss@insiders @tailwindcss/cli@insiders` pulled in `@tailwindcss/oxide-wasm32-wasi` on its own, no flags, and the loader picks it up:

```
require("@tailwindcss/oxide") -> LOADED in 949 ms; exports: Scanner
```

I threw 200 jsx files at it (16600 lines, 79 distinct utilities) and drove the `tailwindcss` core directly with the candidates coming out of the wasm Scanner. Correct CSS, variants included. So the fallback does what it should on this platform.

Out of curiosity I also built the native binding from main on the same board and dropped it into `node_modules/@tailwindcss/oxide/`, to compare the two on identical input. Output is byte for byte identical, so this is only about speed:

| step | wasm32-wasi | native riscv64 |
|------|-------------|----------------|
| require the binding | 949 ms | 70 ms |
| `new Scanner(...)` | 528 ms | 12 ms |
| `scan()` over 200 files | 2399 ms | 87 ms |

About 30x on the scan. I don't think that changes the call: my fixture is 1.6 MB, which is nothing, and one wasm build you maintain beats a growing list of native targets you don't. Just putting the number somewhere findable in case someone turns up later wondering why their riscv64 watch mode feels sluggish.

The other thing I hit is the CLI, and it isn't oxide:

```
Error: Cannot find module '../lightningcss.linux-riscv64-gnu.node'
Require stack:
- node_modules/lightningcss/node/index.js
```

`@tailwindcss/node` pulls `lightningcss@1.33.0`, which has its own optional deps (11 platforms, no riscv64) and no wasm fallback of its own. `@tailwindcss/vite` and `@tailwindcss/postcss` go through `@tailwindcss/node` too, so all four install recipes stop in the same place here. You almost certainly know this part better than I do, so tell me if I'm reading the dependency graph wrong.

I have a riscv64 PR open over there (parcel-bundler/lightningcss#1268), open since June and not moving much, so it isn't something I'd have you wait on.

Happy for you to close this one. I just wanted the riscv64 situation written down before it does, since "works everywhere" is true for oxide now but not quite yet for the CLI path.

PR #20228, opened 2026-06-12
Serve ESM type declarations to ESM importers of `@tailwindcss/postcss`

### LINKED ISSUE #20219: types: export = in @tailwindcss/postcss contradicts exports.import in package.json
## Problem
TypeScript 6.0 (shipped in Deno v2.8.3) changed `export =` in ESM context from a warning (TS1203) to a hard error (TS1192). This causes `deno check` to fail when importing `@tailwindcss/postcss`.

## Root Cause
`@tailwindcss/postcss@4.1.16`'s `package.json` declares ESM support via `exports.import`, but the type declarations in `dist/index.d.ts` use CJS syntax:

```ts
declare const _default: PluginCreator<PluginOptions>;
export = _default;  // CJS, not ESM export default
```

With `moduleResolution: NodeNext`, TS 6.0 sees `exports.import` -> treats `.d.ts` as ESM -> `export =` in ESM -> TS1192 error.

## Reproduce
Reproducible against stock `tsc` with NodeNext resolution.

## Suggested Fix
Change `export = _default` to `export default _default` to match the `exports.import` ESM declaration.

truffle-dev:
## Summary

Fixes #20219.

The `@tailwindcss/postcss` exports map serves the CJS-shaped `dist/index.d.ts` (`export = _default`) for both the `import` and `require` conditions, while the correct ESM declaration file `dist/index.d.mts` (`export { _default as default }`) is built and published but never referenced. Type checkers that treat the import condition as ESM reject the default import with TS1192 — concretely, `deno check` on Deno 2.8.3+ (which ships TypeScript 6.0) fails on `import tailwindcss from "@tailwindcss/postcss"`.

This splits the export entry into per-condition `types` blocks so ESM importers resolve `index.d.mts` and CJS consumers keep `index.d.ts` — the same shape `@tailwindcss/vite` already uses (`"types": "./dist/index.d.mts"`).

Two notes on the issue as filed:

- Stock `tsc` with NodeNext does **not** reproduce the error (declaration file format follows the file extension and package `type`, not the matched condition), so I didn't take the suggested `export =` → `export default` change in `index.d.ts` — that would break CJS consumers, and the correct ESM declarations already ship.
- `@tailwindcss/node` has the same single-`types` exports shape; happy to follow up there if you want parity.

## Test plan

Against `@tailwindcss/postcss@4.3.0` with the published exports map, then again with only this `package.json` change applied to the installed package:

```sh
# Deno 2.8.3 (TypeScript 6.0), nodeModulesDir: auto
deno check main.ts   # before: TS1192 "Module ... index.d.ts has no default export" — after: passes
```

```sh
# typescript@6.0.0-dev (NodeNext), one ESM importer + one .cts require importer
tsc --noEmit         # passes both before and after (no regression for npm consumers)
```

`@arethetypeswrong/cli`:

| | published 4.3.0 | with this change |
|---|---|---|
| node16 (from ESM) | 🎭 Masquerading as CJS | 🟢 (ESM) |
| node16 (from CJS) | 🟢 (CJS) | 🟢 (CJS) |
| bundler | 🟢 | 🟢 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure explicitly referenced `@source` directories are scanned even when ignored by git ([#20214](https://github.com/tailwindlabs/tailwindcss/pull/20214))
 - Ensure `@source` globs ending in `**/*` preserve dynamic path segments to avoid scanning too many files ([#20217](https://github.com/tailwindlabs/tailwindcss/pull/20217))
 - Canonicalization: don't fold `calc(…)` divisions when the result would require high precision (e.g. `w-[calc(100%/3.5)]` → `w-[calc(100%/3.5)]`, not `w-[28.571428571428573%]`) ([#20221](https://github.com/tailwindlabs/tailwindcss/pull/20221))
+- Serve ESM type declarations to ESM importers of `@tailwindcss/postcss` ([#20228](https://github.com/tailwindlabs/tailwindcss/pull/20228))
 
 ### Changed
 
--- a/packages/@tailwindcss-postcss/package.json
+++ b/packages/@tailwindcss-postcss/package.json
@@ -24,9 +24,14 @@
   },
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.mjs",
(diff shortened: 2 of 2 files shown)


RobinMalfait (approved):
Thanks!

PR #20223, opened 2026-06-12
docs: fix 'preceeded' typo in candidate.ts comment

Rohan5commit:
## Summary

Fixes a misspelling in a comment: 'preceeded' → 'preceded'.

## Test plan

Comment-only change. No tests needed.

PR #20234, opened 2026-06-14
Fix typo in css-parser.test.ts: 'with with' → 'with'

Rohan5commit:
Fixes a duplicate word typo in `packages/tailwindcss/src/css-parser.test.ts`.


RobinMalfait:
As mentioned before, don't do this. https://github.com/tailwindlabs/tailwindcss/pull/20223#issuecomment-4688335648

PR #20268, opened 2026-06-22
fix(oxide): extract classes adjacent to [% ... %] template tags (#20233)

### LINKED ISSUE #20233: Class names adjacent to [ and ] delimiters not extracted
[tailwind-tt-repro.zip](https://github.com/user-attachments/files/28919537/tailwind-tt-repro.zip)

  v4.1.8

  tailwindcss CLI (standalone binary)

  N/A — using the standalone Tailwind CLI binary, not Node.js (22.22.1)

  Vivaldi

  Linux

  **Reproduction URL**

  Minimal reproduction attached as a zip: [tailwind-tt-repro.zip]

  template.tx:
```
  <div class="[% IF $is_open %]bg-white/40[% ELSE %]bg-white/10[% END %]"></div>
  <div class="bg-blue-500"></div>
```

  app.css:
```
  @import "tailwindcss";
  @source "./template.tx";
```

  Run `tailwindcss -i app.css -o out.css` — bg-white/40 and bg-white/10 are absent from out.css, but bg-blue-500 is present.

  **Describe your issue**

  The oxide scanner does not treat [ and ] as word boundaries, so class names immediately adjacent to [% %]-style template tags are not extracted.

  Template engines that use [% %] delimiters include Template Toolkit (Perl), Text::Xslate (Perl), Template::Tiny (Perl), and ExpressionEngine (PHP). In these templates, conditional
  class names look like:

  `<div class="[% IF cond %]bg-white/40[% ELSE %]bg-white/10[% END %]">`

  The scanner sees `%]bg-white/40[%` as one token because `[` and `]` are not boundary characters, so neither class is extracted.

  By contrast, ERB/EJS templates using <% %> work correctly because < and > are already treated as boundaries. The {/} boundary fix landed in #17001. Adding [/] as boundaries would fix this consistently.

  Workarounds (both are unpleasant):
  1. `@source inline("bg-white/10 bg-white/20 bg-white/40 ...")` — must be kept in sync manually
  2. A pre-processing script that strips [% ... %] tags to a clean file before scanning — requires a parallel watcher process alongside tailwindcss --watch

ther12k:
Fixes #20233

## Problem
Template Toolkit, Text::Xslate, ExpressionEngine, and similar template engines use `[% ... %]` as their tag delimiters. The scanner was dropping classes adjacent to these tags: neither `[` nor `]` was a valid boundary character, so a candidate like `bg-white[%` failed the after-boundary check even after the tag was identified.

```
[% IF ... %]bg-white/40[% ELSE %]bg-white/10
              ^^^^^^^^^^                   -- not extracted (after-boundary '[' invalid)
                            ^^^^^^^^^^^^^  -- not extracted (before-boundary ']' invalid)
```

### 1. 3-layer `[% ... %]` skip in `extract()`

Implemented in the **outer extractor loop**, not inside the candidate machine, because the inner utility machine can terminate a candidate whose last char immediately precedes `[`, advancing the cursor PAST `[` before any in-machine `[%` check could see it.

- **Top-of-loop `[%` check in `extract()`** — handles input-start and post-whitespace arrivals (cursor enters an iteration parked on `[`).
- **Post-match `[%` check after `candidate_machine.next()` returns** — handles the common case where `next()` returns `Done` with cursor.pos on the `[` that opens a template tag. Uses `continue` to skip the bottom `cursor.advance()` since the tag-skip already moved past `[`.
- **`[%` check in `extract_sub_candidates()`** — prevents spawning an inner `CandidateMachine` at `%` (after `[`) that would otherwise re-extract template keywords like `if`, `cond`, `end`.

The skip is **non-greedy** (first `%]` ends the tag). Nested `[% ... %]` inside the body isn't a thing in any of the target template languages. If no `%]` is found (e.g. a typo), the normal candidate extraction path takes over so the rest of the input still parses.

### 2. Boundary additions (minimum required)

Adding `]` as a valid before-boundary and `[` as a valid after-boundary is required for candidates adjacent to template tags to extract:

- `bg-white[%` — before-boundary `e` (Common), after-boundary `[`. `[` is now After-class.
- `bg-white[% end %]` — same as above.
- `bg-white]%` — before-boundary `]`. `]` is now also valid as before-boundary (it already lived in After for `[class.foo]` syntax, so we special-case the check rather than overwrite the enum).

### 3. Carve-out in `has_valid_boundaries`

`[` as after-boundary is **rejected** when the candidate span itself already contains `[`. This preserves the "one arbitrary value per utility" semantics: `bg-[red][blue]` still extracts nothing.

## Tests

Added `test_template_toolkit_syntax` (9 cases) in `crates/oxide/src/extractor/mod.rs`:

- Issue reproduction: `[% IF ... %]bg-white/40[% ELSE %]bg-white/10`
- Lowercase keywords
- Static class adjacent to a `[% %]` block
- `[color:red]` still extracts (arbitrary property regression guard)
- `bg-[red][blue]` still extracts nothing (carve-out regression guard)
- `bg-red-500[% if x %]` — candidate before tag, no trailing
- `bg-red-500[% x %]bg-blue-500[% y %]text-white` — multi-tag loop
- `[% if x %] bg-red-500` — top-of-loop check at input start
- `bg-red-500 [% if x %]` — post-whitespace arrival path

## Verification

- `cargo test -p oxide` → **115 passed, 0 failed, 12 ignored** (1 new test added, was 114 baseline)
- `cargo fmt --check` → clean
- `cargo clippy --workspace --all-targets` → 9 errors, **0 new** from this diff (verified via `git diff origin/main..HEAD` — my changes don't introduce new patterns). Pre-existing clippy on `origin/main` (verified with the same diff on the parent commit).
- `cargo test` does not exercise this path under the default test config; the new test in `mod.rs` is a unit test that runs under `cargo test -p oxide`.

## Scope

`fix(oxide):` — all changes are isolated to `crates/oxide/src/extractor/`. No public API changes, no breaking changes.

## Files changed

```
 CHANGELOG.md                                    |   1 +
 crates/oxide/src/extractor/boundary.rs          |  42 +++++++-
 crates/oxide/src/extractor/candidate_machine.rs |  10 ++
 crates/oxide/src/extractor/mod.rs               | 137 +++++++++++++++++++++++-
 4 files changed, 186 insertions(+), 4 deletions(-)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `@tailwindcss/vite` doesn't crash during HMR when scanned files or directories are deleted ([#20259](https://github.com/tailwindlabs/tailwindcss/pull/20259))
 - Ensure `text-[--spacing(…)]` generates `font-size` instead of `color` ([#20260](https://github.com/tailwindlabs/tailwindcss/pull/20260))
 - Prevent `@source` patterns from scanning unrelated sibling files and folders ([#20263](https://github.com/tailwindlabs/tailwindcss/pull/20263))
+- Ensure classes adjacent to `[% … %]` template tags (Template Toolkit, Text::Xslate, ExpressionEngine, etc.) are extracted by skipping past the tag in the scanner ([#20233](https://github.com/tailwindlabs/tailwindcss/issues/20233))
 
 ## [4.3.1] - 2026-06-12
 
--- a/crates/oxide/src/extractor/boundary.rs
+++ b/crates/oxide/src/extractor/boundary.rs
@@ -4,7 +4,16 @@ use crate::extractor::Span;
 
 #[inline(always)]
 pub fn is_valid_before_boundary(c: &u8) -> bool {
-    matches!(c.into(), Class::Common | Class::Before)
+    // `]` is also a valid before-boundary so that candidates immediately
+    // following `[% ... %]` template tags (Template Toolkit, Text::Xslate,
+    // ExpressionEngine, etc.) are extracted. Symmetric with `]` already being
+    // a valid after-boundary for `[class.foo]`-style attributes.
+    //
+    // The classification macro generates one class per byte; `]` lives in
+    // `After` (for `[class.foo]`), so we special-case it here rather than
+    // losing the After-class semantics by listing it in `Before` as well.
+    // https://github.com/tailwindlabs/tailwindcss/issues/20233
+    matches!(c.into(), Class::Common | Class::Before) || *c == b']'
 }
 
 #[inline(always)]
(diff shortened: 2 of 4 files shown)


RobinMalfait:
Hey! Thanks for the PR, but I believe this is the wrong approach. This makes things slower for everyone even if you don't use the special `[%` boundary characters, and that's not what you want.

You also mention `[% ... %]` often, but that's not the problem, the problem is `%] ... [%`.

Please don't open PRs that are purely vibe coded. Thanks.

PR #20276, opened 2026-06-25
feat(oxide): add OpenHarmony (aarch64-unknown-linux-ohos) prebuilt binary support

social4hyq:
Add `aarch64-unknown-linux-ohos` as a build target so `@tailwindcss/oxide`
loads natively on OpenHarmony/HarmonyOS devices without falling back to WASM.

## Changes

- **`crates/node/package.json`** — add `aarch64-unknown-linux-ohos` to `napi.targets`; add `@tailwindcss/oxide-openharmony-arm64` to `optionalDependencies`
- **`crates/node/npm/openharmony-arm64/package.json`** — new platform subpackage (`"os": ["openharmony"], "cpu": ["arm64"]`)
- **`pnpm-lock.yaml`** — updated for new workspace package
- **`.github/workflows/release.yml`** — new `build-ohos` job + wire into `release` job `needs` and `Move artifacts` step

## Why no `index.js` changes

`@napi-rs/cli@3.7.0` (already used by this project) natively maps `aarch64-unknown-linux-ohos` → `process.platform === 'openharmony'` with package suffix `openharmony-arm64`. The generated `index.js` will contain the correct branch automatically — no manual edits needed.

## CI design decisions

- **Separate `build-ohos` job (not added to the main matrix)** — the existing build matrix entries use `--use-napi-cross` (zig) which doesn't support the OHOS sysroot; OHOS needs its own linker setup
- **`Boshen/setup-ohos-sdk@v1.0.0`** — same action used by rolldown and lightningcss for their OHOS builds; downloads only the native sysroot instead of the full 2.3 GB SDK
- **Shell wrapper linker** — configured via `CARGO_TARGET_*_LINKER` env var; no `.cargo/config.toml` changes
- **`rustflags: -C link-arg=-fuse-ld=lld -C link-arg=-lm`** — required for OHOS targets (lld is the only supported linker; `-lm` resolves missing math symbols)

## Verification

Built and runtime-tested on HarmonyOS HongMeng Kernel 1.12.0 (aarch64) with LLVM 21.1.8:

```bash
cargo build --lib -p tailwind-oxide --release --target aarch64-unknown-linux-ohos  # ✓
```

Runtime verified via `opencode` (which embeds `@tailwindcss/oxide`) on-device ✓

CI: [social4hyq/tailwindcss — Test OHOS build — run #28161581738](https://github.com/social4hyq/tailwindcss/actions/runs/28161581738) — **all steps passed**, artifact `tailwindcss-oxide.openharmony-arm64.node` (2.6 MB) produced ✓


social4hyq:
Thanks for the review! Addressing all comments in commit 7384552:

**CodeRabbit — `dtolnay/rust-toolchain` pinning** → pinned to commit SHA `67ef31d5b988238dd797d409d6f9574278e20537`.

**CodeRabbit — `--ignore-scripts`** → removed from the `build-ohos` install step.

**Greptile — `version-packages.mjs` syncedWorkspaces** → real bug, fixed by adding `crates/node/npm/openharmony-arm64` to the hardcoded list so it versions in sync with `@tailwindcss/oxide` on every release.

**Greptile — Generated loader may miss OHOS** → `@napi-rs/cli@3.7.0` (already used by this project) natively maps `aarch64-unknown-linux-ohos` to `process.platform === 'openharmony'` with package suffix `openharmony-arm64`, so the generated `index.js` is correct automatically. Confirmed by [CI run](https://github.com/social4hyq/tailwindcss/actions/runs/28161581738): artifact is named `tailwindcss-oxide.openharmony-arm64.node`.

**Greptile — Artifact rename** → not needed; `napi build --platform` derives the filename from the NAPI-RS platform suffix, output is already `tailwindcss-oxide.openharmony-arm64.node`, matching the platform package `main` field.

**Removed `test-ohos-build.yml`** → temporary verification workflow, not for upstream.

RobinMalfait:
Hey! I appreciate the PR, but going to skip this for now. We can add support for it if more people _need_ support for this. Right now I'm a bit skeptical, especially becaues of the 3rd party GitHub actions that are being used. 

We can revisit this when more people need support for ohos. Thanks!

SimboDu123:
oh no!
Hope to quickly adapt to OpenHarmony!

social4hyq:
Hi @RobinMalfait, thanks for taking a look at this PR. I'd like to respectfully ask you to reconsider, with some additional context:

**1. There is real, measurable demand.**

The OpenHarmony platform packages of peer projects are seeing substantial usage:

| Package | Monthly Downloads |
|---------|------------------|
| `@rollup/rollup-openharmony-arm64` | **12.6M** |
| `@rolldown/binding-openharmony-arm64` | **5.6M** |

These are not marginal numbers. OpenHarmony/HarmonyOS has a growing developer ecosystem.

**2. `Boshen/setup-ohos-sdk` is the community standard, not an unknown third-party.**

This same GitHub Action is already used by **Rolldown** ([rolldown/rolldown#5258](https://github.com/rolldown/rolldown/pull/5258)), **Rollup** ([rollup/rollup#6081](https://github.com/rollup/rollup/pull/6081)), **OXC** ([oxc-project/oxc#16105](https://github.com/oxc-project/oxc/pull/16105)). Boshen is a well-known ecosystem contributor and an oxc core member — this action is production-proven across multiple projects.

**3. All peer projects listed above have already merged their OHOS support, and the Tailwind PR is technically sound** (reviewed and greenlit by both CodeRabbit and Greptile, CI verified). Without native OHOS binaries, Tailwind CSS users on that platform can only fall back to WASM, which is significantly slower.

Would you be open to revisiting this?

RobinMalfait:
@social4hyq I understand that, but the problem is that we get asked to add support for a bunch of different platforms without knowing _why_ this is needed. There are no bug reports saying that Tailwind CSS is currently not working on these platforms, or if it is working that it is slower compared to a native version (and how much slower).

Adding support for every platform that exists is not really possible, because what's actually happening here is that we have to support these platforms forever. When 3rd party tools stop providing support for a platform, then we have to support it ourselves.

> Without native OHOS binaries, Tailwind CSS users on that platform can only fall back to WASM, which is significantly slower.

Do you have some data on this? What does "significantly slower" mean on your machine for example?

> The OpenHarmony platform packages of peer projects are seeing substantial usage:

I believe those numbers, I just need to understand what's currently broken or how slow things are.

social4hyq:
@RobinMalfait Fair questions — I went and actually measured this instead of just asserting "slower". Here's what I found on a real HarmonyOS device (HongMeng Kernel 1.12.0, aarch64):

**It's not "N times slower" — the WASM fallback doesn't actually run here.**

I pulled the official `@tailwindcss/oxide-wasm32-wasi@4.1.11` from npm (unmodified) and tried to load it under Node.js 26 on-device: it fails deterministically (3/3 runs) with `UVWASI_EACCES` during `WASI` construction, before any scan logic runs. Root cause: the generated wrapper hardcodes `preopens: { '/': '/' }`, and OpenHarmony's sandbox denies opening `/` directly — this isn't something Tailwind can work around without patching `@napi-rs/wasm-runtime`'s codegen.

<details>
<summary>Repro + full error</summary>

```js
// repro.cjs
const wasm = require('@tailwindcss/oxide-wasm32-wasi/tailwindcss-oxide.wasi.cjs');
const s = new wasm.Scanner({ sources: [] });
console.log(s.scanFiles([{ content: '<div class="flex p-4"></div>', extension: 'html' }]));
```

```
$ node repro.cjs
node:wasi:90
    const wrap = new _WASI(args, env, preopens, stdio);
                 ^

Error: UVWASI_EACCES, uvwasi_init
    at new WASI (node:wasi:90:18)
    at Object.<anonymous> (.../tailwindcss-oxide.wasi.cjs:19:16)
    ...
  errno: 2,
  code: 'UVWASI_EACCES',
  syscall: 'uvwasi_init'
}
```

The offending lines in the generated wrapper (`crates/node/npm/wasm32-wasi/tailwindcss-oxide.wasi.cjs`):

```js
const __rootDir = __nodePath.parse(process.cwd()).root   // => '/'
const __wasi = new __nodeWASI({
  version: 'preview1',
  env: process.env,
  preopens: { [__rootDir]: __rootDir },   // preopens: { '/': '/' }
})
```

</details>

For contrast, the native binding (built from this PR's target, self-signed, `require()`'d directly — no upstream code changes needed beyond this PR) scans 300 synthetic files (15 classes each) in **3.5–15.8ms** across 5 runs.

So today, a Tailwind user on OpenHarmony isn't choosing between "fast native" and "slower WASM" — they're choosing between "native" and "doesn't load at all". That's the concrete bug report you asked for; there just isn't an upstream issue open for it yet because anyone hitting it currently has no native option to compare against, so it just looks like Tailwind/oxide "doesn't support OpenHarmony" rather than "the fallback is broken here."

On the "supporting this forever" concern — totally reasonable, but the ongoing cost here should be close to zero: this PR only adds an isolated `openharmony` branch in `requireNative()`/`getPlatformPackageName()` (doesn't touch any existing Linux/Mac/Windows path), and the target/CI setup mirrors exactly what Rolldown, Rollup, and OXC already maintain with the same `Boshen/setup-ohos-sdk` action — so it's a well-trodden, low-maintenance pattern at this point, not something you'd be pioneering alone.

RobinMalfait:
Instead of adding and maintaining prebuilt native binaries for every platform, we've made the `wasm32-wasi` build of `@tailwindcss/oxide` an automatic fallback in #20383. 

This also includes a fix for the `UVWASI_EACCES` error you ran into when manually checking out the wasm version.

This is not yet released, but is available as part of the `insiders` version. You can already try it by using:

CLI:
```sh
```
PostCSS:
```sh
npm install tailwindcss@insiders @tailwindcss/postcss@insiders
```
Vite:
```sh
```
webpack:
```sh
npm install tailwindcss@insiders @tailwindcss/webpack@insiders
```

Could you give that a try on your platform and let us know how it goes? If it works well for you, we'd like to close this PR in favor of the wasm fallback since it covers every platform at once. Thanks again!

PR #20294, opened 2026-07-01
Add linux-loong64 prebuilt binary

wojiushixiaobai:
Add support for the LoongArch 64-bit architecture.


RobinMalfait:
Hey! I appreciate the PR, but can you share more _why_ we need to add support for this?

I would recommend to open an issue first if you are running into problems/bugs, or open a discussion if you want to talk about why this is needed.

PR #20310, opened 2026-07-05
Ensure `@tailwindcss/postcss` rebuilds when the input CSS changes but its mtime is unchanged

### LINKED ISSUE #20307: `@tailwindcss/postcss`: compiler cache ignores changed input CSS when `from`'s mtime is unchanged — stale output with any preprocessor (Sass) pipeline
**Tailwind version:** 4.3.2 (`tailwindcss` + `@tailwindcss/postcss`, latest at time of filing)
**Node:** 22.x · reproduces on macOS and Linux

## Summary

`@tailwindcss/postcss` keeps a per-entry compiler cache keyed by `opts.from`. On each run it decides between the incremental and full rebuild paths by stat-ing the mtimes of **only** `[...fullRebuildPaths, from]` — i.e. the entry file on disk plus Tailwind's own resolved `@import`/`@config`/`@plugin` graph. The actual **input CSS handed to the plugin is never consulted**. When none of those mtimes changed, the plugin discards the fresh input root and re-emits the cached AST (`root.removeAll(); root.append(cachedPostCssAst.clone().nodes)`).

This assumption — that the input is a pure function of `from`'s own mtime plus Tailwind-visible imports — breaks in any pipeline where the input is *derived*: most commonly Sass, where `entry.scss` inlines dozens of `@use`d partials before PostCSS runs. Editing a partial re-runs Sass and produces different CSS for the same unchanged-on-disk `from` → Tailwind silently serves the previous compile. All changes are lost, not just Tailwind-relevant ones, because the whole root is replaced by the cached AST.

## Minimal reproduction (no Sass needed — 15 lines)

```js
// repro.mjs — npm i postcss tailwindcss@4.3.2 @tailwindcss/postcss@4.3.2
import postcss from 'postcss';
import tailwindPostcss from '@tailwindcss/postcss';
import { writeFileSync } from 'node:fs';

writeFileSync('entry.css', '/* on-disk stand-in; content irrelevant, mtime never changes */');
const processor = postcss([tailwindPostcss()]);
const compile = css => processor.process(css, { from: 'entry.css' }).then(r => r.css);

const first = await compile('@import "tailwindcss/utilities.css"; .first { @apply underline; }');
const second = await compile('@import "tailwindcss/utilities.css"; .second { @apply font-bold; }');

console.log('first compile contains .first  :', first.includes('.first'));   // true
console.log('second compile contains .second:', second.includes('.second')); // false  ❌
console.log('second === first (stale)      :', second === first);            // true   ❌
```

Actual output on pristine 4.3.2:

```
first compile contains .first  : true
second compile contains .second: false
second === first (stale)      : true
```

## Real-world impact

Angular 21 dev server (`@angular/build`, esbuild) + Sass + `.postcssrc.json` with `@tailwindcss/postcss`: the global stylesheet `styles.scss` `@use`s a tree of partials containing `@apply`/`theme()`. Editing any partial triggers a correct Angular rebuild (dart-sass re-runs, fresh CSS goes into PostCSS), but the served bundle stays byte-identical until the developer touches `styles.scss` itself or restarts the dev server. Verified live: rebuild fires within 1s of the edit, output hash unchanged, new rule absent; `touch styles.scss` (whose mtime IS checked) flushes it through. The same applies to webpack sass-loader → postcss-loader chains and any other preprocessor-in-front-of-postcss setup.

Angular-side watching/invalidation was ruled out (sass `loadedUrls` → esbuild watchFiles include the partials; its stylesheet cache invalidates correctly — the fresh input demonstrably reaches the plugin).

## Suggested fix

Include an input fingerprint in the cache-validity check: store (a hash of) the incoming root's CSS per cache entry and take the existing `"full"` rebuild path when it differs from the previous compile. We've been running exactly that as a 2-line local patch of `dist/index.{js,mjs}` — insert after the `let isInitialBuild = context.compiler === null` equivalent:

```js
let inputCss = root.toString();
let inputChanged = context.previousInputCss !== undefined && context.previousInputCss !== inputCss;
context.previousInputCss = inputCss;
// ...
let rebuildStrategy = inputChanged ? 'full' : 'incremental';
```

Single-pass production builds are unaffected (the warm-cache path never triggers); the `root.toString()` cost is only paid on watch-mode recompiles of that entry.

Precedent: the 4.3.2 changelog fixed the analogous blind spot for the CLI watcher ("Ensure `@tailwindcss/cli` in `--watch` mode rebuilds when the input CSS file changes") — the PostCSS plugin needs the equivalent for *content* arriving via `process()` rather than from disk.

lazerg:
## Summary

`@tailwindcss/postcss` chooses between an incremental and a full rebuild by comparing the mtimes of the entry file and its resolved `@import`/`@config`/`@plugin` graph. It never looks at the input CSS itself, so when that CSS is produced by an upstream tool (e.g. Sass) and passed to the plugin via `process()`, it can change while the `from` file's mtime stays the same. The plugin then re-emits its previously cached output and silently drops the change.

This stores the input CSS per cache entry and takes the existing full rebuild path when it differs from the previous compile, mirroring the fix the CLI watcher already has for changed input files.

Fixes #20307

## Test plan

- Added a regression test in `packages/@tailwindcss-postcss/src/index.test.ts` that compiles two different inputs for the same on-disk `from` file (unchanged mtime) and asserts the second compile reflects the new CSS.
- Confirmed it fails on `main` (the second compile returns the stale first output) and passes with the fix.
- Ran the `@tailwindcss/postcss` package tests (all green) and checked formatting with Prettier.


lazerg:
Pushed a follow-up for both nitpicks, swapped `root.toString()` for `root.source.input.css` so it's not re-serializing the AST on every rebuild, and merged the two `run()` test helpers into one shared function.

RobinMalfait (approved):
Thanks!

PR #20313, opened 2026-07-08
Fix: Escape regex special characters in dynamic RegExp construction

vibemasshq-dev:
Hi Tailwind team 👋 
While auditing this repo with VibeMass (an AI code auditing tool), i spotted a minor but genuine issue at line 1376 in `packages/tailwindcss/src/canonicalize-candidates.ts`. 

### The Issue
`variable` extracted from arbitrary CSS values like `text-[color:var(--my-color)]` is passed directly into a RegExp constructor without escaping:

```typescript
// Current
let r = new RegExp(`var\\(${variable}[,)]\\s*`, 'g')

### DIFF
--- a/packages/tailwindcss/src/canonicalize-candidates.ts
+++ b/packages/tailwindcss/src/canonicalize-candidates.ts
@@ -1373,7 +1373,8 @@ function allVariablesAreUsed(
   walk(ValueParser.parse(value), (node) => {
     if (node.kind === 'function' && node.value === 'var') {
       let variable = node.nodes[0].value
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Thanks for the PR, but please add a test that proves that this change is required.

You also mention that `text-[color:var(--my-color)]` is passed directly to the regex, that's not true. It's the `--my-color` part that is passed to the regex, nothing else. Feel free to re-open the PR if you have a realistic test that proves that this is necessary.

PR #20317, opened 2026-07-09
Preserve zero spacing as a length in custom properties

### LINKED ISSUE #20315: --spacing(0) now resolves to a unitless 0, breaking downstream calc()
related https://github.com/cosscom/coss/issues/803

From 4.3.1 onward, --spacing(0) (and *-0 spacing utilities) compile to a bare unitless 0 instead of calc(var(--spacing) * 0). Written straight into a property that's fine, but when the value is assigned to a custom property and later used inside a calc() alongside a percentage or length, the type silently changes from <length> to <number>: something that used to resolve to a valid zero-length now resolves to a unitless 0, so calc(100% + var(--x)) becomes calc(100% + 0) — invalid — and the whole declaration is silently dropped. This broke previously-working CSS on upgrade. Notably 4.3.1 also shipped #20127, which treats calc(1rem + 0) as an invalid expression, so the tool already recognizes this shape as broken — could the zero case fold to 0px (a length) instead of unitless 0 to keep it type-safe?

sjh9714:
## Summary

Fixes #20315.

`--spacing(0)` was optimized to unitless `0` even when assigned to a custom property. That changes the value type from a length to a number when the variable is later used in `calc()`, so this keeps direct declarations compact while emitting `0px` for custom-property assignments.

## Test plan

- `pnpm install --frozen-lockfile`
- `pnpm exec vitest run packages/tailwindcss/src/css-functions.test.ts --hideSkippedTests`
- `pnpm exec prettier --check packages/tailwindcss/src/css-functions.ts packages/tailwindcss/src/css-functions.test.ts`

I also ran the same Vitest command after adding the regression test but before the fix, and confirmed it failed with `--offset: 0` instead of `--offset: 0px`.

### DIFF
--- a/packages/tailwindcss/src/css-functions.test.ts
+++ b/packages/tailwindcss/src/css-functions.test.ts
@@ -128,6 +128,28 @@ describe('--spacing(…)', () => {
       `)
     })
 
+    test('--spacing(…) preserves a length when zero is assigned to a custom property', async () => {
+      expect(
+        await compileCss(css`
+          @theme {
+            --spacing: 0.25rem;
+          }
+
+          .foo {
+            --offset: --spacing(0);
+            width: calc(100% + var(--offset));
+          }
+        `),
+      ).toMatchInlineSnapshot(`
+        "
+        .foo {
+          --offset: 0px;
+          width: calc(100% + var(--offset));
+        }
+        "
+      `)
+    })
+
     test('--spacing(…) optimizes the output when the input is `0` (with an inlined theme value)', async () => {
       expect(
         await compileCss(css`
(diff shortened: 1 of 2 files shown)


RobinMalfait:
Hey! Thanks for the PR, but I fixed it in a different way as part of #20319.

Your PR does solve the issue where you use `--foo: --spacing(0)`, but it doesn't solve the issue if you use `--spacing(0)` in a `calc(…)` directly where we expect a `<length>` to be used instead of a `<number>`. E.g.: `calc(100% - --spacing(0))` without optimizations is correct, with a `0` it becomes invalid, with `0px` it's valid again. This PR would produce a `0` instead of `0px` if you use `--spacing(0)` directly.

PR #20321, opened 2026-07-11
docs: add theming guide links to README

deepshekhardas:
Added links to customizing colors and spacing docs for theming.


RobinMalfait:
Thanks for the PR, but we already link to the documentation so this seems unnecessary.

PR #20323, opened 2026-07-12
Avoid full reload in `@tailwindcss/vite` when editing a scanned file that is not a loaded module

### LINKED ISSUE #20320: @tailwindcss/vite forces full page reload when editing a scanned .tsx that isn't a loaded module (Vite 8)

v4.2.4 (`tailwindcss` and `@tailwindcss/vite` both 4.2.4)

Vite 8.0.11, via `@tailwindcss/vite` 4.2.4

v24.12.0

N/A — reproduced from the dev-server terminal (browser-agnostic). The reload is
also visible in any browser as the page's counter resetting to 0.

macOS 26.5.1 (25F80)

**Reproduction URL**

https://github.com/neemski/tailwindcss-vite-full-reload-repro

`npm install && npm run dev`, click the button counter a few times,  then (without clicking "go to route B") edit `src/DeepComponent.tsx` and watch the terminal (and the counter reset to 0). Full details, root-cause analysis, and a verified fix are in the repo's README.

**Describe your issue**

Editing a `.tsx` source file that Tailwind has **scanned for class candidates
but that Vite has not yet loaded as a JS module** makes `@tailwindcss/vite`'s
`hotUpdate` hook send `{ type: 'full-reload' }` — a full page reload instead of
a Fast Refresh update. No CSS-like file is involved; the changed file is plain
`.tsx`.

The realistic way a file gets into that "scanned but not loaded" state is
**code splitting**. Vite lazy-transforms modules: a `lazy(() => import('./RouteB'))`
reference is resolved, but `RouteB`'s contents (and anything it imports) aren't
transformed until that chunk is fetched — i.e. until you navigate there. So any
component reachable only *through* an un-fetched split boundary has no JS module
in the graph yet, only Tailwind's scan stub. Editing it while you're on a
different route reloads the whole app and loses all client state. With
route-level splitting ([`React.lazy`](https://react.dev/reference/react/lazy), [TanStack Router `autoCodeSplitting`](https://tanstack.com/router/v1/docs/guide/automatic-code-splitting), [Next.js
dynamic imports](https://nextjs.org/docs/pages/guides/lazy-loading)), every un-visited route's subtree is in this state, so it fires
on a large fraction of edits.

**Root cause.** `hotUpdate` has a fallback meant for scanned content that is not
a JS module (e.g. `.html`/`.php` templates), which fires a `full-reload` when:

```js
modules.length > 0 &&
modules.every((mod) => mod.type === 'asset' || mod.id === undefined)
```

Tailwind's scanner registers every scanned source file via `addWatchFile`, which
creates an `asset`-type module-graph node with `id === undefined`. So the
discriminator is just whether the file is *also* present as a real, loaded JS
module:

| State of the file                                              | Client module graph for the file      | Guard  | Result         |
| -------------------------------------------------------------- | ------------------------------------- | ------ | -------------- |
| loaded JS module (imported + transformed)                      | `[{js, defined}, {asset, UNDEFINED}]` | bails  | HMR update ✅  |
| scanned only — not yet transformed (behind an un-fetched split)| `[{asset, UNDEFINED}]`                | passes | full-reload ❌ |

Because the `{js, defined}` node is missing for a file Vite hasn't transformed,
the `.every(...)` check is satisfied and Tailwind treats real `.tsx` source as an
external template, forcing a reload. The payload is sent directly through
`environment.hot.send`, bypassing Vite's logger, so nothing prints in the
terminal — the repro wraps `hot.send` to make it observable.

(A component *directly* `lazy()`-imported by a loaded module does **not**
reproduce — Vite pre-resolves that one hop, so it gets a `{js, defined}` node.
It's components one or more levels deeper inside the un-fetched subtree that hit
this.)

**Not plugin-specific.** Reproduces with `@vitejs/plugin-react-swc`,
`@vitejs/plugin-react` (Babel), or no React plugin at all. A React plugin is
included in the repro only so Fast Refresh is active — that is the baseline this
bug violates.

The fix direction from #19904 (exempting CSS-like files) would not cover this,
since the affected files are `.tsx`. Early-returning from `hotUpdate` for
JS/TS module files stops the spurious reload while newly introduced utility
classes still apply live through the normal `css-update` path (verified against
the repro):

```js
hotUpdate({ file }) {
  if (/\.[cm]?[jt]sx?(?:\?.*)?$/.test(file)) return;
  // …existing fallback…
}
```

This suggests the fallback should exempt any file that resolves to a real JS
module, rather than only CSS-like files.

UditDewan:
## Summary

Fixes #20320

In dev mode, `@tailwindcss/vite`'s `hotUpdate` hook has a fallback that sends a `full-reload` for files that Tailwind scans but that aren't handled by Vite or any plugin (e.g. `.php` or `.html` templates). Without it, edits to those files wouldn't refresh the page at all.

However, the check ("every module for this file is an `asset` and/or has no id") also matches JS/TS source files that Tailwind has scanned but that Vite hasn't loaded as a module yet — lazy routes, unvisited chunks, and so on. The scanner's `addWatchFile` call creates an asset-type module-graph node with `id === undefined` for every scanned file, and for a not-yet-loaded `.tsx` there is no real JS node next to it, so the fallback fires and the whole page reloads, throwing away all client state. In apps with route-level code splitting this happens on almost every edit to a not-currently-loaded component.

This change exempts JS/TS module files (`.js`, `.jsx`, `.ts`, `.tsx`, `.mjs`, `.cjs`, `.mts`, `.cts`) from the fallback. Vite handles these files itself whenever they are actually loaded, and for scanned-but-unloaded files any new candidates still flow through the regular CSS update, so nothing is lost by skipping the reload:

- Editing a scanned-but-unloaded `.ts` file now results in a targeted `{"type":"update"}` for the Tailwind stylesheet instead of `{"type":"full-reload"}`, and the generated CSS includes the new candidates.
- Editing a scanned external `.php` file still triggers a full reload as before.

## Test plan

- Added an integration test that boots a Vite 8 dev server with a `src/unloaded.ts` file that is scanned but never imported. A small fixture plugin wraps `environment.hot.send` to log every HMR payload. The test edits `unloaded.ts` and asserts that an `update` payload is sent (not a `full-reload`) and that the stylesheet regenerates with the new candidate.
- Verified both directions manually against a minimal reproduction (Vite 8.1.4): before the change, editing the unloaded file logs `HOT_SEND {"type":"full-reload"}`; after the change it logs `HOT_SEND {"type":"update", ...}` for `/src/index.css` and the CSS contains the new candidate. Editing a scanned `.php` file still logs `HOT_SEND {"type":"full-reload"}`.

```
pnpm exec vitest run --root=./integrations vite/index.test.ts -t "scanned module files that are not loaded do not trigger a full reload"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure CSS nesting is handled even when Lightning CSS isn't run, like in `@tailwindcss/browser` and Tailwind Play ([#20124](https://github.com/tailwindlabs/tailwindcss/pull/20124))
 - Prevent achromatic theme colors from shifting hue in `color-mix(…)` with polar color spaces like `oklch` ([#19830](https://github.com/tailwindlabs/tailwindcss/issues/19830))
 - Ensure `--spacing(0)` is optimized to `0px` instead of `0` so it remains a `<length>` when used in `calc(…)` ([#20319](https://github.com/tailwindlabs/tailwindcss/pull/20319))
+- Prevent `@tailwindcss/vite` from forcing a full page reload when editing a scanned JS/TS file that is not part of the loaded module graph, like a lazy route or an unvisited chunk ([#20323](https://github.com/tailwindlabs/tailwindcss/pull/20323))
 
 ## [4.3.2] - 2026-06-26
 
--- a/integrations/vite/index.test.ts
+++ b/integrations/vite/index.test.ts
@@ -585,6 +585,128 @@ describe.each(['postcss', 'lightningcss'])('%s', (transformer) => {
   )
 
   test(
+    'scanned module files that are not loaded do not trigger a full reload',
+    {
+      fs: {
+        'package.json': json`{}`,
+        'pnpm-workspace.yaml': yaml`
+          #
+          packages:
+            - project-a
+        `,
+        'project-a/package.json': json`
+          {
+            "type": "module",
+            "dependencies": {
+              "@tailwindcss/vite": "workspace:^",
+              "tailwindcss": "workspace:^"
+            },
+            "devDependencies": {
+              ${transformer === 'lightningcss' ? `"lightningcss": "^1",` : ''}
+              "vite": "^8"
+            }
+          }
+        `,
+        'project-a/vite.config.ts': ts`
+          import tailwindcss from '@tailwindcss/vite'
+          import { defineConfig } from 'vite'
+
+          export default defineConfig({
+            css: ${transformer === 'postcss' ? '{}' : "{ transformer: 'lightningcss' }"},
+            build: { cssMinify: false },
+            plugins: [
+              tailwindcss(),
+              {
+                // Log every payload sent over the HMR channels so the test
+                // can observe (the absence of) full reloads
+                name: 'hot-send-logger',
+                configureServer(server) {
+                  for (let environment of Object.values(server.environments)) {
+                    let send = environment.hot.send.bind(environment.hot)
+                    environment.hot.send = (...args) => {
+                      console.log('HOT_SEND ' + JSON.stringify(args[0]))
+                      return send(...args)
+                    }
+                  }
+                },
+              },
+            ],
+          })
+        `,
+        'project-a/index.html': html`
+          <html>
+            <head>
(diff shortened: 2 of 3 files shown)


RobinMalfait:
Hey! Closed this in favor of #20336 because while the regex does solve _some_ things, it doesn't solve the full picture (e.g. a file with a custom extension that isn't loaded yet)

PR #20329, opened 2026-07-14
Prevent `@tailwindcss/upgrade` from rewriting files inside `node_modules` when run from a workspace subpackage

### LINKED ISSUE #20328: `@tailwindcss/upgrade` rewrites files in `node_modules` when run from a pnpm workspace subpackage
## Summary

When `@tailwindcss/upgrade` is run from a **subpackage of a pnpm workspace**, it traverses `../../node_modules/…` and rewrites files inside the installed `tailwindcss` package in place — applying its v3 → v4 migration transform (`@tailwind utilities;` → `@import 'tailwindcss/utilities' layer(utilities);`) to `node_modules/tailwindcss/utilities.css` (and, in a flat-project variant, to `index.css`).

This looks like the tail of [#19726](https://github.com/tailwindlabs/tailwindcss/issues/19726) — the fix landed there (ignore gitignored files) works for the simple flat-project case, but doesn't cover the workspace-subpackage variant where the tool reaches `node_modules` through `../../`. In our case (a pnpm-workspace v4 project), running the tool as a canonicalizer left `node_modules/tailwindcss/utilities.css` self-referentially importing itself, which detonated much later as an infinite CSS-resolution loop and read like a "cache corruption" problem.

## Reproduction

Repo: https://github.com/benface/tailwindcss-upgrade-node-modules-repro

```sh
git clone https://github.com/benface/tailwindcss-upgrade-node-modules-repro
cd tailwindcss-upgrade-node-modules-repro
pnpm install
cd packages/css
pnpm dlx @tailwindcss/upgrade
```

The tool's own output confirms the write:

```
│ ↳ Migrated stylesheet: `../../node_modules/.pnpm/tailwindcss@4.3.2/node_modules/tailwindcss/utilities.css`
│ ↳ Updated package: `tailwindcss`
```

Diff against a pristine copy from the tarball:

```diff
1c1
< @tailwind utilities;
---
> @import 'tailwindcss/utilities' layer(utilities);
```

`node_modules` is gitignored at the repo root (and even at the subpackage level, tested separately — same result).

## Counter-tests (where the existing fix works correctly)

For contrast, both of these leave `node_modules` untouched:

- Flat project (no workspace), `node_modules` in `.gitignore`, `@import 'tailwindcss'`, run from project root
- Same, using explicit-path imports (`@import 'tailwindcss/utilities.css' layer(utilities) source(none)`)

The pnpm-workspace-subpackage layout is the specific gap.

## Why this matters

1. **`node_modules` shouldn't be a write target.** The files are shared across projects via the pnpm store, are immutable install artifacts elsewhere, and are out of scope for `git status`, so the corruption is invisible until it detonates.
2. **The rewrite creates a self-referential import** — the mutated `utilities.css` imports itself under a `layer(...)` wrapper, causing an infinite CSS-resolution loop.
3. **Symptoms look unrelated to the upgrade run** — the user hits infinite-recursion errors much later, from any command that touches Tailwind, with a clean `git status`.
4. **Recovery requires nuking `node_modules` + reinstalling** (and on Bun, purging the install cache too, per #19726). Not obvious from the symptoms.

## Suggested guard

Skip any file whose resolved absolute path falls under a `node_modules/` segment, regardless of how it was reached (traversal via `../../`, symlink resolution, or `@import` resolution). The existing gitignore-aware filter probably just needs its input paths canonicalized before the check.

## Environment

- macOS (arm64), Node 24, pnpm 11.12.0
- `tailwindcss@4.3.2`
- `@tailwindcss/upgrade@4.3.2` (via `pnpm dlx`)

Happy to test a fix.

benface:
Fixes #20328.

## What changed

When `@tailwindcss/upgrade` is invoked from a subpackage of a workspace (e.g. `pnpm --filter …` or `cd packages/foo && pnpm exec upgrade`), it traverses `../../node_modules/…` and applies its v3 → v4 migration transform (`@tailwind utilities;` → `@import 'tailwindcss/utilities' layer(utilities);`) to files inside the installed `tailwindcss` package itself — a self-referential import that later detonates as an infinite CSS-resolution loop and reads like a "cache corruption" bug (matches the report in #19726 / #20328).

Root cause: `globby`'s `isGitIgnored` only walks `.gitignore` files at or below its `cwd`. When invoked from a subpackage, the workspace-root `.gitignore` (which almost always lists `node_modules`) is never consulted, and `../../node_modules/…` paths come back as **not** ignored.

The fix anchors `isGitIgnored` at the git repository root instead of the subpackage:

- New helper: `gitRoot(cwd)` in `src/utils/git.ts` — thin wrapper over `git rev-parse --show-toplevel`.
- `analyze(stylesheets, { base })` in `src/codemods/css/analyze.ts` now calls `isGitIgnored({ cwd: gitRoot(base) ?? base })`. Falls back to the previous behavior when not inside a git repository or when `git` is unavailable.

## Test

Added an integration test that reproduces the exact scenario: a pnpm workspace with a subpackage that imports `tailwindcss/utilities.css`, upgrade run from the subpackage, then asserts `packages/css/node_modules/tailwindcss/utilities.css` still holds the pristine `@tailwind utilities;` directive.

Verified the test fails on `main` and passes with this change.

The 415 existing upgrade unit tests still pass. Two npm-related snapshot failures in `upgrade-errors.test.ts` (`half-upgraded v3 project to v4 (bun|npm)`) pre-exist this change — they're about newer npm's `npm notice run` output that isn't in the snapshots, unrelated to what this PR touches.

---

Edit by @RobinMalfait: going to run on all [ci-all] so we can make sure it works on Windows as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `--spacing(0)` is optimized to `0px` instead of `0` so it remains a `<length>` when used in `calc(…)` ([#20319](https://github.com/tailwindlabs/tailwindcss/pull/20319))
 - Lazily load `@parcel/watcher` when using the `--watch` flag in `@tailwindcss/cli`, so one-off builds and `--watch --poll` work when `@parcel/watcher` can't be loaded ([#20325](https://github.com/tailwindlabs/tailwindcss/issues/20325))
 - Use explicit platform fonts instead of `system-ui` and `ui-sans-serif` so CJK text respects the page's `lang` attribute on Windows ([#19767](https://github.com/tailwindlabs/tailwindcss/issues/19767), [#19768](https://github.com/tailwindlabs/tailwindcss/issues/19768))
-- Prevent `@tailwindcss/upgrade` from rewriting files inside `node_modules` when run from a workspace subpackage, by anchoring the `.gitignore` lookup at the git repository root instead of the subpackage ([#20328](https://github.com/tailwindlabs/tailwindcss/issues/20328))
+- Prevent `@tailwindcss/upgrade` from rewriting ignored files when run from a subdirectory ([#20328](https://github.com/tailwindlabs/tailwindcss/issues/20328))
 
 ## [4.3.2] - 2026-06-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure `--spacing(0)` is optimized to `0px` instead of `0` so it remains a `<length>` when used in `calc(…)` ([#20319](https://github.com/tailwindlabs/tailwindcss/pull/20319))
 - Lazily load `@parcel/watcher` when using the `--watch` flag in `@tailwindcss/cli`, so one-off builds and `--watch --poll` work when `@parcel/watcher` can't be loaded ([#20325](https://github.com/tailwindlabs/tailwindcss/issues/20325))
 - Use explicit platform fonts instead of `system-ui` and `ui-sans-serif` so CJK text respects the page's `lang` attribute on Windows ([#19767](https://github.com/tailwindlabs/tailwindcss/issues/19767), [#19768](https://github.com/tailwindlabs/tailwindcss/issues/19768))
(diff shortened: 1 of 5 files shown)


benface:
Thanks @greptile-apps — good catch on the self-documentation. Test now declares the root `.gitignore` explicitly (68bce62): even though the harness injects a default with the same content, making it visible in the test's `fs` block ties the assertion directly to the mechanism the fix exercises.

RobinMalfait (approved):
Thanks! Made a few small changes, and split your original commit such that I could see the test fail _before_ we apply the fix. The test did indeed fail and passed afterwards 💪

Once CI passes (running on Linux / macOS / Windows), this will be merged and will be available in the next release. Thanks!

PR #20338, opened 2026-07-16
fix: lazily load @parcel/watcher in @tailwindcss/cli (#20325)

deepshekhardas:
Cherry-pick of PR #20325

Lazily loads \@parcel/watcher\ when using the \--watch\ flag in \@tailwindcss/cli\, so one-off builds and \--watch --poll\ work when \@parcel/watcher\ can't be loaded.


deepshekhardas:
Closing - changes already present in main. I will recreate properly.

PR #20344, opened 2026-07-16
Ensure custom variants using `@scope` generate working CSS

### LINKED ISSUE #18961: Custom variants don't work with @scope

v4.1.13

Next.js 15.5.2

v24.7.0

Chrome (v140)

macOS

**Reproduction URL**

https://play.tailwindcss.com/0NPeZeXJ61

**Describe your issue**

Custom variants don't work with `@scope` rules. When defining a custom variant using `@scope`, the generated code places the rule inside the class instead of the class inside the rule. But `@scope` rules don't seem to support nesting like this as the resulting class has no effect

UditDewan:
## Summary

Fixes #18961

Custom variants defined with an `@scope` at-rule currently emit the `@scope` block nested *inside* the generated utility class instead of wrapping it:

```css
@custom-variant in-checkout {
  @scope (.checkout) {
    @slot;
  }
}
```

```css
/* before */
.in-checkout\:underline {
  @scope (.checkout) {
  }
}
```

Browsers don't apply `@scope` in this nested position, so the utility has no effect. Additionally, when the output goes through Lightning CSS the declarations inside the nested `@scope` are dropped entirely, leaving an empty `@scope (.checkout) {}` block.

Other conditional at-rules used in custom variants (`@media`, `@supports`, `@container`, `@starting-style`) are hoisted above the generated style rule when nesting is resolved. This PR adds `@scope` to the set of hoistable at-rules (and to the droppable-if-empty set, since an empty `@scope` block has no effect), so the same custom variant now produces:

```css
/* after */
@scope (.checkout) {
  .in-checkout\:underline {
  }
}
```

This works with both the block form (`@custom-variant name { @scope (…) { @slot; } }`) and the shorthand form (`@custom-variant name (@scope (…));`), and preserves `to (…)` limits.

## Test plan

- Added tests in `index.test.ts` covering custom variants using `@scope`, with and without a `to (…)` limit (these go through the Lightning CSS optimize step, verifying the output survives minification).
- Added a test in `ast.test.ts` verifying `@scope` is hoisted out of style rules during nesting resolution.
- `vitest run` in `packages/tailwindcss`: 41 test files, 4874 tests passed.


RobinMalfait:
Hey! Thanks for the PR. This was a good start, but the `@scope` rule is much more complex and just hoisting it out would introduce other issues. I continue on top of this PR and kept your commits such that you are still marked as a contributor.

More info: https://github.com/tailwindlabs/tailwindcss/pull/20369

UditDewan:
> Hey! Thanks for the PR. This was a good start, but the `@scope` rule is much more complex and just hoisting it out would introduce other issues. I continue on top of this PR and kept your commits such that you are still marked as a contributor.
> 
> More info: #20369

Thanks!

Issue #13533, opened 2024-04-17
Tailwind HMR does not work with Next.js with symlinks


nahtnam:
**What version of Tailwind CSS are you using?**

3.4.1

**What build tool (or framework if it abstracts the build tool) are you using?**

Next.js 14.2.1 with tailwind 3.4.1

**What version of Node.js are you using?**

v20.9.0

**What browser are you using?**

Chrome

**What operating system are you using?**

macOS

**Reproduction URL**

https://github.com/nahtnam/tailwind-symlink-bug

**Describe your issue**

1. Clone the above repo
2. Start the server
3. Go to the prose and add `prose-sm`, `prose-lg`, or `prose-xl`. It will not generate the class name. If you restart the server it does. You can also do something like `py-[100px]`
4. Remove the folder `src/app/(symlink)` (which contains a symlink)
5. Now add or change the prose sizing and it works fine

angela683h:
You’re encountering a known issue with Tailwind CSS and Next.js where Hot Module Replacement (HMR) does not work correctly with symlinks. This can prevent Tailwind from generating the correct class names on the fly. Here’s a quick summary of what you can do:

Remove symlinks: As you’ve noticed, removing the symlink resolves the issue.
Restart the server: Restarting the server can temporarily fix the problem by regenerating the classes.

I hope it's helpful to you.

Best Regard,
angela683h

cris458carlo:
Hello,
You are encountering an issue with Tailwind CSS 3.4.1 and Next.js 14.2.1 where class names like prose-sm, prose-lg, or prose-xl are not being generated unless you restart the server. This issue seems to be related to the presence of a symlink in your project directory.

Here are a few steps you can try to resolve this issue:

Check Tailwind Configuration: Ensure that your tailwind.config.js is correctly set up to include the paths where your classes are used. Sometimes, symlinks can cause paths to be misinterpreted.
Update Dependencies: Make sure all your dependencies are up to date. Sometimes, issues are resolved in newer versions of the tools you’re using.
Symlink Handling: Since removing the symlink seems to resolve the issue, you might want to avoid using symlinks in your project if possible. Alternatively, you can look into configuring your build tool to handle symlinks more gracefully.
Tailwind JIT Mode: If you’re not already using it, consider enabling JIT mode in Tailwind CSS. JIT mode can sometimes handle dynamic class generation more efficiently.
[...]

PR #20347, opened 2026-07-17
Ensure `@source` files behind symlinks trigger rebuilds in `@tailwindcss/vite`

### LINKED ISSUE #20346: @tailwindcss/vite does not hot-reload new candidates from explicit @source files behind workspace symlinks since v4.3.1

v4.3.3

This regressed between v4.3.0 and v4.3.1 and is still reproducible in v4.3.3.

Vite 8.1.4 with @tailwindcss/vite 4.3.3 and a pnpm workspace.

v24.18.0

N/A 

Linux 6.18.38 x86_64

**Reproduction URL**

https://github.com/ageorgeh/tailwind-vite-symlink-hmr-reproduction

**Describe your issue**

When an explicit `@source` points to a file through a pnpm workspace symlink, Tailwind scans the file correctly at startup, but changes to the real source file do not trigger a rescan.

Steps:

1. Run `pnpm install`.
2. Run `pnpm dev`.
3. Open the displayed URL and confirm the example text is red.
4. Without stopping Vite, change `text-red-500` to `text-green-500` in `packages/repro-package/source.html`.
5. Click “Use text-green-500” in the browser.

Expected: Tailwind detects the new candidate, regenerates the CSS, and the text becomes green.

Actual: `.text-green-500` is not generated and the text is unstyled. Restarting Vite causes the class to be generated.

Changing both `tailwindcss` and `@tailwindcss/vite` to v4.3.0 makes the update work without restarting Vite.

This appears related to #20203. 

LLM reasoning as to why: Before that change, source optimization canonicalized the complete source path, including symlinked path segments. Since v4.3.1, the scanner preserves the symlink path. `@tailwindcss/vite` then passes paths from `scanner.files` to `addWatchFile`, but edits happen at the symlink's real workspace target. No watcher update reaches Tailwind, so the file is never rescanned.

Nic-Polumeyv:
Fixes #20346

## Summary

Since #20203 the scanner preserves symlinks in `@source` paths, so `scanner.files` can contain a path like `app/node_modules/pkg/index.js` while Vite tracks the module and its file change events under the real `packages/pkg/index.js`. The plugin registered only the written path via `addWatchFile`, which Vite's watcher ignores inside `node_modules`, so editing the real file behind a workspace symlink never invalidated the CSS module and new candidates only appeared after restarting the dev server.

This registers the resolved path alongside the written one for scanned files and glob bases, cached per root. `isScannedFile` gets the missing direction too: it already resolved the changed file before comparing against scanned paths, now it also resolves scanned paths before comparing against the changed file, so full reloads for non-module `@source` files keep working when the path goes through a symlink.

## Test plan

Added an integration test with a pnpm workspace where `@source` points through the `node_modules` symlink of a linked workspace package. It fails without the change (the new candidate never appears until the dev server restarts) and passes with it. The full `integrations/vite` suite passes. Also verified the reproduction from #20346 directly: with this build, editing the file behind the symlink updates the page without a restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
+- Ensure `@tailwindcss/vite` rebuilds when a `@source` file behind a symlink changes, such as a file in a linked package in a pnpm workspace ([#20347](https://github.com/tailwindlabs/tailwindcss/pull/20347))
 
 ## [4.3.3] - 2026-07-16
 
--- a/integrations/vite/index.test.ts
+++ b/integrations/vite/index.test.ts
@@ -1609,3 +1609,96 @@ test(
     })
   },
 )
+
+test(
+  '`@source` files behind a workspace symlink trigger rebuilds when the real file changes',
+  {
+    fs: {
+      'package.json': json`{}`,
+      'pnpm-workspace.yaml': yaml`
+        #
+        packages:
+          - project-a
+          - project-b
+      `,
+      'project-a/package.json': json`
+        {
+          "type": "module",
+          "dependencies": {
+            "project-b": "workspace:*",
+            "@tailwindcss/vite": "workspace:^",
+            "tailwindcss": "workspace:^"
+          },
+          "devDependencies": {
+            "vite": "^7"
+          }
+        }
+      `,
+      'project-a/vite.config.ts': ts`
+        import tailwindcss from '@tailwindcss/vite'
+        import { defineConfig } from 'vite'
+
+        export default defineConfig({
+          build: { cssMinify: false },
+          plugins: [tailwindcss()],
+        })
+      `,
+      'project-a/index.html': html`
+        <head>
+          <link rel="stylesheet" href="./src/index.css" />
+        </head>
+        <body>
+          <div>Hello, world!</div>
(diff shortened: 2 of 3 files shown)


Nic-Polumeyv:
Testing this a bit further, the same class shows up in the scanner's other consumers.

The CLI's native `--watch` is affected with the mechanism inverted: `watchDirectories` resolves watch dirs since #20242, so events arrive with real paths and `scanner.scanFiles()` rejects them against the symlink-based source allowlist. In the integration harness, an edit to a plain `@source` file rebuilds while the same edit behind a pnpm workspace symlink does not. `--watch --poll` is fine since it rescans through the source paths themselves.

`@tailwindcss/postcss` registers the symlink path in its `dependency` messages (`app/node_modules/pkg/index.js` in a pnpm workspace). Consumers that skip watching `node_modules` never see changes to the real file, which would bring back #13533 on v4; I haven't verified that end to end.

Happy to fix the CLI the same way in a follow-up, though canonicalizing once in `scan_files` might beat patching each consumer.

RobinMalfait:
Hey! Thanks for the PR, but I took a different approach by moving this logic to the `@tailwindcss/oxide` scanner instead such that all packages (`@tailwindcss/postcss`, `@tailwindcss/vite`, `@tailwindcss/webpack` and `@tailwindcss/cli`) benefit from this. So essentially the same conclusion you arrived on in your follow-up 👍

I started a new PR, but I marked you as a co-author. Closing this in favor of #20356.

Issue #18870, opened 2025-09-03
Include glob path fails when path is also gitignored


SanderFlooris:
**What version of Tailwind CSS are you using?**

`v4.1.12`

**What build tool (or framework if it abstracts the build tool) are you using?**

Vite (`6.2.1`)

**What version of Node.js are you using?**

`v22.19.0`

**What browser are you using?**

`Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:142.0) Gecko/20100101 Firefox/142.0`

**What operating system are you using?**

`macOS Sequoia 15.6.1`

**Reproduction URL**

https://github.com/SanderFlooris/tailwind-4-gitignore-testcase

Instructions on how to configure the repository and how to trigger the bug has been provided in the README.

**Describe your issue**

In a project we are working on we prerender some parts of the page as static HTML files, these HTML files will then dynamically be loaded in our templates to cache sections of the webpage, without caching everything.

However we discovered that Tailwind doesn't seem to pick up the classes properly inside of our `cms_content` directory that contains the prerendered sections, even though we manually specify in our Tailwind configuration that that directory should be scanned.

After some digging around we figured out it was caused by the directory being gitignored, and we are unable to overwrite this. The compiler just skips over the directory entirely when we checked with `DEBUG=* yarn dev`.

minseonkkim:
Hi,

The behavior you’re seeing happens because Tailwind CSS v4, by default, ignores directories listed in .gitignore. Setting an environment variable will resolve this.

For example:

macOS/Linux:
```
export TAILWIND_DISABLE_GITIGNORE=true
```

Windows PowerShell:
```
$env:TAILWIND_DISABLE_GITIGNORE="true"
```

After setting this, rebuild Tailwind. Make sure that the cms_content directory is also included in the content array in your tailwind.config.js.

This should allow Tailwind to properly scan the classes inside your prerendered HTML files.

SanderFlooris:
@minseonkkim I tried your suggestion, but it doesn't seem to resolve the issue.

I also searched for `TAILWIND_DISABLE_GITIGNORE` and `DISABLE_GITIGNORE` with Google and inside the tailwindlabs GitHub organization, but there are no matches.
Are you sure the environment variable name is correct? 🤔 

<img width="1045" height="743" alt="Image" src="https://github.com/user-attachments/assets/e976a226-0d9f-4b46-8d62-db6c557b5a4b" />

[...]

PR #20349, opened 2026-07-18
test(oxide): cover gitignored source globs

cyphercodes:
## Summary

Adds regression coverage for explicit `@source` deep globs that point inside a gitignored directory.

The test mirrors the `storage/app/private/cms_content/**/*.html` shape from #18870 and verifies Tailwind scans matching HTML files without scanning non-matching files in the same ignored tree.

## Test plan

```sh
cargo test -p tailwindcss-oxide it_should_scan_explicit_deep_globs_inside_gitignored_folders -- --nocapture
git diff --check
cargo fmt --check
```

Refs #18870

### DIFF
--- a/crates/oxide/tests/scanner.rs
+++ b/crates/oxide/tests/scanner.rs
@@ -1277,6 +1277,36 @@ mod scanner {
         assert_eq!(files, vec!["src/foo.html"]);
     }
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
Hey! Appreciate the PR, but we don't know yet if this is the behavior we want (since it could result in a big breaking change), so going to close this for now.

PR #20374, opened 2026-07-31
Fix extraction of quoted arbitrary variants in v3

supermadu7:
## Summary

Fixes #18387.

The v3 default extractor did not capture the complete candidate when a standard variant preceded an arbitrary selector containing quoted values, such as `focus-visible:[&:not([aria-selected="true"])]:bg-red-500`.

The quote-aware pattern allowed quote characters in the standard-variant portion, causing extraction to begin at the surrounding `class` or `className` attribute and split the candidate. This change excludes single and double quotes from that portion of the pattern while continuing to allow quotes inside arbitrary variants.

Regression tests cover both single- and double-quoted selectors with preceding `focus-visible` and `hover` variants.

[ci-all]

## Test plan

* `npx jest tests/default-extractor.test.js tests/arbitrary-variants.test.js --runInBand`
* `npm test -- --runInBand`

Results:

* 86 test suites passed
* 1,060 tests passed
* 1 test marked todo
* 4 snapshots passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+### Fixed
+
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
Thanks!

PR #20379, opened 2026-08-03
Prevent `@tailwindcss/vite` crash under Vite's experimental `bundledDev`

### LINKED ISSUE #20378: `@tailwindcss/vite` crashes on every edit under Vite's `experimental.bundledDev`

v4.3.3 (`@tailwindcss/vite` 4.3.3)

Vite 8.2.0, with `experimental.bundledDev: true`

v24.14.0

N/A — the dev server fails before the browser is involved

macOS 26.3

**Reproduction URL**

https://github.com/johanrd/tailwind-vite-bundleddev-repro

**Describe your issue**

With Vite's `experimental.bundledDev: true`, editing any file — JS, CSS or HTML — fails the build:

```
[plugin @tailwindcss/vite:generate:serve]
TypeError: Cannot read properties of undefined (reading 'environments')
    at PluginContextImpl.hotUpdate (@tailwindcss/vite/dist/index.mjs:1:3459)
```

`hotUpdate` iterates `Object.values(server.environments)`, but in bundled dev Vite passes only `{ type, file, modules }` — `server`, `timestamp` and `read` are absent. Vite's `HotUpdateOptions` declares all six as required, which is why the plugin dereferences `server` without guarding.

@sapphi-red of Vite has indicated that [`server` cannot be supported in bundled dev and will not be passed](https://github.com/vitejs/vite/issues/23125#issuecomment-5162509851), and that bundled dev requires changes on the plugin side, so the guard looks like it belongs here.

The loop only inspects environments *other* than the current one, so an early return when `server` is undefined should leave classic dev unaffected.

Fully aware `bundledDev` is experimental and you may not want to support it yet — filing mainly so it's on record, since the failure mode gives no hint about where it comes from.

lazerg:
## Summary

Under Vite's experimental `bundledDev` mode, the `hotUpdate` hook in `@tailwindcss/vite` gets called without a `server`. Vite only passes `{ type, file, modules }` here, but the hook loops over `Object.values(server.environments)`, so editing any file (JS, CSS, or HTML) throws `TypeError: Cannot read properties of undefined (reading 'environments')` and the dev server build fails.

The fix returns early when `server` is missing. Those environment loops only look at environments other than the current one, and the server-level `hot`/`ws` reload channels don't exist in this mode, so bailing out leaves the classic (non-`bundledDev`) dev path untouched.

Fixes #20378

## Test plan

- Added a unit test that calls `hotUpdate` without a `server` and checks it doesn't throw. It fails on the current code and passes with the guard.
- Reproduced with a Vite 8 project using `experimental.bundledDev: true`: before the change, editing any JS/CSS/HTML file crashed the dev server; after it, edits work.
- `pnpm run test` and the `@tailwindcss/vite` integration suite both pass.

[ci-all]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix flattening of `@scope` at-rules ([#20369](https://github.com/tailwindlabs/tailwindcss/pull/20369))
 - Fix standalone declarations in `@scope`, wrap them in `:where(:scope)` ([#20369](https://github.com/tailwindlabs/tailwindcss/pull/20369))
 - Always emit a space for empty fallback values in CSS variables (e.g. `var(--tw-blur,)` → `var(--tw-blur, )`) ([#20373](https://github.com/tailwindlabs/tailwindcss/pull/20373))
+- Prevent `@tailwindcss/vite` from crashing on every edit under Vite's experimental `bundledDev` mode ([#20379](https://github.com/tailwindlabs/tailwindcss/pull/20379))
 
 ## [4.3.3] - 2026-07-16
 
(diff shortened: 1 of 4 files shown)


RobinMalfait (approved):
Thanks! 

Made a few small changes to make TypeScript happy. 
Also added an integration test to verify that this mode also works, and not just _not_ crashes.

PR #20384, opened 2026-08-04
Attempt to fix flaky integration test

RobinMalfait:
One of the integration tests is flaky on CI. This is an attempt to "fix" it. 

It has probably something to do with the availability of the resources provided by GitHub because locally this works flawlessly for 100 runs straight.

## Test plan

- Once we get 5 consecutive positive runs we van merge it.
- No actual code was changed, only the flaky integration test itself.

image


CreatorVishal:
👍

PR #20389, opened 2026-08-05
fix: Drop invalid UTF-8 scanner candidates

### LINKED ISSUE #20368: Scanned candidate emitted with corrupted bytes (`child` → `U+FFFD \x06 U+FFFD \x01`) by a long-lived `next dev` (Turbopack) process — evidence against the scanner's `from_utf8_unchecked` safety assumption
## Summary

A long-running `next dev` (Turbopack) process in a pnpm monorepo entered a state
where Tailwind's generated CSS contained a corrupted copy of one arbitrary-variant
candidate. The bytes of `child` in the candidate `[&>*:last-child]:col-span-2`
(present verbatim in a `.tsx` source file) were replaced by the four characters
`U+FFFD U+0006 U+FFFD U+0001`, producing unparseable CSS. Every route then failed
with lightningcss's `Parsing CSS source code failed`, and the state persisted
in memory until the dev server was restarted. A production `next build` of the
identical tree compiled cleanly every time, before, during, and after.

We do not have a reproduction recipe — this is a forensic report of a corrupted
state captured live. We're filing it because the corruption is demonstrably in
the **candidate string itself, before variant expansion** (see Analysis), which
bears directly on the safety assumption that led to closing #19410
(`from_utf8_unchecked` in the scanner is fine because "the extractor's
implementation should guarantee that the resulting candidates are already
UTF-8"). The `U+FFFD` characters indicate a lossy UTF-8 conversion happened
somewhere downstream of bytes that were **not** valid UTF-8.

## Environment

- tailwindcss / @tailwindcss/postcss: **4.3.3** (`@tailwindcss/oxide-darwin-arm64@4.3.3`)
- next: **16.2.12**, dev server (Turbopack), via the PostCSS plugin (`postcss.config.mjs` → `@tailwindcss/postcss`)
- node: v26.5.0 (nvm), pnpm 11, pnpm workspace monorepo (app at `apps/web`)
- macOS 26.5.2, Apple Silicon (arm64)
- Dev server had been running for a long stretch (hours+) and had served many
  on-demand route compiles; the machine was under heavy load at times
  (parallel Playwright suites against sibling dev servers, several `next dev`
  processes in worktrees of the same repo).

## The source candidate

`apps/web/src/app/(practitioner)/dashboard/page.tsx`:

```tsx
className="grid grid-cols-2 gap-px overflow-hidden rounded-xl border border-inst-hairline bg-inst-hairline sm:grid-cols-3 lg:grid-cols-5 [&>*:last-child]:col-span-2 lg:[&>*:last-child]:col-span-1"
```

## The error (verbatim, from the dev-overlay payload)

Every route on the affected server returned HTTP 500 with this compilation
error (control characters rendered here as `<0x06>` / `<0x01>`; `�` is a literal
U+FFFD REPLACEMENT CHARACTER present in the server's output):

```
./apps/web/src/app/globals.css:2848:51
Parsing CSS source code failed
  2846 |     }
  2847 |   }
> 2848 |   .\[\&\>\*\:last-�\6 �\1 \]\:col-span-2 > :last-�<0x06>�<0x01> {
       |                                               ^
  2849 |     grid-column: span 2 / span 2;
  2850 |   }
  2851 | }

Unexpected token Delim('\u{6}')

Generated code of PostCSS transform of file content of apps/web/src/app/globals.css:
./apps/web/src/app/globals.css:2848:51
[...same excerpt...]

Import trace:
  Client Component Browser:
    ./apps/web/src/app/globals.css [Client Component Browser]
    ./apps/web/src/app/layout.tsx [Server Component]
```

Note the class-name position shows the **CSS-escaped** forms `\6 ` and `\1 `
(escapes of U+0006 / U+0001), while the selector tail shows the same characters
raw — both derived from one candidate whose `child` had become
`U+FFFD U+0006 U+FFFD U+0001`.

## Analysis: the corruption is in the candidate, not the CSS text

The utility `[&>*:last-child]:col-span-2` generates a rule whose selector uses
the candidate twice, through two different code paths:

1. the **escaped class name** — `.\[\&\>\*\:last-child\]\:col-span-2`
   (string-escaping of the candidate), and
2. the **expanded variant** — `> :last-child` (obtained by *parsing* the
   arbitrary-variant body out of the candidate and substituting `&`).

In the corrupted output, **both positions carry the identical 4-character
replacement** of the 5 characters `child`. Random corruption of the generated
CSS text (e.g. in Turbopack's string handling downstream) cannot plausibly hit
two different positions, through two different transforms (one CSS-escaped, one
raw), with the same replacement. The candidate string already contained
`last-�<0x06>�<0x01>` when the compiler parsed and expanded it. The corruption
therefore happened at or before candidate intake — i.e. in the scanner/extractor
or in the transfer of its results.

Two details that may help locate it:

- `U+FFFD` means a **lossy** UTF-8 decode happened after invalid bytes
  appeared. The scanner converts extracted byte slices with
  `String::from_utf8_unchecked` (`crates/oxide/src/scanner/mod.rs`); a slice
  containing invalid UTF-8 would sail through unchecked and only be replaced
  with `U+FFFD` later (e.g. at the napi string boundary). The surviving
  `0x06`/`0x01` are valid single-byte UTF-8, consistent with a raw byte
  pattern like `<invalid> 06 <invalid> 01` at extraction time.
- The affected `.tsx` file is plain ASCII at that position and was not being
  edited when the corruption arose; the persistent on-disk chunk for
  `globals.css` in `.next` was healthy (`…:last-child…` intact), so the
  corrupted candidate lived only in the long-running process's incremental
  state (the plugin holds a persistent `Scanner` whose candidate set
  accumulates across `scan()` calls — once a corrupted candidate enters that
  set, every subsequent rebuild re-emits the broken rule, which matches the
  observed "stuck until restart" behavior).

## What we can and can't provide

- Full captured 500 payload (dev-overlay JSON incl. ANSI variant) — available
  on request; excerpted verbatim above.
- No reproduction recipe: restarting the dev server cleared it, and we have
  not been able to trigger it on demand. Conditions at the time: hours-old
  dev server, many on-demand route compiles, heavy parallel filesystem/CPU
  load on the machine.
- `next build` (non-Turbopack CSS path, fresh scanner) of the same tree:
  always clean.

We understand a report without a repro may not be actionable on its own; we're
filing primarily as concrete field evidence that extracted candidates can
contain invalid UTF-8 in real conditions, contra the invariant cited when
closing #19410 — a cheap `from_utf8` (or debug assertion) at that boundary
would turn this silent state-corruption into a loud, attributable error.

mvanhorn:
## Summary

Replace the three unchecked scanner conversions in `crates/oxide/src/scanner/mod.rs` with checked conversions that omit only extracted slices that are not valid UTF-8. Apply the check in the shared `extract` pipeline so initial scans, incremental `scan_content` calls, and CSS-variable extraction cannot insert invalid strings into scanner state, and apply the same policy to both branches of `get_candidates_with_positions` while preserving byte offsets and the legacy `-[]` restoration. Keep the extractor's byte-oriented CSS identifier classification unchanged: accepting non-ASCII bytes during extraction is useful for valid multibyte code points, while the conversion boundary is the authoritative place to enforce the `String` contract.

The scanner currently converts extracted byte slices with unchecked UTF-8 constructors at the shared extraction boundary and both candidate-with-position branches. A source file containing a stray continuation byte can therefore produce an invalid `String`, violating Rust's string invariant and allowing corrupted candidates to persist in a long-lived scanner. The thread provides a deterministic reproduction using invalid bytes inside an arbitrary value, so the problem no longer depends on reproducing the originally reported Turbopack race. Valid candidates found alongside malformed byte sequences must continue to be returned normally.

Fixes #20368

## Test plan

Not applicable to this change.

AI was used for assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don't scan gitignored directories (e.g. `node_modules` and `.git`) when the project uses a safelist-style `.gitignore` (e.g. `/*` followed by `!/…` negations) ([#20397](https://github.com/tailwindlabs/tailwindcss/discussions/20397))
 - Ensure root `theme('…')` namespace lookups in JavaScript plugins and config files return the full namespace object instead of the value of its `DEFAULT` key ([#20399](https://github.com/tailwindlabs/tailwindcss/pull/20399))
 - Skip ignored directories entirely when computing watch globs (`scanner.globs`), instead of walking their full contents on every rebuild ([#20408](https://github.com/tailwindlabs/tailwindcss/pull/20408))
+- Oxide: drop invalid UTF-8 candidates ([#20389](https://github.com/tailwindlabs/tailwindcss/pull/20389))
 
 ## [4.3.3] - 2026-07-16
 
--- a/crates/oxide/src/scanner/mod.rs
+++ b/crates/oxide/src/scanner/mod.rs
@@ -356,15 +356,14 @@ impl Scanner {
                     let i = s.as_ptr() as usize - offset;
                     let original = &original_content[i..i + s.len()];
                     if original.contains_str("-[]") {
-                        return Some(unsafe {
-                            (String::from_utf8_unchecked(original.to_vec()), i)
-                        });
+                        return String::from_utf8(original.to_vec())
+                            .ok()
+                            .map(|candidate| (candidate, i));
                     }
 
-                    // SAFETY: When we parsed the candidates, we already guaranteed that the byte
-                    // slices are valid, therefore we don't have to re-check here when we want to
-                    // convert it back to a string.
-                    Some(unsafe { (String::from_utf8_unchecked(s.to_vec()), i) })
+                    String::from_utf8(s.to_vec())
+                        .ok()
+                        .map(|candidate| (candidate, i))
                 }
 
                 _ => None,
@@ -617,7 +616,7 @@ where
             a
         })
         .into_iter()
-        .map(|s| unsafe { String::from_utf8_unchecked(s.to_vec()) })
+        .filter_map(|s| String::from_utf8(s.to_vec()).ok())
         .collect()
 }
 
--- a/crates/oxide/tests/scanner.rs
+++ b/crates/oxide/tests/scanner.rs
@@ -1077,6 +1077,102 @@ mod scanner {
     }
 
     #[test]
+    fn it_should_drop_invalid_utf8_candidates() {
+        let dir = tempdir().unwrap();
+        fs::write(dir.path().join("index.html"), b"flex bg-[\x80] block").unwrap();
+
+        let mut scanner = Scanner::new(vec![public_source_entry_from_pattern(
+            dir.path().to_path_buf(),
+            "@source '*.html'",
+        )]);
+
+        let candidates = scanner
+            .scan()
+            .into_iter()
+            .map(String::into_bytes)
+            .collect::<Vec<_>>();
+
+        assert_eq!(candidates, vec![b"block".to_vec(), b"flex".to_vec()]);
+    }
+
+    #[test]
+    fn it_should_not_store_invalid_utf8_candidates_during_incremental_scans() {
+        let dir = tempdir().unwrap();
+        let file = dir.path().join("index.html");
+        fs::write(&file, b"flex bg-[\x80]").unwrap();
+
+        let mut scanner = Scanner::new(vec![public_source_entry_from_pattern(
+            dir.path().to_path_buf(),
+            "@source '*.html'",
+        )]);
+
+        let candidates = scanner
+            .scan_content(vec![ChangedContent::File(file, "html".into())])
+            .into_iter()
+            .map(String::into_bytes)
+            .collect::<Vec<_>>();
+        assert_eq!(candidates, vec![b"flex".to_vec()]);
+
+        let candidates =
+            scanner.scan_content(vec![ChangedContent::Content("block".into(), "html".into())]);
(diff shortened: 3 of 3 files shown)


thejamesgore:
Saw this was created after my fix comment #20368 on the issue so I ran the repro harness against this branch to check it closes what I found there.

On main 4be6110a the four invalid-byte cases fail as before, lone `0x80`, `0xFE`, `0xFF`, and a truncated `E2 80` all come back as ill formed strings. On this branch 517f8ce8 all three pass and the valid-ASCII control still returns its candidates and the full oxide scanner suite is green with 67 passed so this verifabily fixes the soundness issue including the deterministic repro.

Two things worth flagging while you're here though:

1. The fuzzer still can't catch a regression of this. `fuzz_targets/parsing.rs` gates on `if let Ok(s) = std::str::from_utf8(data)` meaning only the `&str` path is ever fuzzed so the raw-byte path that `fs::read` feeds that this PR fixes never gets exercised. Feeding the fuzz target raw bytes would keep this class of bug from coming back silently.

2. The other thing is dropped candidates now disappear with no diagnostic and since this isn't limited to the Turbopack race it means any Latin-1/Windows-1252 source file triggers it and a user with a legacy encoded file will just see utilities silently not generating so a `tracing` WARN event at the conveersation boundary would make that debuggable.

RobinMalfait (approved):
I think figuring out why there are bogus candidates in the first place is more important, but since these candidates can't be represented in CSS anyway, I think it's fine filtering them out.

The comment about other encodings might be true, but I haven't seen an issue related to other encodings either. Going to keep it like this for now. 

A trace/warning could work, but often you're scanning code that you might not even have control over which in turn could result in a lot of noise.

Thanks!

PR #20395, opened 2026-08-07
Fix @tailwindcss/postcss losing watch dependencies after a compile error

### LINKED ISSUE #18370: Angular: Error in PostCSS + tailwind doesn't break the build

andershagbard:
## Summary

With `@tailwindcss/postcss` + `postcss-loader` (e.g. webpack), fixing a
CSS error during `--watch` doesn't resume compilation. Repro:

1. `webpack --watch` with `@tailwindcss/postcss` in `postcss.config.js`.
2. Introduce `@apply some-unknown-class` (or break an `@import`ed
   partial with any error) in a watched CSS file.
3. Fix the error and save again — webpack never recompiles. Not just
   for that file: *any* previously-tracked CSS import stops being
   watched until some unrelated, still-watched file happens to change.

### Root cause

Any compile error currently makes the plugin `throw` inside its
`Once()` hook. That causes the whole `postcss().process()` promise to
**reject** instead of resolve. `postcss-loader` only reads
`result.messages` — and therefore only calls `this.addDependency()` —
from a **resolved** result:

```js
// postcss-loader/dist/index.js
try {
  result = await processor.process(root || content, processOptions);
} catch (error) {
  reportError(this, callback, error);
  return; // <- result.messages is never read here
}
for (const message of result.messages) {
  if (message.type === "dependency") this.addDependency(message.file);
  ...
}
```

So on rejection, none of the dependency messages the plugin pushed
(including from any previously successful build) are ever seen by
webpack, which drops every file in the CSS import graph from its
watcher — not just the file that errored — until some other,
still-watched file happens to change (or the entry file itself is
touched directly).

This isn't new: #17754 fixed exactly this by not throwing
(`console.error(error); root.removeAll()`). #18373 then reinstated the
throw to fix #18370 (errors weren't failing the build), unintentionally
reintroducing this regression as a side effect, while leaving behind a
comment describing the behavior the code no longer has.

### Fix

Gate the behavior on the plugin's existing `optimize` option (already
defaults to `NODE_ENV === 'production'`):

- When `optimize` is enabled (typically production builds), keep
  throwing — this preserves #18370's fix, so broken CSS still fails
  the build.
- Otherwise, report the error via `result.warn()` instead of throwing,
  so `result` still resolves and every dependency message (including
  ones re-derived from the last successful build, for cases where the
  failure happens before this attempt's own import/candidate scanning
  ran) reaches the consumer. The stylesheet keeps serving the last
  known-good output rather than going blank while the error is fixed.

### Test plan

- Added unit tests in `packages/@tailwindcss-postcss/src/index.test.ts`
  asserting dependency messages survive a compile error outside of
  `optimize` mode (and that `optimize` mode still throws). These fail
  against the current `throw`-only behavior and pass with this fix.
- `integrations/postcss/index.test.ts`'s existing `rebuild error
  recovery` test needed a small adjustment: recovered builds now also
  log `Waiting for file changes...` (same as any successful build), so
  a later wait for that exact message needed `process.flush()` first
  to avoid matching a stale occurrence from the recovered build.
- Verified full suites pass: `@tailwindcss/postcss` unit tests (22/22),
  `integrations/postcss/` (31/31), `integrations/webpack/` (8/8,
  including a webpack+postcss-loader watch test).
- Manually reproduced the original bug with `webpack --watch` in a
  real project and confirmed this fixes it.

[ci-all]


daltino (commented):
The fix is well reasoned and the test coverage is solid. Re emitting `context.fullRebuildPaths` (and the scanner's watched files) into `result.messages` on the error path is exactly the right approach since PostCSS loaders like `postcss loader` rely entirely on those messages to maintain their file watch sets, and a thrown exception short circuits that message population entirely. One thing worth double checking: the newly introduced `resolvedInputFile` variable is computed in the error handler but I want to make sure it is actually consumed somewhere in the broader diff (the patch is truncated at line 365). If it ends up being used to guard against re registering the input file as a dependency (mirroring the logic in the success path that intentionally skips the input file), that would be important to verify is consistent with the existing `does not register the input file as a dependency` test case. The test ordering change in `index.test.ts` (flushing and waiting for "Waiting for file changes..." before writing the content file) is a meaningful fix in its own right, eliminating the pre existing race where a second build could be kicked off concurrently while the settled build was still registering its dependencies.

PR #20399, opened 2026-08-08
Make `createCompatConfig` theme mappings safe to spread when a namespace returns a string

ibnlanre:
When building the legacy config compat layer, namespace mappings like `fontSize` → `text`, `boxShadow` → `shadow`, `animation` → `animate`, etc. spread the result of `theme(namespace, {})` directly:

```ts
...(theme('text', {}) ?? {})
```

Some of these namespaces can return a **string** (e.g. `theme('text', {})` → `'1rem'`). Spreading a string produces char-indexed keys (`{ '0': '1', '1': 'r', ... }`) instead of a `{ DEFAULT: ... }` entry, silently corrupting the compat config output.

This PR adds a small `spreadTheme` helper that normalizes the return value before spreading:

- `string` → `{ DEFAULT: value }`
- plain object → `{ ...value }`
- anything else → `{}`

and updates all namespace mappings to use it. Includes unit tests for the helper plus integration tests for the affected namespaces and the string-return regression cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Detect classes in Ruby percent literals using angle brackets or custom delimiters (e.g. `%w<flex>`, `%w|flex|`), including in Slim and Haml templates ([#20387](https://github.com/tailwindlabs/tailwindcss/pull/20387))
 - Preserve whitespace in `--default(…)` values in custom functional utilities (e.g. `--default(box alphabetic)` no longer becomes `boxalphabetic`) ([#20392](https://github.com/tailwindlabs/tailwindcss/pull/20392))
 - Don't scan gitignored directories (e.g. `node_modules` and `.git`) when the project uses a safelist-style `.gitignore` (e.g. `/*` followed by `!/…` negations) ([#20397](https://github.com/tailwindlabs/tailwindcss/discussions/20397))
+- Ensure root `theme('…')` namespace lookups in JavaScript plugins and config files return the full namespace object instead of the value of its `DEFAULT` key ([#20399](https://github.com/tailwindlabs/tailwindcss/pull/20399))
 
 ## [4.3.3] - 2026-07-16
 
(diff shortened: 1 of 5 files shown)


RobinMalfait:
What's a situation/use case where you ran into this kind of issue?

ibnlanre:
> What's a situation/use case where you ran into this kind of issue?

I ran into this while working on [Jumi](https://www.github.com/ibnlanre/jumi), a Tailwind CSS v4 animation plugin I maintain.

I was resolving a project's theme through the compat layer so Jumi could read the animation values, and for certain themes the output just looked… wrong. A value that should come back as `{ DEFAULT: 'spin 1s linear infinite' }` was instead coming back as `{ '0': 's', '1': 'p', '2': 'i', … }`. When a namespace only defines its DEFAULT token (e.g. `@theme { --animate: … }` with no `--animate-*` variants), `theme('animate', {})` returns a plain string, and spreading a string gives you **character-indexed keys** instead of `{ DEFAULT: … }`.

What made me think this could've gone unnoticed for a while is that it's completely silent. The config still resolves, so one would only notice when something (typically, a plugin) actually reads those theme values. I also had several cases of arbitrary class values returning as just `"e"` because of this bug.

RobinMalfait:
> What made me think this could've gone unnoticed for a while is that it's completely silent. 

Yes and no, you would notice it given that some things just don't work if you do this. If you don't notice anything wrong, it means you're not using anything related to this and nothing will be broken.

---

Overal, I think the idea is good, but I don't like the `spreadTheme` function. Instead the  `theme('animate', {})` should use an object no matter what (at least in the JS plugins/configs). If `--animate: …` is set, it should contain the `DEFAULT` key.

This is how Tailwind CSS v3 behaves where a top-level namespace like this should result in an object. 
Luckily, most places already use `theme('animate.DEFAULT')` if they just want to get the default value out. E.g.: https://sourcegraph.com/search?q=context:global+theme%5C%28%27.*%3F%5C.DEFAULT%27%5C%29&patternType=regexp&sm=0

There is a chance that returning an object instead of a string will break, but in that case the fix in user land would be to use `theme('namespace.DEFAULT')`. Right now, before this change, there is no user land workaround we can use here.

I pushed some changes to this branch.

RobinMalfait (approved):
I think this should do it, thanks for the PR!

PR #20405, opened 2026-08-10
Scan git-ignored directories inside explicit glob sources

### LINKED ISSUE #18870: Include glob path fails when path is also gitignored

61021:
Fixes #18870

## Summary

`@source "../storage/app/private/cms_content/**/*.html"` finds nothing when the target directory is ignored from the inside, i.e. a `.gitignore` containing `*` + `!.gitignore`, the way Laravel ships every `storage/` directory. The reporter's testcase also shows the two shapes that already work: pointing `@source` at the exact file, and the same glob once the nested `.gitignore` is deleted.

The asymmetry is in how the walker treats files vs. directories. Glob sources emit an `!*.html` rule, and manually registered rules rank above on-disk `.gitignore` files, so a git-ignored *file* matched by an explicit glob is scanned today (this PR adds a test pinning that). That rule can't save anything inside a git-ignored *directory* though: `*` matches the child directories themselves, the walker prunes them, and no file rule ever runs. An ignore in an ancestor `.gitignore` (say `/storage` in the project root) doesn't trigger the bug, because the normalized source base starts the walk inside the ignored tree. That's what makes the failure look random unless you know where the ignore line lives.

Glob sources now also register a directory re-include (`!*/`) for their base, ordered so that everything that should still win, does:

- registered before the auto source detection rules, so `node_modules` etc. inside a glob source stay pruned unless targeted explicitly (the concern raised in the issue about accidentally walking huge trees)
- registered before the `@source` rules, so `@source not` still excludes directories inside a re-opened source
- on-disk `.gitignore` files rank below manually registered rules, which is what lets the walker descend

Bare directory sources (`@source "../storage/app/private/cms_content"`) are unchanged: they emit no file rules, so `.gitignore` semantics inside them still apply, as pinned by `it_respects_gitignore_in_workspace_root`.

## Test plan

- Five new tests in `crates/oxide/tests/scanner.rs`: the issue's Laravel shape, a glob through an ancestor-ignored directory, git-ignored files matched by a glob (pins existing behavior), `node_modules` staying pruned inside a re-opened source, and `@source not` inside one.
- `cargo test --workspace` passes.
- Reproduced against the reporter's testcase layout (CSS entry in `src/`, relative `../storage/...` sources, `source(none)`): before this change the scanner returns no files from `cms_content`; with it, the pre-rendered HTML is scanned and its candidates extracted.

[ci-all]

### DIFF
--- a/crates/oxide/src/scanner/mod.rs
+++ b/crates/oxide/src/scanner/mod.rs
@@ -725,6 +725,12 @@ fn create_walker(sources: &Sources) -> Option<WalkBuilder> {
         }
     };
 
+    // Bases of glob sources whose directories must stay walkable even when a
+    // `.gitignore` ignores them, otherwise the pattern rules emitted below can
+    // never reach the files they whitelist (e.g. a directory that ignores
+    // itself with `*`, like Laravel's `storage/` directories).
+    let mut reopened_bases: Vec<&PathBuf> = Vec::new();
+
     for source in sources.iter() {
         match source {
             SourceEntry::Auto { base } => {
@@ -752,6 +758,14 @@ fn create_walker(sources: &Sources) -> Option<WalkBuilder> {
                     if let Some(extension) = Path::new(&pattern).extension() {
                         // Extend auto source detection to include the extension
                         emit(base, format!("!*.{}", extension.to_string_lossy()));
+
+                        // The extension rule above whitelists matching files even
+                        // when they are git-ignored, but it cannot stop the walker
+                        // from pruning git-ignored directories on the way to those
+                        // files. Keep directories under this base walkable as well.
+                        if !reopened_bases.contains(&base) {
+                            reopened_bases.push(base);
+                        }
                     }
                 }
             }
@@ -829,6 +843,17 @@ fn create_walker(sources: &Sources) -> Option<WalkBuilder> {
 
     for root in other_roots {
         builder.add(root);
+    }
+
+    // Re-open directories under glob sources. These rules are registered before
+    // the auto source detection rules and the `@source` rules so both still take
+    // precedence, e.g. `node_modules` inside a glob source stays pruned unless
+    // targeted explicitly.
+    for base in reopened_bases {
+        let mut ignore_builder = GitignoreBuilder::new(base);
+        ignore_builder.add_line(None, "!*/").unwrap();
+        let ignore = ignore_builder.build().unwrap();
+        builder.add_gitignore(ignore);
     }
 
     // Setup auto source detection rules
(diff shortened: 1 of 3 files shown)


crates/oxide/src/scanner/mod.rs
```diff
@@ -831,6 +845,17 @@ fn create_walker(sources: &Sources) -> Option<WalkBuilder> {
...
+    // Re-open directories under glob sources. These rules are registered before
+    // the auto source detection rules and the `@source` rules so both still take
+    // precedence, e.g. `node_modules` inside a glob source stays pruned unless
+    // targeted explicitly.
+    for base in reopened_bases {
+        let mut ignore_builder = GitignoreBuilder::new(base);
+        ignore_builder.add_line(None, "!*/").unwrap();
+        let ignore = ignore_builder.build().unwrap();
+        builder.add_gitignore(ignore);
+    }
```
RobinMalfait, line 857: This part opens a new can of worms that I only want to deal with if we can prove that there are issues that can't be solved by simpler glob patterns. More info: https://github.com/tailwindlabs/tailwindcss/issues/18870#issuecomment-5242590261

In this case, let's say you have a setup like this:
```gitignore
# .gitignore
dist/
```
Then you have your CSS file that contains:
```css
@source './src';
@source './src/**/*.html';
```

If you then, for any reason, have a file like `./src/dist/index.js` then this file _will_ be scanned because the `src/` folder was re-opened. It was re-opened for the `.html` use case, not the `.js` use case.

RobinMalfait:
Thanks for the PR!

Were you trying to fix https://github.com/tailwindlabs/tailwindcss/issues/18870, or did you run into an issue yourself with this kind of setup? If you ran into this issue yourself, can you open an issue with a minimal reproduction repo attached so we can take a look?

This PR does improve some of the bugs involved, but it introduces new bugs as well. I just want to make sure that _if_ there is a bug that can't be solved in user land with a different `@source` pattern, that we solve it properly without introducing new bugs or a lot of new complexity.

Thanks again!

61021:
I don't have this setup myself. It was open and still reproduced on main, and I thought it was easy on the top of my head so I took a shot at it, your counter example is right, this branch scans src/dist/index.js and main doesn't. Glob-only doesn't leak; the walk is shared across sources, so a directory re-opened for one source's html files becomes visible to every other source, and file rules can't remember the ancestor was ignored. A proper fix means per-source walks or extension-aware pruning in the walker, not worth it without a live use case. Closing both makes sense.

PR #19632, opened 2026-02-02
Improve performance in bigger projects

RobinMalfait:
This PR improves the performance of Oxide when scanning large codebases. 

The `Oxide` API, looks something like this:

```ts
let scanner = new Scanner({ sources })
let candidates = scanner.scan() // Found candidates
let files = scanner.files // Scanned files
let globs = scanner.globs // Scanned globs
```

The `files` and `globs` are used to tell PostCSS, Vite, webpack etc which files to watch for changes.
The `.scan()` operation extracts the candidates from the source files. You can think of these as potential Tailwind CSS classes.

In all these scenarios we have to walk the file system and find files that match the `sources`.

### 1. Prevent multiple file system walks

The first big win came from the fact that accessing `.files` after a `.scan()` also does an entire walk of the file system (for the given `sources`), which is unnecessary because we just walked the file system.

This is something that's not really an issue in smaller codebases because we have `mtime` tracking. We don't re-scan a file if its `mtime` hasn't changed since the last scan. However, in large codebases with thousands of files, even walking the file system to check `mtime`s can be expensive.

### 2. Use parallel file system walking

Another big win is to use a parallel file system walker instead of a synchronous one. The big problem here is that the parallel build has 20ms-50ms of overhead which is noticeable on small codebases. We don't really know if you have a small or big codebase ahead of time, so maybe some kind of hint in the future would be useful.

So the solution I settled on right now is to use a synchronous walker for the initial scan, and then switch to a parallel walker for subsequent scans (think dev mode). This gives us the best of both worlds: fast initial scan on small codebases, and fast re-scans on large codebases.

Caveat: if you use the `@tailwindcss/cli` we know exactly which files changed so we can just re-scan those files directly without walking the file system at all. But in `@tailwindcss/postcss` we don't know which files changed, so we have to walk the file system to check `mtime`s.

While this improvement is nice, it resulted in an annoying issue related to `mtime` tracking. Since the parallel walker processes files in parallel, the `mtime` was typed as `Arc<Mutex<FxHashMap<PathBuf, SystemTime>>>` so to avoid locking, I decided to only walk the files here and collect their paths. Then later we check the `mtime` to know whether to re-scan them or not.

Initially I just removed the `mtime` tracking altogether. But it did have an impact when actually extracting candidates from those files, so I added it back later.

### 3. Delaying work

I was still a bit annoyed by the fact that we had to track `mtime` values for every file. This seems like annoying overhead, especially when doing a single build (no dev mode).

[...]

Issue #19616, opened 2026-01-29
Big slowdown with large amount of directories


timneutkens:
**What version of Tailwind CSS are you using?**

Latest version (and earlier versions)

**What build tool (or framework if it abstracts the build tool) are you using?**

- PostCSS
- Vite
- Webpack

**What version of Node.js are you using?**

All Node.js versions (issue is unrelated to specific version)

**What browser are you using?**

Unrelated to browser.

**What operating system are you using?**

Unrelated to OS (though likely worse on Windows)

**Reproduction URL**

Reproducible on both the Next.js and Vite playground in the repository:

```
DEBUG=tailwindcss:oxide pnpm vite
```

and

```
DEBUG=tailwindcss:oxide pnpm nextjs
```

Make sure to create a bunch of subdirectories, easy way to do that:

```
cd playgrounds/vite && mkdir -p src/components/{auth,dashboard,settings,profile,notifications,messages,search,navigation,footer,sidebar}/{sub1,sub2,sub3} && for dir in src/components/*/sub*; do echo "export const Component = () => <div className=\"test\">test</div>" > "$dir/index.tsx"; done && find src/components -type f | wc -l
```

**Describe your issue**

On every change the Oxide directory scanning applies twice. Files are checked once but directories are checked twice, causing a roughly 2x slowdown on applications with a lot of directories.

Example Oxide debug log for a single line change in a component:

```
2026-01-29T11:50:10.763136Z  INFO scan_sources: tailwindcss_oxide::scanner: enter
2026-01-29T11:50:10.764183Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/app.tsx"
2026-01-29T11:50:10.764256Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/components"
2026-01-29T11:50:10.764300Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/components/sidebar"
2026-01-29T11:50:10.764344Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/components/sidebar/sub1"
2026-01-29T11:50:10.764401Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/components/sidebar/sub2"
2026-01-29T11:50:10.764454Z  INFO scan_sources: tailwindcss_oxide::scanner: Discovering "src/components/sidebar/sub3"
[...]

PR #20408, opened 2026-08-12
Fix slow Vite rebuilds in projects with large gitignored directories

marickvantuil:
Preface: For full transparency, I used AI (Claude) to help dig into this issue we're having and write some of the explanation below.

Our local Laravel application (with Docker) takes between 2-6s seconds (re)building `app.css` after each Blade file update. After checking with Claude, there seems to be some potential unnecessary traversing of big (gitignored) directories, specifically `vendor` and `storage`.

As far as I understand it, two scans happen in tailwindcss:

- a first scan, the content scan. This one reads the project files (auto-detect). It respects .gitignore, so it correctly skips vendor/ and storage/.

- a second scan, the glob resolution (resolve_globs() in the oxide crate). It builds watch patterns that the Vite plugin registers, so a file save can trigger a CSS rebuild. It walks the project again, but this walk does not respect .gitignore. It only skips a hardcoded list of directory names from `ignored-content-dirs.txt` (node_modules, .git, venv, ...). vendor/ and storage/ are not on that list, so it descends into them and stats every file. These directories can never produce a watch pattern, because the content scan never visited them. That work is perhaps wasted?

Maybe why this went unnoticed: on a native filesystem a stat takes about 1µs, so 57k of them is around 0.06s. A Docker bind mount multiplies that by roughly 100. And the big directory in a typical JS project is `node_modules`, which the name list already catches; `vendor/` is the PHP ecosystem's `node_modules`, and it is not on that list. (#19616 / #19632 fixed the same problem for `scanner.files`; the CLI never reads `scanner.globs`, so it was never affected.)

The fix: the second walk now skips any directory the scan never visited, the same pruning the first walk (line 73) already does.

Measured on our local project: `get_globs()` after `scan()` went from 9.9s to somewhere between 0.2 and 1.0s across runs, with identical output (35 globs). A template save now reloads in about 0.3s instead of 4.

## Test plan

- `cargo test` in `crates/oxide`: 66 tests pass, including the glob assertions for nested and gitignored directories in `tests/scanner.rs`.
- Compared the sorted `get_globs()` output before and after on the project above: identical.

I've also set up a repo that reproduces the issue in a fresh Laravel app:

https://github.com/marickvantuil/tailwind-scan-reproduce

The results:

| Environment | vendor/ | scan | get_globs | syscalls under vendor/ |
| --- | --- | --- | --- | --- |
| GitHub Actions (native FS) | present | 5.0 ms | 48.8 ms | 11,410 |
| GitHub Actions (native FS) | absent | 4.5 ms | 0.9 ms | n/a |
| macOS, Docker bind mount | present | 15.8 ms | 599.4 ms | 11,410 |
| macOS, Docker bind mount | absent | 16.5 ms | 6.5 ms | n/a |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Preserve whitespace in `--default(…)` values in custom functional utilities (e.g. `--default(box alphabetic)` no longer becomes `boxalphabetic`) ([#20392](https://github.com/tailwindlabs/tailwindcss/pull/20392))
 - Don't scan gitignored directories (e.g. `node_modules` and `.git`) when the project uses a safelist-style `.gitignore` (e.g. `/*` followed by `!/…` negations) ([#20397](https://github.com/tailwindlabs/tailwindcss/discussions/20397))
 - Ensure root `theme('…')` namespace lookups in JavaScript plugins and config files return the full namespace object instead of the value of its `DEFAULT` key ([#20399](https://github.com/tailwindlabs/tailwindcss/pull/20399))
-- Skip ignored directories entirely when computing watch globs (`scanner.globs`), instead of walking their full contents on every rebuild ([#TBD](https://github.com/tailwindlabs/tailwindcss/pulls))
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
This makes sense to me! I did add a test (which fails if you undo the fix) such that we don't regress on this behavior.

Thanks!

PR #20419, opened 2026-08-14
Don't emit utilities that resolve a theme value when an unsupported modifier is used

webdevelopersrinu:
## Summary

Modifiers on functional utilities that resolve a named theme value are silently
dropped instead of invalidating the candidate. For example, with the default theme:

- `rounded-sm/[5]` emits the same CSS as `rounded-sm` (the `[5]` is ignored)
- `shadow-sm/foo`, `inset-shadow-sm/foo`, `text-shadow/foo`, and `drop-shadow/foo`
  emit the full shadow CSS with the invalid `foo` modifier ignored

This is inconsistent with how every sibling code path behaves: `rounded/foo`,
`rounded-[4px]/foo`, `rounded-sm/5`, and `drop-shadow-xl/foo` all correctly
produce no output, because those paths check `candidate.modifier`.

This PR adds the missing guards:

- In the generic `functionalUtility` handler, a candidate whose named value
  resolves from the theme now rejects modifiers (except fractions like `w-1/2`,
  where the modifier is part of the resolved value).
- The `shadow`, `inset-shadow`, and `text-shadow` utilities now apply the same
  `if (candidate.modifier && !alpha) return` guard in their default-value,
  arbitrary-value, and named-size branches that `drop-shadow` already applies,
  and `drop-shadow` gets it in its default-value branch too.

Existing tests asserting candidates like `drop-shadow/foo` produce no output were
passing for the wrong reason: they run without a theme, so the theme lookup fails
before the modifier is ever considered. The new tests provide a theme so the
invalid modifier is what invalidates the candidate.

## Test plan

- Added assertions to the `rounded`, `filter`, `shadow`, `inset-shadow`, and
  `text-shadow` tests that compile candidates with invalid modifiers against a
  theme that defines the relevant values, and expect no output. All of them fail
  without the fix.
- `pnpm vitest run packages/tailwindcss/src/utilities.test.ts` — 398 passed.


RobinMalfait (approved):
Thanks!

I updated the tests to make them part of the existing section where we expect the tests to fail. 
I also added a missing case for the `stroke-*` utilities, because `stroke-2/foo` would also compiler otherwise.

Last but not least, I also split up your commit, just so I could easily verify that the tests were indeed failing before applying the fix.

Thanks again!

webdevelopersrinu:
Thank you @RobinMalfait for the quick review and merge.
This was my first contribution to Tailwind CSS and the experience was great.

PR #20420, opened 2026-08-14
Don't space out `and`/`or`/`not` inside function calls in `supports-[…]` variants

webdevelopersrinu:
## Summary

The `supports-[…]` variant works around a Chrome bug where `@supports (a)or(b)`
is invalid by spacing out the `and`, `or`, and `not` keywords. However, the
replacement is applied to the entire value, including the inside of function
calls, where these words can be part of a selector.

For example, `supports-[selector(a:not(.foo))]:flex` generates:

```css
@supports selector(a: not (.foo))
```

The selector `a: not (.foo)` is unparsable, so a condition that is true in
every browser silently becomes false and the utility never applies. The same
happens to class names like `.and` or `.or` inside `selector(…)`.

This PR only spaces out the keywords at the condition level: parens preceded by
an identifier (other than the keywords themselves) start a function call, and
everything inside is left as-is. The Chrome workaround still applies to the
condition itself, e.g. `supports-[(display:grid)or(display:flex)]` still
becomes `@supports (display: grid) or (display: flex)`.

## Test plan

- Added a test covering `selector(a:not(.foo))`, class names `.and`/`.or`
  inside `selector(…)`, the Chrome `(a)or(b)` workaround, and a top-level
  `not(…)` condition.
- `pnpm vitest run packages/tailwindcss/src/variants.test.ts` — 102 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Oxide: drop invalid UTF-8 candidates ([#20389](https://github.com/tailwindlabs/tailwindcss/pull/20389))
 - `@tailwindcss/vite` no longer forces a full page reload for external files (e.g.: `.php` files) ([#20414](https://github.com/tailwindlabs/tailwindcss/issues/20414))
 - Canonicalization: don't merge utilities that reference different theme variables set to CSS-wide keywords like `unset` ([#20417](https://github.com/tailwindlabs/tailwindcss/pull/20417))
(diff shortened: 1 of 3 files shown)


RobinMalfait (approved):
Thanks!

I updated the code to make use of an existing value parser and mutating the AST, instead of using a bunch of regexes.

webdevelopersrinu:
Thank you @RobinMalfait for the review and the refactor to the value parser  that approach is much cleaner than the regex tokenizer, and I learned a lot comparing the two. Happy to see both fixes land

PR #20423, opened 2026-08-20
Fix utility staticValues shadowed by fallback theme namespaces

koreahghg:
## Summary

Some functional utilities resolve their value against two theme namespaces: their own dedicated one, and a shared fallback used by many unrelated utilities (e.g. `leading` uses `['--leading', '--spacing']`, `columns` uses `['--columns', '--container']`, `backdrop-blur` uses `['--backdrop-blur', '--blur']`). These utilities also define `staticValues` for special keywords like `none`/`auto`.

The `staticValues` fallback only kicked in when the theme resolved to nothing at all across *every* key in `themeKeys`. This means a same-named custom theme value defined for a completely different, unrelated utility silently wins over the intended static value:

```css
@theme {
  --spacing-none: 0; /* meant for spacing utilities, e.g. p-none */
}
```

```html
<p class="leading-none">...</p>
<!-- outputs line-height: var(--spacing-none) → 0, instead of the intended line-height: 1 -->
```

This fixes #19722 (`leading-none` shadowed by a custom `--spacing-none`), and also fixes two more previously unreported instances of the exact same root cause:

- `columns-auto` shadowed by a custom `--container-auto` (outputs `columns: 500px` instead of `columns: auto`)
- `backdrop-blur-none` shadowed by a custom `--blur-none` (doesn't properly reset the backdrop blur)

**Fix:** `functionalUtility()` now also resolves the candidate against just the utility's own dedicated namespace (`themeKeys[0]`). A `staticValues` entry is only shadowed if that resolution (not the combined fallback resolution) finds a value — i.e. only when the utility's *own* namespace defines something for that name. A same-named value that only exists in a shared fallback namespace no longer suppresses the static value.

## Test plan

- Added regression tests in `utilities.test.ts` for all three affected utilities (`leading-none`, `columns-auto`, `backdrop-blur-none`), each reproducing the shadowing with a colliding custom theme value and asserting the correct static output.
- `pnpm vitest run packages/tailwindcss/src` — all 5007 tests pass (1 pre-existing skip), no regressions from this change.

### DIFF
--- a/packages/tailwindcss/src/utilities.test.ts
+++ b/packages/tailwindcss/src/utilities.test.ts
@@ -8787,6 +8787,27 @@ test('columns', async () => {
     }
     "
   `)
+
+  // A custom `--container-auto` value must not shadow the `columns-auto`
+  // static value: https://github.com/tailwindlabs/tailwindcss/issues/19722
+  expect(
+    await run(
+      ['columns-auto'],
+      css`
+        @theme {
+          --container-auto: 500px;
+        }
+        @tailwind utilities;
+      `,
+    ),
+  ).toMatchInlineSnapshot(`
+    "
+    .columns-auto {
+      columns: auto;
+    }
+    "
+  `)
+
   expect(
     await run([
       'columns',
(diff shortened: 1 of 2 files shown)


koreahghg:
Closing this — I hadn't seen #19773 and #20394 when I opened this. This PR uses the same approach as #19773 (treating `themeKeys[0]` as the utility's own namespace generically in `functionalUtility`), which was already flagged as an overreach in the discussion on #19722 (it doesn't hold for utilities like `min-h` where more than one key is legitimately utility-specific, not just a generic fallback like `--spacing`).

#20394's scoped, opt-in approach (`fallbackThemeKeys`) looks like the right pattern here. I'll follow up separately, scoped just to the `columns-auto`/`--container-auto` and `backdrop-blur-none`/`--blur-none` cases, which aren't covered by #20394 since those utilities don't go through `spacingUtility()`.

PR #20425, opened 2026-08-20
Don't crash when theme.keyframes is explicitly set to null

koreahghg:
## Summary

Setting `theme.keyframes` (or `theme.extend.keyframes`) to `null` in a JS config crashes the entire build.

`keyframesToRules` in `packages/tailwindcss/src/compat/apply-keyframes-to-theme.ts` checks `'keyframes' in resolvedConfig.theme` before calling `Object.entries(resolvedConfig.theme.keyframes)`. But `resolveConfig` (see `packages/tailwindcss/src/compat/config/resolve-config.ts`) resolves every theme value with `value ?? null`, so a theme key explicitly set to `null` — the documented way to clear a theme key, e.g. `theme.extend.colors: null` — still exists on the object via `in`, just with a `null` value. That means `theme: { keyframes: null }` reaches `Object.entries(null)`, which throws `TypeError: Cannot convert undefined or null to object` and crashes the whole build.

Every sibling module that reads from the resolved theme this way (`container.ts`, `screens-config.ts`, etc.) already guards against this with a falsy/`|| {}` check — this file was the one outlier still using `in`.

## Fix

Swap the `'keyframes' in resolvedConfig.theme` check for a truthy check on `resolvedConfig.theme.keyframes` itself, matching the pattern used elsewhere in `compat/`.

## Test plan

- Added a regression test to `apply-keyframes-to-theme.test.ts`: `theme: { keyframes: null }` no longer throws and correctly results in no keyframes.
- Verified the new test fails with the exact reported crash on the code prior to this fix, and passes after.
- `pnpm vitest run` — all 5008 tests in the `tailwindcss` package pass (5007 existing + 1 new), no regressions.
- `tsc --noEmit` shows no new type errors introduced by this change (pre-existing, unrelated errors exist elsewhere in the monorepo from packages that require a built `@tailwindcss/oxide` native module).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Canonicalization: don't merge utilities that reference different theme variables set to CSS-wide keywords like `unset` ([#20417](https://github.com/tailwindlabs/tailwindcss/pull/20417))
 - Don't generate utilities when a modifier is used that would otherwise be silently ignored (e.g. `rounded-sm/[5]`, `shadow-sm/foo`, `stroke-2/50`) ([#20419](https://github.com/tailwindlabs/tailwindcss/pull/20419))
 - Only normalize top-level `and`, `or`, and `not` keywords in `supports-[…]` variants (e.g. `selector(a: not (.foo))` → `selector(a:not(.foo))`) ([#20420](https://github.com/tailwindlabs/tailwindcss/pull/20420))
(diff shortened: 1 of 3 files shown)


koreahghg:
Thanks for the review!

Re: the suggestion to also cover \`theme.extend.keyframes: null\` — good catch that it wasn't covered, but worth noting that path never actually hit the crash: \`deepMerge\` skips \`null\`/\`undefined\` sources entirely (see \`compat/config/deep-merge.ts\`), so the resolved value there is always \`{}\`, not \`null\`. I've pushed a follow-up commit that:
- corrects the changelog line, which incorrectly implied the \`extend\` path was also crashing
- adds a dedicated regression test for \`theme.extend.keyframes: null\` anyway, so a future change to \`deepMerge\`/\`resolveConfig\` can't silently reintroduce a crash there too

Full suite still green (5009/5010, 1 pre-existing skip).

PR #20427, opened 2026-08-21
Cold start performance improvements

smnbbrv:
## Summary

Hi guys, thanks for you great work!

I work on performance improvements on https://github.com/schoero/eslint-plugin-better-tailwindcss project. Part of issues can be fixed on the tailwind side only.

This is a first fix, I have a bigger one in mind, it would require a small additional public API method - out of scope of this PR.

Please let me know what you think.

## Finding

`getVariantOrder()` re-sorts all parsed variants with the (expensive) variant comparator on every call, and it is called by every compileCandidates() invocation — per build pass, per @apply substitution, and hundreds of times during candidate canonicalization via the variant signature caches. Since parsedVariants is append-only, the computed order only changes when a new variant is parsed, so we can cache the result and invalidate on parsedVariants.size. This makes cold canonicalization (IntelliSense, lint plugins) faster and removes repeated sorting from @apply substitution and incremental rebuilds.

## Test plan

Tested at https://github.com/smnbbrv/better-tailwindcss-bench . The relevant part is the `epbt-now / tw-patched`

**plugin cost** (ms, cold run net of parse baseline)

| codebase   | epbt-now / tw-now | epbt-patched / tw-now | epbt-now / tw-patched | epbt-patched / tw-patched |
| ---------- | ----------------: | --------------------: | --------------------: | ------------------------: |
| mixed      |              6292 |           3965 (-37%) |           5395 (-14%) |               3595 (-43%) |
| repetitive |              5279 |           2425 (-54%) |            4819 (-9%) |               2124 (-60%) |
| unique     |              6096 |           5458 (-10%) |           5507 (-10%) |               5295 (-13%) |

### DIFF
--- a/packages/tailwindcss/src/design-system.ts
+++ b/packages/tailwindcss/src/design-system.ts
@@ -75,6 +75,8 @@ export function buildDesignSystem(
   let variants = createVariants(theme)
 
   let parsedVariants = new DefaultMap((variant) => parseVariant(variant, designSystem))
+  let cachedVariantOrder: Map<Variant, number> | null = null
+  let cachedVariantOrderSize = -1
   let parsedCandidates = new DefaultMap((candidate) =>
     Array.from(parseCandidate(candidate, designSystem)),
   )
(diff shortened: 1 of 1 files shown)


RobinMalfait (approved):
Thanks for the PR!

I wish it was a bit more co-located, and maybe used a `DefaultMap`, but that has other issues where now we would grow the map forever.

I like it, thanks!

PR #20434, opened 2026-08-26
Don't warn about Angular's `::ng-deep` and `:host-context()` when optimizing CSS

### LINKED ISSUE #20433: Optimizer warns on Angular's `::ng-deep` — the `:deep`/`:slotted`/`:global` filter doesn't cover it

v4.3.3 (also reproduces on v4.1.18)

`@tailwindcss/postcss` 4.3.3 with postcss 8.5.6. Originally hit via Angular 21, but the reproduction uses postcss directly — no framework needed.

v22.21.1

N/A — this is build-time console output.

macOS

**Reproduction URL**

https://gist.github.com/weilinzung/ace42ceb95c2f47b6747cc1f34f162bd

```
npm install && npm run build
```

(Tailwind Play can't reproduce this one — the warning comes from `optimizeCss`, which Play doesn't run.)

**Describe your issue**

`optimizeCss` already filters Lightning CSS warnings for the framework-specific deep selectors Vue and Svelte emit, but Angular's `::ng-deep` isn't in the list, so every Angular component stylesheet using it prints a warning block per occurrence.

Given this component stylesheet:

```css
@reference './app.css';

:host {
  ::ng-deep .some-child,
  ::ng-deep .other-child {
    @apply text-sm;
  }
}
```

Actual output:

```
Found 2 warnings while optimizing generated CSS:

Issue #1:
│ :host ::ng-deep .some-child, :host
┆        ^-- 'ng-deep' is not recognized as a valid pseudo-element. Did you mean ':ng-deep' (pseudo-class) or is this a typo?
```

Expected: no warnings, same as `:deep()` / `:slotted()` / `:global()`.

### Why the existing filter misses it

`packages/@tailwindcss-node/src/optimize.ts:67`

```ts
if (/'(deep|slotted|global)' is not recognized as a valid pseudo-/.test(warning.message)) {
```

The message reads `'ng-deep' is not recognized…`, so the pattern's leading `'` never matches — the character before `deep` is `-`.

### Suggested fix

```diff
-    if (/'(deep|slotted|global)' is not recognized as a valid pseudo-/.test(warning.message)) {
+    if (/'(ng-deep|deep|slotted|global)' is not recognized as a valid pseudo-/.test(warning.message)) {
```

`::ng-deep` is Angular's supported deep-selector spelling — the compiler strips it during view-encapsulation shimming (`_shadowDeepSelectors = /(?:>>>)|(?:\/deep\/)|(?:::ng-deep)/g`), so it never reaches a browser. It's the direct analogue of the three pseudo-classes this filter already covers.

Worth noting that `/deep/` and `>>>`, Angular's two other spellings of the same thing, already pass silently because `nonStandard.deepSelectorCombinator` is enabled. `::ng-deep` is the only one that warns, and it's the spelling the Angular docs use — so in practice every Angular codebase hits this.

There's a second, smaller effect: the unrecognized pseudo-element also stops Lightning CSS from keeping the selector list together, so `:host ::ng-deep .a, :host ::ng-deep .b { … }` is emitted as two separate rules rather than one.

For scale, on our workspace (7 Angular apps, 69 stylesheets using `::ng-deep`) a production build prints 61 warning blocks / 430 lines.

### Precedent

- [#20277](https://github.com/tailwindlabs/tailwindcss/pull/20277) added an `@position-try` suppression to this same function in v4.3.2.
- [#10943](https://github.com/tailwindlabs/tailwindcss/pull/10943) already special-cases `::ng-deep` when applying utilities.

Happy to open a PR with the one-liner if that's useful.

weilinzung:
Fixes #20433.

The warning filter in `optimize.ts` already ignores `:deep()`, `:slotted()` and `:global()`. Angular's two deep selectors are the same kind of thing — non-standard pseudo-selectors that the framework's compiler resolves before the CSS reaches a browser — but they aren't covered, so every Angular component stylesheet using them prints a warning block per occurrence.

```
Found 2 warnings while optimizing generated CSS:

Issue #1:
│ :host ::ng-deep .some-child, :host
┆        ^-- 'ng-deep' is not recognized as a valid pseudo-element. Did you mean ':ng-deep' (pseudo-class) or is this a typo?
```

Angular strips both during view-encapsulation shimming — `::ng-deep` via `_shadowDeepSelectors = /(?:>>>)|(?:\/deep\/)|(?:::ng-deep)/g`, and `:host-context()` in the same pass — so neither ever reaches a browser.

Worth noting that `/deep/` and `>>>`, Angular's two other spellings of the deep selector, already pass silently because `nonStandard.deepSelectorCombinator` is enabled. `::ng-deep` is the only one that warns, and it's the spelling the Angular docs use — so in practice every Angular codebase hits this. On the workspace where I ran into it (7 Angular apps), a production build printed 425 `ng-deep` warnings and 2 `host-context` ones.

## Test plan

There's no automated coverage here because the warning path is behind `process.env.NODE_ENV !== 'test'`, so a spy sees nothing under Vitest regardless of the filter — the same reason #20277 shipped without one. Instead I ran the file before and after the change against Lightning CSS 1.33.0 directly, counting emitted warning blocks:

| Input | Before | After |
| --- | --- | --- |
| `:host ::ng-deep .a, :host ::ng-deep .b { … }` | 1 | 0 |
| `:host-context(.dark) .a { … }` | 1 | 0 |
| `:deep(.a) { … }` | 0 | 0 |
| `.a::totally-not-real { … }` — genuine typo | 1 | **1** |

Generated CSS is byte-identical before and after; only the warning is suppressed. Genuine unknown pseudo-selectors still warn, so the typo hint the message exists for is preserved.

Happy to restructure this if you'd prefer the predicate extracted so it can be unit-tested, or to split `ng-deep` and `host-context` into their own block rather than extending the existing regex.

A minimal reproduction of the original issue is at https://gist.github.com/weilinzung/ace42ceb95c2f47b6747cc1f34f162bd.


packages/@tailwindcss-node/src/optimize.ts
```diff
@@ -61,10 +61,16 @@ export function optimize(
...
   result.warnings = result.warnings.filter((warning) => {
-    // Ignore warnings about unknown pseudo-classes as they are likely caused
-    // by the use of `:deep()`, `:slotted()`, and `:global()` which are not
-    // standard CSS but are commonly used in frameworks like Vue.
-    if (/'(deep|slotted|global)' is not recognized as a valid pseudo-/.test(warning.message)) {
+    // Ignore warnings about unknown pseudo-classes and pseudo-elements as they
+    // are likely caused by the use of `:deep()`, `:slotted()`, and `:global()`
+    // in frameworks like Vue, or `::ng-deep` and `:host-context()` in Angular.
+    // None of these are standard CSS, and each is resolved by the framework's
+    // own compiler before the CSS reaches a browser.
```
RobinMalfait:
```suggestion
    // Ignore warnings about unknown pseudo-classes that are used in frameworks
    // such as Vue or Angular and are handled by their own compilers, before
    // reaching the browser.
```

RobinMalfait (approved):
Thanks!

PR #11, opened 2017-11-01
Fix some times typo

mdavis1982:
Changes `some times` to `sometimes` in "Functions and Directives" section of the docs.

PR #20436, opened 2026-08-27
Do not migrate a variant literal behind a conditional

### LINKED ISSUE #20435: Upgrade codemod rewrites `variant="outline"` to `"outline-solid"` inside conditional expressions

`@tailwindcss/upgrade` v4.3.3, migrating a project from v3.4.17 to v4.3.3.

postcss 8.5.26 (the reproduction is a bare PostCSS project; no framework). The bug is in
the upgrade codemod's template migration, so the build tool is incidental.

v25.1.0 (npm 11.6.2)

N/A

macOS 26.6.2 (Darwin 25.6.0, arm64)

**Reproduction URL**

https://github.com/carlos-cubas/tailwind-upgrade-outline-repro

```sh
npm install
npx @tailwindcss/upgrade@4.3.3 --force
git diff
```

Seven files, fourteen lines of TSX. It contains four control cases (the forms #18922
guards, which correctly survive) and four conditional cases plus one comment (which do
not). Verified from a clean clone.

**Describe your issue**

`@tailwindcss/upgrade` rewrites `variant="outline"` to `variant="outline-solid"` when a
conditional expression sits between `variant` and the string literal. These are React prop
values, not class names, so the rewrite silently breaks them.

PR #18922 fixed this for the direct forms, and those still work. The guard it added is a
look-behind regex in `is-safe-migration.ts`:

```js
/variant\s*[:=]\s*\{?['"`]$/
```

Because it only inspects the text immediately preceding the candidate, any conditional
between `variant` and the literal defeats it.

From the reproduction — the four control cases are untouched, the four conditional cases
are all rewritten:

| | Input | After upgrade |
|---|---|---|
| ✅ | `<Button variant="outline" />` | unchanged |
| ✅ | `<Button variant={"outline"} />` | unchanged |
| ✅ | `Button({ variant: "outline" })` | unchanged |
| ✅ | `function c4({ variant = "outline" })` | unchanged |
| ❌ | `Button({ variant: isActive ? "outline" : "ghost" })` | `"outline-solid"` |
| ❌ | `<Button variant={first ? "default" : "outline"} />` | `"outline-solid"` |
| ❌ | `<Button variant={variant ?? "outline"} />` | `"outline-solid"` |
| ❌ | `<Button variant={required ? 'secondary' : 'outline'} />` | `"outline-solid"` |

Comments are also not excluded from candidate extraction. This line is rewritten too, and
`outline-solid: none` is not valid CSS:

```diff
-// this comment mentions `outline: none` as CSS prose
+// this comment mentions `outline-solid: none` as CSS prose
```

**How loud it is depends on the call site.** In our project the codemod hit four call
sites in shadcn/ui-derived components. Three flowed into a `cva`-typed `variant` prop and
`tsc` did reject them:

```
error TS2322: Type '"default" | "outline-solid"' is not assignable to type
'"link" | "default" | "destructive" | "outline" | "secondary" | "ghost"'.
```

So for typed variants this is a noisy break rather than a silent one, which is worth
saying plainly. The fourth was a plain data attribute:

```jsx
<div data-variant={variant ?? "outline"} />   // becomes "outline-solid"
```

Nothing type-checks that, and the sibling `group-data-[variant=ghost]/…` selectors that
read it keep working, so the wrong value just sits there. Comment prose is silent for the
same reason. A project on JavaScript, or with a looser variant type, gets no signal on
any of the four.

**Expected behaviour.** A string literal that is the value of a `variant` prop should not
be migrated regardless of what expression it sits inside, and candidates inside comments
should not be migrated at all.

I appreciate the immediate look-behind cannot see through an arbitrary expression, so this
may want a different check rather than a longer regex — happy to open a PR against
`is-safe-migration.ts` with cases added to the existing table in
`is-safe-migration.test.ts` if you would like it in that form.

HuzaifaChaudary:
Closes #20435

## problem

`@tailwindcss/upgrade` rewrites `variant="outline"` to `variant="outline-solid"` when a conditional sits between `variant` and the string. these are react prop values and not class names so the rewrite breaks them.

#18922 added a guard for the direct forms and those still work. the guard is a look behind that only inspects the text immediately before the literal

```js
/variant\s*[:=]\s*\{?['"`]$/
```

so as soon as a ternary or a nullish branch comes between the two it no longer matches. the four cases from the issue

```jsx
Button({ variant: isActive ? "outline" : "ghost" })
<Button variant={first ? "default" : "outline"} />
<Button variant={variant ?? "outline"} />
<Button variant={required ? 'secondary' : 'outline'} />
```

## fix

find where the `variant` value starts and follow it until whatever ends its own syntax. the candidate is the last thing on the line so the value it sits in is the one that has not ended yet.

- `variant={…}` ends at its matching brace
- `variant="…"` ends at its own quote which is what keeps the match inside a single attribute and away from a neighbouring `:class`
- `{ variant: … }` ends at a comma or at the brace around it

the walk skips strings and balanced groups as it goes so there is no nesting limit. two things are deliberately not part of the value

- an argument passed to a call so `<Button variant={cn("ghost", a ? "shadow" : "x")} />` still migrates `shadow`
- a `variant:` that only appears inside some other attribute text so `<div title="variant: outline" className={b ? "shadow" : "x"}>` still migrates `shadow`

`CONDITIONAL_TEMPLATE_SYNTAX` is byte identical to `main`. the diff only adds one more heuristic below it plus the two helpers it needs.

## tests

21 negatives in `is-safe-migration.test.ts` and 14 positives.

- the four from the issue plus every shape the reviews surfaced
- object literals nested to any depth and braces inside strings
- positives in the other direction because an over broad guard would silently stop migrating a real class

| | tests |
|---|---|
| `is-safe-migration` on `main` | 68 passed |
| `is-safe-migration` here | 92 passed |
| `@tailwindcss/upgrade` on `main` | 415 passed across 28 files |
| `@tailwindcss/upgrade` here | 439 passed across 28 files |

with the source reverted to `main` and the tests kept 17 of the 21 shadcn negatives fail. prettier is clean on both files.

## not covered

the guard is line scoped like every other entry in `CONDITIONAL_TEMPLATE_SYNTAX` so a value wrapped across lines is still not seen

```jsx
<Button variant={
  active ? "outline" : "ghost"
} />
```

that limit is shared by the whole array and by `isMiddleOfString` so lifting it is a separate change rather than something to fold in here.

the issue also notes that a comment mentioning `outline: none` gets rewritten. that one is candidate extraction rather than this safety guard so it is a different fix and i have left it alone.

---

disclaimer: this contribution was prepared with the help of an ai agent. i reviewed the change, reproduced the failing cases against `main` first, and ran the package test suite locally before opening it.


HuzaifaChaudary:
thanks, two of the three were real and are now fixed in `893ab46a`.

**a call in the condition.** `variant={isActive(foo, bar) ? "outline" : "ghost"}` was not guarded. my `[^{},]` ended the match at the comma inside the call. a parenthesised group is now consumed as a unit so its own commas stay inside it.

**a parenthesised branch.** `variant={cond ? ("outline") : "ghost"}` was not guarded either, because the `(` sat between the operator and the literal. parens are now allowed there. covered in both arms.

the guard is now

```js
/variant\s*[:=]\s*\{?(?:(?:[^{},()]|\([^()]*\))*?(?:\?\?|\?|:)\s*)?\(*\s*['"`]$/
```

a bare comma or brace still ends the match, which is what keeps a neighbouring `className` out of reach. `Button({ variant: a ? "ghost" : "x", className: b ? "shadow" : "y" })` still migrates `shadow`, and there is a test for it.

**on the brace comment.** i think that one is a misread. `[^{},]` is a negated class over `{`, `}` and `,`, so both braces were already excluded, not just the opening one. the current class `[^{},()]` excludes both parens as well.

**on multiline values.** correct, and it is not something this change introduces. `currentLineBeforeCandidate` is built by walking back until `\n`, so every entry in `CONDITIONAL_TEMPLATE_SYNTAX` is line scoped, not just this one. making the guards span lines would change behaviour for all of them, so it felt like the wrong thing to fold into a targeted fix. happy to open a separate issue for it if you want that tracked.

tests are now 78 in `is-safe-migration` and 425 across `@tailwindcss-upgrade`, up from 68 and 415 on `main`. the three new negatives fail without the change.

HuzaifaChaudary:
both of these were right, fixed in `3730f8a5`.

**the vue one is the important one, and it was a regression i introduced.** `<div variant="ghost" :class="active ? 'shadow' : 'none'">` was being guarded, so a class that should migrate no longer did. that is worse than the bug this pr set out to fix. vue writes attributes without braces, so my `[^{},()]` gap crossed `"ghost"` and walked straight on into `:class`.

the rule i had wrong is when a quote may be crossed. inside a brace a quoted ternary branch is ordinary, so `variant={first ? "default" : "outline"}` has to cross `"default"`. outside a brace a quote closes the attribute value, so crossing one means leaving the attribute. the guard now splits on exactly that:

```
\{(?:[^{}]|PAREN)*?(?:\?\?|\?|:)\s*     braced, quotes may be crossed, braces may not
(?:[^{},()'"`]|PAREN)*?(?:\?\?|\?|:)\s* bare, a quote ends it
\{                                       the plain variant={"..."} form
```

**nested parens.** `variant={isActive(getState(foo, bar), x) ? "outline"}` now matches too, one level of nesting.

i pulled the pattern out into named parts above the array rather than growing the inline literal. three reviewers have now had to reason about this regex and the one liner was getting hard to read, so the parts are named and commented.

two new tests, one in each direction, and both fail on the previous commit:

- `<Button variant={isActive(getState(foo, bar), x) ? "outline" : "ghost"} />` must not migrate
- `<div variant="ghost" :class="active ? 'shadow' : 'none'">` must still migrate

`is-safe-migration` is 80 tests, `@tailwindcss-upgrade` is 427 across 28 files, up from 68 and 415 on `main`. prettier clean.

thanks for the catch, the vue case is one i should have tested for myself given the whole point of the guard is not to suppress real classes.

HuzaifaChaudary:
all three were right. fixed in `99d920d2`, and i changed the shape rather than patch the pattern again.

three rounds of review have now found holes in this guard, and looking at why, they were all the same underlying mistake: one regex trying to know where a value ends in three different syntaxes at once. every time i widened it for one syntax it either missed another or reached into a neighbouring attribute. so it is now one rule per form, each stopping at what actually ends that form. the array already tests each rule independently so nothing else had to change.

```
variant\s*[:=]\s*\{?['"`]$                                      the plain forms
variant\s*[:=]\s*\{(?:[^{}]|PAREN)*?TAIL                        jsx, a brace ends it
variant\s*=\s*(["'])(?:(?!\1)[^{}])*?TAIL                       attribute, its own quote ends it
variant\s*:\s*(?:[^{},]|PAREN)*?TAIL                            object, a comma or brace ends it
```

that covers the three you found:

- `:variant="active ? 'ghost' : 'outline'"` — the attribute rule. the backreference is what keeps it inside its own value, which is also what stops the `:class` over match from the last round
- `variant: (value ?? "outline")` — the object rule, an unbalanced opening paren is fine there now
- `variant: theme === "dark" ? "outline" : "ghost"` — the object rule too. quotes are ordinary inside an object value, a comma or brace is what ends it

three more negative tests, all failing on the previous commit. `is-safe-migration` is 83, `@tailwindcss-upgrade` is 430 across 28 files, against 68 and 415 on `main`. prettier clean.

happy to keep going if there is another hole, but if the maintainers would rather this whole heuristic move off regexes and onto a real parse of the surrounding expression, that is a bigger change than i should make unasked and i would open a separate issue for it instead.

HuzaifaChaudary:
both fixed in `75e4e2f6`. an object literal is now allowed as a unit inside a condition, the same way a parenthesised group already was, so its braces do not read as the end of the value.

- `:variant="isActive({ x: 1 }) ? 'ghost' : 'outline'"`
- `<Button variant={{ tone: "dark" }.tone === "dark" ? "outline" : "ghost"} />`

both are covered by tests and both fail on the previous commit. `is-safe-migration` is 85, `@tailwindcss-upgrade` is 432 across 28 files, against 68 and 415 on `main`.

### where i think this actually stands

this is the fifth round, and i want to be straight about the trend rather than keep quietly widening things.

@coderabbitai is right that the real answer is scanning the expression with balanced brace, paren and quote state instead of matching it. a regex can only ever handle a bounded nesting depth. `{ a: { b: { c: 1 } } }` inside a condition would still defeat what i just pushed, and i am not going to pretend otherwise by adding a third nesting level.

what i am less sure about is whether that scanner belongs in this pull request. every entry in `CONDITIONAL_TEMPLATE_SYNTAX` is a line scoped regex, and the whole array shares the same limits, so replacing the approach for one entry would leave the file half in each world and would be a much larger diff than the bug being fixed here warrants.

so my suggestion, and it is only a suggestion since this is your file:

- take this as is if the four reported forms plus the eleven the reviews found are worth fixing now. it is strictly better than `main` and it cannot suppress a real class, which is the failure mode i care most about and which has a test in both directions
- or tell me you want the scanner and i will do it properly as a replacement for the array rather than bolted next to it

either is fine by me. i would rather ask than keep adding nesting levels until someone finds level four.

HuzaifaChaudary:
both of these are real, and both are the ceiling i described in the last comment rather than new defects. a brace inside a string and an object nested past one level are the same structural limit: a regex cannot track balanced state, so there is always a next case. i said i would not keep adding levels and i am going to hold to that.

what i did do is measure which direction the remaining gaps fail in, because that is the part that decides whether this is mergeable as it stands.

**they only fail by under guarding.** for both of your cases the guard does not fire, so the prop is rewritten as a class, which is exactly what `main` does today for every conditional form. no case gets worse than it is now.

```
<Button variant={{ tone: "}" }.tone === "}" ? "outline" : "ghost"} />   not guarded
<Button variant={{ a: { b: { c: 1 } } }.a ? "outline" : "ghost"} />     not guarded
```

**nothing over guards.** the failure i actually care about is a real class silently no longer migrating, since that is worse than the bug this fixes. i checked that direction explicitly, including with your new shapes:

```
<div variant="ghost" :class="active ? 'shadow' : 'none'">     still migrates
<Button variant="ghost" className={active ? "shadow" : "x"}/> still migrates
Button({ variant: a ? "g" : "x", className: b ? "shadow" : "y" }) still migrates
<div :class="isActive({ x: 1 }) ? 'shadow' : 'none'">          still migrates
<div className={{ a: "}" }.a ? "shadow" : "x"}>                still migrates
```

so the state of this pr is: fixes the four forms from #20435 plus eleven more the reviews surfaced, cannot suppress a class that migrates today, and leaves a shrinking tail of exotic shapes behaving exactly as they already do on `main`.

that seems like a reasonable place to stop to me, but it is your call and i am happy either way:

- merge as is and let the tail stay a known limit
- or say the word and i will replace these four rules with a small backward scanner that tracks brace, paren and quote state properly. i would keep it to the `variant` rules and leave the other entries in the array alone, so it stays a contained change rather than a rewrite of the heuristic

i would rather do that once, deliberately, than keep going a level at a time.

HuzaifaChaudary:
pushed `518310c2`. both open findings were real. i stopped widening the pattern and replaced it.

### what i verified first

@coderabbitai on a brace inside a string and @greptile-apps on an object nested past one level. i reproduced both against `75e4e2f6` before changing anything.

```
<Button variant={{ tone: "}" }.tone === "}" ? "outline" : "ghost"} />   outline -> outline-solid
<Button variant={{ a: { b: { c: 1 } } }.a ? "outline" : "ghost"} />    outline -> outline-solid
```

both real and both are the depth ceiling i described last time. so i did the scan rather than add a third nesting level.

### something neither review found

while writing probes i found the previous commit **over guarding** which is the direction i claimed was safe.

```
<Button variant={cn("ghost", active ? "shadow" : "none")} />
```

`shadow` migrates on `main` and stopped migrating on `75e4e2f6`. the braced rule let `[^{}]` run straight through the unclosed `cn(` so anything after a call inside a variant prop got swallowed. it was not even consistent because `variant={cn({ className: b ? "shadow" : "x" })}` still migrated. that was a regression i introduced and no review caught it. i want to correct my earlier comment where i said nothing over guards.

### what it does now

walk the value from the prop until whatever ends its own syntax. the candidate is the last thing on the line so the value it sits in is the one that has not ended yet.

- `variant={…}` ends at its matching brace
- `variant="…"` ends at its own quote which is what keeps a match inside one attribute and away from a neighbouring `:class`
- `{ variant: … }` ends at a comma or at the brace around it

strings and balanced groups are skipped while walking so nesting has no limit. two things are deliberately not part of the value. an argument passed to a call because it is passed to that call and not to the prop. and a `variant:` that only appears inside some other attribute text.

`CONDITIONAL_TEMPLATE_SYNTAX` is now byte identical to `main`. the whole diff to that file is one extra heuristic below the loop plus the two helpers it needs. the four composed regexes and the three `String.raw` fragments are gone.

### evidence

each fix reverted on its own and watched to fail.

**source reverted to `main` with the tests kept**

```
× does not replace classes in invalid positions #61 .. #77
Tests  17 failed | 75 passed (92)
```

**source reverted to `75e4e2f6` with the tests kept**

```
× does not replace classes in invalid positions #74   { tone: "}" } brace in a string
× does not replace classes in invalid positions #75   { a: { b: { c: 1 } } } two levels
× replaces classes in valid positions #12             variant={cn("ghost", active ? "shadow" : "none")}
Tests  3 failed | 89 passed (92)
```

**only the in a string check removed**

```
× replaces classes in valid positions #13   <div title="variant: outline" className={b ? "shadow" : "none"}>
Tests  1 failed | 91 passed (92)
```

**with the fix**

```
is-safe-migration        92 passed
@tailwindcss/upgrade    439 passed across 28 files
```

against 68 and 415 on `main`. `prettier --check` is clean on both files and `tsc --noEmit` reports nothing in this file.

### on ci

`CI` and `Integration Tests` have never actually run on this branch. every run on every push is `action_required` because i am a first time contributor, so the checks are waiting for a maintainer to approve the workflow rather than failing. that is what `mergeable_state` is blocked on. i ran `pnpm exec vitest run --hideSkippedTests` for the package and `prettier --check` locally on node 24 which is what the workflow uses.

happy to keep iterating if anything here reads wrong.

HuzaifaChaudary:
@greptile-apps i ran this one rather than argue it, because the claim is specific enough to settle.

that exact line is already a test. it went in with `518310c2` as case #11 of the valid positions block:

```
[`<Button variant={{ a: 1 }} className={b ? "shadow" : "none"} />`, 'shadow', 'shadow-sm'],
```

it passes. 92 passed, 0 failed.

then i applied the change you asked for, `isOpenValue(line, start)` in place of `isOpenValue(line, start + 1)`, and ran it again:

```
FAIL  is-safe-migration.test.ts > replaces classes in valid positions #11
AssertionError: expected 'shadow' to deeply equal 'shadow-sm'
Tests  1 failed | 91 passed (92)
```

case #11 is the line you gave. so the change is the thing that breaks it, which is the opposite way round from the report.

the reason is the `groups.length === 0` return inside `isOpenValue`. the string being scanned is `<Button variant={{ a: 1 }} className={b ? "`. starting at `start + 1` the scan is past the outer brace, so `{ a: 1 }` pushes and pops, and the next `}` finds an empty stack and returns false right there. it never reaches `className={`. starting at `start` instead puts the outer brace on the stack, the outer `}` only pops it back to empty, and the scan then walks on into `className={` and ends with that group open, which is the rejection you predicted.

so including the outer brace is what causes the walk past the prop, not what prevents it.

worth saying the underlying point was fair. that boundary is where this thing would break if it broke, i just had it covered already.

HuzaifaChaudary:
ran both p1s against `36917ce3` rather than reason about them. one does not reproduce, one does, and the one that does is in direct conflict with a test that is already green here.

| case | candidate | result |
|---|---|---|
| `<div :variant="active ? 'outline' : 'ghost'"></div>` | `outline` | **guarded**, stays `outline` |
| `<div :variant="active ? 'ghost' : 'outline'"></div>` | `outline` | guarded |
| `<Button variant={getVariant(cond ? "outline" : "ghost")} />` | `outline` | **migrates to `outline-solid`** |
| `<Button variant={cn("ghost", active ? "shadow" : "none")} />` | `shadow` | migrates to `shadow-sm` |

**first branch bypasses guard, does not reproduce.** the reasoning was that a later outer quote makes `line.includes` see the value as closed. there is no later quote to find. the guard reads `currentLineBeforeCandidate`, so for the first branch the string it sees is

```
<div :variant="active ? '
```

the `"` opened at `variant="` and nothing closes it before the candidate, so `!line.includes(char, start + 1)` is true and the value is correctly treated as open. first and second branch behave the same, and both are guarded above.

**calls expose conditional variant literals, reproduces.** confirmed, `outline` becomes `outline-solid` there.

but look at the last two rows together. they are the same shape, a string literal inside a call inside `variant={...}`, and they need opposite answers:

```jsx
<Button variant={getVariant(cond ? "outline" : "ghost")} />   must NOT migrate
<Button variant={cn("ghost", active ? "shadow" : "none")} />  must migrate
```

nothing in the syntax separates them. it depends on what the callee does with the argument, which the codemod cannot know. `cn` composes class names so its arguments are classes, `getVariant` computes a variant so its argument is a variant.

the current rule picks the `cn` side, and that is not accidental. treating a call group as opaque is exactly the over-guard i introduced earlier in this pull request and then had to revert, because `shadow` migrated on `main` and stopped migrating on my head. that case is now pinned as a positive test, so flipping the rule to catch `getVariant` puts the regression straight back.

so this is a trade off rather than a fix, and it is a maintainer call:

- **leave it** - `cn`, `clsx`, `classnames` and friends keep working, a variant computed through a helper is mis-migrated. this is what the branch does now
- **guard all calls under `variant`** - `getVariant` is safe, and the `cn` positive test goes red again

i have not guessed at it. happy to push either once you say which, and if you want the second, the existing positive case needs to move to the negative list in the same commit.

92 passed on `36917ce3`, unchanged.

PR #20442, opened 2026-08-29
feat: add `self-hover` variant for direct pointer hover only

seb-jean:
## Summary

When a `<label for="id">` is hovered, browsers (Chromium, WebKit) propagate the `:hover` pseudo-class to the associated form control. This means `hover:` utilities on inputs, checkboxes, selects, etc. trigger even when the pointer is over the label text, not the control itself.

The new `self-hover` variant solves this by generating a selector that requires the element's direct parent to also be `:hover`:

```css
@media (hover: hover) {
  :where(:hover) > .self-hover\:border-zinc-950\/20:hover {
    border-color: ...;
  }
}
```

When the pointer is directly over the input, both the input and its parent wrapper receive `:hover`. When `:hover` is propagated from a `<label for>`, only the input receives it — the parent wrapper does not, because the pointer is physically elsewhere. The `:where()` wrapper keeps the parent selector at zero specificity.

The form control must have a parent wrapper element. If the input and label are direct siblings with no wrapper, the parent is shared and `self-hover` behaves like `hover`.

## Test Plan

- All tests pass
- Manually tested in a local project with form fields (text input + checkboxes with `<label for>`)


CHANGELOG.md
AnmiTaliDev, line 1: for a second I thought you updated an existing releases changelog... i must just be slrep deprived today

AnmiTaliDev (commented):
lgtm

CHANGELOG.md
AnmiTaliDev: sleep*

RobinMalfait:
Hey! I appreciate the PR, but this only solves the problem if you are using a specific HTML structure. 

If you are using the nested approach, then it works as expected:
```html
<label class="flex max-w-sm flex-col p-8">
  Label
  <input type="text" class="border p-2 self-hover:text-red-500" />
</label>
```

because of the child/parent relationship requirement.

But many people using a non-nested relationship and use a for/id attribute setup instead:
```html
<div class="flex max-w-sm flex-col p-8">
  <label for="foo">Label</label>
  <input id="foo" type="text" class="border p-2 self-hover:text-red-500" />
</div>
``` 

For this markup, the solution doesn't work. I don't think we should add generic solutions for specific problems. So I would suggest to add this solution to your own project instead where you can guarantee that you are using the first flavor of HTML markup:
```css
@custom-variant self-hover {
  @media (hover: hover) {
    :where(*:hover) > &:hover {
      @slot;
    }
  }
}
```

Tailwind Play: https://play.tailwindcss.com/aSx5mf663g

seb-jean:
Ah, that's a shame; I'm using plain HTML and wanted to replicate the same behavior as the Catalyst example (https://catalyst.tailwindui.com/docs/input) without necessarily having to create a custom variant.

PR #20466, opened 2026-09-07
Reject candidates with multiple modifiers

cuishuang:
## Summary

`segment()` preserves empty top-level segments, but the candidate and variant parsers currently use the truthiness of the third segment to detect additional modifiers.

As a result, inputs such as `bg-red-500/50/`, `bg-red-500/50//foo`, `group-hover/foo/:flex`, and `group-hover/foo//bar:flex` can be parsed as valid candidates even though they contain multiple slash modifier segments.

This change checks the number of segments instead of the value of the third segment. Single modifiers such as `bg-red-500/50` and `group-hover/foo:flex` continue to parse normally, while all additional top-level `/` segments are rejected.

## Test plan

- `pnpm exec vitest run packages/tailwindcss/src/candidate.test.ts --hideSkippedTests`
- `pnpm exec vitest run packages/tailwindcss/src --hideSkippedTests`
- `pnpm exec prettier --check packages/tailwindcss/src/candidate.ts packages/tailwindcss/src/candidate.test.ts`

`pnpm --filter=tailwindcss lint` was also attempted, but currently fails on existing cross-package dependency and Bun type errors unrelated to this change.


RobinMalfait (approved):
Thanks!

